PR #2, opened 2017-04-03
skeleton for S3 integration

ttung:
Depends on #1 

Forgot to convert this out into a PR. :)


blobstore/blobstore.py
```diff
@@ -0,0 +1,42 @@
...
+class BlobStore(object):
+    """Abstract base class for all blob stores."""
+    def __init__(self):
+        pass
+
+    def list(self, prefix=None):
+        """Returns a list of all blob entries in a container that match a given prefix."""
+        raise NotImplementedError()
+
+    def get_presigned_url(self, objname):
```
kislyuk: Would this be a pre-signed read URL or pre-signed write URL? If read, I suggest renaming this to `get_presigned_read_url()` for the avoidance of doubt. If write or both, we'll need a more comprehensive model of the underlying APIs here (since different object stores use different upload models).

blobstore/blobstore.py
```diff
@@ -0,0 +1,42 @@
...
+        """Retrieves a presigned url for blob objname.  Raises BlobNotFoundError if the blob is not present."""
+        raise NotImplementedError()
+
+    def set(self, objname, src_file_handle):
```
kislyuk: This is for unit testing only, right? We wouldn't want to slurp blob data through the upload route in the actual API - we want some sort of smart client behavior based on either a redirect or temporary upload credentials returned by the upload route.

test/s3blobstoretest.py
```diff
@@ -0,0 +1,28 @@
...
+class TestS3BlobStore(unittest.TestCase):
+    def setUp(self):
+        self.aws_access_key = os.environ['AWS_ACCESS_KEY']
+        self.aws_secret_key = os.environ['AWS_SECRET_KEY']
+
+    def tearDown(self):
+        pass
+
+    def test_connect(self):
+        s3blobstore = S3BlobStore("czi-hca-test", self.aws_access_key, self.aws_secret_key)
```
kislyuk: Probably want the bucket name parameterized through an environment variable or config file as well.

blobstore/blobstore.py
ttung: Good idea.

blobstore/blobstore.py
ttung: I didn't realize there was a presigned upload endpoint. Definitely will be tweaking this as I gain better understanding of what's available on S3...

test/s3blobstoretest.py
ttung: 👍

kislyuk:
LGTM, will tweak things as we go.

PR #3, opened 2017-04-20
ignore gcs credential file

ttung:
Ignore the gcs credential file so we don't mistakenly commit credential data to the public repo.


kislyuk:
LGTM. Tests fail for unrelated reason

PR #8, opened 2017-05-13
add flask-failsafe to permit effortless reloads when there's a syntax error

ttung:
Otherwise dss-app crashes each time there's a syntax error.


ttung:
Not sure why travis failed here when it passed here: https://travis-ci.org/HumanCellAtlas/data-store/jobs/231763280

ttung:
This got merged in with #9.

PR #9, opened 2017-05-15
rename test -> tests

ttung:
Rationale: There's a standard python library named "test".  There is not a "tests".


kislyuk:
Huh, didn't know that could cause it to fail. Thanks.

PR #11, opened 2017-05-19
Refactor blobstore code structure. Begin effort to make blobstore drivers initialize lazily.

kislyuk:


blobstore/blobstore.pyi
```diff
@@ -1,20 +0,0 @@
-from typing import *
```
ttung: no type hints?

README.md
```diff
@@ -24,5 +24,19 @@ Run `./dss-api` in this directory.
 #### Running tests
 Run `make test` in this directory.
 
+#### Configuring cloud-specific access credentials
```
ttung: thanks so much for writing this up.  Setting the bucket permissions up right was challenging, but probably not required for this.

PR #12, opened 2017-05-22
Add definitions for describing the file metadata and for describing the file contents of a bundle.

ttung:
`blob_common_metadata` is the base type.
`file_version` is the file object in the bundle list.
`file_metadata` is the file metadata (searchable by file uuid)

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -126,6 +126,73 @@ definitions:
       - uuid
       - name
       - versions
+  blob_common_metadata:
+    type: object
+    description: Common fields between the files list of a bundle and the metadata tracking an unique file.
+    properties:
+      uuid:
+        type: string
+        description: A RFC4122-compliant ID for the file.
+      timestamp:
+        type: string
+        format: date-time
+        description: Timestamp of bundle creation in RFC3339.
+      crc32c:
+        type: string
+        description: CRC-32C of the file contents in hex
+      md5:
+        type: string
+        description: MD5 of the file contents in hex
+      sha1:
+        type: string
+        description: SHA-1 of the file contents in hex
+      sha256:
+        type: string
+        description: SHA-256 of the file contents in hex
+    required:
+      - uuid
+      - timestamp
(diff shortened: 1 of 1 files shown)


dss-api.yml
```diff
@@ -126,6 +126,73 @@ definitions:
...
+    type: object
+    description: Common fields between the files list of a bundle and the metadata tracking an unique file.
+    properties:
+      uuid:
+        type: string
+        description: File unique ID
+      timestamp:
+        type: integer
+        format: int64
+        description: Timestamp of file creation measured in milliseconds since epoch
```
kislyuk: Are we doing milliseconds or ISO timestamps? In the spec I think I saw ISO timestamps. I'm on the fence with maybe a little bit of preference for ISO timestamps since they're human readable and more type inferable.

ttung: ISO timestamps are restricted to second-level accuracy.  We should ideally have more accuracy than that.

Having said that, it seems people do tack on the milliseconds onto ISO timestamps, though at that point they're not ISO 8601 compliant.

Thoughts?

ttung:
Updated the descriptions for the UUID fields.
Changed writer_uid to creator_uid.

dss-api.yml
kislyuk: Doesn't ISO 8601 support milliseconds with an optional decimal? https://www.w3.org/TR/NOTE-datetime seems to indicate that there can be a variable number of fractional digits after the optional dot. [Also found this](https://nbsoftsolutions.com/blog/iso-8601-and-nanosecond-precision-across-languages). I can't see the actual standard since it seems you have to pay money for it (what a mess).

So IMO that shouldn't be the deciding factor. I'm leaning in favor of ISO timestamps.

kislyuk: For documentation purposes we may want to say "RFC 3339 compliant timestamp string"

ttung: You are correct.  Amending.

PR #14, opened 2017-05-22
Flesh out the bundle api.

ttung:
1) Add `list_versions`, which just returns a list of versions.
2) Given a bundle UUID+version, include a list of files (i.e., `file_version`)

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -80,6 +80,28 @@ paths:
           description: OK
   /bundles/{uuid}:
     get:
+      operationId: dss.api.bundles.list_versions
+      parameters:
+        - name: uuid
+          in: path
+          description: Bundle unique ID.
+          required: true
+          type: string
+      responses:
+        200:
+          description: OK
+          schema:
+            type: array
+            items:
+              type: string
+              format: date-time
+              description: Timestamp of bundle creation in RFC3339.
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+  /bundles/{uuid}/{bundle_version}:
+    get:
       parameters:
         - name: replica
           in: query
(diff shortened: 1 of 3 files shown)


dss-api.yml
```diff
@@ -208,6 +230,24 @@ definitions:
...
+        description: Bundle unique ID
+      bundle_timestamp:
+        type: integer
+        format: int64
+        description: Timestamp of bundle creation measured in milliseconds since epoch
+      contents:
+        type: array
+        items:
+          $ref: "#/definitions/file_version"
+      writer_uid:
```
kislyuk: Should we call this "created_by" or "creator_uid"?

dss-api.yml
```diff
@@ -208,6 +230,24 @@ definitions:
...
+  bundle_version:
+    type: object
+    properties:
+      uuid:
+        type: string
+        description: Bundle unique ID
+      bundle_timestamp:
+        type: integer
+        format: int64
+        description: Timestamp of bundle creation measured in milliseconds since epoch
```
kislyuk: Same question as elsewhere, timestamp format - iso or epoch?

dss-api.yml
ttung: I don't like `created_by`, since there's ambiguity about the data type (name?  uid?).  I'll change it to `creator_uid`.

ttung:
I changed this but I don't know how to re-request review... :(

kislyuk:
LGTM thanks!

kislyuk:
Did you mean to merge this into `ttung_rename` or master? I can re-merge into master if the latter.

ttung:
It was intended for master, but in order to show stacked diffs correctly, I had to use ttung_rename as the base.  Github is not great at stacked diffs. :(

ttung:
I fixed it and manually pushed.  Thanks for flagging it, @kislyuk!

PR #15, opened 2017-05-24
Continuous deployment instrumentation

kislyuk:
- Use [Chalice](https://github.com/awslabs/chalice) to enable deployment of the app to AWS Lambda/API Gateway
- Add a Chalice-Connexion adapter to preserve the ability to use Connexion validation
- Add a `deploy` config to Travis CI to perform automatic deployment any time we push to master

In our AWS account, I have separately set up the DNS/SSL/CloudFront to serve the app on https://hca-dss.czi.technology/v1/.

### DIFF
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -17,10 +17,14 @@ from flask_failsafe import failsafe
 
 logging.basicConfig(level=logging.DEBUG)
 
+app = None
+logger = None
+
 @failsafe
 def create_app():
+    global app, logger
     app = connexion.App(__name__)
+    logger = app.app.logger
     resolver = RestyResolver("dss.api", collection_endpoint_name="list")
     app.add_api('../dss-api.yml', resolver=resolver, validate_responses=True)
-
     return app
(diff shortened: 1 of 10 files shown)


dss/__init__.py
```diff
@@ -17,10 +17,14 @@
 
 logging.basicConfig(level=logging.DEBUG)
 
+app = None
```
ttung: why do these need to be globals?

can the request methods just get the current app object via `flask.current_app`?

kislyuk: Good point. I didn't know about `flask.current_app`. That works better as a global accessor. I'll delete this one.

kislyuk: Deleted.

ttung:
What credentials are required to deploy this?

ttung:
The deploy stage failed. Are you deploying using the credentials embedded in `.travis.yml`?  If so, were the AWS role extended with the proper permissions?

kislyuk:
@ttung - Sorry about the breakage. Yes, I need to audit the AWS role and add permissions according to the needs of the deploy stage. I'll do that later today.

kislyuk:
Should be all fixed now. I committed a few Travis fixes directly on master. I'll add some docs or a setup script for getting the IAM policy set up correctly for the CI IAM user.

PR #20, opened 2017-05-30
flesh out GET/HEAD support for /files/<uuid>

ttung:
1. Updated UUID definition.
 2. `replica` is now not required.  It is enforced in code for only GET requests.
 3. Added `timestamp` parameter.
 4. Added response headers.  Unfortunately, it looks like connexion does not verify the minLength and maxLength fields of string headers (see https://github.com/zalando/connexion/blob/master/connexion/decorators/response.py), but if they add the proper support, this should hopefully match.
 5. Use a shared code path for GET and HEAD.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -32,17 +32,97 @@ paths:
       parameters:
         - name: uuid
           in: path
-          description: File unique ID.
+          description: A RFC4122-compliant ID for the file.
           required: true
           type: string
         - name: replica
           in: query
           description: Replica to fetch from.
-          required: true
+          required: false
           type: string
+        - name: timestamp
+          in: query
+          description: Timestamp of file creation in RFC3339.  If this is not provided, the latest version is returned.
+          required: false
+          type: string
+          format: date-time
       responses:
+        200:
+          description: On a HEAD request, returns metadata
+          headers:
+            # edits to here should probably be reflected in the 302 section immediately below.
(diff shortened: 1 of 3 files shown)


ttung:
Changed `writer` --> `creator` following #12

ttung:
fixed copypasta referring to `bundle` when it should be `file`.

dss-api.yml
```diff
@@ -32,17 +32,97 @@ paths:
...
+          description: Timestamp of file creation in RFC3339.  If this is not provided, the latest version is returned.
+          required: false
           type: string
+          format: date-time
       responses:
+        200:
+          description: On a HEAD request, returns metadata
+          headers:
+            # edits to here should probably be reflected in the 302 section immediately below.
+            X-DSS-BUNDLE_UUID:
```
kislyuk: Should we use dashes throughout the header name?

ttung: Explanation: I used dashes for content-type because it's actually dash'ed in common usage.  I'm using _ as a space replacement.

Happy to change it to dashes everywhere if you think that's better.

kislyuk: I think we should change this to use dashes everywhere in header names for consistency. I'll merge this as is for now. If you agree, please open another PR or I can do this myself later.

ttung:
I added support for proper validation of headers and submitted a [PR](https://github.com/zalando/connexion/pull/459).

PR #22, opened 2017-05-31
switch MD5 to S3-etag

ttung:
MD5 is not truly supported by S3

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -68,23 +68,19 @@ paths:
             X-DSS-CRC32C:
               type: string
               description: CRC32C of the file contents in hex.
-              minLength: 8
-              maxLength: 8
+              pattern: "^[A-Za-z0-9]{8}$"
             X-DSS-S3_ETAG:
               type: string
               description: S3 ETag of the file contents.
-              minLength: 32
-              maxLength: 38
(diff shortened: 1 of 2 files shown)


kislyuk:
I think we need to discuss this more in the context of the download integrity control strategy. How does the consumer validate the download? At minimum, the consumer would need the part size (and the part size would need to be constant across the file, or there would need to be an agreed upon parameterization of the part size) in addition to the etag. Ideally, we would provide md5s of all the parts to the consumer, however we probably will not be able to deliver them in a header, since each md5 is at least 24 bytes (base64-encoded), which may go up to 240KB for 10K parts. Headers of this size will probably face problems getting delivered.

kislyuk:
Actually, I think providing md5s of all parts should not be treated as a priority. So what we really need is documentation of the part size or part size ramp that we intend to use.

ttung:
I had already updated the on-disk format spec to reflect how the chunking is happening.  Damn you Google for no subscription mechanism for docs!!

kislyuk:
Aha, thanks. Hmm.. I don't know if the smaller bound of 8MB chunks is optimal. I would suggest 64MB based on my experience. It better amortizes per-part costs on the S3 backend, uploader and downloader.

Otherwise I'm on board with the approach (though we should also clarify that we're using powers of 2, not powers of 10).

PR #23, opened 2017-05-31
use patterns instead of minLength/maxLength.

ttung:
This allows us to further constrain the valid input set. Again, headers are not verified by connexion, but one day.... :)


kislyuk:
This is cool, although it makes the swagger less readable :)

kislyuk:
As a result of the github outage, the CI results are not being delivered, even though CI passed. Merging.

PR #31, opened 2017-06-01
initial cut at PUT for /files/<uuid>

ttung:


tests/test.py
```diff
@@ -33,6 +35,20 @@ def test_file_api(self):
...
         self.assertEqual(res.status_code, requests.codes.found)
 
+        res = self.app.put(
+            '/v1/files/123',
+            data=json.dumps(
+                dict(source_url="s3://foobar",
+                     bundle_uuid=str(uuid.uuid4()),
+                     creator_uid=4321,
+                     content_type="text/html",
+                     )),
```
kislyuk: What do you think about writing a test helper mixin to be able to just say `res = self.assert_response('/v1/files/123', json=dict(...), code=requests.codes.created); self.assertEqual(res.json["foo"], bar)`?

dss/api/files.py
```diff
@@ -32,5 +33,5 @@ def get(uuid: str, replica: str=None, timestamp: str=None):
 def list():
     return dict(files=[dict(uuid="", name="", versions=[])])
 
-def post():
-    pass
+def put(uuid: str):
+    return jsonify(dict(timestamp="2017-06-01T19:21:17.068Z")), 201
```
kislyuk: 201 -> requests.codes.created?

dss-api.yml
```diff
@@ -119,6 +119,59 @@ paths:
...
+              content_type:
+                description: Content-type of the file.
+                type: string
+              timestamp:
+                description: Timestamp of file creation in RFC3339.
+                type: string
+                format: date-time
+            required:
+              - source_url
+              - bundle_uuid
```
kislyuk: I thought the model was to create individual files, then to create/update a bundle with a list of file UUIDs. Has that changed?

dss-api.yml
```diff
@@ -119,6 +119,59 @@ paths:
...
+            type: object
+            properties:
+              source_url:
+                description: Cloud URL for source data.
+                type: string
+                pattern: "^(gs|s3|wasb)://"
+              bundle_uuid:
+                description: A RFC4122-compliant ID for the bundle that contains this file.
+                type: string
+              creator_uid:
```
kislyuk: Shouldn't this be auto-assigned based on the future auth scheme that we adopt for the producer API?

dss-api.yml
```diff
@@ -119,6 +119,59 @@ paths:
...
+                description: User ID who is creating this file.
+                type: integer
+                format: int64
+              content_type:
```
kislyuk: Consider leaving this as optional. The default would then be to copy the content-type of the object referred in source_url.

tests/test.py
ttung: where is `self.assert_response` defined?

dss/api/files.py
ttung: will fix

dss-api.yml
ttung: Yes, that's correct.  However, the UUIDs are assigned ahead of time so we can build the reverse relationships from file UUIDs to bundle UUIDs.

dss-api.yml
ttung: Sure, but until we add that in, I'll leave this in place.

dss-api.yml
ttung: Would it be set as metadata on the object?

tests/test.py
kislyuk: I was thinking of a mixin like this:
```
class DSSTestRequest:
    def assert_response(self, path, **kwargs):
        ... request prep ...
        res = self.app.request(path, method=kwargs.get("method"), ...)
        if "code" in kwargs:
            self.assertEqual(res.status_code, kwargs["code"])
        ... response prep ...
        return res
```

kislyuk: And then:
```
class TestDSS(unittest.TestCase, DSSTestRequest):
    ...
```

dss-api.yml
kislyuk: I think yes - for example, all objects on S3 have content-types. When we do the ingest of the object body from the source URL, we can grab the content-type from its headers.

ttung: Sounds good to me.

tests/test.py
kislyuk: Also, Flask just merged support for easy JSON payload access in test fixtures (https://github.com/pallets/flask/pull/2358). It's not released yet, but it will make this boilerplate easier in the future.

ttung:
Updated:

1. remove content-type.
2. updated documentation to reflect what metadata the method is expecting to be set on the source blob item.
3. test helper changes punted to #35 and #36

kislyuk:
Merging this now so it doesn't go stale. Please update the test with new helpers at your convenience.

ttung:
Test refactor is in #36

PR #35, opened 2017-06-06
skeleton for assert_response

ttung:
This allows us to build simple test statements.  The next diff in this stack will refactor the existing tests to use the new facility.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -3,10 +3,17 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import os, sys, unittest
+import functools
 import json
+import os
+import re
 import requests
+import sys
+import typing
+import unittest
 import uuid
+
+from flask import wrappers
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
 sys.path.insert(0, pkg_root)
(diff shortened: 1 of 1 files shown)


tests/test.py
```diff
@@ -16,6 +21,48 @@
...
     def setUp(self):
         self.app = dss.create_app().app.test_client()
+        self.sre = re.compile("^assert(.+)Response")
+
+    def assertResponse(
+            self,
+            method: str,
+            path: str,
+            expected_code: int,
+            expected_headers: typing.Dict={},
```
kislyuk: for my education, why do you say `typing.Dict` instead of `dict` here?

tests/test.py
```diff
@@ -16,6 +21,48 @@
...
+            json_response_tester: typing.Optional[typing.Callable[[typing.Any], bool]]=None,
```
kislyuk: I would suggest using this pattern instead:
```
def assert_response(..., expect_json=True, **kwargs):
    if "body" in kwargs and isinstance(kwargs["body"], dict):
        headers["content-type"] = "application/json"
        kwargs["body"] = json.dumps(kwargs["body"])
    if expect_json:
        return res.json() # or whatever means we must use to unpack json for now
    else:
        return res
```
This avoids boilerplate and allows me to chain assertions directly or tersely on the return value.

Alternatively, we could do another wrapper around this helper to implement this (e.g. assert_json_response).

tests/test.py
ttung: Will revise.

tests/test.py
ttung: Adopted aspects of your suggested changes.

tests/test.py
```diff
@@ -16,6 +21,62 @@
...
+        self.sre = re.compile("^assert(.+)Response")
+
+    def assertResponse(
+            self,
+            method: str,
+            path: str,
+            expected_code: int,
+            json_request_body: typing.Optional[dict]=None,
+            parse_response_as_json: bool=False,
+            **kwargs) -> typing.Tuple[wrappers.Response, str, typing.Optional[dict]]:
```
kislyuk: Another typing newbie comment: can I just use `(x, y)` instead of `Tuple[x, y]`?

kislyuk:
Is there a missing `wrappers` import? Tests are failing.

tests/test.py
ttung: Doesn't seem like it.

ttung:
Moved import from #36 
Removed the `parse_response_as_json` parameter (instead just try-except it)

PR #36, opened 2017-06-06
refactor existing tests using self.assert*Response

ttung:


ttung:
Updated the tests for bundles as well.

tests/test.py
```diff
@@ -80,40 +80,36 @@ def __getattr__(self, item: str) -> typing.Any:
...
-        self.assertEqual(res.status_code, requests.codes.ok)
-        res = self.app.get("/v1/bundles/123")
-        self.assertEqual(res.status_code, requests.codes.ok)
-        res = self.app.get("/v1/bundles/123/55555")
-        self.assertEqual(res.status_code, requests.codes.bad_request)
-        res = self.app.get("/v1/bundles/123/55555?replica=foo")
-        self.assertEqual(res.status_code, requests.codes.ok)
+        self.assertGetResponse("/v1/bundles", requests.codes.ok)
+        self.assertGetResponse("/v1/bundles/123", requests.codes.ok)
+        self.assertGetResponse("/v1/bundles/123/55555", requests.codes.bad_request)
```
kislyuk: Should we have a default replica? I would imagine we could allow the user to omit it, and do geolocation or other optimizations in the future.

PR #59, opened 2017-06-09
S3 event bus and daemon skeletons

kislyuk:
- Begin daemon deployment instrumentation

- Add private route for pubsub push notifications (just a stub for now; in a future PR it will forward GCE pubsub events to the event bus)

- Add daemon build dep virtualenvs to CI cache

- Add S3 access to lambda policy

- Various chalice/domovoi deployment tweaks


kislyuk:
Added a way to parameterize bucket name in policies.

chalice/.chalice/policy.json
```diff
@@ -14,8 +14,8 @@
       "Effect": "Allow",
       "Action": "s3:*",
       "Resource": [
-        "arn:aws:s3:::hca-dss-test",
-        "arn:aws:s3:::hca-dss-test/*"
+        "arn:aws:s3:::$DSS_S3_TEST_BUCKET",
```
ttung: qq: is this policy written to AWS?

kislyuk: It is written to AWS (by Chalice) only when deploying locally. When deploying on Travis CI, it skips that step (to avoid needing IAM write permissions).

PR #60, opened 2017-06-12
Don't associate the {gcs,s3}_blobstore handle with a container

ttung:
There are methods that involve multiple buckets, so it doesn't make sense to tie the handle to a container.  Verify that the handle is sane through a sensible API.


ttung:
@mikebaumann copy_object takes two bucket parameters.

PR #64, opened 2017-06-13
add copy methods to BlobStore and S3BlobStore

ttung:


dss/blobstore/s3.py
```diff
@@ -51,3 +51,26 @@ def get_metadata(self, container: str, object_name: str):
...
+            src_container: str, src_object_name: str,
+            dst_container: str, dst_object_name: str,
+            **kwargs
+    ):
+        # TODO: note that this implementation does not (yet) copy objects larger
+        # than 5GB.  for such files, we need a multipart copy operation.  see
+        # http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html
+        # and
+        # http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Client.upload_part_copy
+        # for more details.
```
kislyuk: Instead of this warning, I suggest refactoring this to use the managed multipart copy functionality provided by boto3: http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Object.copy

ttung:
Pushed the wrong rev.  Reverting this.

PR #65, opened 2017-06-13
Added a hcablobstore module with a copy method.

ttung:
BlobStore = convenience methods to do generic cloud storage things.
HCABlobStore = convenience methods to do HCA-specific cloud storage things.

For instance, copying while preserving the metadata we expect to preserve from blobs in the staging area: distinctly a HCA-specific functionality.

Also added a test to verify the hca-specific copy method.


.travis.yml
```diff
@@ -34,6 +34,7 @@ env:
   global:
   - AWS_DEFAULT_REGION=us-east-1
   - DSS_S3_TEST_BUCKET=hca-dss-test
+  - DSS_S3_TEST_WRITE_BUCKET=hca-dss-test-write-dst
```
kislyuk: Not sure if I agree with this nomenclature, I find this naming confusing. I would suggest naming a "test staging bucket" for testing transfers across buckets. Otherwise I think our main bucket (DSS_S3_TEST_BUCKET) should be used for both read and write testing.

ttung:
Now this writes to the same bucket, and I also configured the bucket such that blobs with the testoutput/ prefix gets dropped after 5 days.

ttung:
Switched it back to using two buckets.  One is the source for data used in the test.  This is a bucket that the test user can only read from but cannot write to.  This has the convenient property of insulating the test data from bad code.

ttung:
read from: `hca-dss-test-src`
write to: `hca-dss-test`

I considered using permissions to make a ro section of `hca-dss-test`, but that means we're relying on the cloud provider being able to assign different permissions to various prefixes.  This seems like the future-proof way.

PR #68, opened 2017-06-15
update API for set and implement it

ttung:


dss/blobstore/__init__.py
```diff
@@ -17,6 +20,17 @@ def generate_presigned_url(self, object_name: str, method: str):
         """
         raise NotImplementedError()
 
+    def set(
```
kislyuk: `set` seems a little too generic, suggest using `set_from_fileobj` or something like that.

PR #69, opened 2017-06-15
rename bucket to container.

ttung:


dss/blobstore/__init__.py
```diff
@@ -17,6 +20,17 @@ def generate_presigned_url(self, object_name: str, method: str):
         """
         raise NotImplementedError()
 
+    def set(
```
kislyuk: To be updated wrt other branch.

kislyuk:
PR structurally fine but don't agree with the renaming - "container" is a much more generic term, with potential to confuse semantics in our code/docs. Suggest sticking with bucket or going for a more specific term like "blob_store_container". Looked at nomenclatures and google, aws use "bucket"; azure uses "container".

PR #71, opened 2017-06-15
enforce the uuid format.

ttung:


tests/test.py
```diff
@@ -64,7 +64,7 @@ def assertHeaders(
             response: wrappers.Response,
             expected_headers: dict = {}) -> None:
         for header_name, header_value in expected_headers.items():
-            self.assertEqual(response.headers[header_name], header_value)
```
ttung: i ninja'd this unrelated change in.  sorry.

ttung: `assertEqual` is deprecated FWIW.

kislyuk: yeah - assert_equal is not available in 2.7. since this project is switching to 3.6 only, we should go ahead and replace all deprecated methods with their new names.

PR #72, opened 2017-06-15
Separate get and head requests, but map it to the same code path.

ttung:
There's a connexion bug that results in some oddity.  Read the comment in files.py.head().

@Mackey22 needs this for the CLI.


kislyuk:
Thanks! Should we file a bug with connexion?

PR #77, opened 2017-06-16
catch cloud-specific exceptions for `get_metadata()`  and throw a dss exception.

ttung:
This way, we don't have to write cloud-specific error handling code everywhere.

### DIFF
--- a/dss/blobstore/s3.py
+++ b/dss/blobstore/s3.py
@@ -1,11 +1,16 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import typing
-
 import boto3
 import botocore
+import requests
+import typing
 
-from . import BlobStore, BlobStoreCredentialError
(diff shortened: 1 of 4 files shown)


ttung:
Added test, also I implemented this for google cloud as well.

dss/blobstore/s3.py
```diff
@@ -1,11 +1,12 @@
...
 
-import typing
-
 import boto3
 import botocore
+import requests
+import typing
 
-from . import BlobStore, BlobStoreCredentialError
+from . import BlobNotFoundError, BlobStore, BlobStoreCredentialError, \
```
kislyuk: Style nitpick, I'm a fan of the multiline tuple syntax:
```
from . import (a, b, c,
               foo, bar)
```

ttung:
added parens.

PR #78, opened 2017-06-16
Added deployment script to setup AWS Elasticsearch service.

mikebaumann:
If the Elasticsearch domain does not exist then it will create one, otherwise the existing service will be used.
This does not yet remove the indexes from an existing service.
It also does not yet provide the Elasticsearch endpoint to the dss-index Lambda.

Thanks in advance for your review!


daemons/Makefile
```diff
@@ -6,6 +6,7 @@ dss-sync dss-index:
 	git clean -df $@/domovoilib
 	cp -R ../dss ../dss-api.yml $@/domovoilib
 	./build_deploy_config.sh $@ $(STAGE)
+	if [[ $@ == "dss-index" ]]; then ./deploy_aws_elasticsearch.sh $@ $(STAGE); fi
```
ttung: make this multiline pls.

ttung: if this can be made multiline, please do so.  i forget if this is a restriction of makefiles though...

daemons/deploy_aws_elasticsearch.sh
```diff
@@ -0,0 +1,145 @@
...
+function check_elasticsearch_domain_exists() {
+    set +e
+    aws es list-domain-names | grep $elasticsearch_domain_name > /dev/null
+    result=$?
+    set -e
+    return $result
+}
+
+function wait_for_elasticsearch_domain_removed() {
+	timeout_value=$1
```
ttung: indentation is different here.

daemons/deploy_aws_elasticsearch.sh
```diff
@@ -0,0 +1,145 @@
...
+	start_time=`date +'%s'`
+	timeout_expiration=$(($start_time + $timeout_value))
+
+	while [[ `date +'%s'` -lt $timeout_expiration ]]; do
+	    check_elasticsearch_domain_exists
+	    if [[ $? -ne 0 ]]; then
+	        echo Elasticsearch domain has been removed
+	        return 0
+	    fi
+        echo Waiting for Elasticsearch domain to be removed ...
```
ttung: indent on this line and next.

ttung:
Just as a general comment: For `echo` commands, I've usually seen the argument inside double quotes.  Also, BSD sed and Linux sed has different functionality.  Did you test with one or the other?

daemons/Makefile
mikebaumann: Make does require commands to be on one line, which is why I did it this way.

daemons/deploy_aws_elasticsearch.sh
mikebaumann: Ah, a tab versus spaces issue. In the revision, all tabs are replaced by spaces and the indentation is now correct.

daemons/deploy_aws_elasticsearch.sh
mikebaumann: Ditto the above comment response.

mikebaumann:
WRT Tony's general comments ...
Yes, I agree. Better to quote strings passed to echo, and I have done so in the revision.
Also, based on your comment, I tested the sed expression on Linux (Ubuntu), in addition to my Mac (BSD sed) used for development. For this simple substitution of one string literal for another, they both worked the same. It's a good point, though. There are some significant differences between the two.

daemons/deploy_aws_elasticsearch.sh
```diff
@@ -0,0 +1,145 @@
...
+
+setup_elasticsearch_domain $elasticsearch_domain_name
+
+wait_for_elasticsearch_endpoint 300
+
+# TODO Delete existing indexes here via Python script
+
+display_elasticsearch_domain_info
+
+echo Completed Elasticsearch setup
```
kislyuk: General Bash style issue, if you're going to be writing this much shellcode, I would prefer that you not pass variables out-of-band and instead pass them as arguments e.g. rewrite
```
foo() {
    echo $x
}
x=1
foo
```
as
```
foo() {
    local x=$1
    echo $x
}
ix=5
foo $ix
```

daemons/deploy_aws_elasticsearch.sh
```diff
@@ -0,0 +1,145 @@
...
+    aws es create-elasticsearch-domain \
+      --domain-name $elasticsearch_domain_name \
+      --elasticsearch-version $elasticsearch_version \
+      --elasticsearch-cluster-config $elasticsearch_cluster_config \
+      --ebs-options $ebs_options
+}
+
+function update_elasticsearch_domain_policy() {
+    echo Updating policy for Elasticsearch domain with name: $elasticsearch_domain_name
+    sed "s/ELASTICSEARCH_DOMAIN/$elasticsearch_domain_name/g" ./es_policy.json.template > ./es_policy.json
```
kislyuk: Generally it's preferable to use `$(dirname $0)/` instead of relative paths (`./`) to make sure the script can run when invoked from other directories.

daemons/deploy_aws_elasticsearch.sh
mikebaumann: Yes, I agree. I just pushed a commit that corrects that.

daemons/deploy_aws_elasticsearch.sh
mikebaumann: Yes, great point. I'll put that in now.

PR #80, opened 2017-06-16
Dockerize

sampierson:
Setup for Docker (development, mainly)
  - runs as root for now (this makes smaller images).
  - Installs vim, a .vimrc, .bashrc (for color prompt) and awscli.

Setup for docker-compose
  - "." and "~/.aws" are mounted in the container.


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN pip install awscli --upgrade
+
+ADD .dockerfiles/* /root/
+
+WORKDIR /code/data-store
+ADD requirements-dev.txt .
+RUN pip install --requirement requirements-dev.txt
+ADD requirements.txt .
+RUN pip install --requirement requirements.txt
+ADD . /code/data-store
```
kislyuk: These commands now live in two different places, `.travis.yml` and `Dockerfile`. Should we instead put them in a Makefile target and call them from both places?

.env
```diff
@@ -0,0 +1 @@
+COMPOSE_PROJECT_NAME=hca
```
kislyuk: Not sure what your intended nomenclature is here, but this repo is for the HCA DCP DSS (Human Cell Atlas Data Coordination Platform Data Storage System) so calling this "hca" could be overly general.

.dockerfiles/.bashrc
```diff
@@ -0,0 +1,113 @@
...
+# enable programmable completion features (you don't need to enable
+# this, if it's already enabled in /etc/bash.bashrc and /etc/profile
+# sources /etc/bash.bashrc).
+if ! shopt -oq posix; then
+  if [ -f /usr/share/bash-completion/bash_completion ]; then
+    . /usr/share/bash-completion/bash_completion
+  elif [ -f /etc/bash_completion ]; then
+    . /etc/bash_completion
+  fi
+fi
```
kislyuk: This file is just a bunch of bashrc boilerplate, is it possible to include it by reference from elsewhere (like copy it from a known location)?

Dockerfile
sampierson: I'm greatly in favor of DRYness but in this case it would defeat Docker's caching algorithm which is very smart about only performing subsequent steps if files referenced in the previous steps change.  Docker build run faster because of this.

.env
sampierson: I'll change it to hca_dcp so the resulting image will be named hca_dcp_dss.

.dockerfiles/.bashrc
sampierson: I can sed /etc/skel/.bashrc instead.

.env
sampierson: It turns out that docker-compose doesn't support underscores or dashes in COMPOSE_PROJECT_NAME, so options are: hcadcp_dss or dcp_dss.  Which would you prefer?

sampierson: Actually if we use `hca` as the project name and name the individual compose image as `dcp_dss`, we can get to `hca_dcp_dss`.

.dockerfiles/.bashrc
kislyuk: That sounds good!

.env
kislyuk: Oh, I'm actually fine with leaving this as is. I'm just not very familiar with docker-compose, so I thought for a moment the image would be called "hca".

Dockerfile
kislyuk: OK, sounds good.

kislyuk:
Please merge at your convenience.

PR #75, opened 2017-06-16
API for generating a presigned url and test for it.

ttung:
This is the same as #70, which I accidentally closed.  Oops!

PR #81, opened 2017-06-16
Clean up the fetching of bucket env variables.

ttung:
Addressing feedback from #75 .


ttung:
huhwhat i never closed this... :(

PR #85, opened 2017-06-16
implementation of PUT /files/{uuid}

ttung:
Retrieve the file from s3://hca-dss-test-src/test_good_source_data and uploaded it.

I visually confirmed that the file metadata + blob seems to have been correctly formed on S3.

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -1,13 +1,20 @@
 import binascii
+import datetime
 import hashlib
+import io
+import json
+import pyrfc3339
+import re
 import requests
 import uuid
-
 
 from flask import jsonify, make_response, redirect, request
 from werkzeug.exceptions import BadRequest
 
 from .. import get_logger
+from ..blobstore import BlobNotFoundError
+from ..config import Config
+from ..hcablobstore import FileMetadata
 
 
 def head(uuid: str, replica: str=None, timestamp: str=None):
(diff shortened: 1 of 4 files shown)


dss/api/files.py
```diff
@@ -45,8 +52,100 @@ def get(uuid: str, replica: str=None, timestamp: str=None):
...
+    )).lower()
+
+    # does it exist? if so, we can skip the copy part.
+    try:
+        handle.get_metadata(
+            dst_bucket, dst_object_name)
+    except BlobNotFoundError:
+        hca_handle.copy_blob_from_staging(
+            src_bucket, src_object_name,
+            dst_bucket, dst_object_name)
```
kislyuk: Consider rolling these into one operation, by using a put with a precondition (I think the only one you could use for this is `if-unmodified-since`, but I'm not 100% sure).

dss/api/files.py
```diff
@@ -45,8 +52,100 @@ def get(uuid: str, replica: str=None, timestamp: str=None):
...
+        FileMetadata.VERSION: FileMetadata.FILE_FORMAT_VERSION,
+        FileMetadata.BUNDLE_UUID: request_data['bundle_uuid'],
+        FileMetadata.CREATOR_UID: request_data['creator_uid'],
+        FileMetadata.TIMESTAMP: timestamp,
+        FileMetadata.CONTENT_TYPE: metadata['hca-dss-content-type'],
+        FileMetadata.CRC32C: metadata['hca-dss-crc32c'],
+        FileMetadata.S3_ETAG: metadata['hca-dss-s3_etag'],
+        FileMetadata.SHA1: metadata['hca-dss-sha1'],
+        FileMetadata.SHA256: metadata['hca-dss-sha256'],
+    })
```
kislyuk: What's our plan for not requiring the user to compute all of these?

dss/api/files.py
ttung: I think the conditions are only on the source, not the destination.

dss/api/files.py
ttung: Staging service will calculate all this.

Issue #43, opened 2017-06-06
Load sample data bundles


kislyuk:
Integration test to load bundles from https://github.com/HumanCellAtlas/data-bundle-examples and exercise the entire producer & consumer API using them

briandoconnor:
Hey, just FYI we're expecting the sample data bundles to be on a feature branch by Jim by next Tues 6/13 and hoping to get Laura's feedback and tag as a release on 6/16.

https://github.com/HumanCellAtlas/data-bundle-examples

sampierson:
Done, to a certain approximation.
Most of the 19,000 bundles were loaded into the dev data-store.
There were some errors, but these are now covered in separate issues, e.g. #399

PR #88, opened 2017-06-16
Integration test to import example-data-bundles (#43)

sampierson:
Okay teach me some Python.


sampierson:
Test are failing on the flake8 Python 3.4 tests.  Is f"" a 3.6 thing?

ttung:
Let's just junk 3.4 support. :)

tests/integration_test_example_bundles.py
```diff
@@ -0,0 +1,110 @@
...
+            self.upload_bundle(bundle)
+            self.check_bundle(bundle)
+
+    def upload_bundle(self, bundle):
+        for file in bundle.files:
+            self.upload_file(file)
+        self.create_bundle(bundle)
+
+    def upload_file(self, file):
+        self.progress("Uploading file %s" % file.path)
```
ttung: you can also use str.format

tests/integration_test_example_bundles.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def test_example_bundles(self):
+        for bundle in ExampleBundle.all():
+            self.progress("Processing bundle %s" % bundle.path)
+            self.upload_bundle(bundle)
+            self.check_bundle(bundle)
+
+    def upload_bundle(self, bundle):
+        for file in bundle.files:
+            self.upload_file(file)
```
ttung: planning ahead, you will want to save the timestamps from the upload operation for the `create_bundle` operation.

while you could set the timestamp during your file upload, that would mean multiple runners of this will collide.

kislyuk:
Please rebase on master (we had to rewrite history due to a bad merge).

tests/integration_test_example_bundles.py
sampierson: I'm going to switch back to f"" once Python 3.4 builds are stopped.

ttung: happy-dance

tests/integration_test_example_bundles.py
sampierson: You mean save the timestamp in the response from the PUT /files/<uuid> right?

ttung: Yeah.

ttung:
So one weird thing about this diff (as I understand it) is that you're testing against the deployed version and not whichever rev this is on. Is that problematic? If the test goes red in a branch, that tells the branch author nothing about their changes.

sampierson:
Good point Tony.  This is an interesting question though, and  fodder for a much larger discussion, about test/staging environments and what should be done in each.    It's good to have _something_ that's testing our full stack in a close to production style deployment as possible, and this is designed to be a thorough exercising and stress test of the API.
One idea would be to have the CI process push lambdas to a test AWS deployment then test against that.

ttung:
Please remember to squash.

And consider how we could build that CI process where we can tie a commit to the results. :)

ttung:
By the way, what might be a really good flow is for this to run with the deploy flow:

1. deploy to "staging"
2. run integration tests against "staging"
3. if integration tests pass, deploy to "prod"

It means we run this against commits on master, not every single random commit, which seems a bit unscalable.

PR #89, opened 2017-06-17
add urllib3 to requirements.txt

ttung:


ttung:
`make test` was failing for me because urllib3 wasn't installed. ninja-pushing this.

PR #90, opened 2017-06-17
implement BlobStore.list for S3

ttung:


dss/blobstore/s3.py
```diff
@@ -26,8 +26,44 @@ def __init__(self) -> None:
...
+            if 'Contents' not in response:
+                return
+            for item in response['Contents']:
+                yield item['Key']
+
+            if response['IsTruncated']:
+                next_marker = response.get(
+                    'NextMarker', response['Contents'][-1]['Key'])
+            else:
+                return
```
kislyuk: I suggest using a managed boto3 resource here.
```
def list_bucket():
    yield from boto3.resource("s3").Bucket("czi-hca-test").objects.filter(Prefix="abc", Marker="/")
```
The managed resource automatically takes care of pagination.

ttung:
Updated to use the higher-level API, and ready for review.

PR #97, opened 2017-06-19
move file-put tests to its own module.

ttung:
Rationale: file-get tests should refer to an immutable data set. Therefore, we need to point our bucket setup at a different location. Since the client is set up in `setUp`, we need to have a different `setUp` method for read and write tests.


ttung:
Might not need this.  Found an adequate workaround...

PR #101, opened 2017-06-19
API to get files!

ttung:
Removed a bunch of the old tests that merely verified the API was alive and replaced them with somewhat meaningful tests.

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -3,10 +3,12 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 
+import hashlib
 import os
 import sys
-import uuid
 import unittest
+import urllib
+import uuid
 
 import requests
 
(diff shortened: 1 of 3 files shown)


tests/test_file.py
```diff
@@ -56,6 +45,105 @@ def test_file_put(self):
...
+            "",
+            "/v1/files/" + file_uuid,
+            "",
+            urllib.parse.urlencode(
+                (("replica", "aws"),
+                 ("timestamp", timestamp)),
+                doseq=True,
+            ),
+            "",
+        ))
```
kislyuk: Is there a way to use kwargs here to make it more explicit what the different args refer to (query_string, host/netloc, etc)?

ttung: Ok, it's a bit complicated, but I think it's on the whole more readable...

PR #105, opened 2017-06-19
Indexer ready for integration testing

mikebaumann:
Note that the indexer unit test now expects a local Elasticsearch instance to be running, and fails with a message to that effect if it is not. For developers this may be fine, yet we need to consider the effect on the Travis builds.


kislyuk:
please fix the tests, it appears moto needs to be added to requirements-dev.txt

tests/expected_index_document.json
```diff
@@ -0,0 +1,166 @@
...
+                    "description": "suspensions were processed through Drop-Seq to generate single-cell cDNA libraries attached to microbeads. Microbeads were counted, and amplified by PCR, and the 3' end of the cDNA prepared for sequencing using a modified Nextera XT protocol.",
+                    "type": "nucleic acid library construction protocol"
+                }
+            ],
+            "short_label": "retina bipolar 1",
+            "submission_date": "2016-05-25",
+            "uuid": "5aca931f-9c88-4483-ba46-685b70c08a73"
+        }
+    }
+}
```
kislyuk: For all of these data bundle files, I suggest that you pull them from the submodule instead of copying them here.

mikebaumann: Thanks, I have changed the tests to use the files data-bundle-examples submodule.

mikebaumann:
Thanks for your review! I added moto to requirements-dev.txt and worked with you to add Elasticsearch to the Travis build. As noted above, I also changed the index text to use data from the data-bundle-examples submodule, including dynamically creating and removing files in the bundle that are not to be indexed, for test purposes. The manifest (including the non-indexed files) is included in the index document (as per the spec) but the content of the non-indexed files are not included in the index document. The combination of both indexed and non-indexed files in a bundle is tested.

PR #112, opened 2017-06-20
Clean up the bundle interface.

ttung:
1) list operation should belong inside the /bundles uri.
2) timestamp -> version
3) enforce data formats such as version: date-time and replica: (AWS|GC|Azure)


ttung:
Originally /bundles/<uuid> would list the versions of the bundles.  I think it belongs in the general /bundles query interface, to be implemented.

PR #113, opened 2017-06-20
Move version field into the query string.

ttung:
This is to make it consistent with bundles.


ttung:
cc: @Mackey22 @sampierson  this may affect you.

PR #125, opened 2017-06-22
Patch the tags into the metadata set.

ttung:
Patch the tags into the metadata set.

For S3, metadata = metadata + tags.

Other cloud providers do not appear to have the notion of tags, so this is a S3-specific thing.

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -45,6 +45,26 @@ class TestFileApi(unittest.TestCase, DSSAsserts):
         )
         self.assertIn('version', response[2])
 
+    def test_file_put_metadata_from_tags(self):
+        file_uuid = uuid.uuid4()
+        response = self.assertPutResponse(
+            "/v1/files/" + str(file_uuid),
+            requests.codes.created,
+            json_request_body=dict(
+                source_url="s3://hca-dss-test-src/test_good_source_data/metadata_in_tags",
+                bundle_uuid=str(uuid.uuid4()),
+                creator_uid=4321,
+                content_type="text/html",
+            ),
+        )
+        self.assertHeaders(
+            response[0],
+            {
+                'content-type': "application/json",
+            }
+        )
+        self.assertIn('version', response[2])
+
     def test_file_head(self):
         file_uuid = "ce55fd51-7833-469b-be0b-5da88ebebfcd"
         version = "2017-06-16T19:36:04.240704Z"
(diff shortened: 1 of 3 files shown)


sampierson:
I'm not sure this approach will work.  If the metadata does not contain all required keys, then dss.api.files.put() detonates with a KeyError at https://github.com/HumanCellAtlas/data-store/blob/master/dss/api/files.py#L62-L67.  You need to merge the metadata and tags before that code.

sampierson:
Also you're indexing into the file_metadata with short name, e.g. 
"s3_etag" instead of "hca-dss-s3_etag" which was in the version from a few days ago.  Is this what we want now?  If so will have to go back and re-tag all the example-bundle-data.

ttung:
screen shot 2017-06-22 at 8 28 20 am
These are the tags in my test file, which is hca-dss-*

ttung:
This only solves the path in PUT /files.

Are you constructing the on-disk layout without going through PUT /files?

sampierson:
Okay, but per my first comment, I don't see how we are going to get that far.

sampierson:
All I know is my experience from testing.  Can you explain to me how this is going to fix the error I saw:
```
'  File "/var/task/chalicelib/dss/api/files.py", line 94, in put\n'
 "    metadata['hca-dss-sha256'],\n"
 "KeyError: 'hca-dss-sha256'\n")
```

ttung:
Yeah you're right. Let me fix this.

ttung:
Ok this is now ready for review.

tests/test_file.py
```diff
@@ -45,6 +45,26 @@ def test_file_put(self):
         )
         self.assertIn('version', response[2])
 
+    def test_file_put_metadata_from_tags(self):
+        file_uuid = uuid.uuid4()
+        response = self.assertPutResponse(
+            "/v1/files/" + str(file_uuid),
+            requests.codes.created,
+            json_request_body=dict(
+                source_url="s3://hca-dss-test-src/test_bad_source_data/metadata_in_tags",
```
sampierson: Should this fixture be in test_bad_source_data?  I was confused.  It looks like good data to me.  We are expecting a created response code.

ttung: Unclear.  In the long term, is this going to be a supported mode of operation?

ttung: I guess it has to be... All right, I need to rework this PR to make it clear that this is the norm.

ttung:
Updated test fixture location.

PR #127, opened 2017-06-22
[easy] Make linter config more general

kislyuk:


kislyuk:
mypy doesn't like me.

PR #128, opened 2017-06-22
Fix indexer test and corrected handling of 'indexed' value

mikebaumann:
Fixed the  test for the change of the manifest 'files' section from a dict to a list.
Corrected handling of 'indexed' value in the test manifest and the resulting index document.

### DIFF
--- a/tests/expected_index_document.json
+++ b/tests/expected_index_document.json
@@ -1,168 +1,168 @@
 {
-  "state": "new",
-  "manifest": {
-    "version": "0.0.0",
-    "timestamp": "1498001450",
-    "files": [
-      {
-        "name": "assay.json",
-        "uuid": "525564a9-860a-44b8-b301-af0a983e7c95",
-        "timestamp": "1498000186",
-        "content-type": "metadata",
-        "indexed": "True",
-        "sha256": "7a900ce44e7ce7eb3daebef8baf402596c9533f74f5a06f1d191c2e4391d9e06",
-        "sha1": "e9f08d8e7856d09ce2d68c0abb0909615cf44365",
-        "s3-etag": "7edf489368f8a5ce72d7665c97155ba8",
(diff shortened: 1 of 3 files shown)


tests/expected_index_document.json
```diff
@@ -1,168 +1,168 @@
...
+            "uuid": "5aca931f-9c88-4483-ba46-685b70c08a73"
         }
-      ],
-      "short_label": "retina bipolar 1",
-      "submission_date": "2016-05-25",
-      "uuid": "5aca931f-9c88-4483-ba46-685b70c08a73"
     }
-  }
-}
+}
```
kislyuk: Add a newline at the end of the file please.

dss/events/handlers/index.py
```diff
@@ -67,7 +67,7 @@ def create_index_data(s3, bucket, bundle_key, manifest, log):
     index_files = {}
     bucket = s3.Bucket(os.getenv("DSS_S3_TEST_BUCKET"))
     for file_info in files_info:
-        if file_info['indexed'] is True:
+        if file_info['indexed'].lower() == 'true':
```
kislyuk: This does not correspond to the on-disk format of the bundles that I've seen. The on-disk format contains booleans, not strings. What's the rationale for changing it back here?

tests/sample_data_loader.py
```diff
@@ -95,7 +95,7 @@ def create_file_info(bundle_path: str, filename: str, indexed: bool):
     file_info['uuid'] = str(uuid.uuid4())
     file_info['timestamp'] = str(int(os.path.getctime(file_path)))
     file_info['content-type'] = "metadata"  # TODO What should the content type be?
-    file_info['indexed'] = str(indexed)
+    file_info['indexed'] = str(indexed).lower()
```
kislyuk: Why is this cast to str?

PR #129, opened 2017-06-22
Clarifications of the swagger spec.

ttung:
1. Return values are always lower case.  Input values remain case-insensitive.
2. Add patterns in some places where they are missing.
3. Clarify that the checksums are in hex format.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -68,21 +68,21 @@ paths:
               description: Content-type of the file.
               type: string
             X-DSS-CRC32C:
-              description: CRC32C of the file contents in hex.
+              description: CRC-32C (in hex format) of the file contents in hex.
               type: string
-              pattern: "^[A-Za-z0-9]{8}$"
+              pattern: "^[a-z0-9]{8}$"
             X-DSS-S3-ETAG:
-              description: S3 ETag of the file contents.
+              description: S3 ETag (in hex format) of the file contents.
               type: string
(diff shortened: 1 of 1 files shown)


dss-api.yml
```diff
@@ -68,21 +68,21 @@ paths:
...
+              description: CRC-32C (in hex format) of the file contents in hex.
               type: string
-              pattern: "^[A-Za-z0-9]{8}$"
+              pattern: "^[a-z0-9]{8}$"
             X-DSS-S3-ETAG:
-              description: S3 ETag of the file contents.
+              description: S3 ETag (in hex format) of the file contents.
               type: string
-              pattern: "^[A-Za-z0-9]{32}(-([2-9]|[1-8][0-9]|9[0-9]|[1-8][0-9]{2}|9[0-8][0-9]|99[0-9]|[1-8][0-9]{3}|9[0-8][0-9]{2}|99[0-8][0-9]|999[0-9]|10000))?$"
+              pattern: "^[a-z0-9]{32}(-([2-9]|[1-8][0-9]|9[0-9]|[1-8][0-9]{2}|9[0-8][0-9]|99[0-9]|[1-8][0-9]{3}|9[0-8][0-9]{2}|99[0-8][0-9]|999[0-9]|10000))?$"
```
sampierson: Wow what a ridiculous RegEx.  I'm guessing this came from Amazon. I mean yeah technically it's very accurate, but (-[0-9]+) is a whole lot more readable.  This is not a request for change.

PR #131, opened 2017-06-22
method to pull the cloud-provided checksum.

ttung:
Bonus test included.


sampierson:
Does GCS also use etags?  If not, then if the calling code is to be cloud-agnostic, it seems like returning the type of the checksum too would be useful, e.g. { 'sha1': "deadbeef" }

PR #133, opened 2017-06-22
Checksum-based mechanism for determining whether we copy.

ttung:
Now the new approach is:

1. Does the checksum on disk match our expectations?  If so, skip the copy.
2. Do the copy and verify that the copy was done correctly.


ttung:
#136

PR #140, opened 2017-06-23
API to generate a presigned URL on google cloud.

ttung:


dss/blobstore/gcs.py
```diff
@@ -43,6 +44,15 @@ def list(
...
             yield blob_obj.name
 
+    def generate_presigned_GET_url(
+            self,
+            bucket: str,
+            object_name: str,
+            **kwargs) -> str:
+        bucket_obj = self._ensure_bucket_loaded(bucket)
+        blob_obj = bucket_obj.get_blob(object_name)
+        return blob_obj.generate_signed_url(datetime.timedelta(1))
```
kislyuk: Suggest using `datetime.timedelta(days=1)` for explicitness. Also eventually the API will want to expose the lifetime as an input parameter, subject to minimum (0) and maximum bounds.

ttung:
Updated and ready for re-review.

ttung:
Eh, I'll just land. :)

PR #142, opened 2017-06-23
GCS support for fetching an object.

ttung:


dss/blobstore/gcs.py
```diff
@@ -62,6 +62,21 @@ def upload_file_handle(
...
+        :param object_name: the name of the object for which metadata is being
+        retrieved.
+        :return: the data
+        """
+        bucket_obj = self._ensure_bucket_loaded(bucket)
+        blob_obj = bucket_obj.get_blob(object_name)
+        if blob_obj is None:
+            raise BlobNotFoundError()
+
+        return blob_obj.download_as_string()
```
sampierson: (no action required)
Wow Google what a horribly named method.  This is Python so download_as_string() implies to be that a `.decode('utf-8')` or something is happening and it's returning a string, when in fact it is returning bytes.

PR #144, opened 2017-06-23
GCS support of verify_blob_checksum

ttung:


dss/hcablobstore/gcs.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class GCSHCABlobStore(HCABlobStore):
+    def __init__(self, handle: BlobStore) -> None:
+        self.handle = handle
+
+    def verify_blob_checksum(self, bucket: str, object_name: str, metadata: typing.Dict[str, str]) -> bool:
+        """
+        Given a blob, verify that the checksum on the cloud store matches the checksum in the metadata dictionary.  The
+        keys to the metadata dictionary will be the items in ``COPIED_METADATA``.  Each cloud-specific implementation
```
sampierson: COPIED_METADATA -> MANDATORY_METADATA?

tests/test_gcshcablobstore.py
```diff
@@ -0,0 +1,58 @@
...
+                {
+                    HCABlobStore.MANDATORY_METADATA['CRC32C']['keyname']: "e16e07b9",
+                }
+            )
+        )
+
+        self.assertFalse(
+            self.hcahandle.verify_blob_checksum(
+                self.test_src_data_bucket, "test_good_source_data/1",
+                {
```
sampierson: Looks like bad source data not good source data if the checksum is wrong.
Confused.

tests/test_gcshcablobstore.py
```diff
@@ -0,0 +1,58 @@
...
+                    HCABlobStore.MANDATORY_METADATA['CRC32C']['keyname']: "e16e07b9",
+                }
+            )
+        )
+
+        with self.assertRaises(BlobNotFoundError):
+            self.hcahandle.verify_blob_checksum(
+                self.test_src_data_bucket, "test_good_source_data/0/DOES_NOT_EXIST",
```
sampierson: Looks like bad source data not good source data if we're expecting an exception.

tests/test_gcshcablobstore.py
ttung: The checksum in `tests/test_gcshcablobstore.py` is wrong.  The checksum on the actual file is correct.

I'll restructure this test to pull the actual checksums from the cloud provider (rather than from a source file). #156

dss/hcablobstore/gcs.py
ttung: Fixed.

PR #145, opened 2017-06-23
GCS support of copy_blob

ttung:
Add test that apparently was never written


sampierson:
This test is gradually going to create many many files in the destination bucket.  How about a TearDown that deletes the test has completed?

ttung:
I plan on setting a cleanup policy on the bucket to nuke files after a fixed time interval.

There is already one set up on AWS but it is restricted to one prefix. Once we separate the buckets, I'll apply it to the entire output directory.

PR #146, opened 2017-06-23
Fix ES permissions; CI/CD/env infra improvements

kislyuk:
Sorry this is a hodge-podge of stuff. Working on un-screwing up our daemons and deployment infra.

- Update and parameterize Elasticsearch IAM access policies for Lambdas

- Begin centralized IAM policy templates and management scripts

- Begin centralized environment script

- Begin centralized Makefile util library

- Add check to Makefiles to ensure environment script is sourced before running commands

- Add docs for the deployment and CI/CD config process in the main readme

PR #147, opened 2017-06-23
Allow lambda dss-dev read-only access to hca-dss-test-src bucket

sampierson:


kislyuk:
IMO this should live in the source bucket's IAM policy, not in the application lambda's policy.

In other words, the staging service should grant the DSS access to its bucket. DSS should not declare that it has access to the staging service's bucket.

kislyuk:
What we could do though is add a script that applies this bucket policy (to be run from a privileged account) similar to the work in progress https://github.com/HumanCellAtlas/data-store/blob/akislyuk-fix-daemons-2/scripts/authorize_aws_deploy.sh (#146).

PR #148, opened 2017-06-23
Nicer error messages when we can't find a bundle or a file.

ttung:
```
{
  "HTTPStatusCode": 404,
  "exception": "An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.",
  "message": "Cannot find bundle."
}
```


ttung:
you mean for get-files, right?

PR #149, opened 2017-06-23
Basic setup for the files/bundles/their unit tests to communicate with GCS.

ttung:
Unit test in the next diff.

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -4,22 +4,36 @@ from contextlib import contextmanager
 
 from .blobstore import BlobStore
 from .blobstore.s3 import S3BlobStore
+from .blobstore.gcs import GCSBlobStore
 from .hcablobstore import HCABlobStore
 from .hcablobstore.s3 import S3HCABlobStore
+from .hcablobstore.gcs import GCSHCABlobStore
 
 
 class Config(object):
     _S3_BUCKET = None  # type: str
+    _GS_BUCKET = None  # type: str
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


dss/config.py
```diff
@@ -4,23 +4,37 @@
...
 
 class Config(object):
     _S3_BUCKET = None  # type: str
+    _GS_BUCKET = None  # type: str
 
     @staticmethod
     def get_cloud_specific_handles(replica: str) -> typing.Tuple[
             BlobStore, HCABlobStore, str]:
+
+        handle: BlobStore
```
kislyuk: Hmm... I don't understand this syntax, what is this type hint attached to?

ttung: handle.

type hinting is currently broken in python if the same variable is declared in two different paths of an if-statement.

dss/config.py
```diff
@@ -4,23 +4,37 @@
...
         if replica == 'aws':
             handle = S3BlobStore()
             return (
                 handle,
                 S3HCABlobStore(handle),
                 Config.get_s3_bucket()
             )
+        elif replica == 'gcs':
+            pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+            credentials = os.path.join(pkg_root, "gcs-credentials.json")
```
kislyuk: I think we want to always use the `GOOGLE_APPLICATION_CREDENTIALS` environment variable, no?

ttung: This is wrong in multiple places.  I'll fix in a follow-up PR.

PR #151, opened 2017-06-23
introduce the notion of prod/test/test_fixture configs

ttung:
Switch the override mechanism to use the BucketConfigs

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -1,6 +1,7 @@
 import os
 import typing
 from contextlib import contextmanager
+from enum import Enum, unique
 
 from .blobstore import BlobStore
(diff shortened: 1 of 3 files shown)


dss/config.py
```diff
@@ -10,9 +11,15 @@
 from .hcablobstore.gcs import GCSHCABlobStore
 
 
+class BucketConfig(Enum):
+    PROD = 0
+    TEST = 1
+    TEST_FIXTURE = 2
```
kislyuk: You can use the shorthand:
```
BucketConfig = Enum("BucketConfig", "PROD TEST TEST_FIXTURE")
```

PR #154, opened 2017-06-23
Quick API file/bundle full lifecycle test

sampierson:
Upload files, create bundle, retrieve bundle, retrieve files

### DIFF
--- a/tests/infra.py
+++ b/tests/infra.py
@@ -1,3 +1,5 @@
+import os
+import sys
 import json
 import logging
 import re
@@ -13,6 +15,17 @@ def start_verbose_logging():
     for logger_name in logging.Logger.manager.loggerDict:  # type: ignore
         if logger_name.startswith("botocore") or logger_name.startswith("boto3.resources"):
             logging.getLogger(logger_name).setLevel(logging.WARNING)
+
+
+def progress(message):
+    """
+    Print progress commentary if env 'VERBOSE' is set to anything.
+
+    Note it doesn't include a newline so you can produce dots... or "doing...done" lines.
+    """
+    if 'VERBOSE' in os.environ:
+        sys.stdout.write(message)
+        sys.stdout.flush()
 
 
 class DSSAsserts(object):
--- a/tests/infra.py
+++ b/tests/infra.py
@@ -1,5 +1,3 @@
-import os
-import sys
 import json
 import logging
 import re
@@ -15,17 +13,6 @@ def start_verbose_logging():
     for logger_name in logging.Logger.manager.loggerDict:  # type: ignore
         if logger_name.startswith("botocore") or logger_name.startswith("boto3.resources"):
             logging.getLogger(logger_name).setLevel(logging.WARNING)
-
-
-def progress(message):
-    """
-    Print progress commentary if env 'VERBOSE' is set to anything.
-
-    Note it doesn't include a newline so you can produce dots... or "doing...done" lines.
-    """
-    if 'VERBOSE' in os.environ:
-        sys.stdout.write(message)
-        sys.stdout.flush()
 
 
 class DSSAsserts(object):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -46,9 +46,9 @@ class TestApi(unittest.TestCase, DSSAsserts):
         self.get_bundle_and_check_files(bundle)
 
     def upload_files_and_create_bundle(self, bundle):
-        for file in bundle.files:
-            version = self.upload_file(file)
-            file.version = version
+        for s3file in bundle.files:
+            version = self.upload_file(s3file)
+            s3file.version = version
         self.create_bundle(bundle)
 
     def upload_file(self, file):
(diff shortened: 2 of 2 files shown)


ttung:
Rather than hit the live prod server, you should do this against a local instance like in `tests/test_file.py` or `tests/test_bundle.py`.

That way unit test run against the current version.

kislyuk:
I think we need to both run this end-to-end test against a local test server, and also have a Travis cron job to run it against the dev deployment channel (to be clear, we don't yet have staging or prod deployment channels) every few hours.

kislyuk:
I do agree, though, that running this test against the deployed API in the course of the unittest suite (which happens automagically for all `tests/test*.py` files) is not appropriate. What if for some reason the deployed API is broken, and you want the tests to be green before re-deploying? This test will prevent you from getting a green board.

sampierson:
I'll rework this.  On the one hand it's testing the API protocol, and that can be accomplished by hitting local code.  On the other hand hitting the live endpoint, while an important check, is more monitoring than testing and we should think about how we want to build monitoring infrastructure.

sampierson:
Okay how about this.

tests/test_api.py
```diff
@@ -0,0 +1,169 @@
...
+          - GET /bundles/<uuid>
+          - GET /files/<uuid>
+        and checks that data corresponds where appropriate.
+        """
+        bundle = S3TestBundle(self.BUNDLE_FIXTURE)
+        self.upload_files_and_create_bundle(bundle)
+        self.get_bundle_and_check_files(bundle)
+
+    def upload_files_and_create_bundle(self, bundle):
+        for file in bundle.files:
```
ttung: `file` is a keyword.  avoid using that.

tests/test_api.py
```diff
@@ -0,0 +1,169 @@
...
+class TestApi(unittest.TestCase, DSSAsserts):
+
+    def setUp(self):
+        DSSAsserts.setup(self)
+        self.app = dss.create_app().app.test_client()
+
+    BUNDLE_FIXTURE = 'fixtures/test_api/bundle'  # 's3://hca-dss-test-src/fixtures/test_api/bundle'
+
+    def test_creation_and_retrieval_of_files_and_bundle(self):
+        """Test file and bundle lifecycle.
```
ttung: NL before `Test`

tests/test_api.py
```diff
@@ -0,0 +1,169 @@
...
+        Exercises:
+          - PUT /files/<uuid>
+          - PUT /bundles/<uuid>
+          - GET /bundles/<uuid>
+          - GET /files/<uuid>
+        and checks that data corresponds where appropriate.
+        """
+        bundle = S3TestBundle(self.BUNDLE_FIXTURE)
```
ttung: why only S3?

tests/test_api.py
```diff
@@ -0,0 +1,169 @@
...
+        for file in bundle.files:
+            response = self.assertGetResponse(
+                str(UrlBuilder().set(path='/v1/files/' + file.uuid).add_query('replica', 'aws')),
+                requests.codes.found,
+            )
+            self.assertEqual(file.bundle.uuid, response[0].headers['X-DSS-BUNDLE-UUID'])
+            self.assertEqual(file.version, response[0].headers['X-DSS-VERSION'])
+
+    @classmethod
+    def progress(cls, message):
```
ttung: do you want this to go in `tests/infra.py`?

tests/test_api.py
```diff
@@ -0,0 +1,169 @@
...
+        self.assertIs(type(response_data), dict)
+        self.assertIn('version', response_data)
+        bundle.version = response_data['version']
+        self.progress(f"-> {bundle.version}\n")
+
+    @staticmethod
+    def put_bundle_payload(bundle):
+        payload = {
+            'uuid': bundle.uuid,
+            'creator_uid': 0,
```
ttung: let's not use 0.

tests/test_api.py
```diff
@@ -0,0 +1,169 @@
...
+                filedata = next(data for data in file_metadata if data['uuid'] == file.uuid)
+            except StopIteration:
+                self.fail(f"File {file.uuid} is missing from bundle")
+            self.assertEqual(filedata['uuid'], file.uuid)
+            self.assertEqual(filedata['name'], file.name)
+            self.assertEqual(filedata['version'], file.version)
+
+    def check_files_are_associated_with_bundle(self, bundle):
+        for file in bundle.files:
+            response = self.assertGetResponse(
```
ttung: Do you want to do a HEAD request instead? I don't know how big your files are.

tests/test_api.py
sampierson: fixed

tests/test_api.py
sampierson: Because this is just a fixture and the goal of the test is to test bundle lifecycle, not to exhaustively test the URLs that the system can access.

tests/test_api.py
sampierson: changed

tests/test_api.py
sampierson: Sure.  But I'm not testing the return value.  This is not a unit test.

tests/test_api.py
sampierson: No files are downloaded.  We're checking for a 302 response.  Are you saying that using GET causes extra processing on the back-end that we want to avoid?

tests/test_api.py
sampierson: Sure.

sampierson:
ready for review

tests/test_api.py
ttung: Is there value in testing the full lifecycle with GS?

tests/test_api.py
ttung: It doesn't result in processing in the back end, but if it's a large file, there's the transfer time.

tests/test_api.py
ttung: thanks!

tests/infra.py
```diff
@@ -15,6 +17,17 @@ def start_verbose_logging():
             logging.getLogger(logger_name).setLevel(logging.WARNING)
 
 
+def progress(message):
+    """
+    Print progress commentary if env 'VERBOSE' is set to anything.
+
+    Note it doesn't include a newline so you can produce dots... or "doing...done" lines.
+    """
+    if 'VERBOSE' in os.environ:
```
ttung: i would just quickly peruse https://docs.python.org/3/library/logging.html to see if there's anything that's appropriate there.

tests/test_api.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+Functional Test of the API
+"""
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import boto3
```
ttung: the @kislyuk model of imports:

```
python library imports

third-party library imports

internal (to this repo) imports
```

sampierson:
The motivation for this test was that on 6/23 I found a bug when I went through the complete lifecycle of file upload, bundle upload, bundle download and file download.  It wasn't until we exercised the complete lifecycle that we realized the downloaded files didn't refer to the correct bundle.  It appears we don't have another test that exercised the complete lifecycle, and this is something we can test without needing to integrate with other systems, so we should test it.

ttung:
> The motivation for this test was that on 6/23 I found a bug when I went through the complete lifecycle of file upload, bundle upload, bundle download and file download. It wasn't until we exercised the complete lifecycle that we realized the downloaded files didn't refer to the correct bundle. It appears we don't have another test that exercised the complete lifecycle, and this is something we can test without needing to integrate with other systems, so we should test it.

if that's the case, shouldn't we validate all the fields, like `creator_uid`?

sampierson:
test_bundle and test_file should exhaustively test all the attributes returned from individual calls.  This test tests the coordination between the calls.

PR #155, opened 2017-06-24
Sync daemon: copy data ourselves from S3 to GCE

kislyuk:
Disable GCSTS and use manual copying.

This depends on #157 which installs an unreleased version of google-cloud-storage (master of https://github.com/GoogleCloudPlatform/google-cloud-python) that supports reading chunks of filehandle input at a time in managed uploads, unlike the released version (v1.1.1).


daemons/dss-index/app.py
```diff
@@ -1,4 +1,6 @@
-import os, sys, json, boto3, domovoi
+import os, sys, json, boto3, logging
```
ttung: i still don't understand your methodology for formatting imports. :(

daemons/dss-index/app.py
```diff
@@ -11,4 +13,5 @@
 
 @app.s3_event_handler(bucket=s3_bucket, events=["s3:ObjectCreated:*"])
 def dispatch_indexer_event(event, context) -> None:
+    app.log.setLevel(logging.DEBUG)
```
ttung: did you mean to leave this in there?

daemons/dss-sync/app.py
```diff
@@ -12,7 +14,11 @@
...
-    bucket = s3.Bucket(event['Records'][0]["s3"]["bucket"]["name"])
-    obj = bucket.Object(unquote(event['Records'][0]["s3"]["object"]["key"]))
-    sync_blob(source_platform="s3", source_key=obj.key, dest_platform="gcs", logger=app.log)
+    app.log.setLevel(logging.DEBUG)
+    if event.get("Event") == "s3:TestEvent":
+        app.log.info("DSS sync daemon received S3 test event")
+    else:
+        s3 = boto3.resource("s3")
+        bucket = s3.Bucket(event['Records'][0]["s3"]["bucket"]["name"])
+        obj = bucket.Object(unquote(event['Records'][0]["s3"]["object"]["key"]))
```
ttung: is the unquote necessary?

dss/events/handlers/sync.py
```diff
@@ -80,12 +84,22 @@ def sync_blob(source_platform, source_key, dest_platform, logger):
...
+        with closing(http.request("GET", s3_blob_url, preload_content=False)) as fh:
+            gcs_bucket = gcs.get_bucket(gcs_bucket_name)
+            gcs_bucket.blob(source_key, chunk_size=1024 * 1024).upload_from_file(fh)
+        logger.info("Completed transfer of {} from {} to {}".format(source_key, s3_bucket_name, gcs_bucket_name))
     elif source_platform == "gcs" and dest_platform == "s3":
-        s3 = boto3.resource("s3")
         gcs_blob = gcs.get_bucket(gcs_bucket_name).blob(source_key)
-        gcs_blob_url = gcs_blob.generate_signed_url(expiration=int(time.time() + presigned_url_lifetime_seconds))
-        http = urllib3.PoolManager(cert_reqs="CERT_REQUIRED")
+        expires_timestamp = int(time.time() + presigned_url_lifetime_seconds)
```
ttung: strictly speaking, the reading presigned url needs a longer expiration since it got created first.

daemons/dss-index/app.py
kislyuk: Yes, for now I want all logs down to the debug level from the sync and indexer daemons to show up in cloudwatch.

daemons/dss-sync/app.py
kislyuk: It doesn't hurt. Even though we moved away from colons, S3 keys can generally be named almost anything. They get URI-encoded in the event notification, so we should decode them before we can use them.

daemons/dss-index/app.py
kislyuk: Normally I like to group imports into three groups (stdlib, third-party, local/relative) but within those groups, I have multiple imports per line for brevity. Here I accidentally put a third-party import (`domovoi`) among the stdlib imports so I'm breaking it out.

dss/events/handlers/sync.py
kislyuk: Sorry, I don't get it... it got created first, as opposed to what?

daemons/dss-index/app.py
kislyuk: NB this doesn't propagate to other loggers, so we won't get spam from connexion/urllib3 and such.

daemons/dss-index/app.py
kislyuk: Cleaned this up into one import per line. Broke out `boto3` into the third-party group.

kislyuk:
Added a few lines (and test) necessary to copy metadata when copying objects.

kislyuk:
Merging since I believe I've addressed the blocking comments and need this for further integration work.

If you have further feedback, please comment here and/or file an issue. Thanks :)

PR #159, opened 2017-06-26
Changes to indexing and query for the Boston Demo

mikebaumann:
The two key changes here are:

- Indexer: replacing dot with underscore in the filename field of the index document

- Search API, both get (as Andrey provided) and post (as needed for the Boston demo)

Related changes include some refactoring for the indexer and search API to share common constants and code. There is some general code clean-up too, mostly to index.py, which adds to the diffs below.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -10,6 +10,20 @@ basePath: /v1
 produces:
   - application/json
 paths:
+  /search:
+    get:
+      description: |
+        Returns a list of files matching given criteria.
+      summary: Find files and bundles by querying their metadata
+      parameters:
+        - name: query
+          in: query
+          description: Metadata query
+          required: true
+          type: string
+      responses:
+        200:
+          description: OK
   /files:
     get:
       summary: Query files
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -13,8 +13,8 @@ paths:
   /search:
     get:
       description: |
-        Returns a list of files matching given criteria.
-      summary: Find files and bundles by querying their metadata
+        Returns a list of bundles matching the given simple criteria
+      summary: Find bundles by querying their metadata using simple criteria
       parameters:
         - name: query
           in: query
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -24,6 +24,32 @@ paths:
       responses:
         200:
           description: OK
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    post:
+      description: |
+        Accepts Elasticsearch JSON query and returns matching bundle identifiers
+      summary: |
+        Find bundles by searching their metadata with an Elasticsearch query
+      consumes:
+        - application/json
+      parameters:
+        - name: query
+          in: body
+          description: Elasticsearch query
+          schema:
+            type: object
+      produces:
+        - application/json
+      responses:
+        200:
+          description: OK
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
   /files:
     get:
       summary: Query files
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -1,0 +1,22 @@
+import json
+import os
+
+from elasticsearch_dsl import Search
+from flask import request
+
+from .. import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
+from .. import get_logger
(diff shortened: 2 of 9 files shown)


tests/test_indexer.py
```diff
@@ -72,23 +74,44 @@ def test_process_new_indexable_object(self):
...
+                                \"files.assay_json.single_cell.method\": \"Fluidigm C1\" \
+                            } \
+                        }, { \
+                            \"match\": { \
+                                \"files.sample_json.ncbi_biosample\": \"SAMN04303778\" \
+                            } \
+                        }] \
+                    } \
+                } \
+            }"
```
kislyuk: Please rewrite this as a Python dict, instead of multiline string, and then call `es_client.search(body=json.dumps(query))`.

tests/test_search.py
```diff
@@ -0,0 +1,55 @@
...
+                                \"files.assay_json.single_cell.method\": \"Fluidigm C1\" \
+                            } \
+                        }, { \
+                            \"match\": { \
+                                \"files.sample_json.ncbi_biosample\": \"SAMN04303778\" \
+                            } \
+                        }] \
+                    } \
+                } \
+            }"
```
kislyuk: Same here.

kislyuk:
Looks good overall. After this week, let's figure out whether we need POST or GET /query.

dss-api.yml
```diff
@@ -10,6 +10,46 @@ basePath: /v1
 produces:
   - application/json
 paths:
+  /search:
+    get:
+      description: |
+        Returns a list of files matching given criteria.
+      summary: Find files and bundles by querying their metadata
```
mikebaumann: Correct(ed)

dss/api/search.py
```diff
@@ -0,0 +1,41 @@
...
+from elasticsearch_dsl import Search
+from flask import request
+
+from .. import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
+from .. import get_logger
+from ..util import connect_elasticsearch
+
+
+def list():
+    es_client = connect_elasticsearch(os.getenv("DSS_ES_ENDPOINT"), get_logger())  # TODO Use a connection manager
```
mikebaumann: Done

dss/api/search.py
```diff
@@ -0,0 +1,41 @@
...
+                      doc_type=DSS_ELASTICSEARCH_DOC_TYPE).query("match", **query).execute()
+    return {"query": query, "results": format_results(request, response)}
+
+
+def post(query: dict):
+    get_logger().debug("Received posted query: %s", json.dumps(query, indent=4))
+    es_client = connect_elasticsearch(os.getenv("DSS_ES_ENDPOINT"), get_logger())  # TODO Use a connection manager
+    response = Search(using=es_client,
+                      index=DSS_ELASTICSEARCH_INDEX_NAME,
+                      doc_type=DSS_ELASTICSEARCH_DOC_TYPE).update_from_dict(query).execute()
```
mikebaumann: Here update_from_dict is being used to set a fully defined / self-contained Elasticsearch query. In the case of .query("match", **query) a portion of a query is already being specified by "match".

dss/events/handlers/index.py
```diff
@@ -50,33 +43,41 @@ def is_bundle_to_index(key) -> bool:
...
     files_info = manifest['files']
     index_files = {}
-    bucket = s3.Bucket(os.getenv("DSS_S3_TEST_BUCKET"))
+    bucket = s3.Bucket(bucket_name)
     for file_info in files_info:
         if file_info['indexed'] is True:
-            if not file_info["name"].endswith(".json"):
-                log.warning("File %s is marked for indexing but is not of type JSON. It will not be indexed.")
+            if not file_info["name"].endswith(".json"):  # TODO Don't check filename, if file is 'indexed' load json
+                logger.warning("File %s is marked for indexing but is not of type JSON. It will not be indexed.",
```
mikebaumann: Thanks, that is more clear.
I have work in progress elsewhere (outside of this PR) to change to the approach of parsing the file as JSON rather than relying on its suffix, and unit test accordingly.

tests/test_indexer.py
mikebaumann: Wow, it's really nice how easy it is to do that in Python. Thanks!

tests/test_search.py
mikebaumann: Ditto.

PR #162, opened 2017-06-26
Use nc instead of wget to check if ES is up

kislyuk:
For whatever reason, wget flakes out often in the process of checking if ES is up, causing spurious test failures.


ttung:
Why is wget failing and why do you think nc will work better?

kislyuk:
I'm not sure. But I've used nc for this purpose before with success, while I've never seen wget used successfully for it.

kislyuk:
In Travis, the occasional wget behavior that crashes our tests seems to be to exit with an error prematurely, instead of retrying the requisite number of times.

PR #164, opened 2017-06-27
Invert noqa marker on import/path mangle lines where E402 is raised

kislyuk:
It turns out you can avoid E402 either by ignoring the import lines,
or the statement lines. In all our cases, the fixed-function import
hack is much less dangerous to ignore than the varied import lines
where we might make typos, etc.

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -11,12 +11,12 @@ import uuid
 
 import requests
 
-pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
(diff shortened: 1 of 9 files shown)


ttung:
Alternatively we can just remove the hack entirely. It just means we have to run python -m unittest tests.test_sync

kislyuk:
I would much prefer to keep it, I like to be able to execute the test script file directly

ttung:
Ok. I approve this diff. Can't actually do it via email though. :(

PR #165, opened 2017-06-27
get the extras from the function params rather than request.get_json()

ttung:


sampierson:
Is this using a feature of Connexion whereby it breaks out everything you specify in the swagger spec into arguments for your function?

ttung:
> Is this using a feature of Connexion whereby it breaks out everything you specify in the swagger spec into arguments for your function?

Yes.

PR #166, opened 2017-06-27
script to set up test fixtures in a bucket.

ttung:


tests/fixtures/populate.py
```diff
@@ -0,0 +1,180 @@
...
+
+        args.extend([
+            "cp",
+            os.path.join(self.local_root, local_path),
+            "gs://{}/{}".format(self.bucket, remote_path),
+        ])
+        subprocess.check_call(args)
+
+
+def upload(uploader: Uploader):
```
sampierson: Please split the uploader code (which is generically useful) and the script that uploads a specific set of files (which isn't so much).

ttung:
ready for review!

tests/fixtures/cloud_uploader.py
```diff
@@ -0,0 +1,102 @@
...
+        if tags is None:
+            tags = dict()
+        args = [
+            "aws", "s3", "cp",
+            os.path.join(self.local_root, local_path),
+            "s3://{}/{}".format(self.bucket, remote_path),
+            "--metadata",
+            json.dumps(metadata_keys)
+        ]
+        subprocess.check_call(args)
```
kislyuk: Why not use `boto3.resource("s3").upload_file()` and the corresponding boto3 method for tagging instead of spawning a subprocess? It calls the same code on the backend.

tests/fixtures/cloud_uploader.py
```diff
@@ -0,0 +1,102 @@
...
+        self.bucket = bucket
+
+    def reset(self) -> None:
+        args = [
+            "gsutil",
+            "-m",
+            "rm",
+            "-r",
+            "gs://{}/**".format(self.bucket),
+        ]
```
kislyuk: Same here. Why use subprocess calls instead of using the underlying libraries directly?

ttung:
The rationale was:

1. Once we decompose to these subprocess calls, we could theoretically emit a script to do the same work.
2. It's super lowest-common-denominator.

PR #173, opened 2017-06-28
Update the documentation for /files and /bundles endpoints.

ttung:


dss-api.yml
```diff
@@ -176,10 +188,22 @@ paths:
...
+        on the cloud provider must have metadata set reflecting the file checksums and the file content type.
+
+        The metadata fields required are:
+        - hca-dss-content-type: content-type of the file
+        - hca-dss-sha256: SHA-256 checksum of the file
+        - hca-dss-sha1: SHA-1 checksum of the file
+        - hca-dss-s3_etag: S3 ETAG checksum of the file.  See
+        https://stackoverflow.com/questions/12186993/what-is-the-algorithm-to-compute-the-amazon-s3-etag-for-a-file-larger-than-5gb
+        for how the general algorithm for how checksum is calculated.  For files smaller than 64MB, this is the MD5
+        checksum of the file.  For files larger than 64MB but smaller than 640GB, we use 64MB chunks.  For files larger
```
kislyuk: I think for clarity about what "640GB" means, you may want to say "640000MB" instead. I believe you want `1024*1024*64*10000` as the cutoff, not `1024*1024*1024*640`. If we select the former and the client selects the latter, the part sizes are within bounds, but the etags won't match.

dss-api.yml
```diff
@@ -176,10 +188,22 @@ paths:
...
+        Create a new version of a file with a given UUID.  The file content is passed in through a cloud URL.  The file
+        on the cloud provider must have metadata set reflecting the file checksums and the file content type.
+
+        The metadata fields required are:
+        - hca-dss-content-type: content-type of the file
+        - hca-dss-sha256: SHA-256 checksum of the file
+        - hca-dss-sha1: SHA-1 checksum of the file
+        - hca-dss-s3_etag: S3 ETAG checksum of the file.  See
+        https://stackoverflow.com/questions/12186993/what-is-the-algorithm-to-compute-the-amazon-s3-etag-for-a-file-larger-than-5gb
+        for how the general algorithm for how checksum is calculated.  For files smaller than 64MB, this is the MD5
```
kislyuk: s/for how/for/

dss-api.yml
```diff
@@ -269,6 +293,10 @@ paths:
 #            $ref: '#/definitions/Error'
   /bundles/{uuid}:
     get:
+      summary: Retrieve a bundle given an UUID and optionally a version.
```
kislyuk: s/an UUID/a UUID/ ?

dss-api.yml
```diff
@@ -303,7 +331,8 @@ paths:
     put:
       summary: Create a bundle
       description: |
-        Create a new bundle with a given UUID.  The list of files UUID+versions to be included must be provided.
+        Create a new version of a bundle with a given UUID.  The list of files UUID+versions to be included must be
```
kislyuk: s/files/file/ ?

dss-api.yml
```diff
@@ -69,6 +69,14 @@ paths:
               - files
   /files/{uuid}:
     get:
+      summary: Retrieve a file given an UUID and optionally a version.
```
kislyuk: s/an UUID/a UUID/ ?

dss-api.yml
```diff
@@ -128,6 +136,10 @@ paths:
           schema:
             $ref: '#/definitions/Error'
     head:
+      summary: Retrieve a file's metadata given an UUID and optionally a version.
```
kislyuk: s/an UUID/a UUID/ ?

dss-api.yml
```diff
@@ -176,10 +188,22 @@ paths:
...
-        Create a new file with a given UUID.  The file content is passed in through a cloud URL.  The file on the cloud
-        provider must have metadata set reflecting the file checksums and the file content type.
+        Create a new version of a file with a given UUID.  The file content is passed in through a cloud URL.  The file
+        on the cloud provider must have metadata set reflecting the file checksums and the file content type.
+
+        The metadata fields required are:
+        - hca-dss-content-type: content-type of the file
+        - hca-dss-sha256: SHA-256 checksum of the file
+        - hca-dss-sha1: SHA-1 checksum of the file
+        - hca-dss-s3_etag: S3 ETAG checksum of the file.  See
```
kislyuk: We may want to check that these docstrings render properly as Markdown, as they will be consumed in autodocs downstream. Does Swagger UI render Markdown formatting?

dss-api.yml
ttung: Yup, good point.

dss-api.yml
ttung: 👍

dss-api.yml
ttung: I think I tested it and it worked.

dss-api.yml
ttung: UUID starts with a vowel sound, no?

PR #174, opened 2017-06-29
move get_env to tests/infra.py

ttung:


kislyuk:
What's the difference between utils and infra? Can we document it in those files?

Also, I'm a fan of relative imports (`from .infra import foo`). They avoid possible ambiguity with namespaces and can help in refactoring. Is there a reason not to use them here?

kislyuk:
Sorry, never mind about my question about utils vs. infra... I see you've blown away `utils.py` altogether.

ttung:
I'll switch to relative imports in a different PR.

ttung:
Turns out you can't do relative imports when you run a module's file as a standalone script.

PR #183, opened 2017-06-30
Add LICENSE file per project policy

kislyuk:


LICENSE
```diff
@@ -0,0 +1,22 @@
+The MIT License
+
+Copyright (c) 2017 The Human Cell Atlas Consortium, https://humancellatlas.org
```
freeman-lab: let's make it `Copyright (c) 2017 Human Cell Atlas`

freeman-lab:
just suggested the line to use for this and other projects

LICENSE
ttung: Is there a requirement to slap copyright headers on all source files?

kislyuk:
I would like to avoid any such requirement if possible. @freeman-lab?

freeman-lab:
@kislyuk @ttung nope, doesn't need to be in all source files, just in the top level `LICENSE`

PR #188, opened 2017-07-05
Introduce environment.local to manage local settings

kislyuk:
Also, make local server entry point more configurable and run chalice (not connexion) from it.


environment
```diff
@@ -24,3 +24,7 @@ DSS_GS_TEST_BUCKET=czi-hca-test
 DSS_GS_TEST_SRC_DATA_BUCKET=hca-dss-test-src
 GOOGLE_APPLICATION_CREDENTIALS=gcp-credentials.json
 set +a
+
+if [[ -f environment.local ]]; then
```
ttung: should be ${SOURCE}/environment.local

environment.local
```diff
@@ -0,0 +1,11 @@
+# HCA DSS environment variables: local configuration file
```
ttung: I'd just name this `environment.local.template` or something like that.

dss-api
```diff
@@ -1,9 +1,31 @@
...
+import os
+import sys
 import logging
+import argparse
+
+from chalice.cli import CLIFactory, run_local_server
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument("--port", type=int, default=5000)
+parser.add_argument("--no-debug", dest="debug", action="store_false")
```
ttung: i actually don't know what the difference is between debug vs log level, and it's quite possible the next person who fires this up also does not know.

dss-api
```diff
@@ -1,9 +1,31 @@
...
+parser.add_argument("--project-dir", help=argparse.SUPPRESS,
+                    default=os.path.join(os.path.dirname(__file__), "chalice"))
+parser.add_argument("--log-level",
```
ttung: i have this vague notion that there's different logging contexts.  is that something we want to have manipulated at the command-line level?

dss-api
```diff
@@ -1,9 +1,31 @@
 #!/usr/bin/env python
```
ttung: &lt;pedantic&gt;this should be a separate PR.&lt;/pedantic&gt;

dss-api
kislyuk: Great point, added a help string that might clarify this a bit.

dss-api
kislyuk: By logging contexts, you mean logger names, right? The standard idiom is to get the logger to log with on a per-package or per-module basis using `logger = logging.getLogger(__name__)`. This setting affects the default log verbosity for all of them. It's probably not worth exposing ability to set log levels for each of them individually, since that gets kinda messy quickly.

environment
kislyuk: Done, thanks!

environment.local
kislyuk: Will defer to a future change.

dss-api
kislyuk: Sorry 😭

PR #189, opened 2017-07-05
Parameterize chalice & domovoi configs wrt stage name

kislyuk:
Tested to accommodate alphanumeric stage names.


chalice/Makefile
```diff
@@ -2,6 +2,6 @@ include ../common.mk
 
 deploy:
 	git clean -df chalicelib
-	cp -R ../dss ../dss-api.yml ../gcp-credentials.json chalicelib
+	cp -R ../dss ../dss-api.yml "$(GOOGLE_APPLICATION_CREDENTIALS)" chalicelib
```
mikebaumann: The user may have set GOOGLE_APPLICATION_CREDENTIALS a filename other than gcp-credentials.json, as is expected by the chalice config. Consider explicitly copying GOOGLE_APPLICATION_CREDENTIALS to the intended name.

chalice/.chalice/config.json
```diff
@@ -1,12 +1,13 @@
 {
   "version": "2.0",
   "app_name": "dss",
-  "environment_variables": {},
+  "environment_variables": {
```
ttung: do we really want everything to run with the same gcp credentials?

chalice/build_deploy_config.sh
```diff
@@ -16,20 +16,29 @@ get_api_id() {
 }
 
 if [[ $# != 1 ]]; then
-    echo "Usage: $(basename $0) apigateway-stage"
+    echo "Usage: $(basename $0) stage"
     exit 1
 fi
 
-stage=$1
+export stage=$1
```
ttung: env vars should be all caps.

chalice/.chalice/config.json
kislyuk: This value only locates the file on the lambda's filesystem. It doesn't influence which credentials we give to the lambda.

Separately from this, we can (in make deploy/build_deploy_config.sh) parameterize which credentials we give to which stage (dev/staging/prod/custom) and to which lambda (dss-apiserver/dss-sync/dss-index).

(Separately, we have the option of switching to Lambda environment variables to transmit the contents of `gcp-credentials.json` - it takes up the majority of the 4KB env var size allowance, but it does fit (2KB gzipped+b64encoded). But then we'd have to figure out how to pass it to google cloud bindings, which expect a JSON file.)

chalice/build_deploy_config.sh
kislyuk: Filed https://github.com/HumanCellAtlas/data-store/issues/200 to track this.

kislyuk:
Filed https://github.com/HumanCellAtlas/data-store/issues/201 to track consolidation for code reuse.

PR #190, opened 2017-07-06
Add make clean target

kislyuk:


mikebaumann:
Great! Thanks for adding that. I had looked for "make clean" earlier this evening to do exactly this.
Nice to have it there now!

PR #203, opened 2017-07-07
Separate fixtures buckets for S3 and GS.

ttung:
This was a code path that was still using the same fixtures bucket name for both clouds, which is incorrect.  It has basically been a coincidence (i.e., people use the same name for their S3 bucket and their GS fixtures bucket) that this hasn't broken yet. :)

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -23,7 +23,8 @@ class TestDSS(unittest.TestCase, DSSAsserts):
     def setUp(self):
         DSSAsserts.setup(self)
         self.app = dss.create_app().app.test_client()
-        self.test_src_data_bucket = get_env("DSS_S3_TEST_SRC_DATA_BUCKET")
+        self.s3_test_src_data_bucket = get_env("DSS_S3_TEST_SRC_DATA_BUCKET")
+        self.gs_test_src_data_bucket = get_env("DSS_GS_TEST_SRC_DATA_BUCKET")
 
     def test_bundle_get(self):
         self._test_bundle_get("aws")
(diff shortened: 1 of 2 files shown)


ttung:
Anyone?

daemons/invoke_lambda.sh
```diff
@@ -2,16 +2,29 @@
...
-lambda_payload="$(sed "s/testBucket/$DSS_S3_TEST_BUCKET/g" $lambda_input_file)"
-lambda_output="$(aws lambda invoke --function-name $lambda_name --invocation-type RequestResponse --payload "$lambda_payload" --log-type Tail /dev/stdout | sed 's/^null//')"
+BUNDLE_KEY="bundles/$(basename "${bundle_file}")"
+
+aws s3 cp "${bundle_file}" s3://${DSS_S3_TEST_BUCKET}/"${BUNDLE_KEY}"
+BUNDLE_FILE_ETAG=$(aws s3api head-object --bucket ${DSS_S3_TEST_BUCKET} --key "${BUNDLE_KEY}" | jq -r '.ETag | fromjson')
+BUNDLE_FILE_SIZE=$(cat "${bundle_file}" | wc -c)
+
+# the wonky if-else is required because us-east-1 is represented as a null location constraint.  weird, eh?
+DSS_S3_BUCKET_REGION=$(aws s3api get-bucket-location --bucket ${DSS_S3_TEST_BUCKET} | jq -r 'if (.LocationConstraint == null) then "us-east-1" else .LocationConstraint end')
```
kislyuk: What are the circumstances in which a bucket is expected not to be in the same region as our deployment? This seems like overparameterizing the setup. If this bucket is in a remote region, the app is misconfigured, no?

Consider replacing this with simply `aws configure get region`.

ttung: One of the DSS goals was to do intra-cloud replication.  If we do that, we could have buckets in other regions.

kislyuk: Sounds good.

PR #211, opened 2017-07-11
Simplify the ci-cd policy template.

ttung:
Allow all s3:Get* and s3:List* actions.  We're running up against the inline policy limit of 2048 characters, and this helps a lot.


kislyuk:
I didn't know you could do this! Nice.

PR #215, opened 2017-07-13
[easy] Cache the GS bucket data like we do for S3

ttung:


kislyuk:
I'm not sure if I understand how this should be used. It seems the global config object switches between buckets based on the deployment stage (prod/staging/dev), or (special case) to the fixtures buckets. How do I use this to copy data between a fixtures bucket and the DSS application bucket?

Consider setting the global config/implicit state based on the stage, but also allowing the bucket name to be passed explicitly to a BlobStore constructor to override the implicit default, and using that for fixtures buckets.

ttung:
Only the higher level APIs use the global config switcher.  blobstore/hcablobstore take a (bucket, key) tuple for its operations.  Things like /files and /bundles use the Config global to pass parameters to blobstore/hcablobstore.

In your example, copying data between the fixtures bucket and the DSS application bucket is probably done like this:
```
normal_bucket = Config.get_s3_bucket()
with override_bucket_config(BucketConfig.TEST_FIXTURE):
  fixture_bucket = Config.get_s3_bucket()
s3blobstore.copy(normal_bucket, key, fixture_bucket)
```

PR #216, opened 2017-07-13
Add a dev stage

ttung:
1. Dev points to a separate bucket.
2. tests continue running against org-humancellatlas-dss-test

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -7,9 +7,13 @@ import boto3
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'chalicelib')) # noqa
 sys.path.insert(0, pkg_root) # noqa
 
-from dss import create_app
+from dss import Config, create_app
 from dss.events.handlers.sync import sync_blob
 from dss.util import paginate
+
+
+Config.set_config_by_env()
+
 
 def get_chalice_app(flask_app):
     app = chalice.Chalice(app_name=flask_app.name)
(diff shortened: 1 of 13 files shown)


chalice/app.py
```diff
@@ -7,10 +7,14 @@
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'chalicelib')) # noqa
 sys.path.insert(0, pkg_root) # noqa
 
-from dss import create_app
+from dss import Config, create_app
 from dss.events.handlers.sync import sync_blob
 from dss.util import paginate
 
+
+Config.set_config_by_env()
```
kislyuk: Should this happen implicitly when importing dss?

dss/__init__.py
```diff
@@ -11,6 +11,8 @@
 from connexion.resolver import RestyResolver
 from flask_failsafe import failsafe
 
+from .config import BucketConfig, Config
```
kislyuk: Could we name BucketConfig something more differentiated? BucketStage maybe?

dss/config.py
```diff
@@ -12,15 +12,32 @@
 
 
 class BucketConfig(Enum):
-    PROD = 0
-    TEST = 1
-    TEST_FIXTURE = 2
+    ILLEGAL = auto()
```
kislyuk: 🚔

chalice/app.py
ttung: no, because tests would then start writing to the dev bucket.

dss/__init__.py
ttung: Sure.

PR #219, opened 2017-07-13
update deployment scripts to deploy to multiple tiers.

ttung:
1. propagate the correct tier data to the config files generated during deploy.  use the stage information to select the correct set of environment variables.
2. invoke_lambda.sh also formats the request accordingly for the tier.


kislyuk:
Why not export all variables and let the app select among them based on the stage name, instead of bothering with the prefixes? The lambdas all have a "stage" variable (provided by Amazon's API) they can introspect and use to key into whatever config we give them.

ttung:
It prevents coding bugs where people explicitly pull in DSS_S3_BUCKET_XYZ rather than a stage-specific env var.  I already caught instances where I failed to refactor code.

Issue #228, opened 2017-07-15
Cannot deploy to custom stage names


mikebaumann:
It is useful to setup multiple independent DSS deployments within the same AWS account. For example, some long-running deployments with large amounts of data, some short-lived setup/torn-down by various developers, and perhaps some in between. This can be done by specifying different/custom stage names for each deployment, and this was working yet a recent change limits the acceptable stage names to "dev" and "beta" or "prod".

dss/config.py
    @staticmethod
    def set_config_by_env() -> None:
        stage = os.environ['DSS_DEPLOYMENT_STAGE']
        if stage == "dev":
            Config.set_config(BucketStage.DEV)
        elif stage == "beta" or stage == "prod":
            Config.set_config(BucketStage.PROD)
        else:
            raise ValueError("Unknown stage!")

PR #229, opened 2017-07-15
Allow deployments to custom stages.

ttung:
1. Bucket configs are "normal" (DSS_S3_BUCKET), "test" (DSS_S3_BUCKET_TEST), and "test_fixture" (DSS_S3_BUCKET_TEST_FIXTURES)
2. Remove logic to infer which env variable based on stage.

This should address #228.


ttung:
> I think later on we should consider starting a centralized common config yaml file holding the configs for different stages, so we have a more structured way to keep this information than just environment vars.

I think that's a devops concern, since that won't be shared among the various deployments of the blue box.

PR #230, opened 2017-07-15
Describe setup steps more completely in the readme

mikebaumann:
Added more complete documentation of the steps to setup DSS.

A few notes to the reviewer:
Populate.py does not have execute permission (at least not on my system), so the instructions explicitly include 'python' in the command line. Is this the best way to handle this case?

Tony had mentioned not using environment variables in the sample command line for populate.py, to instead have the user type the bucket names explicitly given the buckets will be emptied by populate.py. I didn't find a good way in markdown to within the code block display portions of the command syntax that should be replaced by the user. Also, the risk seems lower than before given the environment variables are for explicit test fixture buckets. Please let me know if you would like something different here.

Would you rather use httpie or curl for the example command?

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,18 @@ environment variable `DSS_GS_BUCKET_TEST_FIXTURES` to the name of that bucket.
 #### Running the DSS API locally
 Run `./dss-api` in the top-level `data-store` directory.
 
+#### Check and install software required to test and deploy
+Check that software packages required to test and deploy are available, and install them if necessary.
+
+Run: `make --dry-run`
+
 #### Populate test data
 
 To run the tests, test fixture data must be setup using the following command.
 **This command will completely empty the given buckets** before populating them with test fixture data, please 
 ensure the correct bucket names are provided.
 
-    python tests/fixtures/populate.py --s3-bucket $DSS_S3_BUCKET_TEST_FIXTURES --gs-bucket $DSS_GS_BUCKET_TEST_FIXTURES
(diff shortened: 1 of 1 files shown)


ttung:
I fixed the permissions issue on populate.py in https://github.com/HumanCellAtlas/data-store/commit/b5373209a0b024469fa7bcee5cf42d373a066ead

I agree, now that the test fixtures are explicitly in their own bucket, the risk is smaller.

README.md
```diff
@@ -49,15 +60,107 @@ environment variable `DSS_GS_BUCKET_TEST_FIXTURES` to the name of that bucket.
...
+
+Run `make test` in the top-level `data-store` directory.
+
+#### Deployment
+
+Assuming the tests have passed above, the next step is to manually deploy.  See the section below for information on CI/CD with Travis if continuous deployment is your goal.
+
+You will need to ensure you are ready for deployment refer to `.travis.yml` to ensure you have all needed requirements.  Currently, the Travis deployment uses Ubuntu Trusty (14.04) with the following packages apt-get installed:
+
+    sudo apt-get install jq moreutils gettext
```
ttung: I think you need these just to run `make tests`

kislyuk: In addition to moving these up to before the test instructions, I recommend removing these instructions from this file and replacing them with an instruction to call `make --dry-run`. This way you don't duplicate the dependency checking functionality that is now built into the makefile. (Lists of dependencies in readmes inevitably get out of date)

PR #242, opened 2017-07-17
provide direct access URLs if directurls=true

ttung:
This is to address #191


ttung:
I don't have a strong opinion about including the direct URLs by default.  I do think this is not a permanent API, and the two cases would be
1. a bundle get operation.
2. a bundle checkout operation that may not immediately succeed (i.e., may tell the requester to try again) gives direct access URLs.

PR #245, opened 2017-07-18
Fixes for post commit review comments for indexer-related code and tests

mikebaumann:
This PR addresses post commit review comments for the following:
https://github.com/HumanCellAtlas/data-store/commit/cf58059

The changes made for each review comment are described there.
These changes include additional test fixture data created by populate.py, and the tests will fail until populate.py is run to add the additional data.


dss/events/handlers/index.py
```diff
@@ -50,21 +57,33 @@ def read_bundle_manifest(s3, bucket_name, bundle_key, logger):
...
+                               bundle_id,
+                               file_info[BundleFileMetadata.NAME],
+                               file_info[BundleFileMetadata.CONTENT_TYPE])
+                continue
+            try:
+                file_key = create_file_key(file_info)
+                file_string = bucket.Object(file_key).get()['Body'].read().decode("utf-8")
+                file_json = json.loads(file_string)
+            except Exception as e:
+                logger.warning(("In bundle %s the file \"%s\" is marked for indexing yet could not be parsed."
```
ttung: use the f-string formatting operation.

dss/events/handlers/index.py
```diff
@@ -50,21 +57,33 @@ def read_bundle_manifest(s3, bucket_name, bundle_key, logger):
...
     index_files = {}
     bucket = s3.Bucket(bucket_name)
     for file_info in files_info:
-        if file_info['indexed'] is True:
-            if not file_info["name"].endswith(".json"):  # TODO Don't check filename, if file is 'indexed' load json
-                logger.warning("File %s is marked for indexing but does not end in .json. It will not be indexed.",
-                               file_info["name"])
+        if file_info[BundleFileMetadata.INDEXED] is True:
+            if file_info[BundleFileMetadata.CONTENT_TYPE] != 'application/json':
+                logger.warning(("In bundle %s the file \"%s\" is marked for indexing yet has content type \"%s\""
```
ttung: use the f-string formatting operation.

dss/events/handlers/index.py
```diff
@@ -50,21 +57,33 @@ def read_bundle_manifest(s3, bucket_name, bundle_key, logger):
...
+                                " This file will not be indexed."),
+                               bundle_id,
+                               file_info[BundleFileMetadata.NAME],
+                               file_info[BundleFileMetadata.CONTENT_TYPE])
+                continue
+            try:
+                file_key = create_file_key(file_info)
+                file_string = bucket.Object(file_key).get()['Body'].read().decode("utf-8")
+                file_json = json.loads(file_string)
+            except Exception as e:
```
ttung: should catch S3 exceptions separately from parsing exceptions.

dss/events/handlers/index.py
```diff
@@ -76,24 +95,34 @@ def create_index_data(s3, bucket_name, manifest, logger):
...
     return index
 
 
+def get_bundle_id_from_key(bundle_key):
+    if bundle_key.startswith("bundles/"):
+        bundle_key = bundle_key[8:]
+    return bundle_key
+
+
 def create_file_key(file_info) -> str:
```
ttung: this abstraction probably belongs elsewhere, since it's used in multiple places.

dss/events/handlers/index.py
```diff
@@ -76,24 +95,34 @@ def create_index_data(s3, bucket_name, manifest, logger):
...
             # As due diligence, additional investigation should be performed.
-            index_filename = file_info["name"].replace(".", "_")
+            index_filename = file_info[BundleFileMetadata.NAME].replace(".", "_")
             index_files[index_filename] = file_json
     index['files'] = index_files
     return index
 
 
+def get_bundle_id_from_key(bundle_key):
+    if bundle_key.startswith("bundles/"):
```
ttung:
```
bundle_prefix = "bundles/"
if bundle.key.startswith(bundle_prefix):
  return bundle_key[len(bundle_prefix):]
raise Exception("This key isn't a bundle!")
```

dss/events/handlers/index.py
```diff
@@ -6,16 +6,22 @@
...
 DSS_BUNDLE_KEY_REGEX = r"^bundles/[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}\..+$"
 
-#
-# Lambda function for DSS indexing
-#
+"""
+Lambda function for DSS indexing
+"""
+
+es_client = None
```
ttung: I like doing this:

```
class ElasticsearchClient:
  _es_client = None

  @staticmethod
  def get():
    if ElasticsearchClient._es_client is None:
      ElasticsearchClient._es_client = connect_elasticsearch(os.getenv("DSS_ES_ENDPOINT"), logger)
    return ElasticsearchClient._es_client
```

mikebaumann:
Thanks for the great recommendations, Tony. I have made all the changes you suggested and pushed them earlier this afternoon. Thank you!

dss/events/handlers/index.py
```diff
@@ -4,33 +4,46 @@
...
-from ...util import connect_elasticsearch
+from ...hcablobstore import BundleMetadata, BundleFileMetadata
+from ...util import connect_elasticsearch, create_blob_key
 
 DSS_BUNDLE_KEY_REGEX = r"^bundles/[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}\..+$"
 
-#
-# Lambda function for DSS indexing
-#
+"""
```
ttung: I don't recall if python docblocks need to be after the class declaration or before.  Might want to check that.

dss/events/handlers/index.py
```diff
@@ -45,26 +58,43 @@ def is_bundle_to_index(key) -> bool:
...
     index_files = {}
     bucket = s3.Bucket(bucket_name)
     for file_info in files_info:
-        if file_info['indexed'] is True:
-            if not file_info["name"].endswith(".json"):  # TODO Don't check filename, if file is 'indexed' load json
-                logger.warning("File %s is marked for indexing but does not end in .json. It will not be indexed.",
-                               file_info["name"])
+        if file_info[BundleFileMetadata.INDEXED] is True:
+            if file_info[BundleFileMetadata.CONTENT_TYPE] != 'application/json':
+                logger.warning((f"In bundle {bundle_id} the file \"{file_info[BundleFileMetadata.NAME]}\""
```
ttung: Do you actually need the `((` or can one `(` work?

dss/events/handlers/index.py
```diff
@@ -45,26 +58,43 @@ def is_bundle_to_index(key) -> bool:
...
+                                " This file will not be indexed.")
+                               )
+                continue
+            try:
+                file_key = create_blob_key(file_info)
+                file_string = bucket.Object(file_key).get()['Body'].read().decode("utf-8")
+                file_json = json.loads(file_string)
+            # TODO (mbaumann) Are there other JSON-related exceptions that should be checked below?
+            except json.decoder.JSONDecodeError as e:
+                logger.warning((f"In bundle {bundle_id} the file \"{file_info[BundleFileMetadata.NAME]}\""
```
ttung: ditto.

dss/events/handlers/index.py
```diff
@@ -45,26 +58,43 @@ def is_bundle_to_index(key) -> bool:
...
+                                " is marked for indexing yet could not be parsed."
+                                f" This file will not be indexed. Exception: {e}"))
+                continue
+            except botocore.exceptions.ClientError as e:
+                logger.warning((f"In bundle {bundle_id} the file \"{file_info[BundleFileMetadata.NAME]}\""
```
ttung: ditto.

dss/events/handlers/index.py
```diff
@@ -76,44 +106,45 @@ def create_index_data(s3, bucket_name, manifest, logger):
             # of the query spec, so go with that, at least for now. Do so by substituting
             # dot for underscore in the key filename portion of the index.
             # As due diligence, additional investigation should be performed.
-            index_filename = file_info["name"].replace(".", "_")
+            index_filename = file_info[BundleFileMetadata.NAME].replace(".", "_")
```
ttung: Is there a longer-term plan for the aliasing issue this creates?

dss/events/handlers/index.py
```diff
@@ -45,26 +58,43 @@ def is_bundle_to_index(key) -> bool:
...
+                                " instead of the required content type \"application/json\"."
+                                " This file will not be indexed.")
+                               )
+                continue
+            try:
+                file_key = create_blob_key(file_info)
+                file_string = bucket.Object(file_key).get()['Body'].read().decode("utf-8")
+                file_json = json.loads(file_string)
+            # TODO (mbaumann) Are there other JSON-related exceptions that should be checked below?
+            except json.decoder.JSONDecodeError as e:
```
ttung: later on, you use `ex` as the exception name.

i personally prefer `ex` because if you want to search for the usage of a variable, `e` is pretty awful. :(

anyway, it should just be consistent.

tests/infra.py
```diff
@@ -79,3 +84,131 @@ def __getattr__(self, item: str) -> typing.Any:
...
+        self.path = object_summary.key
+        self.metadata = object_summary.Object().metadata
+        self.indexed = True if self.metadata['hca-dss-content-type'] == "application/json" else False
+        self.name = os.path.basename(self.path)
+        self.url = f"s3://{bundle.bucket.name}/{self.path}"
+        self.uuid = str(uuid.uuid4())
+        self.version = None
+
+
+class StorageTestSupport(DSSAsserts):
```
ttung: It's unclear to me why this needs to extend `DSSAsserts` or even be composed into a class with the tests.  Can you explain why?

mikebaumann: Yes, initially I also thought this would be a set of test functions, so I'll review how that evolved to the current code. The first issue was the test support code uses the unittest assert* methods, typically obtained by inheriting from unittest.TestCase. If these were just test functions outside of a class, I didn't know of a clean way for the test functions to call assert*. If you know of a clean way to do that please let me know. But that is the main reason this test support code is in a class. It is also a cohesive collection of test support operations that work well together, so grouping them in a class seemed reasonable.
Once they were in a class, I did not originally inherit from DSSAsserts, and just included the StorageTestSupport in the list of classes that were inherited by the tests that needed that support. This worked fine (tests passed), but mypy complained:
$ make test
flake8 dss tests chalice/*.py daemons/*/*.py
mypy --ignore-missing-imports dss tests
tests/infra.py:141: error: "StorageTestSupport" has no attribute "assertPutResponse"
tests/infra.py:151: error: "StorageTestSupport" has no attribute "assertIs"
tests/infra.py:152: error: "StorageTestSupport" has no attribute "assertIn"
tests/infra.py:156: error: "StorageTestSupport" has no attribute "assertPutResponse"
tests/infra.py:163: error: "StorageTestSupport" has no attribute "assertIs"
tests/infra.py:164: error: "StorageTestSupport" has no attribute "assertIn"
tests/infra.py:186: error: "StorageTestSupport" has no attribute "assertGetResponse"
tests/infra.py:196: error: "StorageTestSupport" has no attribute "assertEqual"
tests/infra.py:201: error: "StorageTestSupport" has no attribute "fail"
tests/infra.py:202: error: "StorageTestSupport" has no attribute "assertEqual"
tests/infra.py:203: error: "StorageTestSupport" has no attribute "assertEqual"
tests/infra.py:204: error: "StorageTestSupport" has no attribute "assertEqual"
tests/infra.py:208: error: "StorageTestSupport" has no attribute "assertGetResponse"
tests/infra.py:213: error: "StorageTestSupport" has no attribute "assertEqual"
tests/infra.py:214: error: "StorageTestSupport" has no attribute "assertEqual"
make: *** [mypy] Error 1
I addressed this by changing StorageTestSupport to extend DSSAsserts.
Is there a better way?

Makefile
```diff
@@ -8,6 +8,7 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test: lint mypy
+	tests/fixtures/populate.py --s3-bucket $(DSS_S3_BUCKET_TEST_FIXTURES) --gs-bucket $(DSS_GS_BUCKET_TEST_FIXTURES)
```
mikebaumann: The motivation for this change was I added some tests which required additional test fixture data, and the Travis branch build failed because the test fixture data was not available.
I discussed with with Andrey, and he said make needs to setup the necessary data for tests to pass, and populate.py is our way of doing that now, and we will need to make a faster way. It does take a while (about 30 seconds my system).

tests/infra.py
ttung: I'm not sure why this is behaving differently than DSSAsserts (or test_blobstore.py or test_hcablobstore.py), which also uses composition and ought to have the same issues.  I checked out your branch and could "resolve" it by making self of type `typing.Any`.  Not great. :(

dss/events/handlers/index.py
mikebaumann: "A docstring is a string literal that occurs as the first statement in a module, function, class, or method definition." according to PEP 257 -- Docstring Conventions. This comment was meant to apply to the module overall, so I will put it at the top.

dss/events/handlers/index.py
mikebaumann: The inner parens were needed before conversion to f-strings. They are cleaned-up now.

dss/events/handlers/index.py
mikebaumann: Good mention re. the searchability of "ex". It's consistent now.

dss/events/handlers/index.py
mikebaumann: By aliasing issue, are you referring to index document keys representing file names have an underscore in place of a dot? From what I've read, dot in ES doc keys can be problematic, and are best avoided. The metadata group is independently avoiding dot in key names. This does mean that there is not a simple, reliable reverse translation from the key names in the index document to the actual file names, which I agree is an issue. No, I don't have a plan for how to address that issue.

tests/infra.py
mikebaumann: Thanks very much for investigating and a proposed resolution. I think it is best to not have the new class extend DSSAsserts, at the expense of declaring `self` as `Any`.
This is how I did it, let me know if you have other recommendations:
    def create_bundle(self: Any, bundle: S3TestBundle):
        response = self.assertPutResponse(

dss/events/handlers/index.py
mikebaumann: Done

Makefile
kislyuk: Since we last spoke, I realized that running populate.py in tests is a bad idea since it results in tests racing. In fact I noticed that happen between two branches. So, let's try to figure out something different. Also, if your tests use shared resources like S3 buckets, please be aware of race conditions. In my tests I get around those by randomly generating object key UUIDs.

ttung: I agree that generally that we should be using randomly generated UUIDs.  However, I ran into the problem when I was building the initial producer API tests.

PUT is easy to verify.  Write something and then read it back.  But how does one verify GET without knowing PUT works?  The only way I could think of was to write an authoritative thing by hand and GET'ing it.  I don't think making tests for PUT relying on a known functioning GET and tests for GET relying on a known functioning PUT is particularly good.

Happy to entertain other suggestions if you have any. :)

dss/events/handlers/index.py
ttung: Yeah.  That's exactly what I mean.

One way to solve this is to make something an explicit escape character.  Then you can map things forwards and backwards authoritatively.

Makefile
kislyuk: Test methods are sorted alphanumerically. IMHO you could just test PUT, then test GET which happens to depend on PUT working. It's not ideally decomposed, but I don't see a major problem with that 😁

ttung: Let's go ahead and remove this from `make tests`.  We should endeavor to reduce our reliance on pre-populated fixtures and move towards fully encapsulated tests.

dss/events/handlers/index.py
```diff
@@ -1,36 +1,48 @@
...
+            index_data = create_index_data(s3, bucket_name, bundle_id, manifest, logger)
+            add_index_data_to_elasticsearch(bundle_id, index_data, logger)
+            logger.debug(f"Finished index processing of S3 creation event for bundle: {key}")
         else:
-            logger.debug("Not indexing S3 creation event for key: %s", key)
-    except Exception as e:
-        logger.error("Exception occurred while processing S3 event: %s Event: %s", e, json.dumps(event, indent=4))
+            logger.debug(f"Not indexing S3 creation event for key: {key}")
+    except Exception as ex:
+        logger.error(f"Exception occurred while processing S3 event: {ex} Event: {json.dumps(event, indent=4)}")
```
ttung: TIL you can put full expressions inside f-style formatting strings...

dss/events/handlers/index.py
mikebaumann: I opened the following issue for this:
https://github.com/HumanCellAtlas/data-store/issues/280

However, this issue was not introduced in this PR, nor does this PR do anything to make the matter worse, so the independent existence of this issue should not hold up this PR.

Thanks for raising the visibility of it, and offering a potential solution!

ttung: > However, this issue was not introduced in this PR, nor does this PR do anything to make the matter worse, so the independent existence of this issue should not hold up this PR.

yes, I agree. :)

dss/events/handlers/index.py
mikebaumann: Yes, you can, although it seems the code inside an f-string doesn't get the same checking (at least from PyCharm) that code outside the f-string does.

kislyuk: I think as a project policy we should avoid putting expressions in f-strings. Things can quickly get out of hand otherwise, at minimum with linting and when user-controlled data is concerned.

PR #250, opened 2017-07-19
Sync daemon: parallelize copying of large objects between S3 and GS.

### LINKED ISSUE #195: Sync daemon: copy large objects

kislyuk:
Also introduce a few AWS/boto3 utility functions.

Fixes #195.

### DIFF
--- a/daemons/dss-sync/app.py
+++ b/daemons/dss-sync/app.py
@@ -1,3 +1,28 @@
+"""
+Event handlers in the dss-sync daemon use utility functions in dss.events.handlers.sync.
+
+Events are first received from S3. If the event concerns an object less than 64MB in size (the size of one S3 part), it
+is copied immediately in the process. Otherwise, work to copy parts of the object is dispatched by triggering lambdas
+over the SNS topic ``dss-copy-parts``. After copying its assigned parts, each lambda queries the destination object
+store for the state of the object's parts. If the work is almost complete, the lambdas call another SNS topic to join
+the parts and finalize the upload (``dss-s3-mpu-ready`` for S3, ``dss-gs-composite-upload-ready`` for GS).
+
+All the event handler lambdas above are actually the same lambda distribution (dss-sync), called via different entry
+points by different S3 or SNS events, as seen in decorators below.
+
+An example SNS message sent to the worker lambdas: (TODO: (akislyuk) formalize SNS message schema/protocol)
+
+    {"source_platform": "s3",
+     "source_bucket": "hca-test",
+     "source_key": "hca-dss-sync-test/copy-part/991a4058-4671-43f4-86c7-77ef69157f56",
+     "dest_platform": "gs",
+     "dest_bucket": "hca-test",
+     "dest_key": "hca-dss-sync-test/copy-part/991a4058-4671-43f4-86c7-77ef69157f56",
+     "parts": [{"start": 0, "end": 1048575, "id": 1, "total_parts": 1}],
+     "total_parts": 1,
+     "mpu": None}
+"""
+
 import os
 import sys
 import logging
@@ -35,6 +60,10 @@ def process_new_syncable_object(event, context):
 
 @app.sns_topic_subscriber("dss-gs-composite-upload-ready")
 def compose_upload(event, context):
+    """
+    See https://cloud.google.com/storage/docs/composite-objects for details of the Google Storage API used here.
+    """
+    gs_max_compose_parts = 32
     msg = json.loads(event["Records"][0]["Sns"]["Message"])
     gs = dss.Config.get_cloud_specific_handles("gcp")[0].gcp_client
     gs_bucket = gs.get_bucket(msg["dest_bucket"])
@@ -42,11 +71,11 @@ def compose_upload(event, context):
         try:
             context.log("Composing, stage 1")
             compose_stage2_blob_names = []
-            if msg["total_parts"] > 32:
-                for part_id in range(1, msg["total_parts"] + 1, 32):
-                    parts_to_compose = range(part_id, min(part_id + 32, msg["total_parts"] + 1))
+            if msg["total_parts"] > gs_max_compose_parts:
+                for part_id in range(1, msg["total_parts"] + 1, gs_max_compose_parts):
+                    parts_to_compose = range(part_id, min(part_id + gs_max_compose_parts, msg["total_parts"] + 1))
                     source_blob_names = ["{}.part{}".format(msg["dest_key"], p) for p in parts_to_compose]
-                    dest_blob_name = "{}.part{}".format(msg["dest_key"], ascii_letters[part_id // 32])
+                    dest_blob_name = "{}.part{}".format(msg["dest_key"], ascii_letters[part_id // gs_max_compose_parts])
                     if gs_bucket.get_blob(dest_blob_name) is None:
                         compose_gs_blobs(gs_bucket, source_blob_names, dest_blob_name)
                     compose_stage2_blob_names.append(dest_blob_name)
@@ -61,7 +90,7 @@ def compose_upload(event, context):
         time.sleep(5)
 
 @app.sns_topic_subscriber("dss-s3-mpu-ready")
-def complete_mpu(event, context):
+def complete_multipart_upload(event, context):
     msg = json.loads(event["Records"][0]["Sns"]["Message"])
     mpu = resources.s3.Bucket(msg["dest_bucket"]).Object(msg["dest_key"]).MultipartUpload(msg["mpu"])
     while True:
--- a/daemons/dss-sync/app.py
+++ b/daemons/dss-sync/app.py
@@ -1,11 +1,21 @@
 """
 Event handlers in the dss-sync daemon use utility functions in dss.events.handlers.sync.
 
-Events are first received from S3. If the event concerns an object less than 64MB in size (the size of one S3 part), it
-is copied immediately in the process. Otherwise, work to copy parts of the object is dispatched by triggering lambdas
-over the SNS topic ``dss-copy-parts``. After copying its assigned parts, each lambda queries the destination object
-store for the state of the object's parts. If the work is almost complete, the lambdas call another SNS topic to join
-the parts and finalize the upload (``dss-s3-mpu-ready`` for S3, ``dss-gs-composite-upload-ready`` for GS).
+Events are first received from object storage:
+
+* S3 invokes the dss-sync Lambda via SNS (although the event handler below sees an S3 event notification since Domovoi
+  unpacks the SNS envelope). The event handler then calls ``dss.events.handlers.sync.sync_blob``.
+
+* GS invokes the ``/internal/notify`` route of the dss Lambda via Google Cloud Pub/Sub. The route handler then calls
+  ``dss.events.handlers.sync.sync_blob``.
+
+If the event concerns an object that already exists in the destination bucket, the process stops.
+
+If the event concerns an object less than 64MB in size (the size of one S3 part), it is copied immediately in the
+process. Otherwise, work to copy parts of the object is dispatched by triggering lambdas over the SNS topic
+``dss-copy-parts``. After copying its assigned parts, each lambda queries the destination object store for the state of
+the object's parts. If the work is almost complete, the lambdas call another SNS topic to join the parts and finalize
+the upload (``dss-s3-mpu-ready`` for S3, ``dss-gs-composite-upload-ready`` for GS).
 
 All the event handler lambdas above are actually the same lambda distribution (dss-sync), called via different entry
 points by different S3 or SNS events, as seen in decorators below.
@@ -21,6 +31,7 @@ An example SNS message sent to the worker lambdas: (TODO: (akislyuk) formalize S
      "parts": [{"start": 0, "end": 1048575, "id": 1, "total_parts": 1}],
      "total_parts": 1,
      "mpu": None}
+
 """
 
 import os
--- a/daemons/dss-sync/app.py
+++ b/daemons/dss-sync/app.py
@@ -56,13 +56,13 @@ def compose_upload(event, context):
                     source_blob_names = ["{}.part{}".format(msg["dest_key"], p) for p in parts_to_compose]
                     dest_blob_name = "{}.part{}".format(msg["dest_key"], ascii_letters[part_id // gs_max_compose_parts])
                     if gs_bucket.get_blob(dest_blob_name) is None:
-                        compose_gs_blobs(gs_bucket, source_blob_names, dest_blob_name)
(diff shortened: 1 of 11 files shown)


daemons/dss-sync/app.py
```diff
@@ -24,7 +29,93 @@ def process_new_syncable_object(event, context):
...
     else:
-        s3 = boto3.resource("s3")
-        bucket = s3.Bucket(event['Records'][0]["s3"]["bucket"]["name"])
+        bucket = resources.s3.Bucket(event['Records'][0]["s3"]["bucket"]["name"])
         obj = bucket.Object(unquote(event['Records'][0]["s3"]["object"]["key"]))
-        sync_blob(source_platform="s3", source_key=obj.key, dest_platform="gs", logger=app.log)
+        sync_blob(source_platform="s3", source_key=obj.key, dest_platform="gs", logger=app.log, context=context)
+
+@app.sns_topic_subscriber("dss-gs-composite-upload-ready")
+def compose_upload(event, context):
```
kislyuk: Added a sample event, as well as an exposition docstring at the top of this file.

Also added a todo for myself to protocolize the SNS messages somehow.

daemons/dss-sync/app.py
```diff
@@ -24,7 +29,93 @@ def process_new_syncable_object(event, context):
...
+    msg = json.loads(event["Records"][0]["Sns"]["Message"])
+    gs = dss.Config.get_cloud_specific_handles("gcp")[0].gcp_client
+    gs_bucket = gs.get_bucket(msg["dest_bucket"])
+    while True:
+        try:
+            context.log("Composing, stage 1")
+            compose_stage2_blob_names = []
+            if msg["total_parts"] > 32:
```
kislyuk: This is the maximum number of parts you can join at a time on GS. (There is also a 1024 part total limit, so it decomposes into two passes, 32*32).

Added a variable to hold this number, and also a docstring at the top of this function with a link to the docs.

daemons/dss-sync/app.py
```diff
@@ -24,7 +29,93 @@ def process_new_syncable_object(event, context):
...
+                for part_id in range(1, msg["total_parts"] + 1, 32):
+                    parts_to_compose = range(part_id, min(part_id + 32, msg["total_parts"] + 1))
+                    source_blob_names = ["{}.part{}".format(msg["dest_key"], p) for p in parts_to_compose]
```
kislyuk: I have switched to f-style strings wherever possible. But I don't like using f-style strings with complicated constructs and function calls inside. So I still prefer `.format()` in those cases, if that makes sense.

daemons/dss-sync/app.py
```diff
@@ -24,7 +29,93 @@ def process_new_syncable_object(event, context):
...
+                compose_stage2_blob_names = ["{}.part{}".format(msg["dest_key"], p) for p in parts_to_compose]
+            context.log("Composing, stage 2")
+            compose_gs_blobs(gs_bucket, compose_stage2_blob_names, msg["dest_key"])
+            break
+        except AssertionError:
+            pass
+        time.sleep(5)
+
+@app.sns_topic_subscriber("dss-s3-mpu-ready")
+def complete_mpu(event, context):
```
kislyuk: 👍  Done

dss/util/streaming.py
```diff
@@ -0,0 +1,111 @@
+"""
+Utilities in this module support the S3 chunked upload protocol as described in
```
kislyuk: I actually think chunked uploads are really pivotal in letting us maximize the performance of this.

- Part size starts to increase past 64MB after the object exceeds 300GB in size, up to 500MB for a 5TB object (the max size for S3). 500MB is beyond the limit of what we should reasonably buffer. (I do need to add a TODO to properly segment parts beyond 300GB, though)
- Each lambda copies multiple parts at the same time to maximize the full-duplex use of its network interface and to amortize against TCP startup costs and occasional slow S3 connections. Using streaming lets us crank up that parallelism.
- I believe the real memory usage in Python is at least twice the size of the buffer.
- Lambdas don't have an actual filesystem. The FS is a ramdisk, so storing files on it will count against the available memory. But even if we did have an actual storage volume, I don't think it would be a good idea to write parts to it.
- After writing the code for the S3 streaming upload, and stress testing with files up to 100GB, I haven't observed any problems with it so far. I could add some debug logging in case we run into any in the future, though.

Just say no to bufferbloat 😁

daemons/dss-sync/app.py
```diff
@@ -24,7 +33,97 @@ def process_new_syncable_object(event, context):
...
+    gs_max_compose_parts = 32
+    msg = json.loads(event["Records"][0]["Sns"]["Message"])
+    gs = dss.Config.get_cloud_specific_handles("gcp")[0].gcp_client
+    gs_bucket = gs.get_bucket(msg["dest_bucket"])
+    while True:
+        try:
+            context.log("Composing, stage 1")
+            compose_stage2_blob_names = []
+            if msg["total_parts"] > gs_max_compose_parts:
+                for part_id in range(1, msg["total_parts"] + 1, gs_max_compose_parts):
```
ttung: is there any advantage to the 1-based numbering?

All the extra `1` and `+ 1` makes this harder to read.

dss/events/handlers/sync.py
```diff
@@ -1,20 +1,35 @@
...
 
 from dss import Config
+from dss.util.aws import resources, clients, send_sns_msg, ARN
+from dss.util.streaming import get_pool_manager, S3SigningChunker
+from dss.blobstore.gs import GSBlobStore
 
 presigned_url_lifetime_seconds = 3600
 use_gsts = False
 gsts_sched_delay_minutes = 2
+part_size = {"s3": 64 * 1024 * 1024, "gs": 64 * 1024 * 1024 * 10}
```
ttung: The rationale for the part_size for s3 is obvious.
What is the rationale for the part_size of gs?
How does this guarantee that we stay under the component count limit?

I would prefer some scheme where we figure out what the part size is to accommodate the component, and then apply some rational floor to the part size.  I think it's reasonable to do this in a followup.

Also, I find 640 * 1024 * 1024 easier to read.

daemons/dss-sync/app.py
```diff
@@ -24,7 +33,97 @@ def process_new_syncable_object(event, context):
...
+                    bucket=msg["dest_bucket"],
+                    key=msg["dest_key"],
+                    part_num=part["id"],
+                    mpu_id=msg["mpu"]
+                )
+            elif msg["dest_platform"] == "gs":
+                assert len(msg["parts"]) == 1
+                dest_blob_name = "{}.part{}".format(msg["dest_key"], part["id"])
+                dest_blob = gs.get_bucket(msg["dest_bucket"]).blob(dest_blob_name)
+                upload_url = dest_blob.create_resumable_upload_session(size=part["end"] - part["start"] + 1)
```
ttung: Are there any implications to using reusable upload sessions?

dss/events/handlers/sync.py
```diff
@@ -1,20 +1,35 @@
...
 from google.cloud.client import ClientWithProject
 from google.cloud._http import JSONConnection
 
 from dss import Config
+from dss.util.aws import resources, clients, send_sns_msg, ARN
+from dss.util.streaming import get_pool_manager, S3SigningChunker
+from dss.blobstore.gs import GSBlobStore
 
 presigned_url_lifetime_seconds = 3600
 use_gsts = False
```
ttung: is there a point to keeping the gsts code around?

dss/events/handlers/sync.py
```diff
@@ -27,80 +42,182 @@ class GStorageTransferConnection(JSONConnection):
...
+    mpu = dest.blob.initiate_multipart_upload() if dest.platform == "s3" else None
+
+    with ThreadPoolExecutor(max_workers=4) as executor:
+        for part_id, part_start in all_parts:
+            parts_for_worker.append(dict(id=part_id + 1,
+                                         start=part_start,
+                                         end=min(total_size - 1, part_start + part_size[dest.platform] - 1),
+                                         total_parts=len(all_parts)))
+            if len(parts_for_worker) >= parts_per_worker[dest.platform] or part_id == all_parts[-1][0]:
+                print("Invoking dss-copy-parts with parts", ", ".join(str(p["id"]) for p in parts_for_worker))
```
ttung: why print?

dss/events/handlers/sync.py
```diff
@@ -27,80 +42,182 @@ class GStorageTransferConnection(JSONConnection):
...
+        source.blob.reload()
+        if source.blob.size < part_size["s3"]:
+            sync_gs_to_s3_oneshot(source, dest, logger)
+        else:
+            dispatch_multipart_sync(source, dest, logger, context)
+    logger.info(f"Completed transfer of {source_key} from {source.bucket} to {dest.bucket}")
+
+def compose_gs_blobs(gs_bucket, blob_names, dest_blob_name):
+    blobs = [gs_bucket.get_blob(b) for b in blob_names]
+    print("{} of {} blobs found".format(len([b for b in blobs if b is not None]), len(blob_names)))
```
ttung: why print?

dss/events/handlers/sync.py
```diff
@@ -27,80 +42,182 @@ class GStorageTransferConnection(JSONConnection):
...
+    assert not any(b is None for b in blobs)
+    dest_blob = gs_bucket.blob(dest_blob_name)
+    dest_blob.content_type = blobs[0].content_type
+    print("Composing blobs", blob_names, "into", dest_blob_name)
```
ttung: why print?

dss/util/aws/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        self.__dict__.update(dict(zip(self.fields, arn.split(":", 5)), **kwargs))
+        if "region" not in kwargs and not self.region:
+            self.region = self.get_region()
+        if "account_id" not in kwargs and not self.account_id:
+            self.account_id = self.get_account_id()
+
+    @classmethod
+    def get_region(cls):
+        if cls._default_region is None:
+            cls._default_region = botocore.session.Session().get_config_variable("region")
```
ttung: is this necessarily set?  I don't think it was when I was mucking around with the deploy scripts.

tests/test_sync.py
```diff
@@ -1,28 +1,29 @@
...
 import os
 import sys
+import json
+import hashlib
 import unittest
 import uuid
+from argparse import Namespace
 
 import boto3
+from botocore.vendored import requests
```
ttung: move line down.

daemons/dss-sync/app.py
kislyuk: S3 multipart upload part IDs are integers between [1, 10000]. Although GS part IDs are freeform string keys, the enumeration here is a byproduct of the fact that part IDs are generated in `dispatch_multipart_sync()` without regard to which platform is the destination.

daemons/dss-sync/app.py
kislyuk: Resumable uploads are nice because they reduce the impact of any given request failure.

Resumable uploads don't show up as an object until the whole upload is complete. The only implication I can think of is being billed for abandoned resumable uploads. I couldn't find any details on the retention period or billing policy on abandoned resumable uploads - maybe we should reach out to our GCP insider contacts.

dss/events/handlers/sync.py
kislyuk: It depends on an economics evaluation of the system. It's possible, but unlikely, that we'll end up in a regime where GSTS is substantially cheaper than pushing data around with Lambdas, and doesn't impact QoS much, but to do that we'd probably need payload batching of some sort.

I've filed #274 to remove this code in a separate PR and save it in a branch for future reference.

dss/events/handlers/sync.py
kislyuk: The rationale for the gs part size here was to basically apportion an amount of work that one Lambda can reasonably be expected to complete.

As it stands, there is nothing that guarantees that we stay under the GS 1024 component count limit. There is an implicit limit of 640 GB (ought to be enough for everyone 😁), beyond which GS uploads will fail to close.

I agree with you regarding the strategy to determine part size. It seems clear that a single lambda won't be able to copy (5TB/1024) = 5GB of data before running out of the allotted 5 minutes. If AWS doesn't increase the limit soon, and there is no hidden inefficiency that I overlooked, then I'll want to do Lambda chaining (a Lambda senses that it's about to run out of time and relaunches a copy of itself - a pattern I've successfully used before) on each part's resumable upload.

Filed #275 to track part size optimization and tracking against the part count limits.

Fixed this to read `640 * 1024 * 1024`.

dss/events/handlers/sync.py
kislyuk: Fixed to use app logger.

dss/events/handlers/sync.py
kislyuk: Fixed to use app logger.

dss/events/handlers/sync.py
kislyuk: Fixed to use app logger.

dss/util/aws/__init__.py
kislyuk: I'm pretty sure this is always set in a functioning environment, because if it's not, botocore will throw an error on any AWS operation. Can you recall the details of when it was not set?

tests/test_sync.py
kislyuk: I moved this line to the end of this block, but for my education, can you give the rationale? I'd like to group boto-related imports together...

kislyuk:
@ttung, ready for another review.

daemons/dss-sync/app.py
ttung: Doesn't that mean we should make a function to map between part sequence number <-> part IDs?

(To be fair, that may be premature optimization.)

daemons/dss-sync/app.py
ttung: Does the client (our code, in this case) even know how to deal with resuming an upload?

dss/events/handlers/sync.py
ttung: Isn't that the point of source control? :)

dss/util/aws/__init__.py
ttung: I think I ran `aws configure get region` and it was not set.

tests/test_sync.py
ttung: I thought it was:

```
import stdlib
from stdlib import xyz

import 3rdpartylib
import 3rdpartylib2
from 3rdpartylib import xyz
from 3rdpartylib3 import xyz

import localmodule
from localmodule import xyz
```

at least that's what I've observed.  I could be totally wrong ofc.

kislyuk: ok sounds good. will try and follow that from now on.

PR #251, opened 2017-07-20
Templatize swagger YAML; parameterize API host

kislyuk:


environment
```diff
@@ -25,6 +25,7 @@ DSS_GS_BUCKET=org-humancellatlas-dss-dev
 DSS_GS_BUCKET_TEST=org-humancellatlas-dss-test
 DSS_GS_BUCKET_TEST_FIXTURES=org-humancellatlas-dss-test-fixtures
 GOOGLE_APPLICATION_CREDENTIALS="${DSS_HOME}/gcp-credentials.json"
+API_HOST=hca-dss.czi.technology
```
ttung: doesn't `EXPORT_ENV_VARS_TO_LAMBDA` need to be updated?

kislyuk: You are right, updated. Thanks for the catch.

kislyuk:
README has wrapping changes, and also replacement of the blurb that said the swagger host was hardcoded, with a blurb instructing how to set `API_HOST`.

PR #252, opened 2017-07-20
Enable use of populate.py with moto S3 mock

mikebaumann:
Changed cloud_uploader.py support for S3 and GS to use the APIs directly
rather than running CLI commands. This enables populate.py to be
used in unit tests with the moto S3 mock. It also runs much faster
than it did before.

Populate.py was modified slightly to enable convenient programmatic
use, in addition to CLI use.

This was tested both in the context of moto unit tests as well as
with actual S3 through the CLI. The amount of logging output is 
similar to before, which is useful for verifying operation when
run from the CLI. It currently does, however, increase the amount
of logging from a unit test run, now that it is being used in the 
unit tests as well. We may want to adjust that.

### DIFF
--- a/tests/fixtures/cloud_uploader.py
+++ b/tests/fixtures/cloud_uploader.py
@@ -3,7 +3,14 @@ import os
 import subprocess
 import typing
 
+import boto3
+import logging
+
 from checksumming_io.checksumming_io import ChecksummingSink
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
 class Uploader:
     def __init__(self, local_root: str) -> None:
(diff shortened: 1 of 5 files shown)


tests/fixtures/cloud_uploader.py
```diff
@@ -3,8 +3,15 @@
 import subprocess
 import typing
 
+import boto3
+import logging
```
ttung: logging is a stdlib

tests/fixtures/populate.py
```diff
@@ -114,23 +114,26 @@ def upload(uploader: Uploader):
             },
         )
 
-
-if __name__ == '__main__':
+def populate(s3_bucket: str, gs_bucket: str):
```
ttung: this typing is incorrect.  it should be `typing.Optional[str]` since it can be `None`.

tests/test_indexer.py
```diff
@@ -142,6 +146,10 @@ def normalize_inherently_different_values_in_list(self, expected_list, actual_li
             if isinstance(expected_list[i], dict):
                 self.normalize_inherently_different_values_in_dict(expected_list[i], actual_list[i])
 
+def populate_moto_test_fixture_data():
+    s3_bucket_test_fixtures = (infra.get_env("DSS_S3_BUCKET_TEST_FIXTURES"))
```
ttung: remove the extra ()

ttung: Still need to do this.

ttung:
For future reference, the rev title should reflect the final diff

Issue #248, opened 2017-07-19
CRUD event Registration api


Mackey22:
Update Swagger spec to include the following operations:
* `POST` operation that will register an event
* `PATCH` operation that will update the functionality of an event
* `GET` operation that will give all relevant information pertaining to an event
* `DELETE` operation that will fully delete an event

Store all registered operations in a data store.

Mackey22:
Removing `PATCH` operation from list of required endpoints. New list is as follows:

* `POST` operation that will register an event
* `GET` operation that will give all relevant information pertaining to an event
* `DELETE` operation that will fully delete an event

PR #253, opened 2017-07-20
Update dss-api.yml to include subscriptions endpoints

Mackey22:
See #248 

Only parameter there might be conversation about is replica. Replica isn't required as when getting bundles/files, but it is here because each subscription only lives on one replica, per https://docs.google.com/document/d/193u4w_gQzWNy1OK442KRApKF7hncckowvkNWGGCVRO4/edit#heading=h.ykmrpkea3e15

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -269,6 +269,76 @@ paths:
                 format: date-time
             required:
               - version
+  /subscriptions:
+    get:
+      summary: Retrieve an user's event subscriptions.
+      description: |
+        Return a list of associated subscriptions, which contains the uuid,
+        replica, query, creator_uid, and callback_url.
+      operationId: dss.api.subscriptions.find
+      parameters:
+        - name: replica
+          in: query
+          description: Replica to fetch from. Can be one of aws, gcp, or azure.
+          required: true
+          type: string
+          enum: [aws, gcp, azure]
+      responses:
+        200:
+          description: OK
+          schema:
+            type: object
+            properties:
+              subscriptions:
+                type: array
+                items:
+                  $ref: "#/definitions/Subscription"
+            required:
+              - subscriptions
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    put:
+      summary: Create a event subscription.
+      description: |
+        Create a new event subscription. Every time a new bundle version matches this query,
+        a request is sent to callback_url.
+      parameters:
+        - name: replica
+          in: query
+          description: Replica to write to. Can be one of aws, gcp, or azure.
+          required: true
+          type: string
+          enum: [aws, gcp, azure]
+        - name: extras
+          in: body
+          required: true
+          schema:
+            type: object
+            properties:
+              query:
+                description: Elasticsearch query that will trigger the callback.
+                type: object
+              callback_url:
+                description: Url to send request to when a bundle comes in that matches this query.
+                type: string
+                format: url # Seems that this is possible now because of https://github.com/swagger-api/swagger-core/issues/1045
+            required:
+              - query
+              - callback_url
+      responses:
+        201:
+          description: OK
+          schema:
+            type: object
+            properties:
+              uuid:
+                description: A RFC4122-compliant ID for the subscription.
+                type: string
+                pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
+            required:
+              - uuid
   /subscriptions/{uuid}:
     get:
       summary: Retrieve an event subscription given a UUID.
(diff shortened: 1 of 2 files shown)


dss-api.yml
```diff
@@ -252,7 +252,121 @@ paths:
                 format: date-time
             required:
               - version
-
+  /subscriptions/{uuid}:
+    get:
+      summary: Retrieve an event subscription given a UUID.
+      description: |
+        Given a subscription UUID, return the latest version of that bundle.  If the version is provided, that version of the
```
ttung: copy pasta.

dss-api.yml
```diff
@@ -252,7 +252,121 @@ paths:
...
+              subscription:
+                $ref: "#/definitions/Subscription"
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    put:
+      summary: Create a event subscription.
+      description: |
+        Create a new event subscription. Every time an added bundle matches this query,
```
ttung: bundle+version

dss-api.yml
```diff
@@ -252,7 +252,121 @@ paths:
...
+      description: |
+        Given a subscription UUID, return the latest version of that bundle.  If the version is provided, that version of the
+        bundle is returned instead.
+      parameters:
+        - name: uuid
+          in: path
+          description: Subscription unique ID.
+          required: true
+          type: string
+          pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
```
ttung: A RFC4122-compliant ID for the subscription.

dss-api.yml
```diff
@@ -252,7 +252,121 @@ paths:
...
+        - name: extras
+          in: body
+          required: true
+          schema:
+            type: object
+            properties:
+              query:
+                description: Elasticsearch query that will trigger the callback.
+                type: object
+              creator_uid:
```
ttung: shouldn't need this since this should come in via the auth mechanism.

The producer API is different because the purple box is talking to the blue box on behalf of an user.

dss-api.yml
```diff
@@ -252,7 +252,121 @@ paths:
...
+                description: User ID who is creating this subscription.
+                type: integer
+                format: int64
+              callback_url:
+                description: Url to send request to when a bundle comes in that matches this query.
+                type: string
```
ttung: not sure if there's a format operation for URIs.

dss-api.yml
```diff
@@ -252,7 +252,121 @@ paths:
...
+            required:
+              - query
+              - creator_uid
```
ttung: remove

dss-api.yml
```diff
@@ -252,7 +252,121 @@ paths:
...
+              - callback_url
+      responses:
+        201:
+          description: OK
+          schema:
+            type: object
+            properties:
+              timeAdded:
```
ttung: is this needed or desired?

dss-api.yml
ttung: what if someone wants to modify an event subscription?  or is that not supported?

Mackey22: #248 - Modification removed as a result of 7/20 design meeting.

Mackey22: Changing to: "Every time a new bundle version matches this query."

dss-api.yml
Mackey22: Removing

dss-api.yml
Mackey22: There is, but all it does is check if the url starts with "http:" or "https:" https://github.com/swagger-api/swagger-core/blob/dd54691854df903f8e50b0a3768f67625fad7c13/modules/swagger-models/src/main/java/io/swagger/models/refs/GenericRef.java

Should all of the urls being submitted conform to this?

dss-api.yml
Mackey22: 👍

dss-api.yml
Mackey22: Not necessarily, but I thought it would be nice to return some JSON response because the cli automatically parses and prints the json response. Can change that behavior, but I think it's nice to get some output. 

wrt the contents of the response, I didn't think there was much relevant information to return so the most obvious thing was a similar response to what is returned after PUT bundles/files.

dss-api.yml
Mackey22: 👍

dss-api.yml
Mackey22: 👍

dss-api.yml
Mackey22: Didn't change this for the latest push but can if need be

Mackey22:
Tests being merged in with later prs. In the future, I'll add test stubs for each function I call when stacking prs like this.

PR #254, opened 2017-07-21
[EASY] Add standardized index and type names for queries es index.

Mackey22:
@mikebaumann common place for us to be grabbing our queries index names and doc types. 

There will be two indices, `DSS_ELASTICSEARCH_INDEX_NAME` and `DSS_ELASTICSEARCH_QUERIES_INDEX_NAME`. 

`DSS_ELASTICSEARCH_QUERIES_INDEX_NAME` will only have one type, `DSS_ELASTICSEARCH_QUERIES_DOC_TYPE`, that contains callback info, creator_uid, and query.

`DSS_ELASTICSEARCH_INDEX_NAME` will have two types:
* `DSS_ELASTICSEARCH_DOC_TYPE` will hold all the metadata that the blue box inputs
* `DSS_ELASTICSEARCH_QUERIES_DOC_TYPE` will hold all of the percolate queries. The ids of these percolate queries will match up with the ids in `DSS_ELASTICSEARCH_QUERIES_INDEX_NAME/DSS_ELASTICSEARCH_QUERIES_DOC_TYPE`

Wasn't sure about how efficient keeping the percolate queries in the same index as the documents was, but it seems to be common practice in all of Elasticsearches documentation (see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-percolate-query.html and https://www.elastic.co/blog/percolator-redesign-blog-post).

### DIFF
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -18,14 +18,14 @@ from .config import BucketStage, Config
 # ES index containing all docs
 DSS_ELASTICSEARCH_INDEX_NAME = "hca"
 # ES type within DSS_ELASTICSEARCH_INDEX_NAME with docs
-DSS_ELASTICSEARCH_DOC_TYPE = "docs"
+DSS_ELASTICSEARCH_DOC_TYPE = "doc"
 # ES type within DSS_ELASTICSEARCH_INDEX_NAME with percolate queries
-DSS_ELASTICSEARCH_QUERIES_TYPE = "queries"
+DSS_ELASTICSEARCH_QUERY_TYPE = "query"
 
 # ES index with all registered percolate queries
 DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME = "subscriptions"
 # ES type in DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME with subscriptions
-DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE = "subscription_type"
+DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE = "subscription"
 
 def get_logger():
     try:
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -16,6 +16,8 @@ from .config import BucketStage, Config
 # Constants common to the indexer and query route.
 DSS_ELASTICSEARCH_INDEX_NAME = "hca"
 DSS_ELASTICSEARCH_DOC_TYPE = "hca"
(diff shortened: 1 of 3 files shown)


dss/__init__.py
```diff
@@ -16,6 +16,8 @@
 # Constants common to the indexer and query route.
 DSS_ELASTICSEARCH_INDEX_NAME = "hca"
 DSS_ELASTICSEARCH_DOC_TYPE = "hca"
+DSS_ELASTICSEARCH_QUERIES_INDEX_NAME = "queries"
```
ttung: docblock these!

Mackey22: 👍

mikebaumann: For the index that will contain the subscriptions, I prefer the name "subscriptions" rather than "queries" (both in the constant name and value).

dss/__init__.py
```diff
@@ -16,6 +16,8 @@
...
+DSS_ELASTICSEARCH_QUERIES_DOC_TYPE = "queries_type"
```
mikebaumann: For the queries doc type value, I prefer simply "queries"

dss/__init__.py
Mackey22: 👍

dss/__init__.py
Mackey22: I think it may be better to have separate values because we could accidentally use the wrong variable somewhere and not know until the value changes, which would break things.

dss/__init__.py
```diff
@@ -13,9 +13,19 @@
...
-DSS_ELASTICSEARCH_DOC_TYPE = "hca"
+# ES type within DSS_ELASTICSEARCH_INDEX_NAME with docs
+DSS_ELASTICSEARCH_DOC_TYPE = "docs"
+# ES type within DSS_ELASTICSEARCH_INDEX_NAME with percolate queries
+DSS_ELASTICSEARCH_QUERIES_TYPE = "queries"
+
+# ES index with all registered percolate queries
+DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME = "subscriptions"
+# ES type in DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME with subscriptions
+DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE = "subscription_type"
```
mikebaumann: I prefer 'subscription' to 'subscription_type'. Alternately, we could omit the subscription doc type and just place them in the top-level index. Yet, use of a doc type is both consistent and provides more flexibility in the future.

PR #255, opened 2017-07-21
Subscriptions - setup es_client

Mackey22:
Setup elasticsearch connection. If the needed indices don't exist, create them.


Mackey22:
Closing because a lot has changed in jmackey-subscriptions-yml. 

Deployent scripts don't drop all indexes - see #266

PR #257, opened 2017-07-21
Move es util functions into new es folder

Mackey22:
There were some utility functions in `tests/test_indexer.py` that I should be able to use in my tests.

### DIFF
--- a/tests/es/__init__.py
+++ b/tests/es/__init__.py
@@ -4,8 +4,8 @@ from elasticsearch import Elasticsearch
 
 
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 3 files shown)


tests/es/__init__.py
```diff
@@ -0,0 +1,37 @@
+import logging
+
+from elasticsearch import Elasticsearch
+
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
```
mikebaumann: 'logger' is the preferred name to 'log'

Issue #221, opened 2017-07-13
Return JSON on API route failure


Mackey22:
As of now, most failures return text. Returning JSON will keep failures consistent with successful requests and can encode information that's useful for the user.

Mackey22:
Will comment with places I find text response: 

GET files/:uuid -> Cannot find file!

PR #259, opened 2017-07-23
Capture exceptions and format them to the standard fields for errors.

ttung:
There's now a DSSException type that has all the fields we want.
Additionally, there's a fallback exception handler which captures the uncaught exception and formats it in a safe way (although it's not verified).

This should address #221

### DIFF
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -3,16 +3,19 @@
 """
 DSS description FIXME: elaborate
 """
+import traceback
 
 import os
 import logging
 
-import flask
 import connexion
+import flask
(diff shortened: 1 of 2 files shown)


kislyuk:
It would be nice to update some of the tests to assert on the new error format.

Do you know why the tests are failing?

ttung:
> It would be nice to update some of the tests to assert on the new error format.

Whoops, forgot to PR that change out.

> Do you know why the tests are failing?

Yes, it's fixed on `master`. Re-run `pip install -r requirements.txt`

PR #261, opened 2017-07-24
reorder HEAD and GET

ttung:
1. werkzeug does this thing where it automagically adds HEAD to any route that handles GET, and it annoyingly cannot be turned off.  This causes all HEAD requests actually to be handled by the GET spec.  Previously, I believed that the spec was correctly determined, just the wrong code path was being chosen.  It turns out it's validating the wrong spec, but I didn't know because connexion wasn't actually validating default specs at all. 
 https://github.com/zalando/connexion/commit/6668835420231c15095a71e264b2e2b0e1a68bf3 (ironically authored by me) exposed the issue because now the return values are actually properly validated.
2. move the HEAD first so it's added to the rules first.  This is probably not the right fix (the right fix is for werkzeug to evaluate *all* the rules on a HEAD request, and pick the most specific rule), but it seems to work for now.
3. undoes the (IMO) worse hotfix in https://github.com/HumanCellAtlas/data-store/commit/1ac2ab6a367ed77bb170f7062191b0d43e3b6d02

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -70,73 +70,6 @@ paths:
             required:
               - files
   /files/{uuid}:
-    get:
-      summary: Retrieve a file given a UUID and optionally a version.
-      description: |
-        Given a file UUID, return the latest version of that file.  If the version is provided, that version of the file
-        is returned instead.
-
-        Headers will contain the data store metadata for the file.
-
-        This endpoint will do a HTTP redirect to another HTTP endpoint with the file contents.
-      parameters:
-        - name: uuid
-          in: path
-          description: A RFC4122-compliant ID for the file.
-          required: true
-          type: string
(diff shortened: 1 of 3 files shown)


Mackey22:
I don't know what you mean by validating the wrong spec. Also what would the default spec be?

ttung:
HEAD requests were actually using the GET request's spec.  I didn't realize that because the validation for the `default` response type didn't fail.  Once I fixed that bug upstream, then the validation started failing.

ttung:
> This LGTM but please add a test case that opens up the Swagger YAML and scans it for occurrences of routes where GET precedes HEAD, and fails with an informative message. That way we won't accidentally run into this problem in the future.

Relying on ordering seems fairly brittle, and it's definitely not guaranteed.  We should probably fix werkzeug to test all routes for `HEAD` requests and pick the more specific one.

PR #262, opened 2017-07-24
switch the /files not found message to the standard message format.

ttung:
Added a test for the error reporting.  This is an example to help facilitate discussion of #258 and #259.


kislyuk:
Consider integrating these assertions into [DSSAsserts.assertResponse](https://github.com/HumanCellAtlas/data-store/blob/master/tests/infra.py#L29-L61).

PR #265, opened 2017-07-24
Initial version of authentication middleware

kislyuk:
Tests to come after the notifications API (first API using/exposing this code) lands.

The AUTHORIZED_DOMAINS method is intended as a short-term stand-in until we plug in to an authorization data store of some sort.

Please squash merge this, instead of rebase merging.

cc @Mackey22

### DIFF
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -15,7 +15,7 @@ import requests
 import connexion.apis.abstract
 from connexion.operation import Operation
 from connexion.resolver import RestyResolver
-from connexion.exceptions import OAuthProblem
+from connexion.exceptions import OAuthProblem, OAuthResponseProblem, OAuthScopeProblem
 from flask_failsafe import failsafe
 
 from .config import BucketStage, Config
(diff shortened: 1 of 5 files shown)


dss/__init__.py
```diff
@@ -34,6 +38,26 @@ def get_logger():
...
         return logging.getLogger(__name__)
 
+class OperationWithAuthorizer(Operation):
+    authorized_domains = os.environ["AUTHORIZED_DOMAINS"].split()
+    def oauth2_authorize(self, function):
+        def wrapper(request):
+            if "token_info" in request.context.values:
+                token_info = request.context.values["token_info"]
+                if not int(token_info["expires_in"]) > 0:
+                    raise OAuthProblem(description="Authorization token has expired")
```
ttung: I'm curious how this interacts with the error reporting code, and if compliance with the swagger spec is enforced.

ttung:
If this is blocking integration work, I'd say go ahead and land this and resolve the exceptions issue later.

dss/__init__.py
kislyuk: After rebasing, I found that this does indeed interact with `common_error_handler`. So I special-cased OAuth errors in `common_error_handler`. I don't think the spec is consulted.

PR #268, opened 2017-07-25
Made the error type compliant with RFC 7807.

ttung:
Apparently, there's an RFC for everything under the sun.

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -204,9 +204,15 @@ class TestFileApi(unittest.TestCase, DSSAsserts):
                 requests.codes.not_found
             )
 
+            self.assertHeaders(
+                response[0],
+                {
+                    'content-type': "application/problem+json",
+                }
+            )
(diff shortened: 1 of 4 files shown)


tests/test_file.py
```diff
@@ -204,9 +204,15 @@ def _test_file_get_not_found(self, replica):
                 requests.codes.not_found
             )
 
+            self.assertHeaders(
+                response[0],
+                {
+                    'content-type': "application/problem+json",
+                }
+            )
             self.assertEqual(response[2]['code'], "not_found")
```
Mackey22: nit-picky, but I think it's cleaner to say `response, data, actual_json = AssertGetResponse(...)` and then use `actual_json` instead of `response[2]`. Thoughts?

ttung: Yes. :)

ttung: #278

PR #270, opened 2017-07-25
Apply the @dss_handler decorator to all the dss endpoints.

ttung:


tests/infra.py
```diff
@@ -22,6 +22,21 @@ def get_env(varname):
...
+    status: typing.Optional[int] = None
+    """
+    If this is None, then we not check the status.  For all other values, we test that it matches.
+    """
+
+    expect_stacktrace: typing.Optional[bool] = None
+    """
+    If this is True, then we expect the stacktrace to be present.  If this is False, then we expect the stacktrace to be
+    absent.  If this is None, then we do not test the presence of the stacktrace.
+    """
```
kislyuk: Shouldn't block docstrings be above the variables they are documenting?

ttung: I don't know.  Python docstrings seem to follow the definition, unlike Java.  I followed that convention.

https://www.python.org/dev/peps/pep-0224/

this was rejected but it does put the comment below.

ttung: Also that was part of the previous PR (#269).

kislyuk: Just tested it with Sphinx and indeed, attribute docstrings go under the attribute declarations. Weird.

PR #271, opened 2017-07-25
Initial version of notification support

mikebaumann:
Initial version of notification support provided by the DSS indexer.
Additional testing is needed, yet this PR is being opened now to enable integration with other aspects of the system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test: lint mypy
+	tests/fixtures/populate.py --s3-bucket $(DSS_S3_BUCKET_TEST_FIXTURES) --gs-bucket $(DSS_GS_BUCKET_TEST_FIXTURES)
 	PYTHONWARNINGS=ignore:ResourceWarning coverage run --source=dss -m unittest discover tests -v
 
 deploy:
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test: lint mypy
-	tests/fixtures/populate.py --s3-bucket $(DSS_S3_BUCKET_TEST_FIXTURES) --gs-bucket $(DSS_GS_BUCKET_TEST_FIXTURES)
 	PYTHONWARNINGS=ignore:ResourceWarning coverage run --source=dss -m unittest discover tests -v
 
 deploy:
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -1,5 +1,4 @@
 
-
 """Lambda function for DSS indexing"""
 
 import json
@@ -183,7 +182,7 @@ def process_notifications(bundle_id, subscription_ids, logger):
         try:
             # TODO Batch this request
             subscription = get_subscription(subscription_id, logger)
-            notify(subscription, bundle_id, logger)
+            notify(subscription_id, subscription, bundle_id, logger)
         except Exception as ex:
             logger.error(f"Error occurred while processing subscription {subscription_id} for bundle {bundle_id}. {ex}")
 
@@ -204,12 +203,12 @@ def get_subscription(subscription_id, logger):
         return response['hits']['hits'][0]['_source']
 
 
-def notify(subscription, bundle_id, logger):
+def notify(subscription_id, subscription, bundle_id, logger):
     bundle_uuid, bundle_version = split_bundle_id(bundle_id)
     transaction_id = uuid.uuid4()
     payload = {
         "transaction_id": transaction_id,
-        "subscription_id": subscription['subscription_id'],
+        "subscription_id": subscription_id,
         "query": subscription['query'],
         "match": {
             "bundle_uuid": bundle_uuid,
@@ -221,10 +220,10 @@ def notify(subscription, bundle_id, logger):
     response = requests.post(callback_url, data=payload)
     # TODO (mbaumann) Add webhook retry logic
     if 200 <= response.status_code < 300:
-        logger.info(f"Successfully notified for subscription {subscription['subscription_id']}"
+        logger.info(f"Successfully notified for subscription {subscription_id}"
                     f" for bundle {bundle_id} with transaction id {transaction_id} Code: {response.status_code}")
     else:
-        logger.warning(f"Failed notification for subscription {subscription['subscription_id']}"
+        logger.warning(f"Failed notification for subscription {subscription_id}"
                        f" for bundle {bundle_id} with transaction id {transaction_id} Code: {response.status_code}")
 
 
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -1,4 +1,5 @@
 
+
 """Lambda function for DSS indexing"""
 
 import json
@@ -181,8 +182,8 @@ def process_notifications(bundle_id, subscription_ids, logger):
             # TODO Batch this request
             subscription = get_subscription(subscription_id, logger)
             notify(subscription, bundle_id, logger)
-        except Exception as e:
-            logger.error(f"Error occurred while processing subscription {subscription_id} for bundle {bundle_id}. {e}")
+        except Exception as ex:
+            logger.error(f"Error occurred while processing subscription {subscription_id} for bundle {bundle_id}. {ex}")
 
 
 def get_subscription(subscription_id, logger):
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -11,12 +11,12 @@ import botocore
 import requests
 from elasticsearch.helpers import scan
 
-from dss import (DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE,
+from ... import (DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE,
                  DSS_ELASTICSEARCH_QUERY_TYPE, DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
                  DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE)
-from dss.util import create_blob_key
-from dss.hcablobstore import BundleMetadata, BundleFileMetadata
-from dss.util.es import ElasticsearchClient
+from ...util import create_blob_key
+from ...hcablobstore import BundleMetadata, BundleFileMetadata
+from ...util.es import ElasticsearchClient
 
 DSS_BUNDLE_KEY_REGEX = r"^bundles/[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}\..+$"
 
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -226,4 +226,8 @@ def notify(subscription, bundle_id, logger):
 
 
 def split_bundle_id(bundle_id):
-    return bundle_id[:36], bundle_id[37:]
+    split_index = bundle_id.find('.')
+    if split_index >= 0:
+        return bundle_id[:split_index], bundle_id[(split_index + 1):]
+    else:
+        return bundle_id, None
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -1,4 +1,3 @@
-
 """Lambda function for DSS indexing"""
 
 import json
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -123,11 +123,11 @@ def add_index_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
 
 def create_elasticsearch_index(es_client, logger):
     index_mapping = {
-        "mappings": {
+        'mappings': {
             DSS_ELASTICSEARCH_QUERY_TYPE: {
-                "properties": {
-                    "query": {
-                        "type": "percolator"
+                'properties': {
+                    'query': {
+                        'type': "percolator"
                     }
                 }
             }
@@ -160,11 +160,11 @@ def add_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
 
 def find_matching_subscriptions(index_data, logger):
     percolate_document = {
-        "query": {
-            "percolate": {
-                "field": "query",
-                "document_type": DSS_ELASTICSEARCH_DOC_TYPE,
-                "document": index_data
+        'query': {
+            'percolate': {
+                'field': "query",
+                'document_type': DSS_ELASTICSEARCH_DOC_TYPE,
+                'document': index_data
             }
         }
     }
@@ -189,10 +189,10 @@ def process_notifications(bundle_id, subscription_ids, logger):
 
 def get_subscription(subscription_id, logger):
     subscription_query = {
-        "query": {
-            "ids": {
-                "type": DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
-                "values": [subscription_id]
+        'query': {
+            'ids': {
+                'type': DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                'values': [subscription_id]
             }
         }
     }
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -4,10 +4,12 @@
 import json
 import os
 import re
+import uuid
 from urllib.parse import unquote
 
 import boto3
 import botocore
+import requests
 
 from ... import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
 from ...hcablobstore import BundleMetadata, BundleFileMetadata
@@ -29,6 +31,8 @@ def process_new_indexable_object(event, logger) -> None:
             bundle_id = get_bundle_id_from_key(key)
             index_data = create_index_data(s3, bucket_name, bundle_id, manifest, logger)
             add_index_data_to_elasticsearch(bundle_id, index_data, logger)
+            subscriptions = find_matching_subscriptions(index_data, logger)
+            process_notifications(bundle_id, subscriptions, logger)
             logger.debug(f"Finished index processing of S3 creation event for bundle: {key}")
         else:
             logger.debug(f"Not indexing S3 creation event for key: {key}")
@@ -114,18 +118,30 @@ def add_index_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
     add_data_to_elasticsearch(bundle_id, index_data, logger)
 
 
-def create_elasticsearch_index(logger):
+def create_elasticsearch_index(es_client, logger):
+    index_mapping = {
+        "mappings": {
+            "query": {
+                "properties": {
+                    "query": {
+                        "type": "percolator"
+                    }
+                }
+            }
+        }
+    }
     try:
         es_client = ElasticsearchClient.get(logger)
         response = es_client.indices.exists(DSS_ELASTICSEARCH_INDEX_NAME)
-        if response is False:
+        if not response:
             logger.debug(f"Creating new Elasticsearch index: {DSS_ELASTICSEARCH_INDEX_NAME}")
-            response = es_client.indices.create(DSS_ELASTICSEARCH_INDEX_NAME, body=None)
+            response = es_client.indices.create(DSS_ELASTICSEARCH_INDEX_NAME, body=index_mapping)
             logger.debug(f"Index creation response: {json.dumps(response, indent=4)}")
         else:
             logger.debug(f"Using existing Elasticsearch index: {DSS_ELASTICSEARCH_INDEX_NAME}", )
     except Exception as ex:
         logger.critical(f"Unable to create index {DSS_ELASTICSEARCH_INDEX_NAME}  Exception: {ex}")
+        raise
 
 
 def add_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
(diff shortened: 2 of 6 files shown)


dss/events/handlers/index.py
```diff
@@ -126,25 +132,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+        "subscription_id": subscription['subscription_id'],
+        "query": subscription['query'],
+        "match": {
+            "bundle_uuid": bundle_uuid,
+            "bundle_version": bundle_version
+        }
+    }
+    callback_url = subscription['callback_url']
+    response = requests.post(callback_url, data=payload)
+    if response.status_code == requests.codes.ok:
```
kislyuk: Let's interpret all 200 series responses as successful deliveries.

tests/test_indexer.py
```diff
@@ -52,6 +54,16 @@
...
 
 
+class ElasticsearchTestClient:
+    _es_client = None
+
+    @staticmethod
+    def get(logger):
+        if ElasticsearchTestClient._es_client is None:
+            ElasticsearchTestClient._es_client = connect_elasticsearch(os.getenv("DSS_ES_ENDPOINT"), logger)
+        return ElasticsearchTestClient._es_client
```
kislyuk: Non-blocking comment: this construct seems overwrought and confusing to me. Logger is a global and can just be ingested out-of-band. At first glance, `client.get(logger)` appears to do something with logging, not a database connection manager.

In a test case, I would just use a setUpClass method to initialize a connection manager as a class variable.

Let's revisit this later.

dss/events/handlers/index.py
```diff
@@ -126,25 +132,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+        logger.info(f"Successfully notified for subscription {subscription['subscription_id']}"
+                    f" for bundle {bundle_id} with transaction id {transaction_id} Code: {response.status.code}")
+    else:
+        logger.warning(f"Failed notification for subscription {subscription['subscription_id']}"
+                       f" for bundle {bundle_id} with transaction id {transaction_id} Code: {response.status.code}")
+
+
+def split_bundle_id(bundle_id):
+    return bundle_id[:36], bundle_id[37:]
```
kislyuk: Please split on the actual separator character, not an arbitrary opaque string offset. Also, this one-liner doesn't seem to need a helper function.

dss/events/handlers/index.py
```diff
@@ -126,25 +132,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
 
 def add_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
     try:
         ElasticsearchClient.get(logger).index(index=DSS_ELASTICSEARCH_INDEX_NAME,
                                               doc_type=DSS_ELASTICSEARCH_DOC_TYPE,
                                               id=bundle_id,
-                                              body=json.dumps(index_data, indent=4))
-
+                                              body=json.dumps(index_data),
+                                              refresh=True)
```
kislyuk: How computationally/IO-expensive is refreshing the ES index on every insert at scale? If it is non-trivial, what's your strategy for managing it? Can you give a benchmark of refresh performance as a function of the number of documents inserted?

dss/events/handlers/index.py
```diff
@@ -126,25 +130,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
 
 
 def create_elasticsearch_index(logger):
+    index_mapping = {
+        "mappings": {
+            "query": {
```
Mackey22: "query" should be `DSS_ELASTICSEARCH_QUERY_TYPE`

dss/events/handlers/index.py
```diff
@@ -126,25 +130,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+                "document": index_data
+            }
+        }
+    }
+    response = ElasticsearchClient.get(logger).search(
+        index=DSS_ELASTICSEARCH_INDEX_NAME,
+        body=percolate_document)
+    logger.debug("Found matching subscription count: %i", len(response['hits']['hits']))
+    subscription_ids = set()
+    for hit in response['hits']['hits']:
```
Mackey22: Elasticsearch paginates to 10 results by default. You'll want to make sure you get all results. You can use `for hit in response.scan(): ...` to achieve this.

dss/events/handlers/index.py
```diff
@@ -126,25 +130,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+                "type": DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                "values": [subscription_id]
+            }
+        }
+    }
+    response = ElasticsearchClient.get(logger).search(
+        index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+        body=subscription_query)
+    if len(response['hits']['hits']) == 1:
+        return response['hits']['hits'][0]['_source']
```
Mackey22: More pythonic: 

```    
    id_exists = es_client.exists(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
                                 doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
                                 id=uuid)
    if not id_exists:
        # handle incorrect query
    response = es_client.get(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
                             doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
                             id=uuid)```

File: dss/events/handlers/index.py [hunk 8]
```diff
@@ -126,25 +130,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+
+
+def notify(subscription, bundle_id, logger):
+    bundle_uuid, bundle_version = split_bundle_id(bundle_id)
+    transaction_id = uuid.uuid4()
+    payload = {
+        "transaction_id": transaction_id,
```
Mackey22: For dict keys we want use single quotes

dss/events/handlers/index.py
```diff
@@ -126,25 +130,110 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+        "subscription_id": subscription['subscription_id'],
```
Mackey22: I don't think it's necessary to store `subscription_id` in the subscriptions index. It's already the id of each doc. 
Right now I'm not storing it, so either I can start storing it or you can feed `subscription_id` into this function (you define it on line 191). Let me know what you think.

dss/events/handlers/index.py
mikebaumann: Changed dict keys to use single quotes. For dict values of type string, are single or double quotes preferred?

dss/events/handlers/index.py
mikebaumann: Using refresh when adding index docs at scale is too expensive. It is also unnecessary and has been removed.

dss/events/handlers/index.py
mikebaumann: Changed to use "200 <= response.status_code < 300". Is this too broad, should it be limited to the range of currently defined http codes?

dss/events/handlers/index.py
mikebaumann: Changed to: 
`def split_bundle_id(bundle_id):
    split_index = bundle_id.find('.')
    if split_index >= 0:
        return bundle_id[:split_index], bundle_id[(split_index + 1):]
    else:
        return bundle_id, None`

dss/events/handlers/index.py
```diff
@@ -160,30 +161,28 @@ def add_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
...
-    response = ElasticsearchClient.get(logger).search(
-        index=DSS_ELASTICSEARCH_INDEX_NAME,
-        body=percolate_document)
-    logger.debug("Found matching subscription count: %i", len(response['hits']['hits']))
     subscription_ids = set()
-    for hit in response['hits']['hits']:
+    for hit in scan(ElasticsearchClient.get(logger),
+                    index=DSS_ELASTICSEARCH_INDEX_NAME,
+                    query=percolate_document):
         subscription_ids.add(hit["_id"])
```
Mackey22: ' vs "

dss/events/handlers/index.py
```diff
@@ -126,25 +131,112 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+        "match": {
+            "bundle_uuid": bundle_uuid,
+            "bundle_version": bundle_version
+        }
+    }
+    callback_url = subscription['callback_url']
+    response = requests.post(callback_url, data=payload)
+    if 200 <= response.status_code < 300:
+        logger.info(f"Successfully notified for subscription {subscription['subscription_id']}"
+                    f" for bundle {bundle_id} with transaction id {transaction_id} Code: {response.status.code}")
```
Mackey22: `status.code` should be `status_code`

dss/events/handlers/index.py
```diff
@@ -126,25 +131,112 @@ def add_index_data_to_elasticsearch(bundle_key, index_data, logger) -> None:
...
+    else:
+        logger.warning(f"Failed notification for subscription {subscription['subscription_id']}"
+                       f" for bundle {bundle_id} with transaction id {transaction_id} Code: {response.status.code}")
```
Mackey22: `status.code` should be `status_code`. See https://travis-ci.org/HumanCellAtlas/data-store/builds/257754710 line 295

Mackey22:
A little concerning that there was an error in https://travis-ci.org/HumanCellAtlas/data-store/builds/257754710 (line 925) and the tests still passed.

mikebaumann:
In the description of this PR, I stated additional testing is required and this PR is being made available early for integration purposes. This issue occurred in exactly the area that I still need to create tests for, and the final lines of the whole notification process.

dss/events/handlers/index.py
```diff
@@ -1,21 +1,25 @@
+
+
```
kislyuk: Remove these two leading lines please

dss/events/handlers/index.py
```diff
@@ -1,21 +1,25 @@
...
 from urllib.parse import unquote
 
 import boto3
 import botocore
+import requests
+from elasticsearch.helpers import scan
 
-from ... import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
+from ... import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE, DSS_ELASTICSEARCH_QUERY_TYPE
+from ... import DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE
```
kislyuk: Combine these imports with tuple syntax:
```
from ... import (FOO, BAR,
                 FOO2, BAR2)
```

tests/test_indexer.py
```diff
@@ -10,24 +10,25 @@
...
-from dss.events.handlers.index import process_new_indexable_object, ElasticsearchClient
-from dss.util import create_blob_key
-
-fixtures_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'fixtures'))  # noqa
-sys.path.insert(0, fixtures_root)  # noqa
+import moto
+from dss import BucketStage, Config
+from dss import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
+from dss import DSS_ELASTICSEARCH_QUERY_TYPE
+from dss import DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE
```
kislyuk: Combine these imports please.

tests/test_indexer.py
```diff
@@ -160,6 +171,21 @@ def test_es_client_reuse(self):
...
         self.assertIs(es_client_after_first_call, es_client_after_second_call)
 
+    def test_subscription_notification(self):
+        bundle_key = self.load_test_data_bundle_for_path('fixtures/smartseq2/paired_ends')
+        sample_s3_event = self.create_sample_s3_bundle_created_event(bundle_key)
+        process_new_indexable_object(sample_s3_event, logger)
+
+        ElasticsearchTestClient.get().indices.create(DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME)
+        subscribe_for_notification(smartseq2_paired_ends_query,
+                                   "http://green.box.com/notification",
```
kislyuk: Let's make this into an example.com URL. `green.box.com` might actually exist. Also, I suggest only allowing https:// URLs.

dss/events/handlers/index.py
kislyuk: Why? Why not just use `bundle_id.split(".", 1)`?

dss/events/handlers/index.py
kislyuk: This is fine, I think.

dss/events/handlers/index.py
kislyuk: Or
```
bundle_id, _, bundle_version = bundle_id.partition(".")
return bundle_id, bundle_version
```

dss/events/handlers/index.py
```diff
@@ -1,36 +1,56 @@
...
 from urllib.parse import unquote
 
 import boto3
+import botocore
+import requests
+from elasticsearch.helpers import scan
 
-from ... import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
-from ...util import connect_elasticsearch
+from dss import (DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE,
```
ttung: this screams for a `dss.constants` where you can do:

```
from dss.constants import *
```

Makefile
```diff
@@ -8,6 +8,7 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test: lint mypy
+	tests/fixtures/populate.py --s3-bucket $(DSS_S3_BUCKET_TEST_FIXTURES) --gs-bucket $(DSS_GS_BUCKET_TEST_FIXTURES)
```
ttung: ehhhh why is this here too?

mikebaumann: The branch used for #245 (mbaumann-indexer-review-fixes) had a lot of indexer changes on it needed for the notification work. Due to the time it was taking to get the #245 branch on master, the notification work (mbaumann-notification) was recently branched from mbaumann-indexer-review-fixes (hence the change above). I think the best way to get things into a good state is to finishing landing mbaumann-indexer-review-fixes on master, merge master into mbaumann-notification, then land this branch on master. This is a top priority.

ttung: If this is based on the branch for #245, it shouldn't have the contents of #245 in it.

mikebaumann: It does, except for the changes that have been made to #245 since this branch was created.

mikebaumann: I could merge the changes since then from #245 to this branch, or land #245 on master then merge from there.

ttung: Well, #245 is approved, so the latter approach makes the most sense.

Nevertheless, I believe there is some peculiarity to the way you're using git branches...

dss/events/handlers/index.py
mikebaumann: This proposal requires two round-trips to the server to do the work that should be with one.
No plans to implement.

dss/events/handlers/index.py
mikebaumann: Changed indexer and test to not expect or include the subscription_id in the subscription itself.

dss/events/handlers/index.py
mikebaumann: Yes, a good idea. Let's do that as a separate PR, though (we have a lot, too much, in-flight already :)

dss/events/handlers/index.py
Mackey22: 👍

dss/events/handlers/index.py
```diff
@@ -1,18 +1,22 @@
...
-from ... import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
-from ...hcablobstore import BundleMetadata, BundleFileMetadata
-from ...util import create_blob_key
-from ...util.es import ElasticsearchClient
+from dss import (DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE,
+                 DSS_ELASTICSEARCH_QUERY_TYPE, DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                 DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE)
+from dss.util import create_blob_key
+from dss.hcablobstore import BundleMetadata, BundleFileMetadata
+from dss.util.es import ElasticsearchClient
```
kislyuk: Please use relative imports when importing from the same package. Instead of `from dss`, use `from ...`.

dss/events/handlers/index.py
```diff
@@ -115,24 +121,114 @@ def add_index_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
...
+        raise
 
 
 def add_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
     try:
         ElasticsearchClient.get(logger).index(index=DSS_ELASTICSEARCH_INDEX_NAME,
                                               doc_type=DSS_ELASTICSEARCH_DOC_TYPE,
                                               id=bundle_id,
-                                              body=json.dumps(index_data, indent=4))
+                                              body=json.dumps(index_data))  # Do not use refresh here - too expensive.
```
kislyuk: Can you elaborate in a separate issue on the data consistency strategy here? Is a refresh required before reading data back from the index produces data consistent with the data store?

tests/test_indexer.py
```diff
@@ -10,20 +10,24 @@
...
 import boto3
-import moto
 from botocore.exceptions import ClientError
-from elasticsearch import Elasticsearch
+
 
 import dss
-from dss import BucketStage, Config, DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE
-from dss.events.handlers.index import process_new_indexable_object, ElasticsearchClient
+import moto
```
kislyuk: We group imports into 3 categories:
- stdlib
- third party
- our own libraries
Each group is separated by a blank line.

Since moto is a third-party library, please move it back into the block above.

tests/test_indexer.py
```diff
@@ -78,7 +82,6 @@ def tearDownClass(cls):
             cls.mock_s3.stop()
 
     def setUp(self):
-        Config.set_config(BucketStage.TEST)
```
kislyuk: Why is this line removed?

tests/test_indexer.py
```diff
@@ -159,6 +162,21 @@ def test_es_client_reuse(self):
...
+
+        ElasticsearchClient.get(logger).indices.create(DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME)
+        subscribe_for_notification(smartseq2_paired_ends_query,
+                                   "https://example.com/notification",
+                                   "6112f2a3-8b89-4e54-bbc0-65a98bf8fb8b")
+
+        bundle_key = self.load_test_data_bundle_for_path('fixtures/smartseq2/paired_ends')
+        sample_s3_event = self.create_sample_s3_bundle_created_event(bundle_key)
+        process_new_indexable_object(sample_s3_event, logger)
+        # TODO (mbaumann) Verify notification
```
kislyuk: Please add an issue to track this. Testing the notification end-to-end is an important test case.

dss/events/handlers/index.py
```diff
@@ -115,24 +121,114 @@ def add_index_data_to_elasticsearch(bundle_id, index_data, logger) -> None:
...
+    payload = {
+        "transaction_id": transaction_id,
+        "subscription_id": subscription_id,
+        "query": subscription['query'],
+        "match": {
+            "bundle_uuid": bundle_uuid,
+            "bundle_version": bundle_version
+        }
+    }
+    # TODO (mbaumann) Ensure webhooks are only delivered over verified HTTPS (unless maybe when running a test)
```
kislyuk: Please file an issue for this TODO.

dss/events/handlers/index.py
Mackey22: Fair. If you're worried about 2 round-trips, you can also do:
```
try: 
    response = es_client.get(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
                             doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
                             id=uuid)
except elasticsearch.exceptions.NotFoundError as ex:
    # handle incorrect query
```
I think this is a cleaner way of querying.

dss/events/handlers/index.py
mikebaumann: Opened issue https://github.com/HumanCellAtlas/data-store/issues/282 which includes a brief description of the effect of refresh and a link to the Elasticsearch documentation.

dss/events/handlers/index.py
mikebaumann: Lines 158 and 159 are two blank lines between top level functions, the next of which starts on line 160. This is consistent with PEP 8, which I thought we were using.

dss/events/handlers/index.py
mikebaumann: Added https://github.com/HumanCellAtlas/data-store/issues/283

tests/test_indexer.py
mikebaumann: Added https://github.com/HumanCellAtlas/data-store/issues/284.

tests/test_indexer.py
mikebaumann: This is already being called in setUpClass, which is sufficient, and shouldn't be done again in setUp.

tests/test_indexer.py
mikebaumann: Thank you for describing the expected ordering and spacing. 
Done.

dss/events/handlers/index.py
mikebaumann: Thank you for providing the style guideline.
Done.

dss/events/handlers/index.py
kislyuk: Never mind. The line numbering on the diff render in GitHub is confused. Please just submit smaller diffs in the future.

kislyuk:
Approved, please squash-merge.

dss/events/handlers/index.py
mikebaumann: Yes, that is a nice pattern and will implement it.
I am going to open an new PR for this tomorrow as I need to leave for an interview and this PR is otherwise ready to merge.

PR #277, opened 2017-07-26
Subscriptions API functionality

Mackey22:
* Full endpoint coverage for Subscriptions api
* Full testing of functionality
* Upload files for sample index and query docs for testing
* Manage whitelisted emails with contextmanager and config class

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -301,7 +301,16 @@ paths:
         default:
           description: Unexpected error
           schema:
-            $ref: '#/definitions/Error'
+            allOf:
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem]
+                required:
+                  - code
     put:
       summary: Create a event subscription.
       description: |
@@ -345,14 +354,19 @@ paths:
                 pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
             required:
               - uuid
-        500:
-          description: Internal Server Error
-          schema:
-            $ref: '#/definitions/Error'
         default:
           description: Unexpected error
           schema:
-            $ref: '#/definitions/Error'
+            allOf:
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, elasticsearch_error]
+                required:
+                  - code
   /subscriptions/{uuid}:
     get:
       summary: Retrieve an event subscription given a UUID.
@@ -383,18 +397,19 @@ paths:
             properties:
               subscription:
                 $ref: "#/definitions/Subscription"
-        403:
-          description: Forbidden
-          schema:
-            $ref: '#/definitions/Error'
-        404:
-          description: Subscription Not Found
-          schema:
-            $ref: '#/definitions/Error'
         default:
           description: Unexpected error
           schema:
-            $ref: '#/definitions/Error'
+            allOf:
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, not_found]
+                required:
+                  - code
     delete:
       summary: Delete an event subscription.
       description: |
@@ -428,18 +443,19 @@ paths:
                 format: date-time
             required:
               - timeDeleted
-        403:
-          description: Forbidden
-          schema:
-            $ref: '#/definitions/Error'
-        404:
-          description: Subscription Not Found
-          schema:
-            $ref: '#/definitions/Error'
         default:
           description: Unexpected error
           schema:
-            $ref: '#/definitions/Error'
+            allOf:
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, not_found]
+                required:
+                  - code
   /bundles:
     get:
       summary: Query bundles
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -696,10 +696,10 @@ definitions:
         description: Replica this query is monitoring. One of aws, gcp, or azure.
         type: string
         enum: [aws, gcp, azure]
-      creator_uid:
-        type: integer
-        format: int
-        description: User ID who created this subscription.
+      owner:
+        type: string
+        format: email
+        description: User who created this subscription.
       callback_url:
         description: Url to send request to when a bundle comes in that matches this query.
         type: string
@@ -709,7 +709,7 @@ definitions:
     required:
       - uuid
       - replica
-      - creator_uid
+      - owner
       - callback_url
       - query
   bundle_version:
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -345,6 +345,14 @@ paths:
                 pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
             required:
               - uuid
+        500:
+          description: Internal Server Error
+          schema:
+            $ref: '#/definitions/Error'
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
   /subscriptions/{uuid}:
     get:
       summary: Retrieve an event subscription given a UUID.
@@ -375,6 +383,14 @@ paths:
             properties:
               subscription:
                 $ref: "#/definitions/Subscription"
+        403:
+          description: Forbidden
+          schema:
+            $ref: '#/definitions/Error'
+        404:
+          description: Subscription Not Found
+          schema:
+            $ref: '#/definitions/Error'
         default:
           description: Unexpected error
           schema:
@@ -412,6 +428,14 @@ paths:
                 format: date-time
             required:
               - timeDeleted
+        403:
+          description: Forbidden
+          schema:
+            $ref: '#/definitions/Error'
+        404:
+          description: Subscription Not Found
+          schema:
+            $ref: '#/definitions/Error'
         default:
           description: Unexpected error
           schema:
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -80,7 +80,6 @@ class DSSApp(connexion.App):
 
 class OperationWithAuthorizer(Operation):
     def oauth2_authorize(self, function):
-        @dss_handler
         def wrapper(request):
             authorized_domains = Config.get_allowed_email_domains().split()
             if "token_info" in request.context.values:
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -9,7 +9,6 @@ import os
 import json
 import logging
 
-import connexion
 import flask
 import requests
 import connexion.apis.abstract
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -9,6 +9,7 @@ import os
 import json
 import logging
 
+import connexion
 import flask
 import requests
 import connexion.apis.abstract
@@ -78,16 +79,19 @@ class DSSApp(connexion.App):
         ))
 
 class OperationWithAuthorizer(Operation):
-    authorized_domains = os.environ["AUTHORIZED_DOMAINS"].split()
+    Config.set_config(DeploymentStage.NORMAL)
+
     def oauth2_authorize(self, function):
         def wrapper(request):
+            authorized_domains = Config.get_allowed_emails().split()
             if "token_info" in request.context.values:
                 token_info = request.context.values["token_info"]
+
                 if not int(token_info["expires_in"]) > 0:
                     raise OAuthProblem(description="Authorization token has expired")
                 if json.loads(token_info["email_verified"]) is not True:
                     raise OAuthProblem(description="User email is unverified")
-                if not any(token_info["email"].endswith(f"@{ad}") for ad in self.authorized_domains):
+                if not any(token_info["email"].endswith(f"@{ad}") for ad in authorized_domains):
                     raise Forbidden(description="User email is not authorized to access this resource")
             return function(request)
         return wrapper
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -77,7 +77,13 @@ class DSSApp(connexion.App):
             body=problem,
         ))
 
+
 class OperationWithAuthorizer(Operation):
+    # Can't set authorized_domains here b/c config needs to be set before you
+    # call get_allowed_email_domains or else when this file imported from
+    # test_subscriptions, this will throw an error. Instead, testing_403 is
+    # set to true in the test environment when needed.
+    testing_403 = False
     def oauth2_authorize(self, function):
         def wrapper(request):
             authorized_domains = Config.get_allowed_email_domains().split()
@@ -88,7 +94,7 @@ class OperationWithAuthorizer(Operation):
                     raise OAuthProblem(description="Authorization token has expired")
                 if json.loads(token_info["email_verified"]) is not True:
                     raise OAuthProblem(description="User email is unverified")
-                if not any(token_info["email"].endswith(f"@{ad}") for ad in authorized_domains):
+                if self.testing_403 or not any(token_info["email"].endswith(f"@{ad}") for ad in authorized_domains):
                     raise Forbidden(description="User email is not authorized to access this resource")
             return function(request)
         return wrapper
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -81,7 +81,7 @@ class DSSApp(connexion.App):
 class OperationWithAuthorizer(Operation):
     def oauth2_authorize(self, function):
         def wrapper(request):
-            authorized_domains = Config.get_allowed_emails().split()
+            authorized_domains = Config.get_allowed_email_domains().split()
             if "token_info" in request.context.values:
                 token_info = request.context.values["token_info"]
 
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -79,8 +79,6 @@ class DSSApp(connexion.App):
         ))
 
 class OperationWithAuthorizer(Operation):
-    Config.set_config(DeploymentStage.NORMAL)
-
     def oauth2_authorize(self, function):
         def wrapper(request):
             authorized_domains = Config.get_allowed_emails().split()
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -83,6 +83,7 @@ class OperationWithAuthorizer(Operation):
     # call get_allowed_email_domains or else when this file imported from
     # test_subscriptions, this will throw an error. Instead, testing_403 is
     # set to true in the test environment when needed.
+    # TODO: Remove flag trigger
     testing_403 = False
     def oauth2_authorize(self, function):
         def wrapper(request):
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -80,6 +80,7 @@ class DSSApp(connexion.App):
 
 class OperationWithAuthorizer(Operation):
     def oauth2_authorize(self, function):
+        @dss_handler
         def wrapper(request):
             authorized_domains = Config.get_allowed_email_domains().split()
             if "token_info" in request.context.values:
--- a/dss/api/subscriptions.py
+++ b/dss/api/subscriptions.py
@@ -27,8 +27,6 @@ logger = get_logger()
 
 def get(uuid: str, replica: str):
     owner = request.token_info['email']
-    print("\n\nOWNER!")
-    print(owner)
 
     es_client = ElasticsearchClient.get(logger)
     id_exists = es_client.exists(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
--- a/dss/api/subscriptions.py
+++ b/dss/api/subscriptions.py
@@ -19,12 +19,14 @@ from .. import (get_logger,
                 DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE)
 from ..blobstore import BlobNotFoundError
 from ..config import Config
+from ..error import DSSException, dss_handler
 from ..hcablobstore import FileMetadata, HCABlobStore
 from ..util.es import ElasticsearchClient, get_elasticsearch_index
 
 logger = get_logger()
 
 
+@dss_handler
 def get(uuid: str, replica: str):
     owner = request.token_info['email']
 
@@ -33,9 +35,7 @@ def get(uuid: str, replica: str):
                                  doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
                                  id=uuid)
     if not id_exists:
-        return jsonify(dict(message=f"Subscription {uuid} does not exist",
-                            exception="Placeholder",
-                            HTTPStatusCode=requests.codes.not_found)), requests.codes.not_found
+        raise DSSException(requests.codes.not_found, "not_found", "Cannot find subscription!")
 
     response = es_client.get(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
                              doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
@@ -46,14 +46,12 @@ def get(uuid: str, replica: str):
     source['replica'] = replica
 
     if source['owner'] != owner:
-        return jsonify(dict(
-            message=f"You don't own subscription {uuid}.",
-            exception="Placeholder",
-            HTTPStatusCode=requests.codes.forbidden)), requests.codes.forbidden
+        raise DSSException(requests.codes.forbidden, "forbidden", "Your credentials can't access this subscription!")
 
     return jsonify(source), requests.codes.okay
 
 
+@dss_handler
 def find(replica: str):
     owner = request.token_info['email']
     es_client = ElasticsearchClient.get(logger)
@@ -75,6 +73,7 @@ def find(replica: str):
     return jsonify(full_response), requests.codes.okay
 
 
+@dss_handler
 def put(extras: dict, replica: str):
     uuid = str(uuid4())
     query = extras['query']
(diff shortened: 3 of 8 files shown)


dss-api.yml
```diff
@@ -687,10 +687,10 @@ definitions:
         description: Replica this query is monitoring. One of aws, gcp, or azure.
         type: string
         enum: [aws, gcp, azure]
-      creator_uid:
-        type: integer
-        format: int
-        description: User ID who created this subscription.
+      owner:
```
ttung: wait this isn't coming in via oauth?

dss/api/subscriptions.py
```diff
@@ -3,10 +3,14 @@
 import json
 import re
 import typing
+import uuid as uuid_
```
ttung: suggestion:

```
from uuid import uuid4
```

then just use uuid4()

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
 def get(uuid: str, replica: str):
-    pass
+    owner = request.token_info['email']
+
+    es_client = ElasticsearchClient.get(logger)
+    id_exists = es_client.exists(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                                 doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                                 id=uuid)
+    if not id_exists:
+        return jsonify(dict(message="Subscription {} does not exist".format(uuid),
```
ttung: `raise DSSException(..)`

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
 
 
 def get(uuid: str, replica: str):
```
ttung: `@dss_handler`

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+    if not id_exists:
+        return jsonify(dict(message="Subscription {} does not exist".format(uuid),
+                            exception="Placeholder",
+                            HTTPStatusCode=requests.codes.not_found)), requests.codes.not_found
+
+    response = es_client.get(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                             doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                             id=uuid)
+
+    source = response['_source']
```
ttung: i didn't read the spec, but the key really starts with an underscore?

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+    source['uuid'] = uuid
+    source['replica'] = replica
+
+    if source['owner'] != owner:
+        return jsonify(dict(
```
ttung: `DSSException`

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+            message="You don't own subscription {}.".format(uuid),
+            exception="Placeholder",
+            HTTPStatusCode=requests.codes.forbidden)), requests.codes.forbidden
+
+    return jsonify(source), requests.codes.okay
```
ttung: I don't think you need the `requests.code.okay` part, but I could be wrong.

dss/api/subscriptions.py
ttung: use f-style strings.

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+                             doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                             id=uuid)
+
+    source = response['_source']
+    source['uuid'] = uuid
+    source['replica'] = replica
+
+    if source['owner'] != owner:
+        return jsonify(dict(
+            message="You don't own subscription {}.".format(uuid),
```
ttung: f-style strings

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
-    pass
+    owner = request.token_info['email']
+    es_client = ElasticsearchClient.get(logger)
+
+    search_obj = Search(using=es_client,
+                        index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                        doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE)
+    search = search_obj.query({'match': {'owner': owner}})
+
+    responses = []
```
ttung:
```
responses = [{
    'uuid': hit.meta.id,
    'replica': replica,
    'owner': owner,
    'callback_url': hit.callback_url,
    'query': hit.query.to_dict()}
  for hit in search.scan()]
```

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+    }
+    get_elasticsearch_index(es_client, DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, logger, index_mapping)
+
+    percolate_registration = _register_percolate(es_client, uuid, query)
+
+    if percolate_registration['created']:
+        logger.debug("Percolate query registration succeeded:\n{}".format(percolate_registration))
+    else:
+        logger.critical("Percolate query registration failed:\n{}".format(percolate_registration))
+        return (jsonify(dict(
```
ttung: `DSSException`

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+            exception="Placeholder",
+            HTTPStatusCode=requests.codes.internal_server_error)),
+            requests.codes.internal_server_error)
+
+    extras['owner'] = owner
+    subscription_registration = _register_subscription(es_client, uuid, extras)
+    if subscription_registration['created']:
+        logger.debug("Event Subscription succeeded:\n{}".format(subscription_registration))
+    else:
+        logger.critical("Event Subscription failed:\n{}".format(subscription_registration))
```
ttung: in the case of failure, is it desirable to remove the percolate query?

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+        return (jsonify(dict(
```
ttung: DSSException

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+    if source['owner'] != owner:
+        return jsonify(dict(
+            message="You don't own subscription {}.".format(uuid),
+            exception="Placeholder",
+            HTTPStatusCode=requests.codes.forbidden)), requests.codes.forbidden
+
+    return jsonify(source), requests.codes.okay
 
 
 def find(replica: str):
```
ttung: `@dss_handler`

dss/api/subscriptions.py
```diff
@@ -22,16 +26,151 @@
...
+        return (jsonify(dict(
+            message="Unable to register elasticsearch percolate query {}.".format(uuid),
+            exception="Placeholder",
+            HTTPStatusCode=requests.codes.internal_server_error)),
+            requests.codes.internal_server_error)
+
+    return jsonify(dict(uuid=uuid)), requests.codes.created
 
 
 def delete(uuid: str, replica: str):
```
ttung: `@dss_handler`

dss-api.yml
Mackey22: This is part of the object returned by get

dss/api/subscriptions.py
Mackey22: 👍

dss/api/subscriptions.py
Mackey22: Yeah, it's a little funky. There may be a better way that I can look into

dss/api/subscriptions.py
Mackey22: 👍

dss/api/subscriptions.py
Mackey22: 👍

dss/api/subscriptions.py
Mackey22: Following format established in files.py: 

```
    return jsonify(
        dict(version=version)), requests.codes.created
```

dss/api/subscriptions.py
Mackey22: 👍

dss/api/subscriptions.py
Mackey22: Yeah that makes sense - I'll add that.

Mackey22:
@ttung As a whole, I avoided rebasing against master because this branch is stacked on top of two other ones. Wanted to avoid a huge mess with the DSSExceptions and dss_handler decorator. Do you think I should anyways?

dss-api.yml
kislyuk: Since nobody but the owner can see notifications for now, I question the need for including this field.

ttung:
> @ttung As a whole, I avoided rebasing against master because this branch is stacked on top of two other ones. Wanted to avoid a huge mess with the DSSExceptions and dss_handler decorator. Do you think I should anyways?

Happy to show you how to rebase your entire stack if you want to.  It's a useful thing to be able to do.

Alternatively, you could do that rework on a separate PR.

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
+                           "Unable to register elasticsearch percolate query!")
+
+    extras['owner'] = owner
+    subscription_registration = _register_subscription(es_client, uuid, extras)
+    if subscription_registration['created']:
+        logger.debug(f"Event Subscription succeeded:\n{subscription_registration}")
+    else:
+        logger.critical(f"Event Subscription failed:\n{subscription_registration}")
+
+        # Delete percolate query to make sure queries and subscriptions are in sync.
```
mikebaumann: To help ensure the percolate query gets removed if there is a problem adding the subscription, it seems try/except would be a useful construct here.
You were clearly careful with your error checking and I think this would pretty reliable as it stands, yet try/except is worth considering.

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
 def delete(uuid: str, replica: str):
-    pass
+    owner = request.token_info['email']
+
+    es_client = ElasticsearchClient.get(logger)
+    id_exists = es_client.exists(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                                 doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                                 id=uuid)
+    if not id_exists:
+        raise DSSException(requests.codes.not_found, "not_found", "Cannot find subscription!")
```
mikebaumann: The nice pattern for doing this with one request to the ES service that you recently suggested to me could work well here, too.

dss/config.py
```diff
@@ -61,13 +63,13 @@ def get_storage_schema(replica: str) -> str:
     @staticmethod
     def get_s3_bucket() -> str:
         if Config._S3_BUCKET is None:
-            if Config._CURRENT_CONFIG == BucketStage.NORMAL:
+            if Config._CURRENT_CONFIG == DeploymentStage.NORMAL:
```
mikebaumann: I like the name change to DeploymentStage.

Mackey22: Thanks!

dss/api/subscriptions.py
Mackey22: Yeah I'll convert to that

dss/api/subscriptions.py
Mackey22: Yeah that makes sense.

Are you sure there's an exception thrown if the subscription is not created? It seems logical that it would, but also weird that there would be a `created` field.

dss/api/subscriptions.py
mikebaumann: For Elasticsearch per se, and elasticsearch-py, '_source' is used very prominently and consistently in their docs. I am not aware of another option. The elasticsearch-dsl python library organizes its output differently and '_source' does not actually appear. The two output formats are different and each has its advantages and disadvantages.

chalice/app.py
```diff
@@ -7,12 +7,12 @@
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'chalicelib'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
-from dss import BucketStage, Config, create_app
+from dss import DeploymentStage, Config, create_app
```
ttung: I, too, like the rename, but this clutters the diff up significantly.  This should be done in a separate PR.  #290

dss/__init__.py
```diff
@@ -77,16 +77,19 @@ def common_error_handler(exception):
         )
 
 class OperationWithAuthorizer(Operation):
-    authorized_domains = os.environ["AUTHORIZED_DOMAINS"].split()
+    Config.set_config(DeploymentStage.NORMAL)
```
ttung: Why is this being set here?

dss/config.py
```diff
@@ -97,20 +99,57 @@ def get_gs_bucket() -> str:
         return Config._GS_BUCKET
 
     @staticmethod
-    def _clear_cached_config():
+    def get_allowed_emails() -> str:
+        if Config._ALLOWED_EMAILS is None:
```
ttung: Why is the allowed emails tied to the deployment stage?

dss/config.py
```diff
@@ -97,20 +99,57 @@ def get_gs_bucket() -> str:
         return Config._GS_BUCKET
 
     @staticmethod
-    def _clear_cached_config():
+    def get_allowed_emails() -> str:
```
ttung: should this be `get_allowed_domains()` or maybe `get allowed_email_domains()`?

dss/api/subscriptions.py
mikebaumann: I think an exception would not be thrown if a subscription is not created. It's more that the try block would cover for whatever else could go wrong it the subscription code that could exit the function without cleaning-up the query. It is straightforward and carefully written, so the chance of problems seems low, yet the try/except provides additional assurance and I think is the commonly expected construct when clean-up needs to be done if something goes wrong.

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
+from elasticsearch import Elasticsearch
+from elasticsearch.exceptions import NotFoundError
+from elasticsearch_dsl import Search
 from flask import jsonify, make_response, redirect, request
 from werkzeug.exceptions import BadRequest
 
-from .. import get_logger
-from .. import DSS_ELASTICSEARCH_INDEX_NAME, DSS_ELASTICSEARCH_DOC_TYPE, DSS_ELASTICSEARCH_QUERY_TYPE
-from .. import DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE
+from .. import (get_logger,
```
ttung: If you want to make your code extra reviewable (particularly in the future), you can put a NL after your `(` and everything in subsequent lines will be indented +4, rather than be subject to the contents of the previous line.

Just a comment. Not a requirement.

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
 logger = get_logger()
 
 
+@dss_handler
 def get(uuid: str, replica: str):
-    pass
+    owner = request.token_info['email']
 
+    es_client = ElasticsearchClient.get(logger)
+    id_exists = es_client.exists(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
```
ttung: Is it required to do two round trips here?  Why not just do a get?

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
+                "properties": {
+                    "owner": {
+                        "type": "string",
+                        "index": "not_analyzed"
+                    }
+                }
+            }
+        }
+    }
+    get_elasticsearch_index(es_client, DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, logger, index_mapping)
```
ttung: I don't understand why you're getting an index mapping here. Help?

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
 def get(uuid: str, replica: str):
-    pass
+    owner = request.token_info['email']
 
+    es_client = ElasticsearchClient.get(logger)
+    id_exists = es_client.exists(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                                 doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                                 id=uuid)
+    if not id_exists:
+        raise DSSException(requests.codes.not_found, "not_found", "Cannot find subscription!")
```
ttung: You should update your entry in the swagger spec to expect this possible return value.

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
+    response = es_client.get(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                             doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                             id=uuid)
+
+    source = response['_source']
+    source['uuid'] = uuid
+    source['replica'] = replica
+
+    if source['owner'] != owner:
+        raise DSSException(requests.codes.forbidden, "forbidden", "Your credentials can't access this subscription!")
```
ttung: You should update your entry in the swagger spec to expect this possible return value.

dss/api/subscriptions.py
```diff
@@ -3,35 +3,174 @@
...
+    }
+    get_elasticsearch_index(es_client, DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, logger, index_mapping)
+
+    percolate_registration = _register_percolate(es_client, uuid, query)
+
+    if percolate_registration['created']:
+        logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
+    else:
+        logger.critical(f"Percolate query registration failed:\n{percolate_registration}")
+        raise DSSException(requests.codes.internal_server_error,
```
ttung: You should update your entry in the swagger spec to expect this possible return value.

dss/api/subscriptions.py
ttung: 👍

dss/api/subscriptions.py
```diff
@@ -3,35 +3,169 @@
...
+                        "type": "string",
+                        "index": "not_analyzed"
+                    }
+                }
+            }
+        }
+    }
+    get_elasticsearch_index(es_client, DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME, logger, index_mapping)
+
+    try:
```
Mackey22: @mikebaumann I'm going under the assumption that an exception is thrown if a document is not added. The example they give in the Elasticsearch documentation (https://www.elastic.co/guide/en/elasticsearch/guide/master/create-doc.html) of a document not being added responds with top-level objects of error and status. Status is 409, which translates nicely to Elasticsearch's Python Bindings (https://elasticsearch-py.readthedocs.io/en/master/exceptions.html) ConflictError. I think it's reasonable to assume Elasticsearch would throw this in this situation and another exception if for any other reason a doc can't be stored, so I'm catching all with ElasticsearchException.

tests/test_subscriptions.py
```diff
@@ -0,0 +1,157 @@
...
+                                 id=uuid_)
+        registered_query = response['_source']
+        self.assertEqual(self.sample_percolate_query, registered_query)
+
+    def test_get(self):
+        find_uuid = self._put_subscription()
+        url = str(UrlBuilder()
+                  .set(path="/v1/subscriptions/" + str(find_uuid))
+                  .add_query("replica", "aws"))
+        with override_email_config(DeploymentStage.TEST):
```
ttung: Why is it necessary to override email config to the test stage?  Shouldn't it be defaulting to the test stage in tests anyway?

tests/test_subscriptions.py
```diff
@@ -0,0 +1,157 @@
...
+        uuid_ = self._put_subscription()
+
+        es_client = ElasticsearchClient.get(logger)
+        response = es_client.get(index=DSS_ELASTICSEARCH_INDEX_NAME,
+                                 doc_type=DSS_ELASTICSEARCH_QUERY_TYPE,
+                                 id=uuid_)
+        registered_query = response['_source']
+        self.assertEqual(self.sample_percolate_query, registered_query)
+
+    def test_get(self):
```
ttung: should test your negative code paths like:

1. can't find it.
2. not the owner.

tests/test_subscriptions.py
```diff
@@ -0,0 +1,157 @@
...
+        with override_email_config(DeploymentStage.TEST):
+            _, _, json_response = self.assertGetResponse(
+                url,
+                requests.codes.okay,
+                headers=self._get_auth_header())
+        self.assertEqual(self.sample_percolate_query, json_response['subscriptions'][0]['query'])
+        self.assertEqual(self.callback_url, json_response['subscriptions'][0]['callback_url'])
+        self.assertEqual(NUM_ADDITIONS, len(json_response['subscriptions']))
+
+    def test_delete(self):
```
ttung: ditto re negative code paths.

chalice/app.py
Mackey22: Sounds good. After you merge I'll rebase against that and that should take care of any conflicts

dss/__init__.py
Mackey22: Unnecessary. I'll remove it.

dss/api/subscriptions.py
Mackey22: 👍

dss/api/subscriptions.py
Mackey22: 👍

dss/api/subscriptions.py
Mackey22: I'll add a comment. Short explanation is that when I was testing locally, Elasticsearch was preprocessing inputs by splitting strings on punctuation. So for `john@example.com`, if I searched for people with the email address `jill@example.com`, `john@example.com` would show up because elasticsearch matched example w/ example. By including `"index": "not_analyzed"`, Elasticsearch leaves all owner inputs alone.

dss/config.py
Mackey22: Will change to `allowed_email_domains()`

dss/config.py
Mackey22: We're using Service Account credentials when testing on Travis but, for the moment, only allowing email addresses ending in `chanzuckerberg.com ucsc.edu broadinstitute.org ebi.ac.uk` when live.

tests/test_subscriptions.py
Mackey22: 👍

tests/test_subscriptions.py
Mackey22: Yep, this is unnecessary. I'll need to do this for negative code path 403 tho.

tests/test_subscriptions.py
Mackey22: 👍

dss/api/subscriptions.py
mikebaumann: That sounds right, @Mackey22. Thanks for reviewing this.

chalice/app.py
ttung: I merged.  Please rebase!

Mackey22: :/

Mackey22: That actually wasn't nearly as bad as I expected it to be

ttung: Yeah, git's merge algorithms are pretty good.  It's good practice to do these no-op changes separately. :)

dss-api.yml
```diff
@@ -345,6 +345,14 @@ paths:
                 pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
             required:
               - uuid
+        500:
+          description: Internal Server Error
+          schema:
+            $ref: '#/definitions/Error'
```
ttung: You should look at GET /files to see how you should pattern this.  It should include the allowed error codes.

Mackey22: K. Looks like I changed a couple of the paths and just forgot about this one.

Mackey22: Think I misunderstood. You want all of the error types to look like this:

```        default:
          description: Unexpected error
          schema:
            allOf:
              - $ref: '#/definitions/Error'
              - type: object
                properties:
                  code:
                    type: string
                    description: Machine-readable error code.  The types of return values should not be changed lightly.
                    enum: [unhandled_exception, not_found]
                required:
                  - code```
Where enum lists all of the possible error types? So there will just be one error listing for each endpoint?

File: dss/__init__.py [hunk 33]
```diff
@@ -9,6 +9,7 @@
 import json
 import logging
 
+import connexion
```
kislyuk: I don't think this import is needed, please remove it.

dss/api/subscriptions.py
```diff
@@ -3,35 +3,176 @@
...
+        response = es_client.get(index=DSS_ELASTICSEARCH_SUBSCRIPTION_INDEX_NAME,
+                                 doc_type=DSS_ELASTICSEARCH_SUBSCRIPTION_TYPE,
+                                 id=uuid)
+    except NotFoundError as ex:
+        raise DSSException(requests.codes.not_found, "not_found", "Cannot find subscription!")
+
+    source = response['_source']
+
+    if source['owner'] != owner:
```
kislyuk: It's unclear in this notation which variable contains the owner in the request vs. the owner in the database document. Can you change it to be more explicit?

environment
```diff
@@ -26,7 +26,10 @@ DSS_GS_BUCKET_TEST=org-humancellatlas-dss-test
 DSS_GS_BUCKET_TEST_FIXTURES=org-humancellatlas-dss-test-fixtures
 GOOGLE_APPLICATION_CREDENTIALS="${DSS_HOME}/gcp-credentials.json"
 API_HOST=hca-dss.czi.technology
-AUTHORIZED_DOMAINS="chanzuckerberg.com ucsc.edu broadinstitute.org ebi.ac.uk"
+TOKENINFO_URL=https://hca-dss.czi.technology/internal/tokeninfo
+DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS="chanzuckerberg.com ucsc.edu broadinstitute.org ebi.ac.uk"
+DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_TEST="human-cell-atlas-travis-test.iam.gserviceaccount.com"
+DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_EMPTY=""
```
kislyuk: As discussed offline - I don't think this variable or the NO_AUTH config stage is necessary

dss/api/subscriptions.py
Mackey22: 👍  Changed `owner` to `authenticated_user_email` and `source` to `stored_metadata`

dss/__init__.py
Mackey22: 👍

environment
Mackey22: Considering this and alternatives.

dss/config.py
```diff
@@ -97,20 +99,57 @@ def get_gs_bucket() -> str:
...
+    try:
+        Config._CURRENT_CONFIG = temp_config
+        yield
+    finally:
+        Config._CURRENT_CONFIG = original_config
+        Config._clear_cached_bucket_config()
+
+
+@contextmanager
+def override_email_config(temp_config: DeploymentStage):
```
ttung: This doesn't appear to be used / necessary any more.  Can this be nuked?

Alternatively, could this be used to fake the 403 instead of the flag?

Mackey22: Yeah this can be nuked. We didn't want to use this to fake the 403 b/c a NO_AUTH deployment stage isn't the best. Gonna nuke this then merge

PR #278, opened 2017-07-26
simpler return values from assert*Response

ttung:
Thanks to @mackey22 for the suggestion.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -47,7 +47,7 @@ class TestApi(unittest.TestCase, DSSAsserts):
         self.create_bundle(bundle)
 
     def upload_file(self, bundle_file):
-        response = self.assertPutResponse(
+        resp_obj = self.assertPutResponse(
             f"/v1/files/{bundle_file.uuid}",
             requests.codes.created,
             json_request_body=dict(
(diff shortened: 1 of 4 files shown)


tests/test_api.py
```diff
@@ -56,21 +56,19 @@ def upload_file(self, bundle_file):
                 source_url=bundle_file.url
             )
         )
-        response_data = json.loads(response[1])
-        self.assertIs(type(response_data), dict)
-        self.assertIn('version', response_data)
-        return response_data['version']
+        self.assertIs(type(response_json), dict)
```
Mackey22: Looks like you wanted the middle return value here. Are they the same thing?

tests/test_api.py
```diff
@@ -91,12 +89,11 @@ def put_bundle_payload(bundle):
         return payload
 
     def get_bundle_and_check_files(self, bundle):
-        response = self.assertGetResponse(
+        _, _, response_json = self.assertGetResponse(
```
Mackey22: Same as above

tests/test_api.py
ttung: The text of the response is response[1].
The text of the response parsed into python structures from JSON is response[2].

I removed the call to `json.loads()`, so I just use response[2] directly.

tests/test_api.py
ttung: ditto.

ttung:
Changed to a typed NamedTuple.  Ready for review.

PR #279, opened 2017-07-26
Get DSS running in Docker again

sampierson:
requirements-dev.txt now calls requirements.txt so we need both before
running pip.
Add in sourcing of environment before execution of python script.
Ignore the 10,000 files in daemons - they're not useful in the image and
just slow down image building.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@
 #
 FROM python:3.6
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-  vim
+  vim jq moreutils gettext
 RUN pip install awscli --upgrade
 
 RUN sed 's/#force_color_prompt=yes/force_color_prompt=yes/' /etc/skel/.bashrc > /root/.bashrc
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,16 +18,13 @@ RUN sed 's/#force_color_prompt=yes/force_color_prompt=yes/' /etc/skel/.bashrc >
 ADD .dockerfiles/.vimrc /root/
 
 WORKDIR /code/data-store
-ADD requirements-dev.txt .
+ADD requirements-dev.txt requirements.txt ./
```
ttung: is the trailing `/` intentional?  this was just previously `.`

sampierson: From the Dockerfile reference: "If multiple <src> resources are specified, either directly or due to the use of a wildcard, then <dest> must be a directory, and it must end with a slash /."

ttung: Ok cool, I've never used docker. :)

Just wanted to make sure it was intentional.

sampierson: I had to figure it out the hard way :)  Not the most intuitive behavior.

PR #285, opened 2017-07-27
Basic framework to specify work that can be broken up.

ttung:
Three major components:

1. There is an abstract class `ChunkedTask` that specifies the work.  Calls to `run_one_unit()` will run one unit of work.  Calls to `get_state()` will serialize the state for subsequent execution.
2. There is an abstract class `ChunkedTaskRuntime` that specifies the environment for the work.  Calls to `get_remaining_time_in_millis()` should return the amount of time left in this current execution environment.
3. There is a class `ChunkedTaskRunner` that orchestrates the execution.

Add unit tests to test ChunkedTaskRunner.

### DIFF
--- a/dss/events/chunkedtask/__init__.py
+++ b/dss/events/chunkedtask/__init__.py
@@ -1,0 +1,3 @@
+from .base import Task, Runtime
+from .constants import TIME_OVERHEAD_FACTOR
+from .runner import Runner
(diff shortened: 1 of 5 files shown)


ttung:
I do not know why this magically closed.  :(  GITHUB!!

PR #287, opened 2017-07-27
Implementation of ChunkedTask tailored for AWS Lambda.

ttung:
The unit test fakes a more restrictive environment in order to run the test in a reasonable time (~15-20s).  There will be a test that runs against the standard `AWSRuntime` I will build that will run on the cron'ed test run.

### DIFF
--- a/tests/test_chunkedtask.py
+++ b/tests/test_chunkedtask.py
@@ -4,15 +4,22 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import itertools
+import json
 import os
 import sys
+import time
 import typing
 import unittest
+from uuid import uuid4
+
+import boto3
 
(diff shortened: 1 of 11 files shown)


tests/test_chunkedtask.py
```diff
@@ -4,15 +4,23 @@
...
+
+import boto3
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
 from dss.events import chunkedtask
+from dss.events.chunkedtask import awsconstants
+from dss.events.chunkedtask._awsimpl import AWS_FAST_TEST_CLIENT_NAME
+from dss.util.aws import ARN
```
kislyuk: It looks like you ended up not using this :)

PR #292, opened 2017-07-28
Basic framework to specify work that can be broken up.

ttung:
Three major components:

1. There is an abstract class `ChunkedTask` that specifies the work.  Calls to `run_one_unit()` will run one unit of work.  Calls to `get_state()` will serialize the state for subsequent execution.
2. There is an abstract class `ChunkedTaskRuntime` that specifies the environment for the work.  Calls to `get_remaining_time_in_millis()` should return the amount of time left in this current execution environment.
3. There is a class `ChunkedTaskRunner` that orchestrates the execution.

Add unit tests to test ChunkedTaskRunner.

### DIFF
--- a/dss/events/chunkedtask/base.py
+++ b/dss/events/chunkedtask/base.py
@@ -1,0 +1,43 @@
+import typing
+
+TaskStateType = typing.TypeVar('TaskStateType')
+RuntimeStateType = typing.TypeVar('RuntimeStateType')
+
+
+class Task(typing.Generic[TaskStateType]):
+    def run_one_unit(self) -> bool:
+        """
+        In implementations of `Task`, this should run one unit of work.  Returns true if there's more work to do.
+        """
+        raise NotImplementedError()
+
+    @property
+    def expected_max_one_unit_runtime_millis(self) -> int:
+        """In implementations of `ChunkedTask`, this should return the expected maximum runtime for one unit of work."""
+        raise NotImplementedError()
+
+    def get_state(self) -> TaskStateType:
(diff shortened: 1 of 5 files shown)


tests/test_chunkedtask.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self.rounds_remaining -= 1
+
+        return self.rounds_remaining > 0
+
+
+class TestChunkedTaskRunner(unittest.TestCase):
+    def test_workload_resumes(self):
+        initial_state = (1, 1, 25)
+        expected_max_one_unit_runtime_millis = 10  # we know exactly how long we'll take.  we're so good at guessing!
```
kislyuk: :grin:

dss/events/chunkedtask/base.py
```diff
@@ -0,0 +1,43 @@
+import typing
+
+TaskStateType = typing.TypeVar('TaskStateType')
+RuntimeStateType = typing.TypeVar('RuntimeStateType')
+
+
+class Task(typing.Generic[TaskStateType]):
+    def run_one_unit(self) -> bool:
```
kislyuk: Consider using an iterator for ChunkedTask, instead of returning True or False.

ttung: In #323, I allow the task to return a value, so return values actually become significant.

ttung:
> This looks good overall. I wonder why use a custom serialization scheme instead of base64-encoded pickle or json.dumps(object.dict)?

#325

PR #295, opened 2017-07-29
Connexion error handler overload: translate response into Flask format

kislyuk:
@ttung, I believe this is necessary when testing - @Mackey22 ran into this problem.

### DIFF
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -12,6 +12,7 @@ import logging
 import flask
 import requests
 import connexion.apis.abstract
+from connexion.apis.flask_api import FlaskApi
 from connexion.lifecycle import ConnexionResponse
 from connexion.operation import Operation
 from connexion.resolver import RestyResolver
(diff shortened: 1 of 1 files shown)


dss/__init__.py
```diff
@@ -68,12 +69,12 @@ def common_error_handler(exception):
             problem['status'] = exception.code
             problem['code'] = exception.__class__.__name__
             problem['title'] = exception.description
-        return ConnexionResponse(
+        return FlaskApi.get_response(ConnexionResponse(
```
ttung: Does this mean `error.py` needs the same change?

ttung:
This does argue for the hilarious intentional-mismatched handler though.

dss/__init__.py
kislyuk: I think `error.py` doesn't need it because it's deeper in the middleware stack, and has a function above it that will translate its ConnexionResponse to Flask format.

`common_error_handler` on the other hand sits outside the connexion middleware stack, immediately on the Flask stack, it seems.

Issue #297, opened 2017-07-31
get_files returning 500 instead of 400 when a required arg is missing


Mackey22:
Testing files endpoint with uuid defined but no given replica returns 500 response code: 

```
{u'code': u'unhandled_exception',
 u'stacktrace': u'Traceback (most recent call last):\n  File "/var/task/chalicelib/dss/error.py", line 22, in wrapper\n    return func(*args, **kwargs)\nTypeError: get() missing 1 required positional argument: \'replica\'\n',
 u'status': 500,
 u'title': u"get() missing 1 required positional argument: 'replica'"}
```

Will update with more similar responses as I come across them.

PR #298, opened 2017-07-31
Handle parameter mismatches and return 400.

ttung:
Right now, parameter mismatches raise a generic TypeError.  We can't catch that because it could be an internal TypeError, and not necessarily tied to the parameters users send us.

Instead, we attempt to bind the functions to the args, and if that fails, then we know it's the arguments a user sends us.  If it passes, and some *other* TypeError is raised, we still send that to the generic exception handler.

Add test for this case.

Addresses issue #297, and thanks to @Mackey22 for finding this.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -203,7 +203,7 @@ paths:
                   code:
                     type: string
                     description: Machine-readable error code.  The types of return values should not be changed lightly.
-                    enum: [unhandled_exception, not_found]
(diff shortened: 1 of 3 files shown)


kislyuk:
😮 
Sounds like another connexion issue...

ttung:
> Sounds like another connexion issue...

I don't think it's a connexion issue per se.  It's a philosophical mismatch, where connexion is about approximate error information, and we're trying to provide more precise error information. :)

PR #299, opened 2017-08-01
Utility function to find the next N missing parts of a multipart upload.

ttung:
Allows a user to specify the starting part number from where we search and the number of results.

Adds a ton of tests to verify the results.

### DIFF
--- a/dss/blobstore/s3.py
+++ b/dss/blobstore/s3.py
@@ -196,3 +196,49 @@ class S3BlobStore(BlobStore):
                 multipart_chunksize=64 * 1024 * 1024,
             ),
         )
+
+    def find_next_missing_parts(
+            self,
+            bucket: str,
+            key: str,
+            upload_id: str,
+            part_count: int,
+            search_start: int=1,
+            return_count: int=1) -> typing.Sequence[int]:
+        """
+        Given a `bucket`, `key`, and `upload_id`, find the next N missing parts of a multipart upload, where
+        N=`return_count`.  If `search_start` is provided, start the search at part M, where M=`search_start`.
+        `part_count` is the number of parts expected for the upload.
+
+        Note that the return value may contain fewer than N parts.
+        """
+        if part_count < search_start:
+            raise ValueError("")
+        result = list()
+        while True:
+            kwargs = dict(Bucket=bucket, Key=key, UploadId=upload_id)  # type: dict
+            if search_start > 1:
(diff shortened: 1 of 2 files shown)


dss/blobstore/s3.py
```diff
@@ -188,3 +188,46 @@ def copy(
...
+
+                # have we met our requirements?
+                if len(result) == return_count or search_start == part_count:
+                    return result
+
+                search_start += 1
+
+                if parts_resp['IsTruncated'] and search_start == parts_resp['NextPartNumberMarker']:
+                    # finished examining the results of this batch, move onto the next one
+                    break
```
kislyuk: IMHO using an S3 MPU resource with a parts iterator works better here by removing a bunch of complexity and using boto3's implementation.

http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.MultipartUpload.all

ttung: I don't think a parts iterator would be that much better, since there would still need to be logic to *invert* the results (we're looking for missing parts).  also, `all()` forces you to start at part=1, and `filter()` forces you to get the entire set of results or build your own paging logic (I'm less sure about how filter is implemented, but I can't find its source...)

kislyuk: Sounds good.

dss/blobstore/s3.py
```diff
@@ -186,3 +186,46 @@ def copy(
...
+            self,
+            bucket: str,
+            key: str,
+            upload_id: str,
+            part_count: int,
+            search_start: int=1,
+            return_count: int=1) -> typing.Sequence[int]:
+        """
+        Given a `bucket`, `key`, and `upload_id`, find the next N missing parts of a multipart upload, where
+        N=`return_count`.  If `start_search` is provided, start the search at part M, where M=`search_start`.
```
kislyuk: s/start_search/search_start/

Also, `part_count` and `return_count` are not documented, is that intentional?

Also, if the function gets called with part_count < search_start (and there are not enough missing parts), it will loop forever. Suggest adding an assertion that search_start <= part_count.

ttung: Will fix.

`part_count` < `search_start` is indeed an illegal state.  will fix.

ttung: btw, `return_count` is documented, so just `part_count`.

PR #302, opened 2017-08-02
Subscription notification test and fix

mikebaumann:
Added HTTP server to verify subscription notifications for matched documents.

Corrected notification payload content type to be application/json.

### DIFF
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -263,19 +263,19 @@ class TestIndexer(unittest.TestCase, DSSAsserts, StorageTestSupport):
             json_request_body=dict(
                 query=query,
                 callback_url=callback_url),
-            headers=self._get_auth_header()
+            headers=self.get_auth_header()
         )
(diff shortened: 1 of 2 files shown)


tests/test_indexer.py
```diff
@@ -185,6 +253,32 @@ def load_test_data_bundle(self, bundle: S3TestBundle):
...
+            requests.codes.created,
+            json_request_body=dict(
+                query=query,
+                callback_url=callback_url),
+            headers=self._get_auth_header()
+        )
+        uuid_ = resp_obj.json['uuid']
+        return uuid_
+
+    def _get_auth_header(self, real_header=True):
```
kislyuk: IMHO "real" here is loose nomenclature. I'd make it more explicit with `get_auth_header(token=None)`. If token is None, then a real token will be generated. That way I can pass a fake token easily and it will be more obvious to the reader what I'm doing.

PR #306, opened 2017-08-02
Add a `get_all_metadata()` method to retrieve all AWS metadata for an object

ttung:


dss/blobstore/s3.py
```diff
@@ -108,30 +108,43 @@ def get(self, bucket: str, object_name: str) -> bytes:
...
+    ) -> str:
+        """
+        Retrieves the cloud-provided checksum for a given object in a given bucket.
+        :param bucket: the bucket the object resides in.
+        :param object_name: the name of the object for which checksum is being retrieved.
+        :return: the cloud-provided checksum
+        """
+        response = self.get_all_metadata(bucket, object_name)
+        # hilariously, the ETag is quoted.  Unclear why.
+        return response['ETag'][1:-1]
```
kislyuk: Consider using `response.strip('"')`

ttung: That's a great idea.

ttung: Updated.

PR #310, opened 2017-08-03
Run unit tests in parallel.

ttung:
Create test targets for all the tests.  Then created a fast_test target that includes all of them.

Now you can run make -j fast_test and they are run in parallel and so so fast.


kislyuk:
Ooh fancy!

PR #311, opened 2017-08-03
Implement AWS multipart copy as a chunked task.

ttung:
Added a simple unit test that does a multipart copy and verifies that the output etag matches.


kislyuk:
Does this need rebasing? Appears to have a bunch of extra stuff in it.

ttung:
rebased.

PR #314, opened 2017-08-04
Add a work_complete_callback() for runtimes to be notified when work is complete.

ttung:
This way, the AWS runtime can log a message indicating that its work is complete.


kislyuk:
Tests failed, please take a look.

ttung:
#320 changed the message format so it's versioned.

I won't land this until that's approved and can land so master won't break.

PR #315, opened 2017-08-04
Add a simple logging facility.

ttung:
- Logs to cloudwatch.
- Creates the group and stream, if necessary.
- Retrieves the uploadSequenceToken to log to existing streams.
- Adds test to verify most of the functionality.  The key functionality that is not tested is the creation of a logging group.  Rationale: There's a hard limit on the number of test groups.  If this test goes wild, it might do bad things to the account.

### DIFF
--- a/dss/util/aws/__init__.py
+++ b/dss/util/aws/__init__.py
@@ -1,6 +1,6 @@
 import json
 import botocore
-from . import resources, clients
+from . import clients, resources, logging
 
 class ARN:
     fields = "arn partition service region account_id resource".split()
--- a/dss/util/aws/logging.py
+++ b/dss/util/aws/logging.py
@@ -1,0 +1,56 @@
+import time
+
+import boto3
+import botocore.exceptions
+
+
+def log_message(log_group_name: str, log_stream_name: str, message: str):
+    """Logs a message to cloudwatch."""
+
+    logs_client = boto3.client("logs")
+
+    # An astute reader might notice that the exception handling is a bit wonky.  This is because exception classes for
+    # boto are dynamically created, and we cannot actually catch them. :(
+
+    def get_sequence_token():
+        # try to get the upload sequence token
+        paginator = logs_client.get_paginator('describe_log_streams')
+        for page in paginator.paginate(logGroupName=log_group_name, logStreamNamePrefix=log_stream_name):
+            for log_stream in page['logStreams']:
+                if log_stream['logStreamName'] == log_stream_name:
+                    return log_stream.get('uploadSequenceToken', None)
+
+        return None
+
+    while True:
+        try:
+            logs_client.create_log_group(logGroupName=log_group_name)
+        except botocore.exceptions.ClientError as ex:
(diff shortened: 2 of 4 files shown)


ttung:
I looked at watchtower, and it's a lot more than I need.  I think the only mode that makes sense to use it in independent from python's logging module (to avoid clobbering the logging format for the entire app) and in synchronous mode (so we don't have to manage the `atexit` logic).

kislyuk:
I have two concerns:
- What's the rationale for not using the stdlib logging module? The logging format can be adjusted on a per-logger basis.
- What's the rationale for not using threaded/asynchronous logging? I'm concerned about the synchronous RTT delays introduced by the per-message overhead.

If you don't want to plug in to the complexity of logging/threading/atexit, then maybe this will help: simply using `print()` in a lambda results in an asynchronous delivery of a cloudwatch logs message to an `/aws/lambda/<fn>` log group (that log delivery is managed by the lambda host). Is that not sufficient?

ttung:
> What's the rationale for not using the stdlib logging module? The logging format can be adjusted on a per-logger basis.

The only thing I see for adjusting the logging format is https://docs.python.org/3/library/logging.html#logging.basicConfig, and that's a module-level method.  Is there something else?

> What's the rationale for not using threaded/asynchronous logging? I'm concerned about the synchronous RTT delays introduced by the per-message overhead.

I agree that this is not great. :(

> If you don't want to plug in to the complexity of logging/threading/atexit, then maybe this will help: simply using print() in a lambda results in an asynchronous delivery of a cloudwatch logs message to an /aws/lambda/<fn> log group (that log delivery is managed by the lambda host). Is that not sufficient?

My reluctance for using this approach is that it's pretty hard to trace.  I found a `None` getting printed for each of my lambda invocations.  It took me forever to trace it down to a print in one of the libraries we were pulling in (chalice or domovoi).  I considered doing a logging shim that just annotates the log messages so we can ignore the messages that lack the annotation and just treat those as noise.

The second reason is that approach makes it difficult to aggregate logs for a logical concept (i.e., chunked_worker or sync) that may span lambdas.

kislyuk:
> a module-level method. Is there something else?

This is slightly messier than it would ideally be, but it works:
```
import logging
logging.basicConfig(level=logging.INFO)
logger.propagate = False
handler = logging.StreamHandler()
handler.setFormatter(logging.Formatter('DSS_LOG:%(name)s:%(pathname)s:%(lineno)s:%(levelname)s: %(message)s'))
logger.addHandler(handler)

logger.info("log log log")
```

> I considered doing a logging shim that just annotates the log messages so we can ignore the messages that lack the annotation and just treat those as noise.

I prefer that approach.

kislyuk:
(With my example above, you get a log channel ID, logger ID, file and line number, so identifying the message should be very unambiguous. I agree it would be ideal to control the log group name as well, but I don't see a way to properly do that without threads/buffers/atexit).

ttung:
Abandoning in favor of watchtower.

PR #316, opened 2017-08-04
Fix intermittent test failure related to file blob removal

mikebaumann:
The indexer test test_indexed_file_access_error formerly removed
the a file blob associated with a test bundle to verify the indexer
handles this case as expected. However, the file blob was for a commonly used
test file, and intermittent test failures were occurring when run using a shared test bucket.
The test was changed to instead create a bundle manifest entry such that the file blob
reference is invalid. The code and test semantics are the same.

### DIFF
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -44,7 +44,7 @@ from tests.infra import DSSAsserts, StorageTestSupport, S3TestBundle, start_verb
 # The moto mock has two defects that show up when used by the dss core storage system.
 # Use actual S3 until these defects are fixed in moto.
 # TODO (mbaumann) When the defects in moto have been fixed, remove True from the line below.
-USE_AWS_S3 = bool(os.environ.get("USE_AWS_S3", True))
+USE_AWS_S3 = bool(os.environ.get('USE_AWS_S3', True))
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -53,7 +53,7 @@ logger.setLevel(logging.INFO)
 
 start_verbose_logging()
 for logger_name in logging.Logger.manager.loggerDict:  # type: ignore
-    if logger_name.startswith("elasticsearch"):
+    if logger_name.startswith('elasticsearch'):
         logging.getLogger(logger_name).setLevel(logging.WARNING)
 
 
@@ -72,7 +72,7 @@ for logger_name in logging.Logger.manager.loggerDict:  # type: ignore
 
 class TestIndexer(unittest.TestCase, DSSAsserts, StorageTestSupport):
 
-    http_server_address = "127.0.0.1"
+    http_server_address = '127.0.0.1'
     http_server_port = 8729
 
     @classmethod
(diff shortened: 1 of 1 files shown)


tests/test_indexer.py
```diff
@@ -327,6 +347,65 @@ def get_search_results(cls, query, expected_hit_count):
...
+    def get_bundle_id(self):
+        return f'{self.bundle_id}.{self.bundle_version}'
+
+    def add_file(self, bucket_name, name, indexed, file_id):
+        # Add the existing file to the bundle manifest
+        s3 = boto3.resource("s3")
+        file_manifest_string = s3.Object(bucket_name=bucket_name,
+                                         key=f'files/{file_id}').get()['Body'].read().decode('utf-8')
+        file_manifest = json.loads(file_manifest_string, encoding="utf-8")
+        file_uuid, _, file_version = file_id.partition(".")
```
ttung: `file_uuid, file_version = file_id.split(".")`

tests/test_indexer.py
```diff
@@ -152,9 +155,9 @@ def test_indexed_file_unparsable(self):
                                                          files=smartseq2_paried_ends_indexed_file_list)
 
     def test_indexed_file_access_error(self):
-        bundle_key = self.load_test_data_bundle_for_path('fixtures/smartseq2/paired_ends')
-        filename = "project.json"
-        deleteFileBlob(bundle_key, filename)
+        filename = "inaccessible_file.json"
```
ttung: rename this var to inaccessible_file_name or something like that.

tests/test_indexer.py
```diff
@@ -327,6 +347,65 @@ def get_search_results(cls, query, expected_hit_count):
                 time.sleep(0.5)
 
 
+class BundleBuilder:
```
ttung: docblock

tests/test_indexer.py
mikebaumann: This doesn't seem to work due to two variables on the left side and three substrings from splitting on '.'
```
file_uuid, file_version = '6011caee-ebde-4b2b-82f1-bb75d303fa41.2017-08-07T150159.427496Z'.split(".")
  File "<input>", line 1, in <module>
ValueError: too many values to unpack (expected 2)
```
Although it works if `maxsplit=1` is specified:
```
file_uuid, file_version = '6011caee-ebde-4b2b-82f1-bb75d303fa41.2017-08-07T150159.427496Z'.split(".", maxsplit=1)
```
@kislyuk used partition in https://github.com/HumanCellAtlas/data-store/pull/291 so I used it here as well.

tests/test_indexer.py
mikebaumann: Yes, that is a better name. Done.

tests/test_indexer.py
mikebaumann: Added a class docstring. It is brief, given this is a test class used within a single test file.

mikebaumann:
As requested in the summary comment, I also changed the use of quotes as string delimiters to be consistent. This test file now uses single quote consistently as the string delimiter. Please let me know if we have more specific conventions regarding the use of single or double quote as string delimiters.

tests/test_indexer.py
ttung: Ahh. `file_uuid, file_version = file_id.split(".", 1)` should work, but I'll leave it to your discretion.

ttung:
> As requested in the summary comment, I also changed the use of quotes as string delimiters to be consistent. This test file now uses single quote consistently as the string delimiter. Please let me know if we have more specific conventions regarding the use of single or double quote as string delimiters.

I don't recall where we landed on what used single or double quotes, but I believe it was dictionary string keys = single quotes, other strings = double quotes.

The safe way to make this change is to do a predecessor diff that fixes the style of the existing code, and then your changes stacked on top of that.

tests/test_indexer.py
```diff
@@ -327,6 +347,72 @@ def get_search_results(cls, query, expected_hit_count):
...
+    """
+    Utility class for creating a new bundle for testing purposes.
```
ttung: `Creating a new bundle manifest`

tests/test_indexer.py
ttung: IMO, we should consider making this a general purpose class.  bundle.py would probably benefit from it.

kislyuk:
Closing because this was merged as #329. Please reopen if you need this for any reason.

PR #317, opened 2017-08-05
Add hca service accounts to allowed email domains for cli testing

Mackey22:
When testing python bindings and cli in Travis, I get Forbidden errors because the requests are coming from an unauthorized domain. This will make testing possible.


kislyuk:
This should probably be stage (deployment)-specific in the future

PR #320, opened 2017-08-07
Clean up task scheduling and logging.

ttung:
1. Task scheduling happens through a `schedule_task()` method.
2. Work scheduling uses a versioned message now.  Therefore, it's future-proofed against message format changes.
3. Log each key event in the task lifecycle into a unique CloudWatch stream.
4. Updated test to watch these logged events rather than some hacky print statement at the end of a job.
5. Use watchtower to log messages.

### DIFF
--- a/dss/events/chunkedtask/aws.py
+++ b/dss/events/chunkedtask/aws.py
@@ -1,4 +1,10 @@
+import json
 import logging
+import os
+import typing
+import uuid
+
+import boto3
 
 from . import _awsimpl, awsconstants
 from .runner import Runner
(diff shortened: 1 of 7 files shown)


dss/events/chunkedtask/aws.py
```diff
@@ -12,23 +18,98 @@
...
+        awsconstants.REQUEST_VERSION_KEY: awsconstants.CURRENT_VERSION,
+        awsconstants.TASK_ID_KEY: task_id,
+        awsconstants.STATE_KEY: state,
+    }
+
+    sts_client = boto3.client('sts')
+    accountid = sts_client.get_caller_identity()['Account']
+
+    sns_client = boto3.client('sns')
+    region = os.environ['AWS_DEFAULT_REGION']
```
kislyuk: I think `boto3.Session().region_name` is preferable. The env var is not always set, but that snippet will always follow the full boto3 config ingestion protocol.

ttung: Fixed.

Issue #318, opened 2017-08-07
Split chunkedtask's aws test task / runners to its own file


ttung:
See https://github.com/HumanCellAtlas/data-store/pull/312#discussion_r131244430

PR #321, opened 2017-08-07
[easy] Move test task and runtime to its own file.

ttung:
Addresses #318

### DIFF
--- a/dss/events/chunkedtask/_awstest.py
+++ b/dss/events/chunkedtask/_awstest.py
@@ -1,0 +1,56 @@
+import itertools
+import typing
+
+from ._awsimpl import AWSRuntime
+from .constants import TIME_OVERHEAD_FACTOR
+from .base import Task
+
+# The rest of this file is for unit tests.  The reason they are in this file is because they need to be deployed to the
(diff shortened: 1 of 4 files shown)


dss/events/chunkedtask/_awstest.py
```diff
@@ -0,0 +1,56 @@
...
+    spuriously break at some future point in time if lambda lifetimes become longer.
+
+    This should only be used for the fast test.
+    """
+    def __init__(self, context, task_id: str) -> None:
+        super().__init__(context, AWS_FAST_TEST_CLIENT_NAME, task_id)
+        self.time_remaining_iterator = itertools.chain(
+            [int(AWS_FAST_TEST_EST_TIME_MS * TIME_OVERHEAD_FACTOR) + 1],
+            itertools.repeat(0)
+        )
```
kislyuk: Aside: this can probably be rewritten as a comprehension without the use of itertools

ttung: I know there are wonky ways to construct infinite generators, but those tend to rely on esoteric knowledge of python.

Do you have some other approach?

PR #322, opened 2017-08-08
[easy] Refactor out the code to write out the file metadata.

ttung:
This is required because for async copy, this will happen separately.

This partially addresses #136

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -11,7 +11,7 @@ from flask import jsonify, make_response, redirect, request
 from werkzeug.exceptions import BadRequest
 
 from .. import DSSException, dss_handler
-from ..blobstore import BlobNotFoundError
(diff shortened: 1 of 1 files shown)


dss/api/files.py
```diff
@@ -186,10 +186,30 @@ def put(uuid: str, extras: dict, version: str=None):
         FileMetadata.SHA256: metadata['hca-dss-sha256'],
     })
 
+    write_file_metadata(handle, dst_bucket, uuid, version, document)
```
ttung: > Also, I'm slightly confused that you're not removing corresponding code from elsewhere. Where does the metadata write happen now?

the call is here, @kislyuk 

the diff here is not the best way to represent this code movement, unfortunately. :(

ttung:
> Worth writing a unit test for this function?

It is covered by the tests/test_file.py tests, I believe.

PR #326, opened 2017-08-08
endpoint to serve the application secret.

ttung:
This is the worst-kept secret!


tests/test_application_secrets.py
```diff
@@ -0,0 +1,32 @@
...
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+import dss
+from tests.infra import DSSAsserts, start_verbose_logging
+
+start_verbose_logging()
+
+
+class TestFileApi(unittest.TestCase, DSSAsserts):
```
Mackey22: Class name?

ttung: LOL that's embarrassing.

Mackey22:
Hmm build failed for some reason. Other than that lgtm

kislyuk:
IMHO it's better not to declare this as part of the official API and instead shim it in the same way `/` is shimmed in here: https://github.com/HumanCellAtlas/data-store/blob/3bbfb8c79b7005b2b0f64d05b2c36482f0f450f2/chalice/app.py#L52-L56

PR #327, opened 2017-08-08
PUT /files should use random payload

ttung:
This is to ensure that we exercise the file copy path.

To ensure that we exercise the identical-data path, we upload the same data twice.

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -4,6 +4,7 @@
 import hashlib
 import os
 import sys
+import tempfile
 import unittest
 import uuid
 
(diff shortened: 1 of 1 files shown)


tests/test_file.py
```diff
@@ -24,30 +27,43 @@ def setUp(self):
...
-            }
-        )
-        self.assertIn('version', resp_obj.json)
+        tempdir = tempfile.gettempdir()
+        self._test_file_put("s3", self.s3_test_bucket, S3Uploader(tempdir, self.s3_test_bucket))
+        self._test_file_put("gs", self.gs_test_bucket, GSUploader(tempdir, self.gs_test_bucket))
+
+    def _test_file_put(self, scheme: str, test_bucket: str, uploader: Uploader):
+        src_key = generate_test_key()
+        src_data = ("".join(chr(random.randint(0, 255)) for ix in range(1024))).encode("latin-1")
```
kislyuk: Fancy, why not just os.urandom?

ttung: Switched to `os.urandom(..)`

PR #332, opened 2017-08-10
Start up an elasticsearch instance per test instance.

ttung:
This way the test does not rely on any global state.  Adds a new required argument (only for tests) that DSS_TEST_ES_PATH be set to a valid elasticsearch binary.

### DIFF
--- a/dss/util/es.py
+++ b/dss/util/es.py
@@ -1,6 +1,12 @@
 import json
 import os
-from urllib.parse import SplitResult, parse_qs, urlencode, urlparse, urlunsplit
+import random
+import socket
+import subprocess
+import tempfile
+import time
+import typing
+from urllib.parse import parse_qs, urlencode, urlparse
 
 import boto3
 from botocore.auth import SigV4Auth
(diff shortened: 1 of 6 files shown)


dss/util/es.py
```diff
@@ -34,28 +39,68 @@ def __call__(self, r):
         return r
 
 
+class ElasticsearchServer:
+    def __init__(self, startup_timeout_seconds: int=30) -> None:
+        elasticsearch_binary = os.getenv("DSS_TEST_ES_PATH")
+        self.tempdir = tempfile.TemporaryDirectory()
+        while True:
+            port = random.randint(1024, 65535)
```
sampierson: If you're using random port IDs It would be a good idea, once you have picked a new port number, to connect to it and ensure you get and ECONNREFUSED before attempting to use it.

ttung: Good idea, done.

ttung:
Ok sorry, this is really ready for review now.  I was just lucky in my initial tests.  I believe I've removed all the cross-test interactions (mostly ES nodes talking to each other).

dss/util/es.py
```diff
@@ -34,36 +40,99 @@ def __call__(self, r):
...
+                continue
+
+            proc = subprocess.Popen(
+                [
+                    elasticsearch_binary,
+                    "-E", f"http.port={port}",
+                    "-E", f"transport.tcp.port={transport_port}",
+                    "-E", f"path.data={self.tempdir.name}",
+                ],
+            )
```
kislyuk: Here is a cute way to make this a bit simpler:

`elasticsearch -E http.port=0 -E transport.tcp.port=0 -E path.data=$(mktemp -d)`

0 means let the kernel allocate the port. But I couldn't find a nice way to read back the port number, short of parsing the log.

PR #336, opened 2017-08-14
Test to expose a bug and a hilarious bugfix.

ttung:
1. next_part is not advanced.
2. there's an off-by-one where if we run out of items in the work queue and we haven't copied the last chunk, then we don't search for more uncopied parts.

### DIFF
--- a/dss/events/chunkedtask/awscopyclient.py
+++ b/dss/events/chunkedtask/awscopyclient.py
@@ -27,7 +27,7 @@ class S3CopyTask(Task[dict, typing.Any]):
     """
     This is a chunked task that does a multipart copy from one blob to another.
     """
-    def __init__(self, state: dict) -> None:
+    def __init__(self, state: dict, fetch_size: int=100) -> None:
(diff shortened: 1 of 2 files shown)


dss/events/chunkedtask/awscopyclient.py
```diff
@@ -146,6 +148,7 @@ def run_one_unit(self) -> typing.Optional[typing.Any]:
             PartNumber=part_id,
             UploadId=self.upload_id,
         )
+        self.next_part = part_id + 1
```
ttung: where it's not advanced.

dss/events/chunkedtask/awscopyclient.py
```diff
@@ -103,9 +105,9 @@ def run_one_unit(self) -> typing.Optional[typing.Any]:
                     self.upload_id,
                     self.part_count,
                     self.next_part,
-                    100))
+                    self.fetch_size))
 
-        if len(self.queue) == 0 or self.next_part == self.part_count:
+        if len(self.queue) == 0 or self.next_part > self.part_count:
```
ttung: off-by-one.

PR #338, opened 2017-08-14
Sleep per loop while we search for the completion marker.

ttung:
Don't hammer filterLogEvents.


ttung:
I'm going to land this to make tests more reliable.  Feel free to make post-land comments.

PR #344, opened 2017-08-14
[easy] Move the clients list setup inside a method.

ttung:
This is so that circular module dependencies can resolve.


ttung:
> Unrelated to this change, get_payload in this file is probably better named unpack_payload or parse_payload.

👍

PR #347, opened 2017-08-14
Update PUT /files to be async for AWS.

ttung:
1. PUT /files for AWS will now spawn a lambda to execute the copy.
2. tests are uploaded to use the `upload_file()` primitive that waits for completion.


kislyuk:
Please check the Travis failure.

kislyuk:
If the file is under 64 MB, can we just copy it in process and return 200 instead of 202?

ttung:
1. Fixed travis failure.
2. Small files uploaded synchronously in #362

kislyuk:
Why not make `upload_file_wait` a method on a unittest.TestCase/DSSAsserts mixin? It would make it a bit clearer why you're passing the instance as a first arg.

PR #360, opened 2017-08-16
Make the copy type explicit.

ttung:
This will make the code to copy small objects synchronously cleaner.

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -3,6 +3,7 @@ import io
 import json
 import re
 import typing
+from enum import Enum, auto
 
 import iso8601
 import requests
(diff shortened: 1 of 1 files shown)


dss/api/files.py
```diff
@@ -96,6 +97,11 @@ def get_helper(uuid: str, replica: typing.Optional[str]=None, version: str=None)
 
 @dss_handler
 def put(uuid: str, extras: dict, version: str=None):
+    class CopyMode(Enum):
+        NO_COPY = auto()
+        COPY_INLINE = auto()
+        COPY_ASYNC = auto()
```
kislyuk: You're not a fan of the more concise `CopyMode = Enum("CopyMode", "NO_COPY COPY_INLINE COPY_ASYNC")?

ttung: PyCharm doesn't seem to deal with that...

PR #362, opened 2017-08-17
Upload small files synchronously.

ttung:
1. Files smaller or equal to `dss.api.files.SYNC_COPY_LIMIT` will be copied synchronously.
2. Added a flag to `upload_file_wait` indicating whether we should expect a synchronous copy, an asynchronous copy, or if we don't care.
3. Added a test case to upload a file larger than `dss.api.files.SYNC_COPY_LIMIT` and validate whether it was done synchronously.

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -13,11 +13,12 @@ from werkzeug.exceptions import BadRequest
 
 from .. import DSSException, dss_handler
 from ..blobstore import BlobAlreadyExistsError, BlobNotFoundError, BlobStore
+from ..blobstore.s3 import S3BlobStore
 from ..config import Config
 from ..events.chunkedtask import aws
 from ..events.chunkedtask import awscopyclient
 from ..hcablobstore import FileMetadata, HCABlobStore
-from ..util.aws import get_s3_chunk_size
+from ..util.aws import get_s3_chunk_size, AWS_MIN_CHUNK_SIZE
 
 
 @dss_handler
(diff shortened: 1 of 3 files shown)


dss/api/files.py
```diff
@@ -13,12 +13,16 @@
...
 from ..blobstore import BlobAlreadyExistsError, BlobNotFoundError, BlobStore
+from ..blobstore.s3 import S3BlobStore
 from ..config import Config
 from ..events.chunkedtask import aws
 from ..events.chunkedtask import awscopyclient
 from ..hcablobstore import FileMetadata, HCABlobStore
 from ..util.aws import get_s3_chunk_size
 
+SYNC_COPY_LIMIT = 64 * 1024 * 1024
+"""Files small than this size are copied synchronously in the PUT /files handler.."""
```
kislyuk: s/.././

dss/api/files.py
```diff
@@ -13,12 +13,16 @@
...
 from .. import DSSException, dss_handler
 from ..blobstore import BlobAlreadyExistsError, BlobNotFoundError, BlobStore
+from ..blobstore.s3 import S3BlobStore
 from ..config import Config
 from ..events.chunkedtask import aws
 from ..events.chunkedtask import awscopyclient
 from ..hcablobstore import FileMetadata, HCABlobStore
 from ..util.aws import get_s3_chunk_size
 
+SYNC_COPY_LIMIT = 64 * 1024 * 1024
```
kislyuk: Consider parameterizing this by baseline part size

ttung: #368

PR #363, opened 2017-08-17
update documentation for running tests.

ttung:


kislyuk:
Can we set the default to `elasticsearch` in dss/util/es if the environment variable is not set?

PR #366, opened 2017-08-17
[easy] Rename {test-,}awscopyclient to {test-,}s3copyclient

ttung:
It's a misnomer.  AWSRuntime actually makes sense since it's tied to AWS.  S3CopyTask should theoretically be able to run anywhere (and it does, in the unit tests).


ttung:
> I'm assuming you're still working on gscopyclient? 😁

Yeah, much struggling (but I think I'm mostly there).

PR #369, opened 2017-08-21
Add SNS permissions for DSS API Server lambda

kislyuk:


iam/policy-templates/dss-lambda.json
```diff
@@ -28,6 +28,13 @@
     },
     {
       "Effect": "Allow",
+      "Action": "sns:Publish",
+      "Resource": [
+        "arn:aws:sns:*:$account_id:dss*"
```
ttung: can we make this dss-*-$stage?

kislyuk: Done and also updated for the other lambdas' policies

PR #374, opened 2017-08-22
Search route fixes and improvements

mikebaumann:
Added basic pagination support to include all matched search results.
Corrected result format for bundle_url.
Updated search route to use current conventions for route handlers.
Improved search exception handling and tests.


ttung:
This PR is far too large to review properly.  You should consider breaking this up into individual changes, each with a clear thesis.

mikebaumann:
I will work on dividing this, and contact you directly for any questions.
Currently the amount of production code changed by this PR is three small methods in one file (search.py), and the rest is new unit tests for this same functionality. Yet, it could/should be divided to enable easier, focused reviews. Thanks.

tests/test_search.py
```diff
@@ -1,67 +1,200 @@
...
     @classmethod
     def setUpClass(cls):
         cls.es_server = ElasticsearchServer()
         os.environ['DSS_ES_PORT'] = str(cls.es_server.port)
+        dss.Config.set_config(dss.DeploymentStage.TEST)
+        cls.app = dss.create_app().app.test_client()
+        cls.index_document = cls.load_sample_document("sample_index_doc.json")
+
+    @classmethod
+    def load_sample_document(cls, file_name):
```
ttung: why is this a separate method?

tests/test_search.py
```diff
@@ -1,67 +1,200 @@
...
+                        "match": ["SomethingInvalid", "xxx"]
+                    }
+                },
+                requests.codes.internal_server_error
+            )
+        ]
+
+        self.populate_search_index(self.index_document, 1)
+        for query_data in invalid_query_data:
+            with self.subTest("Invalid Queries"):
```
ttung: what's the point of this self.subTest?

if you're going to do this, why not:

```
with self.subTest("Invalid Queries", bad_query=query_data):
  do_bad_things_with_bad_query(..)
```

tests/test_search.py
```diff
@@ -1,67 +1,200 @@
...
+        try:
+            os.environ['DSS_ES_ENDPOINT'] = "bogus"
+            response = self.assertPostResponse(
+                "/v1/search",
+                json_request_body=(self.smartseq2_paired_ends_query),
+                expected_code=requests.codes.internal_server_error)
+            self.assertEqual(response.json['code'], "elasticsearch_error")
+            self.assertEqual(response.json['title'], "Elasticsearch operation failed")
+        finally:
+            os.environ.pop('DSS_ES_ENDPOINT')
```
ttung: this does not restore the state.  prefer that you save and restore that env var.

tests/test_search.py
```diff
@@ -1,67 +1,200 @@
...
+        self.populate_search_index(self.index_document, 1)
+        for query_data in invalid_query_data:
+            with self.subTest("Invalid Queries"):
+                self.assertPostResponse(
+                    "/v1/search",
+                    json_request_body=(query_data[0]),
+                    expected_code=query_data[1])
+
+    def test_search_returns_X_results_when_X_documents_match_query(self):
+        test_matches = [0, 1, 9, 10, 11, 1000, 5000]
```
ttung: this needs a concise explanation.

tests/test_search.py
```diff
@@ -1,67 +1,200 @@
...
     @classmethod
     def tearDownClass(cls):
         cls.es_server.shutdown()
 
     def setUp(self):
-        dss.Config.set_config(dss.DeploymentStage.TEST)
-        self.app = dss.create_app().app.test_client()
-
-        create_elasticsearch_index(logging.getLogger(__name__))
+        elasticsearch_delete_index("_all")
```
ttung: can these two statements race?  or is `elasticsearch_delete_index` a synchronous operation?

tests/test_search.py
```diff
@@ -1,67 +1,200 @@
...
+    def verify_search_results(self, query, expected_result_length=0, bundles=[]):
+        timeout = 5
+        timeout_time = timeout + time.time()
+        while True:
+            response = self.assertPostResponse(
+                "/v1/search",
+                json_request_body=(query),
+                expected_code=requests.codes.ok)
+            search_response = response.json
+            if len(search_response['results']) >= expected_result_length:
```
ttung: is it an error if the number of results is greater than the expected result length?

tests/test_search.py
```diff
@@ -1,67 +1,200 @@
...
+                            doc_type=DSS_ELASTICSEARCH_DOC_TYPE,
+                            id=bundle_id,
+                            body=index_document)
+            bundles.append((bundle_id, bundle_url))
+        return bundles
+
+    def verify_search_results(self, query, expected_result_length=0, bundles=[]):
+        timeout = 5
+        timeout_time = timeout + time.time()
+        while True:
```
ttung: you can restructure this as a while-else statement:

```
while time.time() < timeout_time:
  do things
  if success_condition:
    break
else:
  self.fail("elasticsearch failed to return all results")
```

dss/api/search.py
```diff
@@ -1,47 +1,53 @@
...
-    for hit in response:
-        result = {
-            'bundle_id': hit.meta.id,
-            'bundle_url': bundles_url_base + hit.meta.id,
-            'search_score': hit.meta.score
-        }
-        result_list.append(result)
+    result_list = [{
+        'bundle_id': hit.meta.id,
+        'bundle_url': bundles_url_base + hit.meta.id.replace(".", "?version=", 1),
```
Bento007: I'm not sure what you're asking.The meta.id represents the full bundle Id and the first dot is the separator between the UUID and the version. The dot in the version is retained as you show above.

tests/infra.py
```diff
@@ -21,10 +21,10 @@
 def start_verbose_logging():
     logging.basicConfig(level=logging.INFO)
     for logger_name in logging.Logger.manager.loggerDict:  # type: ignore
-        if logger_name.startswith("botocore") or logger_name.startswith("boto3.resources"):
+        if logger_name.startswith("botocore") or logger_name.startswith("boto3.resources")\
```
Bento007: I'll add it.

tests/test_search.py
Bento007: I'll add it.

tests/test_search.py
Bento007: I'll add one.

tests/test_search.py
Bento007: I'll fix that.

tests/test_search.py
Bento007: I'll fix it

tests/test_search.py
Bento007: i'll add an additional error condition.

dss/api/search.py
```diff
@@ -1,47 +1,53 @@
...
+    except ElasticsearchDslException:
+        raise DSSException(requests.codes.bad_request,
+                           "elasticsearch_query_error",
+                           "Invalid query")
+    except ElasticsearchException:
+        raise DSSException(requests.codes.internal_server_error,
+                           "elasticsearch_error",
+                           "Elasticsearch operation failed")
+
+def format_results(request, search_obj):
```
mikebaumann: Now that `find()` is gone, I have inlined `format_results()`.

tests/__init__.py
```diff
@@ -0,0 +1,28 @@
+import datetime
+
+
+class IndexSearchTestSupport:
```
mikebaumann: As we identified test data/functions used for both index and search tests, we wanted to put it in a common location, and (right or wrong) selected tests/__init__.py. At the time, putting it in IndexSearchTestSupport seemed a useful way to organize and access it. Yet, it would be simpler and more standard to just make them module variables/functions.
I will do that.

tests/test_search.py
mikebaumann: At one point we considered loading multiple, different index documents to test with, and in that case this method made sense. Presently, however, we only load one so I have inlined this method.

tests/test_search.py
Bento007: Since both create and delete index operations must wait for the primary shard within elasticsearch, the operations should happen synchronously.

https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-index.html
https://discuss.elastic.co/t/race-condition-when-removing-indexing-and-counting-data-by-using-the-java-api/2190

dss/api/search.py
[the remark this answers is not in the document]
ttung: Oh, I missed the maxreplace arg. My bad!

tests/__init__.py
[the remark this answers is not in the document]
ttung: I think this is a good thing to have, but I would:

1. move it to a different file.
2. consider making it XYZMixin (I plan on renaming DSSAsserts and friends to have a Mixin suffix)

tests/test_search.py
ttung: I think methods with only a single caller are fine if:

1. There is a high probability that someone else will call it.
2. The method is long and encapsulates a clearly defined purpose.

Use your discretion. :)

tests/test_search.py
```diff
@@ -1,60 +1,205 @@
...
+                        body=self.index_document,
+                        refresh=True)
+
+        response = self.assertPostResponse(
+            "/v1/search",
+            json_request_body=smartseq2_paired_ends_query,
+            expected_code=requests.codes.ok)
+        search_response = response.json
+        self.assertDictEqual(search_response['query'], smartseq2_paired_ends_query)
+        self.assertEqual(len(search_response), 2)
```
ttung: what's the point of this assert?

is it possible `search_response` has more fields in the future?

tests/test_search.py
```diff
@@ -1,60 +1,205 @@
...
+        self.populate_search_index(self.index_document, 1)
+        for query_data in invalid_query_data:
+            with self.subTest("Invalid Queries", bad_query=query_data):
+                self.assertPostResponse(
+                    "/v1/search",
+                    json_request_body=(query_data[0]),
+                    expected_code=query_data[1])
+
+    def test_search_returns_X_results_when_X_documents_match_query(self):
+        """The number of documents indexed is equal to the number of search results returned.
```
ttung: Suggestion:

```
This test creates N identical documents.  It executes a search query that matches the document and verifies all the documents are returned.  N is varied across a variety of values.
```

Also:

1. What's the performance impact of this test?  How long do unit tests take now?
2. Do we need all the cases?

tests/test_search.py
```diff
@@ -1,60 +1,205 @@
...
+
+            es_client.index(index=DSS_ELASTICSEARCH_INDEX_NAME,
+                            doc_type=DSS_ELASTICSEARCH_DOC_TYPE,
+                            id=bundle_id,
+                            body=index_document)
+            bundles.append((bundle_id, bundle_url))
+        return bundles
+
+    def verify_search_results(self, query, expected_result_length=0, bundles=[]):
+        timeout = 5
```
ttung: I'd make this an optional param.

tests/test_search.py
```diff
@@ -1,60 +1,205 @@
...
+                break
+            elif len(search_response['results']) > expected_result_length:
+                self.fail("elasticsearch more results than expected.")
+            else:
+                time.sleep(0.5)
+        else:
+            self.fail("elasticsearch failed to return all results.")
+
+        self.assertDictEqual(search_response['query'], query)
+        self.assertEqual(len(search_response), 2)
```
ttung: is this a requirement?

tests/test_search.py
Bento007: It will allows the tests to fail faster if the incorrect number of keys are not present. We can decided to include more fields in the future. If we add more fields in the future it will remind the tester to add that new field to the test cases.

tests/test_search.py
Bento007: 1. This test takes 24s to run. All unit tests take about 7min
2. All of the test cases are needed especially as we continue to implement pagination. 0 and 1 for edge cases. 9, 10, 11 also edge cases since our original search implementation only returned 10 results max. 1000 is the max number of results currently return per call to scan. 5000 to test if all results are returned when multiple calls to scan are required.

tests/test_search.py
Bento007: done

tests/test_search.py
Bento007: It's not a requirements written anywhere, but it is the current behavior of our code.

tests/test_search.py
ttung: If the correct behavior of this does not rely on the number of keys in search_response, I would leave this assertion out.

tests/test_search.py
ttung: Awesome, it's good to know this does not substantially increase the runtime of tests. :)

tests/__init__.py
```diff
@@ -0,0 +1,26 @@
+import datetime
+
+smartseq2_paired_ends_query = \
```
ttung: i still don't think this belongs here.

tests/test_search.py
```diff
@@ -1,60 +1,201 @@
...
     @classmethod
     def setUpClass(cls):
         cls.es_server = ElasticsearchServer()
         os.environ['DSS_ES_PORT'] = str(cls.es_server.port)
+        dss.Config.set_config(dss.DeploymentStage.TEST)
+        cls.app = dss.create_app().app.test_client()
+        with open(os.path.join(os.path.dirname(__file__), "sample_index_doc.json"), "r") as fh:
+            cls.index_document = json.load(fh)
+
+    @classmethod
```
ttung: remove.

tests/test_search.py
```diff
@@ -1,60 +1,201 @@
...
+                requests.codes.internal_server_error
+            )
+        ]
+
+        self.populate_search_index(self.index_document, 1)
+        for query_data in invalid_query_data:
+            with self.subTest("Invalid Queries", bad_query=query_data):
+                self.assertPostResponse(
+                    "/v1/search",
+                    json_request_body=(query_data[0]),
```
ttung: so if you do the `subTest` thing, i think you should be using `bad_query` inside the with statement.

or you could just use the for loop and no with statement.

tests/test_search.py
```diff
@@ -1,60 +1,201 @@
...
+                        id=bundle_id,
+                        body=self.index_document,
+                        refresh=True)
+
+        response = self.assertPostResponse(
+            "/v1/search",
+            json_request_body=smartseq2_paired_ends_query,
+            expected_code=requests.codes.ok)
+        search_response = response.json
+        self.assertEqual(len(search_response), 2)
```
ttung: Still think you should remove this.  It has no bearing on the fields you care about.

Bento007: removed

tests/test_search.py
Bento007: removed

tests/test_search.py
Bento007: Is this what you mean?
```python
       for bad_query in invalid_query_data:
            with self.subTest("Invalid Queries", bad_query=bad_query):
                self.assertPostResponse(
                    "/v1/search",
                    json_request_body=(bad_query[0]),
                    expected_code=bad_query[1])
```

tests/__init__.py
Bento007: where do we put reused variables, in infra.py?

tests/test_search.py
ttung: Well, you could do:
```
for query_data in invalid_query_data:
    with self.subTest("Invalid Queries", bad_query=query_data):
        self.assertPostResponse(
           "/v1/search",
           json_request_body=(bad_query[0]),
           expected_code=bad_query[1])
```

or you could just do:

```
for bad_query in invalid_query_data:
    self.assertPostResponse(
       "/v1/search",
       json_request_body=(bad_query[0]),
       expected_code=bad_query[1])
```

tests/__init__.py
ttung: I don't know.  search_queries.py?

I just know it doesn't belong in `__init__.py`

tests/test_search.py
Bento007: The first suggestion does not add bad_query into the scope. I also don't want to removed subtest, because that will ensure that all tests run if some fail.  I'll modified the subtest description and variable names. Other than that I don't see any benefit in modifying it further.

Bento007:
Changes are in.

tests/test_search.py
ttung: Ohh, I did not realize that.  Well, that's a bummer.

PR #375, opened 2017-08-23
Removed GET /search route

mikebaumann:
Removed the GET /search route, as requested by kislyuk.

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -1,5 +1,4 @@
 import json
-import os
 
 from elasticsearch_dsl import Search
 from flask import request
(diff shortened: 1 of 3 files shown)


dss/api/search.py
```diff
@@ -1,24 +1,19 @@
...
-    # TODO (mbaumann) Use a connection manager
-    es_client = ElasticsearchClient.get(get_logger())
-    query = json.loads(request.values["query"])
-    response = Search(using=es_client,
-                      index=DSS_ELASTICSEARCH_INDEX_NAME,
-                      doc_type=DSS_ELASTICSEARCH_DOC_TYPE).query("match", **query).execute()
-    return {"query": query, "results": format_results(request, response)}
+    raise DSSException(requests.codes.not_implemented,
+                       "simple_search_not_implemented",
+                       "Simple search is not implemented. Use POST /search with Elasticsearch query.")
```
kislyuk: Why keep this route? Why reference "simple search"? That's terminology internal to us. Please remove the route altogether.

mikebaumann: OK, thanks for the clarification. I'll remove the GET /search route all together.

PR #379, opened 2017-08-24
Manage wheels that Chalice can't find

kislyuk:
Chalice introduced a wheel-based dependency management system in https://github.com/aws/chalice/pull/388. Unfortunately, there is no way to make it fail if some wheels are not found, and no way to instruct it to grab wheels from a custom directory, so we have to pre-process the dependencies by calling the Chalice component responsible for this and analyzing its output.


scripts/find_missing_wheels.py
```diff
@@ -0,0 +1,38 @@
...
+args = parser.parse_args()
+
+with TemporaryDirectory() as td:
+    compat_wheels, missing_wheels = DependencyBuilder(OSUtils())._download_dependencies(td, args.pip_reqs)
+    need_wheels = [w for w in missing_wheels if not os.path.exists(os.path.join(args.wheel_dir, w.identifier))]
+    if need_wheels:
+        if args.build_wheels:
+            if platform.system() != "Linux":
+                parser.exit(f"{parser.prog}: Expected to run on a Linux system.")
+            os.makedirs(args.wheel_dir)
```
ttung: should we nuke the existing wheel_dir so we blow away wheels that are no longer required?

kislyuk: I like this idea. Updated.

PR #380, opened 2017-08-25
Change indexer to use blobstore abstraction

Bento007:
Change indexer to use blobstore abstractions and updated associated s3 test cases. Added stub for gcp file indexer.


dss/events/handlers/index.py
```diff
@@ -17,32 +15,52 @@
...
-            bundle_id = get_bundle_id_from_key(key)
-            index_data = create_index_data(s3, bucket_name, bundle_id, manifest, logger)
-            add_index_data_to_elasticsearch(bundle_id, index_data, logger)
-            subscriptions = find_matching_subscriptions(index_data, logger)
-            process_notifications(bundle_id, subscriptions, logger)
-            logger.debug(f"Finished index processing of S3 creation event for bundle: {key}")
-        else:
-            logger.debug(f"Not indexing S3 creation event for key: {key}")
+        bucket_name = event['Records'][0]["s3"]["bucket"]["name"]
+        blobstore = S3BlobStore()
```
ttung: I would encourage you to call `Config.get_cloud_specific_handles(..)`

dss/events/handlers/index.py
```diff
@@ -17,32 +15,52 @@
...
         raise
 
 
+def process_new_gs_indexable_object(event, logger) -> None:
+    try:
+        # This function is only called for GS creation events
+        # TODO retreive bucket_name and key from gs_event
+        bucket_name = None
+        key = None
+        blobstore = GSBlobStore(os.environ["GOOGLE_APPLICATION_CREDENTIALS"])
```
ttung: ditto

dss/events/handlers/index.py
```diff
@@ -17,32 +15,52 @@
...
+        process_new_indexable_object(blobstore, bucket_name, key, "gcp", logger)
+    except Exception as ex:
+        logger.error(f"Exception occurred while processing GS event: {ex} Event: {json.dumps(event, indent=4)}")
+        raise
+
+
+def process_new_indexable_object(blobstore, bucket_name, key, replica, logger) -> None:
```
ttung: you could actually type most of these parameters.

ttung: Also, if you're passing in replica here, you can just get the BlobStore object inside `process_new_indexable_object`

dss/events/handlers/index.py
Bento007: I just recently found that code. Wish i knew about it sooner. I'll make this change. 👍

Bento007:
I also added blobstore to test_indexer and infra. I intended to make this it's own PR, but I forgot to create new branch and added the changes to this branch. Sorry about that.

dss/events/handlers/index.py
```diff
@@ -6,43 +6,58 @@
...
-        else:
-            logger.debug(f"Not indexing S3 creation event for key: {key}")
+        bucket_name = event['Records'][0]["s3"]["bucket"]["name"]
+        process_new_indexable_object(bucket_name, key, "aws", logger)
     except Exception as ex:
         logger.error(f"Exception occurred while processing S3 event: {ex} Event: {json.dumps(event, indent=4)}")
         raise
 
 
+def process_new_gs_indexable_object(event, logger) -> None:
```
ttung: Did I miss the place where this is called?

dss/events/handlers/index.py
```diff
@@ -52,19 +67,18 @@ def is_bundle_to_index(key) -> bool:
     return result is not None
 
 
-def read_bundle_manifest(s3, bucket_name, bundle_key, logger):
-    manifest_string = s3.Object(bucket_name, bundle_key).get()['Body'].read().decode("utf-8")
+def read_bundle_manifest(blobstore, bucket_name, bundle_key, logger):
```
ttung: https://github.com/HumanCellAtlas/data-store/issues/390

tests/infra.py
```diff
@@ -132,39 +132,29 @@ def __getattr__(self, item: str) -> typing.Any:
...
-        self.path = object_summary.key
-        self.metadata = object_summary.Object().metadata
+        self.metadata = bundle.blobstore.get_user_metadata(bundle.bucket, object_summary)
         self.indexed = True if self.metadata['hca-dss-content-type'] == "application/json" else False
-        self.name = os.path.basename(self.path)
-        self.url = f"s3://{bundle.bucket.name}/{self.path}"
+        self.name = os.path.basename(object_summary)
+        self.path = object_summary
         self.uuid = str(uuid.uuid4())
+        self.url = f"s3://{bundle.bucket}/{self.path}"
```
ttung: If `TestFile` is not to be cloud-specific, then this is not the right source url.

tests/test_indexer.py
```diff
@@ -74,7 +73,12 @@ class TestIndexer(unittest.TestCase, DSSAsserts, StorageTestSupport):
 
     @classmethod
     def setUpClass(cls):
+        cls.replica = "aws"
```
ttung: is this intended? if you're making this cloud-agnostic, shouldn't this test both code paths?

Bento007: That is intended for now. I'm working on making a base class for test_indexer to test any cloud service we use. That will be in my next PR.

tests/infra.py
Bento007: You're absolutely right.

tests/test_indexer.py
ttung: Cool, sgtm.

dss/events/handlers/index.py
Bento007: This will be a separate PR

dss/events/handlers/index.py
Bento007: I'll remove for now and add it in my next PR, since test coverage is not there. Otherwise it won't pass the code coverage.

Bento007:
Ready for review!

tests/infra.py
```diff
@@ -132,39 +133,30 @@ def __getattr__(self, item: str) -> typing.Any:
...
-        self.path = object_summary.key
-        self.metadata = object_summary.Object().metadata
+        self.metadata = bundle.blobstore.get_user_metadata(bundle.bucket, object_summary)
         self.indexed = True if self.metadata['hca-dss-content-type'] == "application/json" else False
-        self.name = os.path.basename(self.path)
-        self.url = f"s3://{bundle.bucket.name}/{self.path}"
+        self.name = os.path.basename(object_summary)
+        self.path = object_summary
         self.uuid = str(uuid.uuid4())
+        self.url = f"{Config.get_storage_schema(replica)}://{bundle.bucket}/{self.path}"
```
ttung: I believe that as a group, we've decided not to allow complicated expressions (anything more than a variable or a variable+index) inside a f-string.

Bento007: I'll switch to using string concatenation instead of f-string.

PR #382, opened 2017-08-26
Refactor GS event notifications

kislyuk:
* Switch from a poor API (https://cloud.google.com/storage/docs/object-change-notification) to a better API (https://cloud.google.com/storage/docs/pubsub-notifications) for GS event notifications. Rip out the associated event forwarder in the `/internal/notify` API route. (The old API is impossible to automate from a clean slate account, and provides no way to list bucket notification subscriptions.)
* Add an event relay Google Cloud Function that forwards relevant GS events to SNS events.
* Add initial infrastructure for deploying Google Cloud Functions.
* Update deployment documentation.
* Add placeholder for indexer daemon GS event entry point.


scripts/deploy_gcf.py
```diff
@@ -0,0 +1,111 @@
...
+}
+
+try:
+    print(gcf_conn.api_request("POST", f"/{gcf_ns}", data=gcf_config))
+except google.cloud.exceptions.Conflict:
+    print(gcf_conn.api_request("PUT", f"/{gcf_ns}/{args.gcf_name}", data=gcf_config))
+
+sys.stderr.write("Waiting for deployment...")
+sys.stderr.flush()
+while gcf_conn.api_request("GET", f"/{gcf_ns}/{args.gcf_name}")["status"] == "DEPLOYING":
```
ttung: do we need a ceiling on how long this can wait?

scripts/deploy_gcf.py
```diff
@@ -0,0 +1,111 @@
...
+    sys.stderr.write(".")
+    sys.stderr.flush()
+    time.sleep(1)
+sys.stderr.write("done\n")
+
+res = gcf_conn.api_request("GET", f"/{gcf_ns}/{args.gcf_name}")
+print(res)
+assert gcf_conn.api_request("GET", f"/{gcf_ns}/{args.gcf_name}")["status"] == "READY"
+
+deploy_blob.delete()
```
ttung: this should be in a `finally` clause.

daemons/Makefile
```diff
@@ -17,4 +17,7 @@ dss-chunked-task:
 	./build_deploy_config.sh $@ $(DSS_DEPLOYMENT_STAGE)
 	cd $@; domovoi deploy --stage $(DSS_DEPLOYMENT_STAGE)
 
-.PHONY: dss-chunked-task dss-sync dss-index
+dss-gs-event-relay:
+	$(DSS_HOME)/scripts/deploy_gcf.py $@ --entry-point "dss_gs_bucket_events_$(subst -,_,$(DSS_GS_BUCKET))"
```
ttung: IIRC, `test_sync.py` uses the test bucket. this will deploy the script against the dev bucket.

kislyuk: `test_sync.py` is a unit test - it doesn't call live lambdas, calling the daemon entry points directly in-process instead. So it shouldn't care about which bucket the deployed event bus is connected to.

scripts/deploy_gcf.py
kislyuk: Will add.

scripts/deploy_gcf.py
kislyuk: Will fix.

daemons/Makefile
ttung: Got it, thanks.

kislyuk:
@ttung ready for another look.

kislyuk:
Test strategy: I will write an integration test that gets called from a Travis cron job. This integration test will verify, in dev, sync and indexing of large bundles.

PR #383, opened 2017-08-27
Add initial set of managed wheels

kislyuk:


ttung:
The same wheel shows up multiple times, in different subdirectories.  I assume that is because each library is saving its own dependent wheels.  This seems unnecessary, given we specify the entire set of libraries in requirements.txt.

kislyuk:
Yes, this is not harmful in its current state, but could cause problems if different packages pull in different versions of dependencies in their graphs. I'll see if I can take just the one wheel for each package itself and throw away the rest.

kislyuk:
(in other words the dependencies pulled in through subgraphs here actually don't respect the pins set in requirements.txt)

kislyuk:
In https://github.com/HumanCellAtlas/data-store/pull/387.

PR #387, opened 2017-08-28
Discard wheel dep subgraphs

kislyuk:


scripts/find_missing_wheels.py
```diff
@@ -19,6 +19,13 @@ def build_wheel(wheel_identifier, wheels_dir):
     subprocess.check_output(["pip", "download", wheel_identifier], cwd=wd)
     if glob.glob(os.path.join(wd, "*.tar.gz")):
         subprocess.check_output("pip wheel *.tar.gz && rm -f *.tar.gz", shell=True, cwd=wd)
+    # The sub-dependency wheels downloaded and built here do not obey the pins set by pip freeze in the overall
+    # requirements.txt. We only keep the top level wheel. Since requirements.txt contains the whole flattened dependency
+    # graph, each missing sub-dependency wheel is handled at top level in the overall scan.
+    for wheel_filename in os.listdir(wd):
+        if not re.match(wheel_identifier.replace("-", "_").replace("==", "-"), wheel_filename):
```
ttung: Why not just `str.find()`?  `re` could have unexpected interactions.

kislyuk: Good point (although wheel names are sanitized to alphanumeric + `_`).

I'm going to use the `if substring not in string:` form if you don't mind.

kislyuk: Actually, scratch that, I want to anchor at 0, so I'll use `if str.find(sub) != 0`.

kislyuk:
Removing complex expression from f-string, in the interest of holding myself accountable :)

Issue #232, opened 2017-07-17
Testing - create staging deployment in HCA account


briandoconnor:
A followup to #186 we want to deploy a new system after the async upload is added.  This will be a stable data store system for Jim and others to use for query development.  For the data being loaded see https://github.com/HumanCellAtlas/data-bundle-examples/issues/9

kozbo:
@briandoconnor will you have time for this by the end of the week?

PR #392, opened 2017-08-29
Fix deploy stage parameterization

kislyuk:
There were a few loose ends in the config files. I found these while deploying the "staging" stage.

Connected to #232


iam/policy-templates/ci-cd.json
```diff
@@ -97,8 +97,8 @@
       ],
       "Resource": [
         "arn:aws:lambda:*:$account_id:function:dss-*",
-        "arn:aws:es:*:$account_id:domain/dss-index-dev",
-        "arn:aws:es:*:$account_id:domain/dss-index-dev/*",
+        "arn:aws:es:*:$account_id:domain/dss-*",
```
ttung: Is this desirable?

kislyuk: I think yes, because we want CI/CD to be able to deploy any of the stages.

ttung: If so, then we probably need to fix the other bucket-specific/stage-specific perms.

kislyuk: Backed out changes to this file, since CI/CD IAM perms are outside the scope of this PR. let's address those separately

Issue #393, opened 2017-08-29
Speed up async copy


kislyuk:
@sampierson and I tested it, and it seems to currently copy data at about 1 GB/minute. Let's look at parallelizing multipart copy.

kozbo:
It looks like all the connected PR's are merged or closed. Can this be moved to Done?

ttung:
No, there's one more piece here needed.

Issue #319, opened 2017-08-07
individual daemons per chunked worker client


ttung:
This is a better way to go, because multiple developers building their own clients will stomp over each other trying to deploy to `-dev` to pass unit tests.

PR #396, opened 2017-08-30
Add a simple logging facility.

ttung:
- Logs to cloudwatch.
- Creates the group and stream, if necessary.
- Retrieves the uploadSequenceToken to log to existing streams.
- Adds test to verify most of the functionality.  The key functionality that is not tested is the creation of a logging group.  Rationale: There's a hard limit on the number of test groups.  If this test goes wild, it might do bad things to the account.

Connects to #319 
Connects to #393

### DIFF
--- a/dss/util/aws/logging.py
+++ b/dss/util/aws/logging.py
@@ -1,0 +1,50 @@
+import time
+
+import boto3
+import botocore.exceptions
+
+
+def log_message(log_group_name: str, log_stream_name: str, message: str):
+    """Logs a message to cloudwatch."""
+
+    logs_client = boto3.client("logs")
+
+    def get_sequence_token():
+        # try to get the upload sequence token
+        paginator = logs_client.get_paginator('describe_log_streams')
+        for page in paginator.paginate(logGroupName=log_group_name, logStreamNamePrefix=log_stream_name):
+            for log_stream in page['logStreams']:
+                if log_stream['logStreamName'] == log_stream_name:
+                    return log_stream.get('uploadSequenceToken', None)
+
+        return None
+
+    while True:
+        try:
+            logs_client.create_log_group(logGroupName=log_group_name)
+        except logs_client.exceptions.ResourceAlreadyExistsException:
+            pass
+        try:
+            logs_client.create_log_stream(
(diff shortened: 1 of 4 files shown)


iam/policy-templates/ci-cd.json
```diff
@@ -78,7 +78,10 @@
       "Action": [
         "logs:*"
       ],
-      "Resource": "arn:aws:logs:*:$account_id:log-group:chunked_task_worker*"
+      "Resource": [
+        "arn:aws:logs:*:$account_id:log-group:chunked_task_worker*",
+        "arn:aws:logs:*:$account_id:log-group:test_logging*"
```
kislyuk: should log groups start with `dss-` per our resource naming convention?

tests/test_logging.py
```diff
@@ -0,0 +1,78 @@
...
+        starttime = time.time()
+        while time.time() < starttime + 30:
+            response = logs_client.filter_log_events(
+                logGroupName="test_logging", logStreamNames=[stream_name])
+
+            for event in response['events']:
+                if event['message'] == message2:
+                    break
+            else:
+                continue
```
kislyuk: I have seen issues where too many requests to the logging client API can cause `503 Slow Down` from AWS, since the cloudwatch logs reader API is substantially rate limited. So consider adding a `time.sleep(1)` here.

dss/util/aws/logging.py
```diff
@@ -0,0 +1,56 @@
...
+import botocore.exceptions
+
+
+def log_message(log_group_name: str, log_stream_name: str, message: str):
+    """Logs a message to cloudwatch."""
+
+    logs_client = boto3.client("logs")
+
+    # An astute reader might notice that the exception handling is a bit wonky.  This is because exception classes for
+    # boto are dynamically created, and we cannot actually catch them. :(
```
kislyuk: You can access these exceptions via `boto3.client("logs").exceptions.ResourceAlreadyExistsException` and `botocore.errorfactory.InvalidSequenceTokenException` (and their identities are the same across multiple client instances).

There are some services (like EC2) that don't expose their exceptions in the AWS API service model, but logs does.

https://github.com/boto/botocore/pull/1005 :)

PR #402, opened 2017-09-01
Parameterize ES index name by stage

kislyuk:
This is needed to be able to run staging and dev using the same ES cluster, which I'd like to keep doing for now. (Prod will get its own cluster)

Connected to #232

### DIFF
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -26,14 +26,14 @@ from .error import DSSException, dss_handler
 # CONSTANTS COMMON TO THE INDEXER AND QUERY ROUTE.
 
 # ES index containing all docs
-DSS_ELASTICSEARCH_INDEX_NAME = "hca"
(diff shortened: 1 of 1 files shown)


dss/__init__.py
```diff
@@ -26,14 +26,14 @@
 # CONSTANTS COMMON TO THE INDEXER AND QUERY ROUTE.
 
 # ES index containing all docs
-DSS_ELASTICSEARCH_INDEX_NAME = "hca"
+DSS_ELASTICSEARCH_INDEX_NAME = "hca-" + os.environ["DSS_DEPLOYMENT_STAGE"]
```
mikebaumann: Because the DSS is intended to be project independent, we should use a more general index base name than "hca". Suggestions are welcome. "index"?,  "primary"? ...
Yet, that may be best done as a different PR.

kislyuk: yeah I thought that too. "dss-docs" might do.

PR #403, opened 2017-09-01
Make Elasticsearch indexes replica specific

mikebaumann:
Change Elasticsearch index names to be replica specific to enable replica
independent indexing, search, subscription/notification.
This commit is focused on replica specific indexing, with changes
to search and subscription for all tests to pass with replica specific indexes.
Additional changes are required to make the search route replica aware,
and additional GCP-specific testing is required for subscriptions,
which will be done as additional PRs.

### DIFF
--- a/daemons/dss-index/app.py
+++ b/daemons/dss-index/app.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os
 import sys
@@ -8,7 +9,7 @@ pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'domovoilib')
 sys.path.insert(0, pkg_root)  # noqa
 
 import dss
-from dss.events.handlers.index import process_new_s3_indexable_object
+from dss.events.handlers.index import process_new_s3_indexable_object, process_new_gs_indexable_object
 
 app = domovoi.Domovoi()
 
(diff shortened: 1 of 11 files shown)


mikebaumann:
I welcome suggestions for making it cleaner/simpler, yet please keep in mind there are two indexes that need to be made replica specific, the one for the index docs and the one for the subscriptions.

kislyuk:
Refactoring it now.

kislyuk:
My refactor is in https://github.com/HumanCellAtlas/data-store/pull/405 vs. this branch.

ttung:
I thought we discussed standing up a separate instance for each replica.  I can't envision any scenario where it makes sense to make one instance serve multiple replicas.

kislyuk:
I think I'd prefer to have all replica specific indexes reside in the same ES cluster. This makes it easier to manage ES connections.

ttung:
When would we reuse connections for ES?

kislyuk:
ES queries will go to replica-specific indexes. A lambda may service many requests over its lifetime, each directed to a particular replica. Communicating with multiple ES instances requires setting up and caching a connection for each of them. I would rather have one ES instance serving the entire stage, until we go to replica-specific API service deployments embedded in their replicas' clouds (i.e. deploy the DSS API services on GCP).

ttung:
What's the cost of caching multiple connections?

kislyuk:
IMHO the main cost is code complexity.

ttung:
I don't think the code would be that more complicated, and splitting it in the future is much more complicated than doing it now.

However, I'll defer to you guys on this.  Maybe you can ask the ES contractor. :)

ttung:
Let's put it this way: would you put all the data for different deployments on the same S3 bucket?

Also, you have to have code to namespace the indices.  That is no different than code to namespace the instances.

Bento007:
Closed by mistake

PR #404, opened 2017-09-01
Separate lambdas for each chunked-worker client.

ttung:
Automatically builds makefile rules from the client list, and rewrites the config for each separate client.

Connects to #319 
Connects to #393

### DIFF
--- a/iam/policy-templates/dss-chunked-task-s3_copy_write_metadata-lambda.json
+++ b/iam/policy-templates/dss-chunked-task-s3_copy_write_metadata-lambda.json
@@ -1,0 +1,43 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "logs:CreateLogGroup",
+        "logs:CreateLogStream",
+        "logs:DescribeLogStreams",
+        "logs:PutLogEvents"
+      ],
+      "Resource": "arn:aws:logs:*:*:*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": "sns:Publish",
+      "Resource": [
+        "arn:aws:sns:*:$account_id:dss-*-$stage"
+      ]
+    },
+    {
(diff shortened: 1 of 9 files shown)


daemons/dss-chunked-task/clients.py
```diff
@@ -0,0 +1,12 @@
...
+import os
+import sys
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from dss.events.chunkedtask import aws
+
+for name in aws.get_clients().keys():
+    print(f"dss-chunked-task-{name}")
```
kislyuk: Consider moving this entry point into /scripts/.

iam/policy-templates/dss-chunked-task-s3_copy_write_metadata-lambda.json
```diff
@@ -0,0 +1,40 @@
...
+    {
+      "Effect": "Allow",
+      "Action": "sns:Publish",
+      "Resource": [
+        "arn:aws:sns:*:$account_id:dss-*-$stage"
+      ]
+    },
+    {
+      "Effect": "Allow",
+      "Action": "s3:*",
```
kislyuk: Consider enumerating permissions on a need basis here (to be fair, we may need to do it in other policies too). For example, to limit exposure to catastrophic data loss, we may want to deny delete access to all daemons except a specially managed deleter daemon.

daemons/dss-chunked-task/clients.py
ttung: fixed.

iam/policy-templates/dss-chunked-task-s3_copy_write_metadata-lambda.json
ttung: done.

PR #407, opened 2017-09-03
Allow chunked tasks to fork off work.

ttung:
This is the megadiff version.  I'll PR separate versions if you think that's easier to read. :)

### DIFF
--- a/dss/events/chunkedtask/_awstest.py
+++ b/dss/events/chunkedtask/_awstest.py
@@ -122,3 +122,23 @@ class SupervisorTask(Task[dict, bool]):
 
     def check_success_marker(self):
         raise NotImplementedError()
+
+
+AWS_SUPERVISOR_TEST_CLIENT_NAME = "supervisortest"
+
+
+class AWSSupervisorTask(SupervisorTask):
+    def __init__(self, state: dict, runtime: AWSRuntime) -> None:
+        super().__init__(state, runtime)
+
+    def check_success_marker(self):
+        # don't pound the filter logs API to a pulp.
+        if (self.last_checked is not None and
+                time.time() < self.last_checked + 1):
+            time.sleep(1)
+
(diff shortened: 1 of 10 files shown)


tests/chunked_worker.py
```diff
@@ -0,0 +1,103 @@
...
+
+        task_id = str(uuid.uuid4()) if new_task else None
+        self.scheduled_work.append((task_class, state, task_id))
+        return task_id
+
+    def work_complete_callback(self, result: typing.Any):
+        self.result = result
+
+
+def task_lifecycle(
```
kislyuk: Can we name this something more action-y like `advance_task_lifecycle()` or `process_task_updates()`?

dss/events/chunkedtask/_awstest.py
```diff
@@ -54,3 +83,62 @@ def run_one_unit(self) -> typing.Optional[bool]:
...
+
+class AWSSupervisorTask(SupervisorTask):
+    def __init__(self, state: dict, runtime: AWSRuntime) -> None:
+        super().__init__(state, runtime)
+
+    def check_success_marker(self):
+        # don't pound the filter logs API to a pulp.
+        if (self.last_checked is not None and
+                time.time() < self.last_checked + 1):
+            time.sleep(1)
```
kislyuk: One unfortunate concern to watch out for is the CloudWatch Logs API limits for the number of concurrent readers and their request rates. As the number of supervisor tasks increases (e.g. if you spawn a supervisor task for each object copy >64MB), using CloudWatch Logs for inter-lambda IPC may not scale. This is something you could stress-test for in an ad hoc test.

http://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html

ttung: This is only for the unit test.  The production code path does not rely on cloudwatch.

PR #410, opened 2017-09-04
A method to drive a task to completion.

ttung:
1. A queue to track all the work that still needs to be run.
2. Each iteration of the loop removes a unit to run, constructs a runtime object and a task object, and runs it.  If there's more work to be run, save the state into the queue.

Connects to #393

### DIFF
--- a/tests/chunked_worker.py
+++ b/tests/chunked_worker.py
@@ -1,0 +1,68 @@
+import collections
+import typing
+import uuid
+
+from dss.events.chunkedtask import Runner, Runtime, Task
+
+
+def run_task_to_completion(
+        initial_task: typing.Type[Task],
+        initial_state: typing.Any,
+        runtime_builder: typing.Callable[[typing.Mapping[str, typing.Any]], Runtime],
+        task_builder: typing.Callable[[typing.Type[Task], typing.Any, Runtime], Task],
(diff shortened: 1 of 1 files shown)


tests/chunked_worker.py
```diff
@@ -0,0 +1,68 @@
...
+    the method, and various callables must be provided to complete the task.
+
+    :param initial_task: This is the initial task that we're starting with.
+    :param initial_state: This is the initial state of the task we're starting with.
+    :param runtime_builder: This is a callable that is called to generate a Runtime object.  It is called with a
+                            dictionary that stores a mapping between task uuids and their results.
+    :param task_builder: This is a callable that produces a Task object, given a task's class, the state, and the
+                         runtime.
+    :param runtime_result: This is a callable that we execute against a runtime object to get the result of the task.
+                           If the task is not complete, the callable should return None.
```
kislyuk: In addition to the comment in the other PR about `task_lifecycle`, consider also making `runtime_result` something more action-y. `get_runtime_result`?

ttung: Renamed `task_lifecycle` to `run_task_to_completion`.

Renamed `runtime_result` to `runtime_result_getter` (to match `runtime_scheduled_getter`)

PR #426, opened 2017-09-06
Enable parallel deploy

kislyuk:


.travis.yml
```diff
@@ -33,12 +37,12 @@ after_success:
 - bash <(curl -s https://codecov.io/bash)
 deploy:
   - provider: script
-    script: make deploy
+    script: make -jO deploy
```
mikebaumann: What does "-j0" do?
After reading the following excerpt from the GNU make 4 manual the effect of the value 0 wasn't completely clear to me:

> If the ‘-j’ option is followed by an integer, this is the number of recipes to execute at
> once; this is called the number of job slots. If there is nothing looking like an integer after
> the ‘-j’ option, there is no limit on the number of job slots. The default number of job
> slots is one, which means serial execution (one thing at a time).

kislyuk:
That's an O not 0.

‘-O[type]’‘--output-sync[=type]’

Ensure that the complete output from each recipe is printed in one
uninterrupted sequence. This option is only useful when using the --jobs option
to run multiple recipes simultaneously (see Parallel Execution
<https://www.gnu.org/software/make/manual/make.html#Parallel>) Without this
option output will be displayed as it is generated by the recipes.

With no type or the type ‘target’, output from the entire recipe of each
target is grouped together. With the type ‘line’, output from each line in
the recipe is grouped together. With the type ‘recurse’, the output from an
entire recursive make is grouped together. With the type ‘none’, no output
synchronization is performed. See Output During Parallel Execution
<https://www.gnu.org/software/make/manual/make.html#Parallel-Output>.

kislyuk:
Actually, you're right, those options have to be specified separately. Updated.

.travis.yml
```diff
@@ -25,6 +25,10 @@ install:
 - pip install -r requirements-dev.txt
 - wget -q ${ES_DOWNLOAD_URL}
 - tar -xzf elasticsearch-${ES_VERSION}.tar.gz
+- wget -q http://us-east-1.ec2.archive.ubuntu.com/ubuntu/pool/main/m/make-dfsg/make_4.1-6_amd64.deb
+- mkdir make4
+- dpkg -x make*.deb make4
+- export PATH=$(pwd)/make4/usr/bin:$PATH
 script:
 - set -eo pipefail
 - make test
```
ttung: I think you should consider making this -j -O.

.travis.yml
```diff
@@ -33,12 +37,12 @@ after_success:
 - bash <(curl -s https://codecov.io/bash)
 deploy:
   - provider: script
-    script: make deploy
+    script: make -j -O deploy
```
ttung: I don't think you should go unlimited.  My guess is that -j4 is probably a realistic limit.

.travis.yml
kislyuk: That's coming in a separate PR :)

.travis.yml
kislyuk: Switched to `-j4`, thanks.

Issue #348, opened 2017-08-14
Native search, subscription and notification support for Google


mikebaumann:
Support search, subscription and notification specific to Google.

Tasks:
Change search route to be replica aware (< 1 day)
Change subscription route to be replica aware (< 1 day)
Change notification to be replica aware ( <1 day)
Deployment testing (manual, < 1 day)

PR #432, opened 2017-09-11
Added Support for Replica in Search API

Bento007:
Added support for replica in search API.

Connects to #348

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -1,5 +1,4 @@
 import json
-import typing
 
 import requests
 from elasticsearch.exceptions import ElasticsearchException
@@ -12,7 +11,7 @@ from .. import Config, Replica, ESIndexType, dss_handler, get_logger, DSSExcepti
 from ..util.es import ElasticsearchClient
 
 @dss_handler
-def post(json_request_body: dict, replica: typing.Optional[str]=None):
+def post(json_request_body: dict, replica: str):
     es_query = json_request_body['es_query']
     get_logger().debug("Received posted query. Replica: %s Query: %s", replica, json.dumps(es_query, indent=4))
     if replica is None:
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -1,4 +1,5 @@
 import json
+import typing
 
 import requests
 from elasticsearch.exceptions import ElasticsearchException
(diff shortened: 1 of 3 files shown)


dss/api/search.py
```diff
@@ -10,14 +11,12 @@
 from .. import Config, Replica, ESIndexType, dss_handler, get_logger, DSSException
 from ..util.es import ElasticsearchClient
 
-# TODO Adding replica as a search parameter and including tests for gcp
-# will be done in a different PR.
-replica = "aws"
-
 @dss_handler
-def post(json_request_body: dict):
+def post(json_request_body: dict, replica: typing.Optional[str]=None):
```
ttung: why not make it required?

mikebaumann: Trent and I debated about whether it should be required or not, and looked at files.py and bundles.py to see if we could infer a convention regarding this. Some of the files and bundles operations treat the replica as optional and default to "aws" (accompanied by a todo to use the local replica when the time comes), and others require the replica option be specified. I proposed making it optional, thinking along the lines of minimizing the number of options a user needs to specify and providing reasonable default values. Of course, the option to specify a replica is available if they choose. 
The subscription route requires the replica parameter, which makes sense there. 
We are happy to make this a required parameter if you think that is a better fit with our existing conventions and future direction.

ttung: I think for search, it should be required.

Rationale: Someone searching for something wants to search for data close to their compute.  The reason it can be optional for get bundle/file is that people are unlikely to just divine the url.  They have to search for something to get the initial url.

If it makes you feel better, I am happy to make get bundle/file require a replica for consistency's sake. :)

Bento007: This is a problem of customizability vs simplicity. The harder we make it to use the less likely it will be used. Since this is a key feature we should make it as easy to use as possible. Making replica optional with a default value simplifies life for the user. If the user want to customize and optimize they can later choose to specify the replica.

kislyuk: I agree with @Bento007, replica should not be required. We should make a request geolocator that assigns the replica based on the network CIDR the request is coming from.

Bento007: Sounds like a future feature we will need to implement on data-store-cli.

ttung: Oh, I see what you're saying.  You're saying search should have the replica field be optional?

I think it makes more sense for the front end to assign a default replica than it does for the backend.  Having the back end automatically determine the replica means the same search could yield different results to different users.

PR #433, opened 2017-09-11
Added Test Cases for GCP Subscribe

Bento007:
Added test cases for GCP subscribe.

Connects to #348

### DIFF
--- a/tests/test_subscriptions.py
+++ b/tests/test_subscriptions.py
@@ -37,7 +37,7 @@ def setUpModule():
 
 def tearDownModule():
     ESInfo.server.shutdown()
-
+    os.unsetenv('DSS_ES_PORT')
 
 class TestSubscriptionsBase(DSSAsserts):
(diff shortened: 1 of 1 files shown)


tests/test_subscriptions.py
```diff
@@ -28,15 +28,21 @@
...
-        os.environ['DSS_ES_PORT'] = str(cls.es_server.port)
+class ESInfo:
+    server = None
+
+def setUpModule():
+    ESInfo.server = ElasticsearchServer()
+    os.environ['DSS_ES_PORT'] = str(ESInfo.server.port)
+
+def tearDownModule():
+    ESInfo.server.shutdown()
```
ttung: i wonder if it makes sense to unset os.environ['DSS_ES_PORT']

Bento007: That sounds like a sensable idea.

Issue #427, opened 2017-09-07
Logging functions using f-strings can cause exceptions if % are substituted into f-string.


Bento007:
check over f-strings in logs to avoid substituting a variable that includes a %. This may cause the logging expression to interpret the % as a variable substitution. This was observed when integrating the google could support for indexer.

from file dss/events/handlers/index.py line 42.
**Erroneous example:**
`logger.error(f"Exception occurred while processing GS event: {ex} Event: %s", json.dumps(event, indent=4))`

after the f-string is evaluation this string is created: 

> Exception occurred while processing GS event: 403 GET https://www.googleapis.com/storage/v1/b/fake/o/bundles%2F1e7366fb-3634-4d8d-8bf6-2a0aaf1868a8.2017-09-07T222143.145620Z: tsmith12@human-cell-atlas-mbaumann-dev.iam.gserviceaccount.com does not have storage.objects.get access to fake/bundles/1e7366fb-3634-4d8d-8bf6-2a0aaf1868a8.2017-09-07T222143.145620Z.

The %2F in the url is evaluated as a variable substitution by logger which results in an exception being thrown because the incorrect number or type of parameters was provided. Here is how to avoid it.

**Fixed Example:**
`logger.error("Exception occurred while processing GS event: %s Event: %s", ex, json.dumps(event, indent=4))`

run test_indexer.TestGCPIndexer.test_debug_message_logged_when_indexing_an_invalid_bundle_name to see fix.

ttung:
So another possibility we could do is:

logger.error("%s", f"abcabcabc")

This standardizes us on the f-string formatting operations, which we are trending towards.

The downside is that we do the string formatting operation regardless of whether the logging is enabled.

PR #436, opened 2017-09-11
first argument logger.XXX should be a formatting string

ttung:
If the variable contains a '%', it may be treated as a formatting operation.

Connects to #427


Bento007:
It does reduce the amount of refactoring we have to do, but it feels more like a hack than a fix. Other than refactoring, is there a reason we don't use logger.error("something %s", var)?

ttung:
I feel f-style strings are a lot easier to read. I can be persuaded not to do this though. :)

PR #442, opened 2017-09-12
mandate replica.

ttung:
see https://github.com/HumanCellAtlas/data-store/pull/432#discussion_r138255838


ttung:
Before I land this, I want to make sure that this is not going to throw wrinkle in integration efforts.  cc: @kislyuk @sampierson

PR #445, opened 2017-09-12
Smoketest: Add large file and subscription API ops

kislyuk:


scripts/smoketest.py
```diff
@@ -45,6 +46,9 @@ def run(command, **kwargs):
 
 sample_id = str(uuid.uuid4())
 bundle_dir = "data-bundle-examples/10X_v2/pbmc8k"
+with open(os.path.join(bundle_dir, "large_file"), "wb") as fh:
+    fh.write(os.urandom((1024 * 1024 * 64) + 1))
```
ttung: Are you explicitly disconnecting this script from the rest of the DSS code base?

kislyuk: No, I don't intend to do that. Do you have suggestions for what I should use here?

ttung: Let me make this threshold a constant first.

ttung: https://github.com/HumanCellAtlas/data-store/pull/447

kislyuk: Addressed, thanks. Please re-review

tests/smoketest.py
```diff
@@ -62,6 +71,13 @@ def run(command, **kwargs):
 run("jq -n '.es_query.query.match[env.k]=env.v' | http --check-status https://${API_HOST}/v1/search > res.json",
     env=dict(os.environ, k="files.sample_json.uuid", v=sample_id))
 
+for replica in "aws", "gcp":
+    res = run(f"hca put-subscriptions --callback-url https://example.com/ --query '{{}}' --replica {replica}",
+              runner=check_output)
+    sub_id = json.loads(res.decode())["uuid"]
```
Bento007: We should also uploaded a file and check subscriber is notified.

Bento007: maybe just change the order of the tests to create a subscription earlier and deleted it later.

kislyuk: That's outside the scope of this PR. Let's address that in a separate PR (which will be much more complicated, since checking that a subscriber is notified will require either standing up an HTTPS test endpoint, or at least looking at server logs). Can you please approve this one so I can merge?

Bento007: As we talk about, modify the --query '{{}}' to a valid query.

kislyuk: `{}` (which is what that f-string compiles to) is a (syntactically) valid query, so the test passes.

We should probably make it more semantically meaningful in the future, in a separate PR.

PR #449, opened 2017-09-13
Suggested Changes to #445 expand smoketest 

Bento007:
I moved put_subscribe to the beginning. This way we will be able to smoke test the notification system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ $(test_srcs): %.py :
 	python -m unittest $@
 
 smoketest:
-	scripts/smoketest.py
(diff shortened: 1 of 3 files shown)


kislyuk:
Please rebase.

I like the idea of testing notifications, unfortunately it's not trivial to set up a test callback endpoint in a way that would avoid tests stepping on each other. Instead, we could look for a webhook delivery log message with a unique random string provided in the callback URL.

kislyuk:
From sprint review: @kislyuk to take a look at unblocking this PR.

PR #452, opened 2017-09-13
Handle parameter and body validation errors with our code.

ttung:
Connexion immediately calls its internal error handler when parameter or body validation fails.  We want to manage the error handling so the error message fits our formatting.

### DIFF
--- a/dss/__init__.py
+++ b/dss/__init__.py
@@ -3,16 +3,17 @@
 """
 DSS description FIXME: elaborate
 """
-import traceback
-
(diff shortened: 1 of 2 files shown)


Bento007:
Will we need to added additional test cases to our other test suites to verify the additional checking?

ttung:
> Will we need to add additional test cases to our other test suites to verify the additional checking?

I don't think so.  As long as we verify it in one place, it should be consistently applied.

Issue #457, opened 2017-09-15
Search result bundle_url should include replica parameter


mikebaumann:
Now that DSS search is replica specific, the result bundle_url should include the replica parameter.
Current result is like this:
`"bundle_url": "https://hca-dss-4.ucsc-cgp-dev.org/v1/bundles/1b992a05-6b82-4d3a-9f38-ae431c7234c3?version=2017-09-15T153830.201872Z"`

Instead, it should be like this:
`"bundle_url": "https://hca-dss-4.ucsc-cgp-dev.org/v1/bundles/1b992a05-6b82-4d3a-9f38-ae431c7234c3?version=2017-09-15T153830.201872Z&replica=gcp"`

PR #458, opened 2017-09-15
Added replica to bundle_url

Bento007:
Connects to #457 

added replica to bundle_url.

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -26,7 +26,7 @@ def post(json_request_body: dict, replica: str):
         bundles_url_base = request.host_url + 'v1/bundles/'
         result_list = [{
             'bundle_id': hit.meta.id,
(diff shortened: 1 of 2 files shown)


dss/api/search.py
```diff
@@ -26,7 +26,7 @@ def post(json_request_body: dict, replica: str):
         bundles_url_base = request.host_url + 'v1/bundles/'
         result_list = [{
             'bundle_id': hit.meta.id,
-            'bundle_url': bundles_url_base + hit.meta.id.replace(".", "?version=", 1),
+            'bundle_url': bundles_url_base + hit.meta.id.replace(".", "?version=", 1) + f"&replica={replica}",
```
ttung: Out of the scope if this diff, but this should be using `util.UrlBuilder`

Bento007: Added a TODO to use `util.UrlBuilder`

Issue #355, opened 2017-08-15
Enable parameterized ES index names for test-specific indices


mikebaumann:
Description from Andrey:
... I want us to commit to refactoring the codebase to make it easy to parameterize the index names on the fly, so that each test ends up initializing and using a unique index instead of launching multiple heavyweight instances of ES (which might have other exciting interactions among themselves!)

This will enable faster, reliable execution of ES-related tests.

Bento007:
By appending the name of the test module with the index we can created test specific indexes.  This is currently in progress but is dependent on several PRs ahead of it.

kozbo:
From Tony: you may need to change the timeout

Bento007:
This issue has been fixed in the associated PRs. Once those are accepted this issues will be closed.

PR #460, opened 2017-09-15
added test specific index to indexer

Bento007:
- Added the environment variable `TEST_MODULE_NAME` to test cases that use an elasticsearch index. 

- Modified `get_es_index_name()` to append the `index`  with `TEST_MODULE_NAME` if `DSS_DEPLOYMENT_STAGE == Test`. 

We can now run the test cases in parallel to speed up execution time.

Connects to #355

### DIFF
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -26,6 +26,7 @@ sys.path.insert(0, pkg_root)  # noqa
 
 import dss
 from dss import Config, DeploymentStage
+from dss.config import IndexSuffix
 from dss.events.handlers.index import process_new_s3_indexable_object, process_new_gs_indexable_object
 from dss.hcablobstore import BundleMetadata, BundleFileMetadata, FileMetadata
 from dss.util import create_blob_key, UrlBuilder
@@ -71,7 +72,7 @@ class ESInfo:
     server = None
 
 def setUpModule():
-    os.environ['TEST_MODULE_NAME'] = __name__.split('.')[-1]
+    IndexSuffix.name = __name__.split('.')[-1]
     HTTPInfo.port = findOpenPort()
     HTTPInfo.server = HTTPServer((HTTPInfo.address, HTTPInfo.port), PostTestHandler)
     HTTPInfo.thread = threading.Thread(target=HTTPInfo.server.serve_forever)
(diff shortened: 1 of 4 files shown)


ttung:
I'm not sure how I feel about using environment variables to pass data around inside a process.  Can you think of a different way to do this?

Bento007:
I'm open to suggestions. The biggest constraint is I don't want to make an major modifications to the productions code. This was the cleanest way of accomplishing this.

Bento007:
@ttung no longer using environment variables. Please take a look before the sprint end.

tests/test_indexer.py
```diff
@@ -71,6 +72,7 @@ class ESInfo:
     server = None
 
 def setUpModule():
+    IndexSuffix.name = __name__.split('.')[-1]
```
ttung: you could consider `rsplit(".", 2)` or `rpartition(".")` or you could leave it as is.

tests/test_search.py
```diff
@@ -38,29 +39,30 @@ class ESInfo:
...
         cls.app = dss.create_app().app.test_client()
         with open(os.path.join(os.path.dirname(__file__), "sample_index_doc.json"), "r") as fh:
             cls.index_document = json.load(fh)
 
     def setUp(self):
         dss.Config.set_config(dss.DeploymentStage.TEST)
         self.app = dss.create_app().app.test_client()
-        elasticsearch_delete_index(self.dss_index_name)
-        create_elasticsearch_index(self.dss_index_name, logger)
+        elasticsearch_delete_index(f"*{IndexSuffix.name}")
```
ttung: why was the `create_elasticsearch_index(..)` call moved to the specific test?

tests/test_search.py
```diff
@@ -132,9 +134,10 @@ def test_search_returns_N_results_when_N_documents_match_query(self):
         """Create N identical documents. A search query is executed to match the document. All documents created must be
         present in the query results. N is varied across a variety of values.
         """
-        test_matches = [0, 1, 9, 10, 11, 1000, 5000]
+        test_matches = [0, 1, 9, 10, 11, 1000, 10001]
```
ttung: why this change?

Bento007: normal searching only returns up to the first 10000 results. This check that we can exceed that limit using paging.

tests/test_indexer.py
Bento007: i'll use rsplit, it'll be a little faster

tests/test_search.py
Bento007: The index is created when the first bundle is indexed. Since no bundles are indexed the index is never created. In all other test cases bundles are indexed before searching.

ttung: What's the harm of leaving the create_index call in `setUp`?

Bento007: it adds unnecessary steps, otherwise it's harmless.

PR #461, opened 2017-09-16
Running tests in parallel

Bento007:
This reduces the time it take to execute our tests suite.

connects to #460

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
 - export PATH=$(pwd)/make4/usr/bin:$PATH
 script:
 - set -eo pipefail
-- make test
+- make -j4 test
 - if [[ $TRAVIS_BRANCH == staging ]]; then source environment.staging; fi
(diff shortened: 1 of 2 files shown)


Bento007:
It's not required for the tests to run in parallel. Each test case that uses ES will spin up it's own instance for now. I think that would be a good idea for a future PR.

ttung:
> It's not required for the tests to run in parallel. Each test file that uses ES will spin up it's own instance for now. I think that would be a good idea for a future PR.

If that's the case, then why #460 ?

Bento007:
> If that's the case, then why #460 ?
The test cases where all using the same indexes before #460, which causes conflict between test suites.  Some test would clear the index before running which would mess up running test cases expecting something to be there. 

We should still try to run a single instance of elasticsearch, but that's a different problem.

Bento007:
added #483 to only use one ES instance for testing.

ttung:
I'd drop the test_slow rule, as it's not clear why one would want to invoke that.

Also (possibly out of the scope of this diff), but it may make sense to make the individual test_*.py targets phony.

Issue #459, opened 2017-09-15
Logic error in query registration


kislyuk:
```
Traceback (most recent call last):
  File "/var/task/chalicelib/dss/api/subscriptions.py", line 103, in put
    percolate_registration = _register_percolate(es_client, uuid, query)
  File "/var/task/chalicelib/dss/api/subscriptions.py", line 170, in _register_percolate
    refresh=True)
  File "/var/task/elasticsearch/client/utils.py", line 73, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/var/task/elasticsearch/client/__init__.py", line 298, in index
    _make_path(index, doc_type, id), params=params, body=body)
  File "/var/task/elasticsearch/transport.py", line 312, in perform_request
    status, headers, data = connection.perform_request(method, url, params, body, ignore=ignore, timeout=timeout)
  File "/var/task/elasticsearch/connection/http_requests.py", line 90, in perform_request
    self._raise_error(response.status_code, raw_data)
  File "/var/task/elasticsearch/connection/base.py", line 125, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)
elasticsearch.exceptions.RequestError: TransportError(400, 'mapper_parsing_exception', 'No field mapping can be found for the field with name [assay.rna.library_contruction]')

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/task/chalicelib/dss/error.py", line 31, in wrapper
    return func(*args, **kwargs)
  File "/var/task/chalicelib/dss/api/subscriptions.py", line 106, in put
    logger.critical(f"Percolate query registration failed:\n{percolate_registration}")
UnboundLocalError: local variable 'percolate_registration' referenced before assignment
```

kislyuk:
Query used:
```
{
  "query": {
    "bool": {
        "must": [{
            "match": {
                "assay.rna.library_contruction": "smart-seq2"
            }
        }, {
            "match": {
                "assay.seq.paired_end": "yes"
            }
        }, {
          "match": {
            "sample.donor.species.ontology": "9606"
          }
        }, {
          "match": {
            "assay.rna.spikein": "ERCC"
          }
        }],
        "should": [{
          "match": {
            "assay.single_cell.cell_handling": "FACs"
          }
        }, {
          "match": {
            "assay.single_cell.cell_handling": "mouth-pipetting"
          }
        }]
    }
  }
}
```

PR #463, opened 2017-09-18
Correcting Logic Error

Bento007:
- Removed undefined values from log messages
- Added useful information to log messages
- Added tests cases to verify exception.
- connects to #459

### DIFF
--- a/tests/test_subscriptions.py
+++ b/tests/test_subscriptions.py
@@ -107,6 +107,36 @@ class TestSubscriptionsBase(DSSAsserts):
         registered_query = response['_source']
         self.assertEqual(self.sample_percolate_query, registered_query)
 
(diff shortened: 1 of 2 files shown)


tests/test_subscriptions.py
```diff
@@ -107,6 +107,36 @@ def test_put(self):
...
+            url,
+            requests.codes.internal_server_error,
+            json_request_body=dict(
+                es_query=es_query,
+                callback_url=self.callback_url),
+            headers=self._get_auth_header()
+        )
+        self.assertIn("No field mapping can be found for the field with name [assay.fake_field]",
+                      resp_obj.json['stacktrace'])
+        self.assertEqual(resp_obj.json['code'], 'elasticsearch_error')
```
ttung: consider using the expected_error kwargs to `assert*Response`

PR #473, opened 2017-09-20
Add support for signing webhook payloads with HMAC

kislyuk:
Also:

- check against local resource access attacks in redirect URLs

- require https webhook URLs in production

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -332,7 +332,7 @@ paths:
                   - code
   /subscriptions:
     get:
-      summary: Retrieve an user's event subscriptions.
+      summary: Retrieve a user\'s event subscriptions.
       description: |
         Return a list of associated subscriptions, which contains the uuid,
         replica, query, creator_uid, and callback_url.
@@ -400,6 +400,14 @@ paths:
                 description: Url to send request to when a bundle comes in that matches this query.
                 type: string
                 format: url # Seems that this is possible now because of https://github.com/swagger-api/swagger-core/issues/1045
+                pattern: "http://.+"
+              secret:
+                description: |
+                  A key used to sign notifications sent to `callback_url`. The signature will be constructed according
+                  to https://tools.ietf.org/html/draft-cavage-http-signatures and transmitted in the Authorization
+                  header.
+                type: string
+                pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
             required:
               - es_query
               - callback_url
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -400,14 +400,16 @@ paths:
                 description: Url to send request to when a bundle comes in that matches this query.
                 type: string
                 format: url # Seems that this is possible now because of https://github.com/swagger-api/swagger-core/issues/1045
-                pattern: "http://.+"
-              secret:
+              hmac_secret_key:
                 description: |
                   A key used to sign notifications sent to `callback_url`. The signature will be constructed according
                   to https://tools.ietf.org/html/draft-cavage-http-signatures and transmitted in the Authorization
                   header.
                 type: string
-                pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
+              hmac_key_id:
+                description: |
+                  A key ID to use for `hmac_secret_key`.
+                type: string
             required:
               - es_query
               - callback_url
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -223,6 +223,7 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
     }
     callback_url = subscription['callback_url']
 
+    # FIXME wrap all errors in this block with an exception handler
     assert urlparse(callback_url).scheme in {"http", "https"}
     if os.environ["DSS_DEPLOYMENT_STAGE"] not in {"dev", "staging"}:
         hostname = urlparse(callback_url).hostname
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -224,16 +224,17 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
     callback_url = subscription['callback_url']
 
     # FIXME wrap all errors in this block with an exception handler
-    assert urlparse(callback_url).scheme in {"http", "https"}
+    assert urlparse(callback_url).scheme in {"http", "https"}, "Unexpected scheme for callback URL"
     if os.environ["DSS_DEPLOYMENT_STAGE"] == "prod":
         hostname = urlparse(callback_url).hostname
         for family, socktype, proto, canonname, sockaddr in socket.getaddrinfo(hostname, port=None):
-            assert ipaddress.ip_address(sockaddr[0]).is_global  # type: ignore
-        assert urlparse(callback_url).scheme == "https"
+            msg = "Callback hostname resolves to forbidden network"
+            assert ipaddress.ip_address(sockaddr[0]).is_global, msg  # type: ignore
+        assert urlparse(callback_url).scheme == "https", "Unexpected scheme for callback URL"
 
     auth = None
     if "hmac_secret_key" in subscription:
-        auth = HTTPSignatureAuth(key=subscription["hmac_secret_key"].encode(),
+        auth = HTTPSignatureAuth(key=subscription['hmac_secret_key'].encode(),
                                  key_id=subscription.get("hmac_key_id", "hca-dss:" + subscription_id))
     response = requests.post(callback_url, json=payload, auth=auth)
 
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -224,13 +224,18 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
     callback_url = subscription['callback_url']
 
     # FIXME wrap all errors in this block with an exception handler
-    assert urlparse(callback_url).scheme in {"http", "https"}, "Unexpected scheme for callback URL"
+    if os.environ["DSS_DEPLOYMENT_STAGE"] == "prod":
+        allowed_schemes = {'https'}
+    else:
+        allowed_schemes = {'https', 'http'}
+
+    assert urlparse(callback_url).scheme in allowed_schemes, "Unexpected scheme for callback URL"
+
     if os.environ["DSS_DEPLOYMENT_STAGE"] == "prod":
         hostname = urlparse(callback_url).hostname
         for family, socktype, proto, canonname, sockaddr in socket.getaddrinfo(hostname, port=None):
             msg = "Callback hostname resolves to forbidden network"
             assert ipaddress.ip_address(sockaddr[0]).is_global, msg  # type: ignore
-        assert urlparse(callback_url).scheme == "https", "Unexpected scheme for callback URL"
 
     auth = None
     if "hmac_secret_key" in subscription:
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -219,15 +219,12 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
             "bundle_version": bundle_version
         }
     }
-    #verify_signature(payload_body)
-    #signature = 'sha1=' + OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new('sha1'), ENV['SECRET_TOKEN'], payload_body)
-    #return halt 500, "Signatures didn't match!" unless Rack::Utils.secure_compare(signature, request.env['HTTP_X_HUB_SIGNATURE'])
-
     callback_url = subscription['callback_url']
 
     auth = None
-    if "hmac_secret" in subscription:
-        auth = HTTPSignatureAuth(key=subscription["hmac_secret"], key_id="hca-dss:" + subscription_id)
+    if "hmac_secret_key" in subscription:
+        auth = HTTPSignatureAuth(key=subscription["hmac_secret_key"].encode(),
+                                 key_id=subscription.get("hmac_key_id", "hca-dss:" + subscription_id))
     response = requests.post(callback_url, json=payload, auth=auth)
 
     # TODO (mbaumann) Add webhook retry logic
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -7,6 +7,7 @@ import uuid
 from urllib.parse import unquote
 
 import requests
+from requests_http_signature import HTTPSignatureAuth
 from elasticsearch.helpers import scan
 
 from dss import Config, ESIndexType, ESDocType, Replica
@@ -223,7 +224,12 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
     #return halt 500, "Signatures didn't match!" unless Rack::Utils.secure_compare(signature, request.env['HTTP_X_HUB_SIGNATURE'])
 
     callback_url = subscription['callback_url']
-    response = requests.post(callback_url, json=payload)
+
+    auth = None
+    if "hmac_secret" in subscription:
+        auth = HTTPSignatureAuth(key=subscription["hmac_secret"], key_id="hca-dss:" + subscription_id)
+    response = requests.post(callback_url, json=payload, auth=auth)
+
     # TODO (mbaumann) Add webhook retry logic
     if 200 <= response.status_code < 300:
         logger.info(f"Successfully notified for subscription {subscription_id}"
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -4,7 +4,9 @@ import json
 import os
 import re
 import uuid
-from urllib.parse import unquote
+import ipaddress
+import socket
+from urllib.parse import urlparse, unquote
 
 import requests
 from requests_http_signature import HTTPSignatureAuth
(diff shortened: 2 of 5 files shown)


dss/events/handlers/index.py
```diff
@@ -218,9 +221,21 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
...
     }
-    # TODO (mbaumann) Ensure webhooks are only delivered over verified HTTPS (unless maybe when running a test)
     callback_url = subscription['callback_url']
-    response = requests.post(callback_url, json=payload)
+
+    assert urlparse(callback_url).scheme in {"http", "https"}
+    if os.environ["DSS_DEPLOYMENT_STAGE"] not in {"dev", "staging"}:
+        hostname = urlparse(callback_url).hostname
+        for family, socktype, proto, canonname, sockaddr in socket.getaddrinfo(hostname, port=None):
+            assert ipaddress.ip_address(sockaddr[0]).is_global  # type: ignore
```
Bento007: What is this line checking for?

kislyuk: It is checking that the webhook subscription isn't trying to make us send a notification to a local network address (not publicly routable, like https://127.0.0.1).

dss-api.yml
```diff
@@ -343,6 +343,16 @@ paths:
                 description: Url to send request to when a bundle comes in that matches this query.
                 type: string
                 format: url # Seems that this is possible now because of https://github.com/swagger-api/swagger-core/issues/1045
+              hmac_secret_key:
+                description: |
+                  A key used to sign notifications sent to `callback_url`. The signature will be constructed according
+                  to https://tools.ietf.org/html/draft-cavage-http-signatures and transmitted in the Authorization
+                  header.
+                type: string
+              hmac_key_id:
```
Bento007: Should hmac_key_id be requried?

dss-api.yml
```diff
@@ -343,6 +343,16 @@ paths:
                 description: Url to send request to when a bundle comes in that matches this query.
                 type: string
                 format: url # Seems that this is possible now because of https://github.com/swagger-api/swagger-core/issues/1045
+              hmac_secret_key:
```
Bento007: Should hmac_secret_key be a required parameter?

dss/events/handlers/index.py
```diff
@@ -218,9 +221,21 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
...
+        assert urlparse(callback_url).scheme == "https"
+
+    auth = None
+    if "hmac_secret_key" in subscription:
+        auth = HTTPSignatureAuth(key=subscription["hmac_secret_key"].encode(),
```
ttung: why would we have a secret key without a keyid?

tests/test_indexer.py
```diff
@@ -445,8 +447,17 @@ def _get_version(self):
...
     _payload = None
+    hmac_secret_key = "ribos0me"
 
     def do_POST(self):
+        HTTPSignatureAuth.verify(requests.Request("POST", self.path, self.headers),
+                                 key_resolver=lambda key_id, algorithm: self.hmac_secret_key.encode())
+        try:
+            HTTPSignatureAuth.verify(requests.Request("POST", self.path, self.headers),
+                                     key_resolver=lambda key_id, algorithm: self.hmac_secret_key[::-1].encode())
+            raise Exception("Expected AssertionError")
```
ttung: should just be `self.fail(..)`

tests/test_indexer.py
```diff
@@ -445,8 +447,17 @@ def _get_version(self):
 class PostTestHandler(BaseHTTPRequestHandler):
     _response_code = 200
     _payload = None
+    hmac_secret_key = "ribos0me"
 
     def do_POST(self):
+        HTTPSignatureAuth.verify(requests.Request("POST", self.path, self.headers),
+                                 key_resolver=lambda key_id, algorithm: self.hmac_secret_key.encode())
+        try:
```
ttung: with `self.assertRaises`

dss/events/handlers/index.py
ttung: Are there any AWS resources that authenticate based solely on source ip?

kislyuk: Not that I know of, but I think it should be pretty unambiguously not allowed to try to send webhook deliveries to host-local or local-subnet endpoints in prod. I haven't found an effective attack vector but it seems quite dangerous. I would like to restrict this condition to prod only but since we don't have any official nomenclature for the prod deployment stages, I check if we're not in dev or staging. I suppose this breaks custom stages, so maybe I should change this to "prod" even though that's error prone if we ever deploy a production deployment that is not called "prod".

dss-api.yml
```diff
@@ -343,6 +343,16 @@ paths:
...
+                description: |
+                  A key used to sign notifications sent to `callback_url`. The signature will be constructed according
+                  to https://tools.ietf.org/html/draft-cavage-http-signatures and transmitted in the Authorization
```
mikebaumann: Thanks for including the reference. Nice/helpful description.

dss/events/handlers/index.py
```diff
@@ -218,9 +221,21 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
             "bundle_version": bundle_version
         }
     }
-    # TODO (mbaumann) Ensure webhooks are only delivered over verified HTTPS (unless maybe when running a test)
     callback_url = subscription['callback_url']
-    response = requests.post(callback_url, json=payload)
+
+    assert urlparse(callback_url).scheme in {"http", "https"}
```
mikebaumann: If a webhook notification fails, it should be clear from the Lambda log why it failed. The caller already performs that logging, yet using `assert` only throws an `AssertionError` with no message indicating the cause of the error / notification failure. I know there is sometimes information that should not be logged for security reasons, yet it seems in the cases where assert is used here, the general cause could be provided without logging sensitive information. Please throw exceptions with a message that will providing meaningful/helpful logging.

dss/events/handlers/index.py
```diff
@@ -218,9 +221,21 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
...
+    if os.environ["DSS_DEPLOYMENT_STAGE"] not in {"dev", "staging"}:
```
mikebaumann: Is the intention here to ensure that for production deployments the address type is global and the scheme is "https"? If so, can/should we test for that directly using `os.environ["DSS_DEPLOYMENT_STAGE"] in {"prod"}`?
One reason I ask is that at UCSC we use custom stage names extensively (e.g. in my environment "DSS_DEPLOYMENT_STAGE" is set to "mbaumann", the custom stage name to which I deploy often), which then causes test_indexer.py to fail.
I suppose eventually/ideally there would a specific configuration setting indicating if this webhook validation should be performed, yet I don't think that is a required at this point.

dss/events/handlers/index.py
mikebaumann: My understanding from ttung is our convention is to use single quotes for dict key names, and double quotes for all other strings.

tests/test_indexer.py
```diff
@@ -286,7 +287,8 @@ def subscribe_for_notification(self, es_query, callback_url):
             requests.codes.created,
             json_request_body=dict(
                 es_query=es_query,
-                callback_url=callback_url),
+                callback_url=callback_url,
+                hmac_secret_key=PostTestHandler.hmac_secret_key),
```
mikebaumann: Do we expect that in production all subscriptions will always use HMAC?
If so, then I'm OK with only testing this case.
Otherwise, I think there should be unit tests for both the case where HMAC is configured in the subscription, and the case in which it is not configured in the subscription.

dss/events/handlers/index.py
```diff
@@ -218,9 +221,21 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
...
+        hostname = urlparse(callback_url).hostname
+        for family, socktype, proto, canonname, sockaddr in socket.getaddrinfo(hostname, port=None):
+            assert ipaddress.ip_address(sockaddr[0]).is_global  # type: ignore
+        assert urlparse(callback_url).scheme == "https"
```
mikebaumann: The conditions asserted here can/should also be checked at the time the subscription is registered, and the registration should be rejected if these conditions are not me. It is preferable to not allow an invalid subscription into the system than to allow a subscription for which the notifications will always fail.
@Bento007/I can help with these changes if you agree with the above.

tests/test_indexer.py
mikebaumann: @Bento007/I can help by adding the test for the case where HMAC is not configured, if you agree this case should be tested.

dss/events/handlers/index.py
```diff
@@ -4,9 +4,12 @@
 import os
 import re
 import uuid
-from urllib.parse import unquote
+import ipaddress
+import socket
+from urllib.parse import urlparse, unquote
 
 import requests
+from requests_http_signature import HTTPSignatureAuth
```
mikebaumann: @Bento007 observed that `requests_http_signature` is a recently developed package related to this PR. Should it also be reviewed? @Bento007 has an interest in security and could be a good candidate to do the review. This could also help develop broader knowledge within the team regarding our security mechanisms.

kislyuk: Sure, review as much as you like.

dss/events/handlers/index.py
kislyuk: Actually, they should be checked in both places.

tests/test_indexer.py
kislyuk: Yes, I need to add a test for that.

dss/events/handlers/index.py
mikebaumann: Yes, that is what I intended to communicate. Thanks!

dss/events/handlers/index.py
kislyuk: Yes, I will change it to read `{"prod"}`. The difficulty is that we have not "officially" deployed prod yet, so we just need to nail down that we will in fact name it `prod`.

dss/events/handlers/index.py
kislyuk: Will fix.

tests/test_indexer.py
kislyuk: This is not a unittest.TestCase subclass.

tests/test_indexer.py
kislyuk: This is not a unittest.TestCase subclass.

Bento007: unittest.TestCase is inherited at run time through it's children.

kislyuk: I'm very confused about what you mean, @Bento007...

Bento007: My mistake, I thought this was a modification to `TestIndexerBase` which is an abstract class that uses unittest.TestCase asserts.

dss/events/handlers/index.py
kislyuk: Changed the stage check to compare against "prod".

dss-api.yml
kislyuk: No, in cases where the user doesn't want to deal with HMAC-based authentication.

dss/events/handlers/index.py
kislyuk: Key IDs exist here to aid the client in identifying the key. When the user doesn't need to identify their HMAC key, they might omit this parameter.

dss/events/handlers/index.py
kislyuk: Added more informative assert messages.

dss/events/handlers/index.py
kislyuk: Changed to compare to "prod".

dss/events/handlers/index.py
kislyuk: In Python, there is no convention about which quoting styles to use where. It's important to be consistent though, so I've changed this to single quotes.

tests/test_indexer.py
kislyuk: Test added.

dss/events/handlers/index.py
kislyuk: Filed #550 to deal with this later.

dss/events/handlers/index.py
```diff
@@ -218,9 +221,23 @@ def notify(subscription_id: str, subscription: dict, bundle_id: str, logger):
             "bundle_version": bundle_version
         }
     }
-    # TODO (mbaumann) Ensure webhooks are only delivered over verified HTTPS (unless maybe when running a test)
     callback_url = subscription['callback_url']
-    response = requests.post(callback_url, json=payload)
+
+    # FIXME wrap all errors in this block with an exception handler
+    assert urlparse(callback_url).scheme in {"http", "https"}, "Unexpected scheme for callback URL"
```
ttung: suggestion:

```
if os.environ["DSS_DEPLOYMENT_STAGE"] == "prod":
    allowed_schemes = {'https'}
else:
    allowed_schemes = {'https', 'http'}

assert urlparse(callback_url).scheme in allowed_schemes, "Unexpected scheme for callback URL"
```

ttung: Also these assertions are untested.

tests/test_indexer.py
```diff
@@ -195,6 +196,25 @@ def test_indexed_file_access_error(self):
                                                          excluded_files=[inaccesssible_filename.replace(".", "_")])
 
     def test_subscription_notification_successful(self):
+        PostTestHandler.verify_payloads = True
```
ttung: There's a lot of code duplication among these test cases.  For instance, `test_subscription_notification_successful` and `test_unsigned_ subscription_notification_successful` only differ in two places.  Please consider simplifying.

Also, `PostTestHandler.verify_payloads` is global state that needs to be unset if the test fails for some reason.

tests/test_indexer.py
```diff
@@ -461,6 +490,7 @@ def do_POST(self):
     def reset(cls):
         cls._response_code = 200
         cls._payload = None
+        cls.authenticated_requests = 0
```
ttung: where is this used?

tests/test_indexer.py
kislyuk: I have tried to make these into one test with no code duplication, but because `self.verify_notification` assumes that only one notification is ever active at a given time, I'm unable to do so.

kislyuk: I've fixed `PostTestHandler.reset()` to reset verify_payloads instead of the other variable (which was unused, so that was a bug).

tests/test_indexer.py
kislyuk: Sorry, that was unused and not cleaned up. Fixed.

tests/test_indexer.py
ttung: Why not have:

```
    def test_subscription_notification_successful(self):
        self._test_subscription_notification_successful_helper(
            True,
            dict(hmac_secret_key=PostTestHandler.hmac_secret_key, hmac_key_id="test"),
        )

    def test_unsigned_subscription_notification_successful(self):
        self._test_subscription_notification_successful_helper(
            False,
            dict(),
        )

    def _test_subscription_notification_successful_helper(self, verify_payloads, subscribe_kwargs):
        PostTestHandler.verify_payloads = verify_payloads
        bundle_key = self.load_test_data_bundle_for_path("fixtures/smartseq2/paired_ends")
        sample_event = self.create_sample_bundle_created_event(bundle_key)

        ElasticsearchClient.get(logger).indices.create(self.subscription_index_name)
        subscription_id = self.subscribe_for_notification(smartseq2_paired_ends_query,
                                                          f"http://{HTTPInfo.address}:{HTTPInfo.port}",
                                                          **subscribe_kwargs)

        bundle_key = self.load_test_data_bundle_for_path("fixtures/smartseq2/paired_ends")
        sample_event = self.create_sample_bundle_created_event(bundle_key)
        prefix, _, bundle_id = bundle_key.partition("/")
        self.verify_notification(subscription_id, smartseq2_paired_ends_query, bundle_id)
```

dss/events/handlers/index.py
kislyuk: Thanks, I used your suggestion and added unit tests for `notify()`.

tests/test_indexer.py
kislyuk: To address your code duplication concern, I added a utility function to delete a subscription, and unified these tests into one test that duplicates less code.

tests/test_indexer.py
```diff
@@ -194,29 +195,73 @@ def test_indexed_file_access_error(self):
...
+        def _notify(subscription, bundle_id="i.v"):
+            notify(subscription_id=subscription["id"], subscription=subscription, bundle_id=bundle_id, logger=logger)
+        with self.assertRaisesRegex(requests.exceptions.InvalidURL, "Invalid URL 'http://': No host supplied"):
+            _notify(subscription=dict(id="", es_query={}, callback_url="http://"))
+        with self.assertRaisesRegex(AssertionError, "Unexpected scheme for callback URL"):
+            _notify(subscription=dict(id="", es_query={}, callback_url=""))
+        with self.assertRaisesRegex(AssertionError, "Unexpected scheme for callback URL"):
+            _notify(subscription=dict(id="", es_query={}, callback_url="wss://localhost"))
+        try:
+            environ_backup = os.environ
```
ttung: technically, you should have this statement *before* the try, but if this statement causes an exception, i guess you're screwed anyway.

tests/test_indexer.py
```diff
@@ -194,29 +195,73 @@ def test_indexed_file_access_error(self):
...
+            os.environ = environ_backup
+
+    def delete_subscription(self, subscription_id):
+        self.assertDeleteResponse(
+            str(UrlBuilder().set(path=f"/v1/subscriptions/{subscription_id}").add_query("replica", self.replica)),
+            requests.codes.ok,
+            headers=self.get_auth_header()
+        )
+
     def test_subscription_notification_successful(self):
```
ttung: `test_subscription_notification_successful` still has a fair amount of internal duplication.  and now that you've collapsed it into one method, it should be even easier to dedup.

```
    for verify_payloads, subscribe_kwargs in ((True, dict(hmac_secret_key=PostTestHandler.hmac_secret_key)),
                                              (False, dict()):
        PostTestHandler.verify_payloads = verify_payloads
        bundle_key = self.load_test_data_bundle_for_path("fixtures/smartseq2/paired_ends")
        sample_event = self.create_sample_bundle_created_event(bundle_key)

        ElasticsearchClient.get(logger).indices.create(self.subscription_index_name)
        subscription_id = self.subscribe_for_notification(smartseq2_paired_ends_query,
                                                          f"http://{HTTPInfo.address}:{HTTPInfo.port}",
                                                          **subscribe_kwargs)

        bundle_key = self.load_test_data_bundle_for_path("fixtures/smartseq2/paired_ends")
        sample_event = self.create_sample_bundle_created_event(bundle_key)
        prefix, _, bundle_id = bundle_key.partition("/")
        self.verify_notification(subscription_id, smartseq2_paired_ends_query, bundle_id)
```

kislyuk: Deduplicated code per your suggestion.

PR #475, opened 2017-09-22
Replace the inline chunkedtask code with the chained-aws-lambda library.

ttung:
1. Lots of red lines to blow away the old code (and tests).
2. New deployment mechanism that generates a top-level app.py and IAM policy template.
3. Customization of the IAM policy template to allow the proper S3 bucket access.

Connects to #393


dss/api/files.py
```diff
@@ -3,6 +3,8 @@
 import json
 import re
 import typing
+
+from chainedawslambda import aws
```
kislyuk: The name of this import is not very specific, I'm just concerned it might be a bit confusing.

requirements.txt
```diff
@@ -8,6 +8,7 @@ botocore==1.6.1
 cachetools==2.0.1
 certifi==2017.7.27.1
 cffi==1.10.0
+chained-aws-lambda==0.0.3
```
kislyuk: https://pypi.python.org/pypi/chained-aws-lambda publishes 0.0.3 but there are no tags in https://github.com/chanzuckerberg/chained-aws-lambda, making it hard to review the release. Consider using https://github.com/HumanCellAtlas/data-store-cli/blob/master/common.mk to cut releases with automatic tagging and annotation.

(Also, not urgent or blocking, but the docs in that repo need more updating/cleanup)

daemons/Makefile
```diff
@@ -13,19 +13,28 @@ dss-sync dss-index:
...
+	cp -rf chained-aws-lambda $@
+	chained_aws_lambda.py '$$account_id' \
+	  --policy-path $@/policy.json.template \
+	  --app-path $@/app.py \
+	  --lambda-name dss-$*-$(DSS_DEPLOYMENT_STAGE)
+
+	cat chained-aws-lambda/.chalice/config.json | jq '.app_name="dss-$*"' > $@/.chalice/config.json
+	# grant access to $DSS_S3_BUCKET and $DSS_S3_BUCKET_TEST.  in build_deploy_config.sh, this gets replaced with the
+	# actual bucket names.
+	cat $@/policy.json.template | jq -f s3_enable.filter | sponge $@/policy.json.template
```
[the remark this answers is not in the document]
ttung: Rationale: the library generates the policy statements the library itself requires, and then any consumers of the library adds to the policy as necessary.

daemons/Makefile
```diff
@@ -13,19 +13,28 @@ dss-sync dss-index:
...
 	shopt -s nullglob; for wheel in $@/vendor.in/*/*.whl; do unzip -q -o -d $@/vendor $$wheel; done
-	cp -R ../dss ../dss-api.yml $@/domovoilib
+	cp -R ../dss $@/domovoilib
 	cp "$(GOOGLE_APPLICATION_CREDENTIALS)" $@/domovoilib/gcp-credentials.json
-	./build_deploy_config.sh $@ $(DSS_DEPLOYMENT_STAGE)
-	cd $@; DSS_CHUNKED_TASK_CLIENT_NAME=$* domovoi deploy --stage $(DSS_DEPLOYMENT_STAGE)
+	policy_template=$@/policy.json.template ./build_deploy_config.sh $@ dev
+	cd $@; domovoi deploy --stage dev
```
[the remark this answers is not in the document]
ttung: Bug!  Good catch.

dss/api/files.py
ttung: What about just importing chainedawslambda.aws and use the full namespace?

requirements.txt
ttung: I bumped to v0.0.5 which has a tag.

I agree, the docs are in terrible shape. :(

dss/api/files.py
kislyuk: 👍

PR #487, opened 2017-09-27
added issues template for creating new issues

Bento007:
This should improve the quality of our issue reporting. We should evolve this document over time.


ttung:
The only suggestion I have is to tell people to enumerate dependent/dependency relationships, i.e., between PRs and between issues.

Issue #399, opened 2017-08-31
Timeouts on API requests


ttung:

ttung:
Looked at a bunch of logs, still pending.

kozbo:
blocked on #460

PR #489, opened 2017-09-27
Monitor and return an error message if we exceed the time permitted.

ttung:
1. Move the work of processing a request into a separate thread.
2. If the time permitted runs out, we return a standardized error message.
3. Attach the stacktraces of the remaining threads for debugging purposes.

Connects to #399 
Depends on #507, #508

### DIFF
--- a/tests/test_exptime.py
+++ b/tests/test_exptime.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import os
+import sys
+import unittest
+
+import requests
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from tests.infra import DSSAssertMixin, DSSUploadMixin, ExpectedErrorFields
+from tests.infra.server import ThreadedLocalServer
+
+
+class TestExptime(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
+    @classmethod
+    def setUpClass(cls):
+        cls.app = ThreadedLocalServer()
+        cls.app.start()
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.app.shutdown()
+
+    def test_exptime(self):
+        self.assertGetResponse(
+            "/internal/slow_request",
+            requests.codes.gateway_timeout,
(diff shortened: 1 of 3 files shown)


kislyuk:
What do you think about spawning another process to send ourselves a signal on a timer instead (like SIGUSR1) and setting up the signal handler to perform the shutdown and error report?

I bring this up because you previously said that you consider Python threading to be flaky under some circumstances. So the signal-on-a-timer solution would avoid relying on that.

ttung:
> I bring this up because you previously said that you consider Python threading to be flaky under some circumstances. So the signal-on-a-timer solution would avoid relying on that.

The issues I've encountered with python is around multiprocessing.  I've not done much with python threading because of the GIL. :)

Generically, I've encountered issues with signals and exec in a multithreaded environment, but those are general unix problems.

kislyuk:
This approach looks fine to me but I don't see why we shouldn't apply it by default to all methods. Can you change it to be a default middleware?

ttung:
> This approach looks fine to me but I don't see why we shouldn't apply it by default to all methods. Can you change it to be a default middleware?

I'm not sure what you mean.  Applying it by default to all methods in chalice would require mucking around with code "internal" (i.e., prefixed with underscore) to chalice or subclassing the Chalice class which is brittle IMO.  `dispatch()` in `chalice/app.py` is the trampoline for all the connexion-powered endpoints, so it handles everything defined in `dss-api.yml`, so all those endpoints are covered by this change.

tests/test_exptime.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import os
+import sys
+import unittest
+
+import requests
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
```
Bento007: what does `# noqa` mean?

ttung: one is not supposed to have code statements before all the imports, so this is flagged as a lint error.

kislyuk: @Bento007, I'm not sure if you were asking about what `#noqa` means or why the tag is on this line. The noqa tag suppresses linter (flake8) errors on this line, which would normally cause tests to fail.

PR #492, opened 2017-09-28
make PUT /files idempotent.

ttung:
This makes it easier to manage API clients that have to do retries.  Idempotence means that they can just blindly try until they succeed.

It is idempotent as long as the parameters are identical.

This depends on #491.

Connects to #399

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -240,6 +240,17 @@ paths:
               - bundle_uuid
               - creator_uid
       responses:
+        200:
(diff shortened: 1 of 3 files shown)


ttung:
Now returns 200 when the file exists with identical content.  Updated `dss-api.yml` to reflect this change.

kislyuk:
LGTM but the test failed, it seems for an unrelated reason.

PR #499, opened 2017-10-02
Add chained-aws-lambda to the set of lambdas that the ci-cd user can interact with

ttung:


ttung:
Ok this can't possibly break the build, so I'm landing this.

Issue #490, opened 2017-09-28
dss in unit test should be spawned as it is on lambda / dss-api script.


ttung:
Right now, we're invoking a flask test client.  We should invoke it through chalice to match how it is launched on lambda or the dss-api script.  Here's an example of how it can be done: https://github.com/aws/chalice/blob/master/tests/integration/test_local.py

PR #507, opened 2017-10-06
Start the server up for unit tests.

ttung:
The infrastructure for managing requests that run too long live at the chalice layer.  However, the unit tests currently just invoke the flask app directly.  To build unit tests against that, we need to start up a chalice server and run all the unit tests against that.

Connects to #490 
Depends on #506

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -818,4 +818,4 @@ definitions:
     properties:
       files:
         type: string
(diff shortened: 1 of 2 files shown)


kislyuk:
I was hoping we could have avoided running the server out-of-process... :(

PR #508, opened 2017-10-06
Send a content-length header to prevent a deadlock.

ttung:
If we don't send a content-length, the python requests library will sit there and continue to try to read until the network connection is severed.  This is happening because (after some of my other changes), there's a library that's polluting the http.server globals (I haven't tracked this down) and enabling HTTP/1.1, which supports resumes.

### DIFF
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -452,6 +452,7 @@ class PostTestHandler(BaseHTTPRequestHandler):
 
     def do_POST(self):
         self.send_response(self._response_code)
+        self.send_header("Content-length", "0")
         self.end_headers()
         length = int(self.headers['content-length'])
         if length:


kislyuk:
We should bisect the tree to check which library brought this uncool behavior in.

Bento007:
Is there an open issue to resolve this?

ttung:
It's not a specific library.  Let me try to figure out why this is happening.

ttung:
This behavior got added in chalice at https://github.com/aws/chalice/commit/ab76cb4b13af7846bd7085e3d00834dbd41d3f5d#diff-01f6949cdd7c8ec1fea7ca7554dace6b

Removed at https://github.com/aws/chalice/commit/73050523d4ded9c786a1d592022ba273a67e089a#diff-01f6949cdd7c8ec1fea7ca7554dace6b

We can upgrade chalice to solve the issue, I suspect.

ttung:
It makes a lot of sense that it's a chalice issue, since my PR chain is about adding chalice.  Unfortunately, that means we are no closer to resolving the random hang issue.

kislyuk:
Approved and yes, let's upgrade Chalice.

Issue #471, opened 2017-09-19
Manifest response should include file size


caaespin:
I think the current manifest response should include the file sizes of the files in the bundle. This would be really helpful when calling the system and creating an index for various purposes.

ttung:
That sounds fine to me. Send me a PR and I can approve.

xbrianh:
Including size in manifests requires population of a size metadata fields in dcp-cli and staging-service.

kislyuk:
Update from sprint review: need to decide which code path is responsible for injecting size into the file metadata in the blue box. Consider possible solutions with @ttung

PR #509, opened 2017-10-06
Include size in test fixture metadata tags

xbrianh:
Adding filesize tag in test fixture metadata, along with
checksums and content type, as first step in including
filesize in file and bundle manifests.

connects to #471


tests/fixtures/populate_lib.py
```diff
@@ -14,21 +14,22 @@ def upload(uploader: Uploader):
...
-            "hca-dss-content-type": "text/plain",
+            "hca-dss-content-type": "text/plain"
         }
     )
     uploader.upload_file(
         "9cdc9050cecf59381fed55a2433140b69596fc861bee55abeafd1f9150f3e2da",
         "test_good_source_data/incorrect_case_checksum",
         {
             "hca-dss-content-type": "text/plain",
+            "hca-dss-size": "8685",
```
ttung: I'd have `upload_file` compute the size.  There's no compelling reason we need to do it manually.

tests/fixtures/datafiles/011c7340-9b3c-4d62-bf49-090d79daf198.2017-06-20T214506.766634Z
```diff
@@ -1 +1 @@
-{"format": "0.0.1", "version": "2017-06-20T214506.766634Z", "files": [{"name": "LICENSE", "uuid": "ce55fd51-7833-469b-be0b-5da88ebebfcd", "version": "2017-06-16T193604.240704Z", "content-type": "text/plain", "indexed": false, "crc32c": "e16e07b9", "s3-etag": "3b83ef96387f14655fc854ddc3c6bd57", "sha1": "2b8b815229aa8a61e483fb4ba0588b8b6c491890", "sha256": "cfc7749b96f63bd31c3c42b5c471bf756814053e847c10f3eb003417bc523d30"}], "creator_uid": 12345}
\ No newline at end of file
+{"format": "0.0.1", "version": "2017-06-20T214506.766634Z", "files": [{"name": "LICENSE", "uuid": "ce55fd51-7833-469b-be0b-5da88ebebfcd", "version": "2017-06-16T193604.240704Z", "content-type": "text/plain", "size": 11358, "indexed": false, "crc32c": "e16e07b9", "s3-etag": "3b83ef96387f14655fc854ddc3c6bd57", "sha1": "2b8b815229aa8a61e483fb4ba0588b8b6c491890", "sha256": "cfc7749b96f63bd31c3c42b5c471bf756814053e847c10f3eb003417bc523d30"}], "creator_uid": 12345}
```
ttung: The format of these files should increment.

tests/fixtures/cloud_uploader.py
```diff
@@ -88,6 +84,16 @@ def upload_file(
             multipart_chunksize=chunk_sz,
         )
 
+        if metadata_keys is None:
```
ttung: You only need to set one (tags vs metadata_keys), not both.

I personally prefer tags because they are immutable, whereas metadata keys are mutable.

tests/fixtures/cloud_uploader.py
```diff
@@ -128,8 +134,11 @@ def upload_file(
             *args,
             **kwargs) -> None:
         logger.info("%s", f"Uploading {local_path} to gs://{self.bucket.name}/{remote_path}")
+        fp = os.path.join(self.local_root, local_path)
         blob = self.bucket.blob(remote_path)
-        blob.upload_from_filename(os.path.join(self.local_root, local_path))
+        blob.upload_from_filename(fp)
         if metadata_keys:
```
ttung: I think you have to make this unconditional.  If the caller doesn't specify metadata_keys, you should still set the size.

tests/fixtures/populate_lib.py
```diff
@@ -14,14 +14,14 @@ def upload(uploader: Uploader):
         "cfc7749b96f63bd31c3c42b5c471bf756814053e847c10f3eb003417bc523d30",
         "test_good_source_data/0",
         {
-            "hca-dss-content-type": "text/plain",
```
ttung: please revert the changes in this file.

tests/fixtures/cloud_uploader.py
xbrianh: Then size metadata will be included on manifests in fixtures/datafiles/*, which we may not want?

tests/fixtures/cloud_uploader.py
xbrianh: Cool that tags are immutable :)

btw, tags and metadata_keys args were never populated simultaneously in `populate_lib.py`, as far as I can tell. The test `test_file_put_metadata_from_tags` passes an empty dict to metadata_keys, which is the design pattern I was following along with. This is not an argument in favor of my change, just a statement of rationale.

tests/fixtures/cloud_uploader.py
ttung: Oh you are right. Hmmm, I’m not sure what the best way to proceed is. I forgot the manifests also get uploaded through this.

My preferred option is to add an optional flag to upload_file that turns on setting the size field, but I’m open to ideas.

xbrianh: Mostly I'm agnostic... The other option seems to be including size when other fields are present, which involves fewer changes but is not explicit. Are there use cases for metadata fields *without* size?

ttung: No, there is no use case for metadata fields without size.  Let's not dwell on this.  Document this behavior in `upload_file` and let's move on.

tests/fixtures/cloud_uploader.py
```diff
@@ -128,8 +133,12 @@ def upload_file(
...
```
ttung:
```
>>> a = {}
>>> if a:
...     print("Hello")
... 
```

i would suggest `if metadata_keys is not None:`

xbrianh: Uff, perhaps should check for len(metadata_keys) given my log comment.

ttung:
```
>>> len(None)
  File "<stdin>", line 1, in <module>
TypeError: object of type 'NoneType' has no len()
>>> 
```

xbrianh: Maybe overthinking, but passing in `metadata_keys={}` means that `metadata_keys is not None` evaluates as True.

Perhaps:
def upload_file(...):
    if metadata_keys is None:
        metadata_keys = dict()
    ...
    if len(metadata_keys):
        # Add size tag when other metadata is present
        sz = os.stat(fp).st_size
        metadata_keys['hca-dss-size'] = str(sz)
    blob.metadata = metadata_keys
    blob.patch()

ttung: Yeah, that's fine.  If you want to do that, I fully support it. :)

Issue #511, opened 2017-10-06
Create Issue and PR template


Bento007:
Create a template for issues and pull requests to ensure important information is included and all best practice steps are followed. This will be helpful for new team members getting started and existing team members to follow our best practices.

PR #512, opened 2017-10-06
Minimal suitable issue and PR template

kislyuk:
This is an update for #487.

Connects to #511


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
...
+<!-- If your issue is about a feature, please add the "enhancement" tag. -->
+<!-- Check https://waffle.io/HumanCellAtlas/data-store for other tags in use and decorate your issue as appropriate. -->
+
+<!-- If you are filing a bug, please summarize the expected behavior, actual behavior, and steps to reproduce the bug.
+Uncomment the headings below:
+### Expected behavior
+### Actual behavior
+### Steps to reproduce the behavior -->
+
+<!-- If you are on the core dev team for this repo, mention your issue in daily scrum, sprint review, or to the PM. -->
```
Bento007: I suggest we add "the new issue must also be assigned to someone or yourself."

kislyuk: 👎 it's often useful to file something for later triage. All issues in the sprint should be assigned, but that doesn't necessarily hold true for the backlog.

kislyuk: Changing to obtain consensus

Bento007: Then I suppose all issue in the backlog should go to the scrummaster or PM do determine priority.

kozbo: Two comments:
First, it doesn't add a lot of functionality to assign all un-triaged issues to the scrum-master. I have worked that way in the past and issues get ignored either way. The key is to have a process to scrub the backlog on a regular schedule.
Second, I think we should add some information here about which build the problem was discovered in. I suggest branch, hash, build date?

kislyuk: @kozbo I like your ideas, can you please edit the issue template accordingly?

PR #515, opened 2017-10-07
added total matching results to search response.

Bento007:
Search will now return the `count` which is equal to the number of results matching the `es_query`. The search testcase have been updated to check the total count is equal to number of documents indexed.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -84,6 +84,9 @@ paths:
               results:
                 description: Results matching the `es_query`.
                 type: array
+              count:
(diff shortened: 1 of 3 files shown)


dss-api.yml
```diff
@@ -84,6 +84,9 @@ paths:
               results:
                 description: Results matching the `es_query`.
                 type: array
+              count:
```
ttung: consider making this `total_hits` as that is considerably less ambiguous.

Bento007: sounds good

PR #517, opened 2017-10-07
Added response partial to search.

Bento007:
When a search is executed and additional results available Partial (206) is returned with a link in the header with the next url. When all results are returned OK (200) is returned and no link is in the response header.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -74,11 +74,11 @@ paths:
           type: string
       responses:
         200:
-          description: OK all results retrieved
+          description: All results retrieved.
           schema:
             $ref: '#/definitions/SearchResult'
         206:
-          description: OK page of result retrieved
+          description: Page of result retrieved.
           schema:
             $ref: '#/definitions/SearchResult'
           headers:
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -78,13 +78,13 @@ paths:
           schema:
             $ref: '#/definitions/SearchResult'
         206:
-          description: Page of result retrieved.
+          description: A single page of results was retrieved.
           schema:
             $ref: '#/definitions/SearchResult'
           headers:
             Link:
               type: string
-              description: URL to retrieve paginated results.
+              description: URL to retrieve the next page of results.
         default:
           description: Unexpected error
           schema:
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -74,23 +74,17 @@ paths:
           type: string
       responses:
         200:
-          description: OK
+          description: OK all results retrieved
           schema:
-            type: object
-            properties:
-              es_query:
-                description: Elasticsearch query used to produce the results.
-                type: object
-              results:
-                description: Results matching the `es_query`.
-                type: array
-              total_hits:
-                description: The total number of matching results found.
-                type: integer
+            $ref: '#/definitions/SearchResult'
+        206:
+          description: OK page of result retrieved
+          schema:
+            $ref: '#/definitions/SearchResult'
           headers:
             Link:
               type: string
-              description: URL to retrieve paginated results in [RFC 5988](https://tools.ietf.org/html/rfc5988) format.
+              description: URL to retrieve paginated results.
         default:
           description: Unexpected error
           schema:
@@ -832,6 +826,18 @@ definitions:
         type: integer
         format: int
         description: User ID who created this bundle manifest.
+  SearchResult:
+    type: object
+    properties:
+      es_query:
+        description: Elasticsearch query used to produce the results.
+        type: object
+      results:
+        description: Results matching the `es_query`.
+        type: array
+      total_hits:
+        description: The total number of matching results found.
+        type: integer
   Error:
     type: object
     properties:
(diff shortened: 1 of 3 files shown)


dss-api.yml
```diff
@@ -71,20 +71,17 @@ paths:
           type: string
       responses:
         200:
-          description: OK
+          description: OK all results retrieved
```
ttung: I think you can omit the "OK"

dss-api.yml
```diff
@@ -71,20 +71,17 @@ paths:
...
-            properties:
-              es_query:
-                description: Elasticsearch query used to produce the results.
-                type: object
-              results:
-                description: Results matching the `es_query`.
-                type: array
+            $ref: '#/definitions/SearchResult'
+        206:
+          description: OK page of result retrieved
```
ttung: ditto.

tests/test_search.py
```diff
@@ -76,8 +77,8 @@ def test_search_post(self):
             path=url,
             json_request_body=dict(es_query=smartseq2_paired_ends_query),
             expected_code=requests.codes.ok)
-        next_url = self.get_next_url(search_obj.response.headers['Link'])
-        self.assertEqual(next_url, '')
+        next_url = self.get_next_url(search_obj.response.headers)
+        self.assertEqual(next_url, None)
```
ttung: `assertIsNone`

tests/test_search.py
```diff
@@ -333,20 +335,20 @@ def get_search_results(self, es_query, url_params=None):
         search_obj = self.assertPostResponse(
```
ttung: I feel you can combine this with:

```
while True:
    search_for_stuff(..)
    if search_obj.response.status_code == 200:
        break
```

Or am I completely off base?

dss-api.yml
Bento007: Sounds good, will fix.

dss-api.yml
Bento007: Will Fix

tests/test_search.py
Bento007: will fix

tests/test_search.py
Bento007: That's work. Thanks!

dss-api.yml
```diff
@@ -71,20 +71,17 @@ paths:
...
-            properties:
-              es_query:
-                description: Elasticsearch query used to produce the results.
-                type: object
-              results:
-                description: Results matching the `es_query`.
-                type: array
+            $ref: '#/definitions/SearchResult'
+        206:
+          description: Page of result retrieved.
```
ttung: A single page of results were retrieved. (?)

dss-api.yml
```diff
@@ -71,20 +71,17 @@ paths:
...
+          schema:
+            $ref: '#/definitions/SearchResult'
           headers:
             Link:
               type: string
-              description: URL to retrieve paginated results in [RFC 5988](https://tools.ietf.org/html/rfc5988) format.
+              description: URL to retrieve paginated results.
```
ttung: URL to retrieve the next page of results (?)

dss/api/search.py
```diff
@@ -70,21 +70,15 @@ def post(json_request_body: dict, replica: str, per_page: int, _scroll_id: typin
         } for hit in page['hits']['hits']]
 
         # TODO: (tsmith12) if page returns 0 hits, then all results have been found. delete search id
-        # TODO: (tsmith12) if all results found return request.code.ok.
         if len(result_list) < per_page:
```
ttung: I would consider doing this:

```
if len(result_list) < per_page:
    result_code = requests.codes.ok
else:
    result_code = requests.codes.partial
response = make_response(jsonify({'es_query': es_query, 'results': result_list}), result_code)
```

tests/fixtures/cloud_uploader.py
```diff
@@ -11,7 +11,7 @@
 
 logging.basicConfig(level=logging.ERROR)
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
```
ttung: stuff that somehow snuck into this PR. :(

tests/test_search.py
ttung:
```
url = self.build_url(url_params)
while True:
    search_response = search_for_stuff(url, ...)
    if search_obj.response.status_code == 200:
        break
    url = self.get_next_url(search_obj.response.headers)
```

dss-api.yml
Bento007: Will fix

dss-api.yml
Bento007: Will fix

tests/test_search.py
Bento007: Will fix.

dss/api/search.py
Bento007: The problem is I only want to add the `Link` in the header when a next url is returned. I intentionally isolated `response.headers['Link'] = build_link_header({next_url: {"rel": "next"}})` into the `else` to prevent `Link` from appearing in the header.

ttung:
Is link url an absolute url or just the path component btw?

Bento007:
It is an absolute URL. Do we have a convention we are using already for urls that I should conform to?

ttung:
> It is an absolute URL. Do we have a convention we are using already for urls that I should conform to?

I think absolute URL makes sense.  But since there's no server running (as of this PR), what is the host and scheme?

Bento007:
> what is the host and scheme?

It's `http://localhost`

PR #520, opened 2017-10-09
Update ISSUE_TEMPLATE.md

kozbo:
Add field to specify the version of the code where a bug was observed
Connects to #511


kislyuk:
Please make sure to either fill out or delete the headings in the PR template. If you don't have a test plan, then the header shouldn't be there.

PR #548, opened 2017-10-11
Allow users of the URLBuilder API to set the query.

ttung:
Also adds a `has_query()` method, which returns True if there's a query field with a given name.


ttung:
Actually, `UrlBuilder` has no unit tests.

Added now.

PR #555, opened 2017-10-13
Corrected quoting of jq command in smoketest.py

mikebaumann:
Corrected quoting to resolve the following error occurring during smoketest.py execution, which was indirectly causing smoketest to fail:

jq: error: syntax error, unexpected IDENT, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:
.uuid=4ec21a01-46cb-4130-9a1d-2da3c1c85809

### Test plan
This problem was seen in the following Travis cron/smoketest builds:
https://travis-ci.org/HumanCellAtlas/data-store/jobs/286195792
https://travis-ci.org/HumanCellAtlas/data-store/builds/287704452

The fix was verified in a local (UCSC) deployment of the current master branch.

### DIFF
--- a/tests/smoketest.py
+++ b/tests/smoketest.py
@@ -54,7 +54,7 @@ bundle_dir = "data-bundle-examples/10X_v2/pbmc8k"
 with open(os.path.join(bundle_dir, "async_copied_file"), "wb") as fh:
     fh.write(os.urandom(ASYNC_COPY_THRESHOLD + 1))
 
-run(f"cat {bundle_dir}/sample.json | jq .uuid=\"{sample_id}\" | sponge {bundle_dir}/sample.json")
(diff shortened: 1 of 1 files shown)


kislyuk:
@ttung this is why i preferred passing it through env. now we have super ugly quotes with no intuitive way to figure out the right number of backslashes :(

ttung:
:(

My bad, ok let's go back to the env way.

Bento007:
This would also work, though it's not much more appealing to look at.
`run(r"""cat {bundle_dir}/sample.json | jq .uuid=\"{sample_id}\" | sponge {bundle_dir}/sample.json""".format(bundle_dir=bundle_dir, sample_id=sample_id))`

kislyuk:
Thanks @mikebaumann for looking into this. I edited this to do it a slightly different/more readable way in the linked commit.

PR #556, opened 2017-10-17
Include filesize in manifests.

xbrianh:
Lookup size during file put, report in file/bundle manifests.

Connects to #471

### DIFF
--- a/dss/blobstore/__init__.py
+++ b/dss/blobstore/__init__.py
@@ -87,6 +87,20 @@ class BlobStore:
         """
         raise NotImplementedError()
 
+    def get_size(
+            self,
+            bucket: str,
+            object_name: str
+    ) -> int:
+        """
+        Retrieves the filesize
+        :param bucket: the bucket the object resides in.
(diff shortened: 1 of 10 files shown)


dss/blobstore/__init__.py
```diff
@@ -87,6 +87,20 @@ def get_user_metadata(
         """
         raise NotImplementedError()
 
+    def get_size(
```
ttung: pls add unit test for this.

xbrianh:
Note: in test_file.py `upload_file` was nested under `_test_file_put`
Moved to class level so I could use it too.

ttung:
I think this is fine now, but if we roll this out, the existing data will fail to satisfy GET /bundle/<uuid>.

There needs to be a conversation on #data-store-eng about how this can roll out successfully, i.e., what are the obstacles to clearing out the data store (dev *and* staging).  It may also be worth investing in building a tool that automates this process.

In the meanwhile, you can safely update the write path (i.e., PUT /bundle/<uuid>).

PR #557, opened 2017-10-17
Consolidate all the open port seeking methods.

ttung:


ttung:
> It makes sense and cleaner. I'm not sure if it'll be any faster but it couldn't be any slower than the current code.

It's not for performance reasons.  Purely for codebase health. :)

PR #558, opened 2017-10-17
Fix timeouts for invocations on the command line.

ttung:
When we use the chalice app in tests, we override the timeout.  When we use the chalice app on lambda, the timeout is valid.

However, in the current version of chalice we're using, calling `get_remaining_time_in_millis()` throws an exception when not run in a lambda. 

This diff:

1. Sets the timeout in config.
2. Overrides the timeout set in config due to a chalice bug (see https://github.com/aws/chalice/pull/579)

### DIFF
--- a/chalice/.chalice/config.json
+++ b/chalice/.chalice/config.json
@@ -11,5 +11,6 @@
       "environment_variables": {
       }
     }
-  }
+  },
+  "lambda_timeout": 300
 }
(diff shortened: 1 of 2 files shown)


kislyuk:
Sorry about the delay with this review. This PR does not seem to stack correctly - it has a number of extraneous changes. Could you rebase it to stack correctly?

Also, Chalice 1.0.2+ should not crash when get_remaining_time_in_millis() is called in local mode.

chalice/app.py
```diff
@@ -85,6 +85,8 @@ def wrapper(*args, **kwargs):
                 future = executor.submit(method, *args, **kwargs)
                 time_remaining_s = chalice_app._override_exptime_seconds  # type: typing.Optional[float]
                 if time_remaining_s is None:
+                    import pdb
+                    pdb.set_trace()
```
kislyuk: One last change: please deactivate the pdb tracer :)

PR #562, opened 2017-10-17
Deployment stage + magic methods to invoke.

ttung:
If we don't nip this in the bud, we'll have the codebase littered with `os.environ["DSS_DEPLOYMENT_STAGE"] == "ababab"` and that would be awful.

Depends on #561

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -1,7 +1,7 @@
 import os
 import typing
 from contextlib import contextmanager
-from enum import Enum, auto
+from enum import Enum, EnumMeta, auto
 
 from cloud_blobstore import BlobStore
 from cloud_blobstore.s3 import S3BlobStore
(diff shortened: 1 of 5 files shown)


dss/config.py
```diff
@@ -12,6 +12,30 @@
...
+class DeploymentStage(Enum, metaclass=DeploymentStageMeta):
+    """
+    Represents the current deployment stage.  Through the `DeploymentStageMeta` metaclass, we provide the magic methods
+    IS_{STAGE}, which return True if the current deployment stage matches `STAGE`.
+
+    e.g., IS_PROD() will return True iff the current deployment is a production deployment.
+    """
+    PROD = "prod"
+    STAGING = "staging"
+    DEV = "dev"
```
kislyuk: Why use a metaclass when you can just use getattr to dynamically generate the method? That seems a bit simpler.

ttung: __getattr__ does not work as a classmethod. This works like DeploymentStage.IS_PROD()

kislyuk: ah I see. OK LGTM then.

Issue #212, opened 2017-07-11
Provide multiple search output options


mikebaumann:
Currently DSS search always formats results as a list of bundles.
Other output formats should be supported, including:

- A format that includes the matched index document
- A (relatively) "raw" format that could return the results of queries such:

> Please give me a list of all the contact emails and titles for all the projects, and how many samples and files each has.

and other queries specifically written to return a subset of the available data

Bento007:
I need more information on how we want the output format and the different options we want to support. @mikebaumann @sampierson @briandoconnor @dshiga

kislyuk:
Update from sprint review: Began work on an optional parameter to include raw indexed metadata from the search hit.

PR #564, opened 2017-10-17
Multiple result views supported for search

Bento007:
Connects to #212

### Test plan
- Added smoketest case to use format raw.
- updated test search to validate new raw format

### Release notes
Bundle metadata is return with the search results when format=raw. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -48,6 +48,12 @@ paths:
                 type: object
             required:
               - es_query
+        - name: format
+          in: query
+          description: Specifies the output format. Raw will return the json meta data in the results.
+          required: false
+          type: string
+          enum: [raw]
         - name: replica
           in: query
           description: Replica to search.
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -50,7 +50,7 @@ paths:
               - es_query
         - name: format
           in: query
-          description: Specifies the output format. Raw will return the json meta data in the results.
+          description: Specifies the output format. `raw` will return the json meta data in the results.
           required: false
           type: string
           enum: [raw]
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -75,7 +75,7 @@ def post(json_request_body: dict,
                       'search_score': hit['_score']
                       }
             if format == 'raw':
-                result.update({'meta_data': hit['_source']})
+                result['meta_data'] = hit['_source']
             result_list.append(result)
 
         # TODO: (tsmith12) if page returns 0 hits, then all results have been found. delete search id
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -22,7 +22,11 @@ class PerPageBounds:
 
 
 @dss_handler
-def post(json_request_body: dict, replica: str, per_page: int, _scroll_id: typing.Optional[str] = None) -> dict:
+def post(json_request_body: dict,
+         replica: str,
+         per_page: int,
+         _scroll_id: typing.Optional[str] = None,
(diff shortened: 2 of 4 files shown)


dss/api/search.py
```diff
@@ -63,14 +67,24 @@ def post(json_request_body: dict, replica: str, per_page: int, _scroll_id: typin
 
         # TODO: (tsmith12) allow users to retrieve previous search results
         _scroll_id = page['_scroll_id']
-        result_list = [{
-            'bundle_id': hit['_id'],
-            'bundle_url': _build_bundle_url(hit, replica),
-            'search_score': hit['_score']
-        } for hit in page['hits']['hits']]
+
+        if format == 'raw':
```
ttung: I would consider switching this to a simple for-loop because you can remove a fair amount of the duplicated code.

dss-api.yml
```diff
@@ -48,6 +48,12 @@ paths:
                 type: object
             required:
               - es_query
+        - name: format
+          in: query
+          description: Specifies the output format. Raw will return the json meta data in the results.
```
ttung: You may want to quote it `` `raw` `` or do it in some way that indicates that the value should be `raw` and not `Raw`.

dss/api/search.py
```diff
@@ -63,14 +67,23 @@ def post(json_request_body: dict, replica: str, per_page: int, _scroll_id: typin
...
+            result = {}
+            if format == 'raw':
+                result.update({'meta_data': hit['_source']})
+            else:
+                result.update(
+                    {'bundle_id': hit['_id'],
+                     'bundle_url': _build_bundle_url(hit, replica),
+                     'search_score': hit['_score']
+                     })
+                result_list.append(result)
```
ttung: I think you need to indent this left.

dss/api/search.py
```diff
@@ -63,14 +67,23 @@ def post(json_request_body: dict, replica: str, per_page: int, _scroll_id: typin
 
         # TODO: (tsmith12) allow users to retrieve previous search results
         _scroll_id = page['_scroll_id']
-        result_list = [{
-            'bundle_id': hit['_id'],
-            'bundle_url': _build_bundle_url(hit, replica),
-            'search_score': hit['_score']
-        } for hit in page['hits']['hits']]
+
+        result_list = [typing.Any]
```
ttung: Whoa, I did not know you could do this.

Although, it could be more specific, i.e., `result_list = []  # type: typing.Sequence[dict]`

Bento007: I figured out i could do that after it gave me errors for not defining the list type. I'll change it your your way.

Bento007:
Whoops copy paste error. I completely dropped the other fields!

Bento007:
I was totally was overthinking it.

dss/api/search.py
```diff
@@ -70,15 +70,12 @@ def post(json_request_body: dict,
 
         result_list = []  # type: typing.List[dict]
         for hit in page['hits']['hits']:
-            result = {}
+            result = {'bundle_id': hit['_id'],
+                      'bundle_url': _build_bundle_url(hit, replica),
+                      'search_score': hit['_score']
+                      }
             if format == 'raw':
                 result.update({'meta_data': hit['_source']})
```
ttung: Would just do:

`result['meta_data'] = hit['_source']` as it is faster and simpler.

Bento007: will fix

PR #568, opened 2017-10-19
Updated error when a bogus endpoint entered for elastic search.

### LINKED ISSUE #547: Bug: updated error for search with bogus endpoint
### Expected behavior
When an invalid endpoint is detected, search should response with status code `not_found` and messaged `invalid endpoint`

### Actual behavior
When an invalid endpoint is detects, search responses with status code `interal_server_error` and message `Elasticsearch Internal Server Error`

Bento007:
Fixes #547 

### Test plan
Updated expected error for test_search.

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -103,6 +103,11 @@ def post(json_request_body: dict,
             raise DSSException(requests.codes.not_found,
                                "elasticsearch_context_not_found",
                                "Elasticsearch context has returned all results or timeout has expired.")
+        elif ex.status_code == 'N/A':
+            get_logger().error("%s", f"Elasticsearch Invalid Endpoint. Exception: {ex}")
+            raise DSSException(requests.codes.not_found,
+                               "not_found",
+                               "Elasticsearch Invalid Endpoint")
         else:
             get_logger().error("%s", f"Elasticsearch Internal Server Error. Exception: {ex}")
             raise DSSException(requests.codes.internal_server_error,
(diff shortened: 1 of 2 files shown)


tests/test_search.py
```diff
@@ -198,9 +198,9 @@ def test_elasticsearch_exception(self):
             self.assertPostResponse(
                 path=url,
                 json_request_body=dict(es_query=smartseq2_paired_ends_query),
-                expected_code=requests.codes.internal_server_error,
-                expected_error=ExpectedErrorFields(code="internal_server_error",
-                                                   status=requests.codes.internal_server_error))
+                expected_code=requests.codes.not_found,
```
ttung: Shouldn't this be a client error (maybe 400?)?

Bento007: The error is actually cause on the server side when the DSS_ES_ENDPOINT is removed. Maybe the error should be "Elasticsearch Misconfigured"? Either way, not_found seems like the appropriate error in my opinion.

ttung: If it's a server error, look for a 5xx error code.

Bento007: How about 503 service_unavailable?

dss/api/search.py
```diff
@@ -105,8 +105,8 @@ def post(json_request_body: dict,
                                "Elasticsearch context has returned all results or timeout has expired.")
         elif ex.status_code == 'N/A':
             get_logger().error("%s", f"Elasticsearch Invalid Endpoint. Exception: {ex}")
-            raise DSSException(requests.codes.not_found,
-                               "not_found",
+            raise DSSException(requests.codes.service_unavailable,
+                               "service_unavailable",
                                "Elasticsearch Invalid Endpoint")
```
ttung: I think you may wish to expand on this human-readable message a bit.

PR #569, opened 2017-10-19
Update post search output format parameter definition

kislyuk:


ttung:
I think you need to update the unit test.

PR #574, opened 2017-10-23
size metadata written on file/bundle put

xbrianh:
Lookup size during file/bundle put.

Backward compatible size metadata on write path.
Read path to come later.
Expects get_size() from cloud_blobstore.

Connects to #471

### Test plan
Unit tests after read/write path finished.


xbrianh:
Will get_size be provided by cloud_blobstore, or should it be included here?

ttung:
> Will get_size be provided by cloud_blobstore, or should it be included here?

cloud_blobstore should provide this functionality.  please submit a PR to https://github.com/chanzuckerberg/cloud-blobstore

xbrianh:
Tests pass. How likely is this to cause mayhem upon merge?

ttung:
> Tests pass. How likely is this to cause mayhem upon merge?

One way to find out. :)

Issue #235, opened 2017-07-17
Remove index document when a bundle is removed


mikebaumann:
Although this is not a common operation for the final system, is useful for development/test.
This will keep the index cleaner and more consistent with the data that is present.

mikebaumann:
As pointed out by @ttung in #578, this specific feature is not needed because it is covered by:
1. Administrative delete
2. DSS reset admin tool (to be developed)
In addition, the reindexing service (to be developed) should detect and remove any index document for which the corresponding bundle does not exist in the storage bucket, thereby removing stale data from the index and restoring consisteny.

PR #578, opened 2017-10-23
Remove deleted bundles from index

xbrianh:
De-index bundles upon s3 deletion events.
Require that index exists, and that bundle is indexed.

Connects #235

### Test plan
Unit test for succesful de-indexing upon mock deletion event.


dss/events/handlers/index.py
```diff
@@ -59,6 +59,44 @@ def process_new_indexable_object(bucket_name: str, key: str, replica: str, logge
...
+                                               ESDocType.doc.name,
+                                               bundle_id)
+
+        # TODO Is this bad? deletes should be rare
+        ElasticsearchClient.get(logger).indices.refresh(index_name)
+
+        # TODO (brianh): Notification on removal?
+        logger.debug(f"Finished de-index processing of {replica} removal event for bundle: {key}")
+    else:
+        logger.debug(f"Not de-indexing {replica} removal event for key: {key}")
```
Bento007: This is a really confusing way of saying key not in index. Maybe f"Unable to remover, {key} not in {replica} index."

daemons/dss-index/app.py
```diff
@@ -10,20 +10,24 @@
...
 
 s3_bucket = dss.Config.get_s3_bucket()
 
-@app.s3_event_handler(bucket=s3_bucket, events=["s3:ObjectCreated:*"])
+@app.s3_event_handler(bucket=s3_bucket, events=["s3:ObjectCreated:*", "s3:ObjectRemoved:Delete"])
 def dispatch_s3_indexer_event(event, context) -> None:
     app.log.setLevel(logging.DEBUG)
     if event.get("Event") == "s3:TestEvent":
         app.log.info("DSS index daemon received S3 test event")
     else:
```
Bento007: Why not do elif "ObjectRemoved" in event["Records"][0]["eventName"]:?

tests/test_indexer.py
```diff
@@ -417,6 +445,10 @@ def setUpClass(cls):
     def process_new_indexable_object(self, event, logger):
         process_new_gs_indexable_object(event, logger)
 
+    def test_process_removed_object(self):
```
Bento007: Maybe `@self.skip("de-index for s3 only")`. Also add a todo to add delete event for GS.

daemons/dss-index/app.py
xbrianh: No conditional in original code. Only two paths: ObjectRemoved checked explicitly, assume ObjectCreated otherwise.

dss/events/handlers/index.py
```diff
@@ -59,6 +59,44 @@ def process_new_indexable_object(bucket_name: str, key: str, replica: str, logge
...
+
+
+def bundle_is_indexed(index_name: str, bundle_id: str, logger) -> bool:
+    if not ElasticsearchClient.get(logger).indices.exists(index_name):
```
mikebaumann: It is not necessary to check if the index exists here. This case is covered by the following statement, which will return false if the index does not exist or the document does not exist in the index. Removing the check for the index will save an unnecessary round-trip to the ES server.

tests/test_indexer.py
```diff
@@ -28,7 +28,8 @@
 import dss
 from dss import Config, DeploymentStage
 from dss.config import IndexSuffix
-from dss.events.handlers.index import process_new_s3_indexable_object, process_new_gs_indexable_object, notify
+from dss.events.handlers.index import process_new_s3_indexable_object, process_s3_removed_object
```
mikebaumann: These can be imported in a single statement without exceeding the line length limit, as follows:
```
from dss.events.handlers.index import (process_new_s3_indexable_object, process_s3_removed_object,
                                       process_new_gs_indexable_object, notify)
```

tests/test_indexer.py
```diff
@@ -115,14 +116,19 @@ def tearDown(self):
...
-        bundle_key = self.load_test_data_bundle_for_path("fixtures/smartseq2/paired_ends")
-        sample_event = self.create_sample_bundle_created_event(bundle_key)
-        self.process_new_indexable_object(sample_event, logger)
+        bundle_key = self.insert_and_process_new_indexable_object()
         search_results = self.get_search_results(smartseq2_paired_ends_query, 1)
         self.assertEqual(1, len(search_results))
         self.verify_index_document_structure_and_content(search_results[0], bundle_key,
                                                          files=smartseq2_paried_ends_indexed_file_list)
 
+    def test_process_removed_object(self):
```
mikebaumann: Please add two additional tests:

1.  The key is not a valid bundle key
2. The key is a valid bundle key, but the bundle is not indexed.

dss/events/handlers/index.py
```diff
@@ -59,6 +59,44 @@ def process_new_indexable_object(bucket_name: str, key: str, replica: str, logge
...
+        logger.error("Exception occurred while processing S3 event: %s Event: %s", ex, json.dumps(event, indent=4))
+        raise
+
+
+def process_removed_object(key: str, replica: str, logger) -> None:
+    index_name = Config.get_es_index_name(ESIndexType.docs, Replica[replica])
+    bundle_id = get_bundle_id_from_key(key)
+    if bundle_is_indexed(index_name, bundle_id, logger):
+        logger.info(f"Received {replica} removal event for bundle which will be de-indexed: {key}")
+        ElasticsearchClient.get(logger).delete(index_name,
```
mikebaumann: nit. The Elasticsearch-py documentation [here](https://elasticsearch-py.readthedocs.io/en/master/api.html) recommends using keyword arguments for all calls for consistency and safety, and that that is (generally) what we have been doing.

ttung:
Your life is going to be significantly complicated by the fact that admin delete is not yet in place.  You should be de-indexing if a tombstone is dropped, but none of that is in place.

Personally, I wouldn't even bother with this until tombstoning is done.

tests/test_indexer.py
```diff
@@ -115,14 +116,19 @@ def tearDown(self):
...
+        bundle_key = self.insert_and_process_new_indexable_object()
+        sample_event = self.create_sample_bundle_removed_event(bundle_key)
+        self.process_removed_object(sample_event, logger)
+        search_results = self.get_search_results(smartseq2_paired_ends_query, 0)
```
mikebaumann: The search that you are performing to verify that the document is no longer there should also be performed before removal to ensure that the document is there and is found by the search. The document being loaded sometimes changes, and it could change in a way that the search no longer matches it, and in that case, as the test stands now, de-indexing could fail but the test would still pass. Let me know if you have questions or need clarification regarding this case.

mikebaumann:
To avoid having stale data in the index, it seems useful to remove the index for a bundle:

1. When the bundle is removed from the storage bucket directly (more of a development use case)
2. When an administrative delete is performed, including creation of a tombstone.

Although the event/path to the code to perform the de-indexing is different in these two cases, the code to actually perform the de-indexing is, I believe, the same for both cases. According to the [existing indexing spec](https://docs.google.com/document/d/1oNgk15q90a6O33vUszUTq9bwwe3gp349bYBIO0BdpPo/edit#heading=h.uktg6yyby05k), the admin deleted bundle is simply removed from the index.
I am not sure how finishing this PR is complicated by not having admin delete implemented yet, but I may be overlooking something. To me, this PR seems like a useful step forward.

ttung:
> To avoid having stale data in the index, it seems useful to remove the index for a bundle:
> 1. When the bundle is removed from the storage bucket directly (more of a development use case)
> 2. When an administrative delete is performed, including creation of a tombstone.
> Although the event/path to the code to perform the de-indexing is different in these two cases, the code to actually perform the de-indexing is, I believe, the same for both cases. According to the existing indexing spec, the admin deleted bundle is simply removed from the index.
I am not sure how finishing this PR is complicated by not having admin delete implemented yet, but I may be overlooking something. To me, this PR seems like a useful step forward.

It's not clear to me when we would ever want to remove a bundle directly.  We definitely want to trigger removal on the arrival of a tombstone.  It's not clear we want to trigger a removal when a bundle is removed.

Having said that, I do agree that this PR has value.  I'm just concerned we might have to rip a significant chunk of it back out.

Incidentally, a development use case that has a lot of value is resetting a deployment.  I know @xbrianh needs that to roll out a change in the blue box metadata schema that requires nuking all the data.

mikebaumann:
Last time we needed to reset the deployments @sampierson started a util to do that, available here:
https://github.com/HumanCellAtlas/data-store/blob/spierson-reset-dss/scripts/reset-dss.sh
It removes /bundles and /files, and if this PR is implemented and works as intended that should be all that is needed.

PR #585, opened 2017-10-25
Include size in manifests GET

xbrianh:
Size metadata read on file/bundle get

Merge requires nuking/re-initialization of existing dss deployments.

Connects to #471


ttung:
I think this looks fine.  In addition to nuking the dev data store, the test artifacts need to be populated.

Finally, you should open up discussions for how to upgrade staging.

PR #599, opened 2017-11-01
close the test connection.

ttung:
For reasons unclear to me, the standalone test is now failing without this.  It appears that it is waiting for the test connection to make its request, and that never happens, and it does not answer subsequent requests.


ttung:
There was some weird ipv6 stuff going on with travis.  #600 worked around the core issue, and this was sort of an optional fix that still makes sense to land.

PR #601, opened 2017-11-02
Clean up daemons directory

ttung:
1. Remove dss-chunked-task as it is no longer used.
2. Update .gitignore to point to chained-aws-lambda instead.

### Test plan
Deployed all the daemons and run all the tests.


ttung:
Removed `daemons/dss-chunked-task/.chalice/config.json`

mweiden:
@ttung I'm assuming travis is upset because of the IAM permission scramble from earlier

PR #602, opened 2017-11-02
Update queries to reflect new test fixtures.

ttung:


ttung:
Travis keeps on killing the jobs early, but I have successfully run this multiple times on my laptop, so I'm just going to land it.

PR #606, opened 2017-11-02
Additional updates to index, search and subscriptions tests for metadata v3

mikebaumann:
Updates to index, search and subscription tests to be consistent with metadata version 3.
This was prompted by recently run of populate.py which pulled in current
metadata (version 3) from data-bundle-examples.

A mismatch between the updated query and some of the existing sample index documents was causing test_search.py to a run for a long time, in which case Travis terminated the build.
Restoring consistency resolved this problem.


ttung:
> Those older meta version documents will come in handy.

How come?  Why not just remove them and whoever needs it can pull it out of git history?

mikebaumann:
Yes, as tests/sample_unversioned_index_doc.json is currently unused, I have removed it and we can retrieve it from the history if/when needed for new tests.

Issue #605, opened 2017-11-02
Remove hca-dss-content-type tag requirement for files to be stored.


sampierson:
As discussed in document https://docs.google.com/document/d/1TqihrgXjct9aDmTJO52_gE2WlpFysB1OkG9C8exmWTw/edit we are now using in-line media-type parameters to distinguish between DCP file types.

Please remove the requirement that all files stored in the HCA should have a hca-dss-content-type tag.

sampierson:
Tags containing content-type should only be used behinds the scenes for storage mediums that don't have a a content-type concept.

PR #621, opened 2017-11-06
Use the cloud-native content-type attached to the src data blob.

ttung:
This removes the usage of the hca-dss-content-type field.

Connects to #605 

### Test plan
Tested by populating test fixtures buckets and running `DSS_GS_BUCKET_TEST_FIXTURES=org-humancellatlas-dss-test-fixtures-ttung DSS_S3_BUCKET_TEST_FIXTURES=org-humancellatlas-dss-test-fixtures-ttung make tests/test_file.py`.

The travis run is likely to fail, unfortunately, until I run this on the main test fixtures bucket.

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -130,6 +130,7 @@ def put(uuid: str, json_request_body: dict, version: str=None):
 
     metadata = handle.get_user_metadata(src_bucket, src_object_name)
     size = handle.get_size(src_bucket, src_object_name)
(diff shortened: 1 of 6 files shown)


xbrianh:
looks like dss-api.yml needs some love too:
https://github.com/HumanCellAtlas/data-store/blob/e1161e9c4f16c125efd101ebca43e9565fcf4c25/dss-api.yml#L252

PR #624, opened 2017-11-07
Use application/octet-stream instead of binary/octet-stream.

ttung:


mweiden:
@ttung I'm coming late to this. Can you explain the motivation for this change?

ttung:
@sampierson  says binary/octet-stream is not a valid mime type.

mweiden:
@ttung I noticed that it's not present in [RFC2046](https://www.ietf.org/rfc/rfc2046.txt), but wasn't sure if there were changes later.

Issue #464, opened 2017-09-18
Race conditions on S3 blobs during API requests


ttung:
This is most likely due to race conditions (s3 eventual consistency).

kozbo:
What PR's are attached to this issue?

ttung:
#566 is a partial fix for this.

PR #626, opened 2017-11-08
spin-loop for PUT /bundles on missing file

ttung:
We've observed that there are race conditions on S3 where a `PUT /files` has completed but the metadata files are not available on the subsequent `PUT /bundles` request.  We added logic to make it easier for the client to identify this situation (and retry) in #566, but it is desirable to attempt the retry on the server as well.

Connects to #464

### Test plan
Verified that the existing paths still work.

Added test that sets up a missing file, starts a `PUT /bundles` request, and then uploads the missing file.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -13,8 +13,9 @@ import typing
 from concurrent.futures import ThreadPoolExecutor, TimeoutError
 from http.cookies import SimpleCookie
 
-import chalice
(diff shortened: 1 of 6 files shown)


ttung:
https://github.com/ttung/py-nestedcontext is the source for nestedcontext for those who are curious.

PR #629, opened 2017-11-08
remove dependency on application/json

ttung:


Bento007:
I think that's too open. There was a reason we wanted only application/json to be index.  @mikebaumann what is your opinion?

Issue #630, opened 2017-11-08
Overly strict parsing of content-type in metadata indexer


kislyuk:
@sampierson encountered this problem when ingesting a bundle through the blue box. The indexer requires the content-type to be exactly "application/json", but modifiers like "application/json; foo=bar" should be admissible.

@Bento007 interest

```
>aegea logs /aws/lambda/dss-index-staging --start-time=-1h| grep 3153fc1e-4136-454d-a67d-24a0c73cd632
[...]

PR #631, opened 2017-11-09
less strict parsing of content-type in metadata indexer

Bento007:
Fix for #630

### DIFF
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -82,7 +82,7 @@ def create_index_data(handle: BlobStore, bucket_name: str, bundle_id: str, manif
     index_files = {}
     for file_info in files_info:
         if file_info[BundleFileMetadata.INDEXED] is True:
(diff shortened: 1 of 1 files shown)


sampierson:
This solution looks fine for now, but using the MediaType class out of dcp-cli would be a more robust approach.  I think I should extract into a separate repo, so that we can use it across several apps.  I already need to share it between dcp-cli and upload-service.

dss/events/handlers/index.py
```diff
@@ -20,7 +20,6 @@
 
 DSS_BUNDLE_KEY_REGEX = r"^bundles/[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}\..+$"
 
-
```
ttung: PEP8 requires two NL here.

Bento007:
@sampierson once you separate it, open another issues to use it.

PR #633, opened 2017-11-11
Prevent smoketest from clobbering the current virtualenv

hannes-ucsc:
smoketest installs the CLI into the currently active virtualenv or even the global site-dist if no virtualenv is active. The CLI's dependencies might differ from those of data-store creating potential breakage in data-store after smoketest.py runs. This PR changes smoketest to install the CLI into a temporary virtualenv and remove it afterwards.

### Test plan

Run smoketest.py manually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,5 @@ npm-debug.log
 
 # Local environment configuration
 /environment.local
+
+/smoketest-*.tmp
--- a/tests/smoketest.py
+++ b/tests/smoketest.py
@@ -3,9 +3,8 @@
 This script runs a basic integration test of the DSS. It is invoked by Travis CI from a periodic cron job.
 """
 
-import os, sys, argparse, time, uuid, json
+import os, sys, argparse, time, uuid, json, shutil, tempfile
 from subprocess import check_call, check_output, CalledProcessError
-from tempfile import TemporaryDirectory
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
@@ -13,6 +12,8 @@ sys.path.insert(0, pkg_root)  # noqa
 from dss.api.files import ASYNC_COPY_THRESHOLD
 
 parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--no-clean', dest='clean', action='store_false',
+                    help="Don't remove the temporary working directory on exit.")
 args = parser.parse_args()
 
 def GREEN(message=None):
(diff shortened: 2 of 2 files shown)


tests/smoketest.py
```diff
@@ -43,45 +42,58 @@ def run(command, runner=check_call, **kwargs):
     run("git clone --depth 1 --recurse-submodules https://github.com/HumanCellAtlas/data-store-cli")
 
 run("http --check-status https://${API_HOST}/v1/swagger.json > data-store-cli/swagger.json")
-run("pip install -r data-store-cli/requirements.txt")
-run("python -c 'import sys, hca.dss.regenerate_api as r; r.generate_python_bindings(sys.argv[1])' swagger.json",
-    cwd="data-store-cli")
-run("find data-store-cli/hca -name '*.pyc' -delete")
-run("pip install --upgrade --no-deps .", cwd="data-store-cli")
+venv = tempfile.mkdtemp()
```
ttung: use `with tempfile.TemporaryDirectory() as venv:`

it provides cleanup for free.

tests/smoketest.py
```diff
@@ -43,45 +42,58 @@ def run(command, runner=check_call, **kwargs):
...
-        run("http -v --check-status https://${API_HOST}/v1/bundles/$(jq -r .bundle_uuid upload.json)?replica=gcp")
-        break
-    except SystemExit:
-        time.sleep(1)
-else:
-    parser.exit(RED("Failed to replicate bundle from AWS to GCP"))
-run("hca dss download --replica gcp $(jq -r .bundle_uuid upload.json)")
+    run(f"cat {bundle_dir}/sample.json | jq .uuid=env.sample_id | sponge {bundle_dir}/sample.json",
+        env=dict(os.environ, sample_id=sample_id))
+    run(f"{venv_bin}hca dss upload "
```
ttung: Does executing {venv_bin}/hca use that virtualenv even if one doesn't activate that venv?

hannes-ucsc: Yes. Running a binary from the venv's bin activates that venv on the fly for the duration of the process.

tests/smoketest.py
hannes-ucsc: Will do.

hannes-ucsc:
I piggy-backed edc3934 onto this PR to also prevent smoketest from dirtying the working copy. I can pull it out into another PR if requested.

.gitignore
```diff
@@ -44,3 +44,5 @@ npm-debug.log
 
 # Local environment configuration
 /environment.local
+
+/smoketest-*.tmp
```
ttung: Why is this necessary since you have the `with tempfile.TemporaryDirectory...`

hannes-ucsc: I added `dir=` to that, placing the temp dir into the build directory. This makes the tempdir easier to find for post-mortems but I forgot to actually make clean-up optional by going back to  

```
cleanup=True
…
try:
    …
finally:
    if cleanup:
        rmdir(…)
```

Let me push that.

hannes-ucsc:
Running smoketest now leaves no temp files lying around in the source tree, but can be optionally directed to skip cleaning up its temporary working directory (via the --no-clean options). This can be useful for post-mortems.

tests/smoketest.py
```diff
@@ -39,49 +40,70 @@ def run(command, runner=check_call, **kwargs):
...
-        res = json.load(fh2)
-        print(json.dumps(res, indent=4))
-        assert len(res["results"]) == 1
-
-    res = run(f"hca dss put-subscriptions --callback-url https://example.com/ --es-query '{{}}' --replica {replica}",
-              runner=check_output)
-    sub_id = json.loads(res.decode())["uuid"]
-    run(f"hca dss get-subscriptions --replica {replica}")
-    run(f"hca dss delete-subscriptions --replica {replica} {sub_id}")
+workdir = tempfile.mkdtemp(dir=os.getcwd(), prefix="smoketest-", suffix='.tmp')
```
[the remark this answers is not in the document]
hannes-ucsc: I disagree. Considering the way the build currently pollutes the source tree with hundreds of files, smoketest would be the least of our problems if the source tree were on an NFS. IOW, what you propose as a fix for the NFS scenario doesn't go far enough anyway. So if I had to weigh that against the convenience of quickly finding smoketest's working directory after running with `--no-clean`, I'd chose the latter.

tests/smoketest.py
```diff
@@ -39,49 +40,70 @@ def run(command, runner=check_call, **kwargs):
...
+                  "--callback-url https://example.com/ "
+                  "--es-query '{}' "
+                  f"--replica {replica}",
+                  runner=check_output)
+        sub_id = json.loads(res.decode())["uuid"]
+        run(f"{venv_bin}hca dss get-subscriptions --replica {replica}")
+        run(f"{venv_bin}hca dss delete-subscriptions --replica {replica} {sub_id}")
+finally:
+    if args.clean:
+        shutil.rmtree(workdir)
```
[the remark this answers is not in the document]
hannes-ucsc: Of course. I keep forgetting that we're not on Python 2.

tests/smoketest.py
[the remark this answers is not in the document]
ttung: We should build the deployment stuff out of directory as well, but I don't want to spend all day debating this.

Go ahead and make the `tempfile.TemporaryDirectory(..)` change and land this.

hannes-ucsc:
@ttung's instructions were a little ambiguous but I resolved them in my favor, leaving `dir=` in place but changing `mkdtemp` to `TemporaryDirectory` and `shutil.rmtree(workdir)` to `workdir.cleanup`.

Issue #594, opened 2017-10-30
Index bundles according to set of schema major versions


Bento007:
Place the index document representing the metadata for a bundle in an index specific to the set of schema major versions the bundle contains.

see epic #289

mikebaumann:
This should ensure that the bundle metadata conforming to the schemas is able to be written to the ES index without field type conflicts occurring. This is based on the assumption/commitment that a change to the field type of a given field name would cause the major semantic version number of that schema and all containing schemas to be incremented. 
The bundle metadata may contain additional data not currently defined by the schemas, and this could still potentially conflict, and there will still need to be a way to handle these conflicts, which may include use of ES `ignore_malformed` and/or removing conflicting data from indexing, but it seems much more acceptable to omit searching for data not defined by the schema than it does to omit search of data that is defined by the schema.
The associated index could be identified an index name consisting of an ordered list of the schema names and major numbers the bundle contains. This would be a good first pass. If schema major numbers change rarely, this may be sufficient. If instead this results in excessive proliferation of indexes, the indexes could  be consolidated (e.g. an index with a smaller set of schema major numbers could be merged be merged into another index with a larger set of the same schema name and major numbers) at the time of indexing or reindexing.

mikebaumann:
This approach is compatible with a variety of ways of setting the ES mappings appropriate for DSS use. Hopefully, a small set of common index settings and default mapping templates will be sufficient and can be applied consistently to all DSS indexes. Yet, the indexing approach described above is also amenable to using the schema to setup the indexes if that were to become necessary for some reason. In any case, setting-up the index mappings is a different issue/story that needs to be done and is not covered here.

PR #646, opened 2017-11-16
New indexes are alias when created and access by alias name.

Bento007:
Rather than accessing indexes directly they are now accessed through an alias name. This helps with reindexing and indexing by major version. For now the alias will only point to a single index. 

Connected to #594

### Test plan
Updated units test to reflect change. Will need further refactoring once indexing by major version is implemented.

### Deployment Notes
Do not merge until reindex is supported.

### Release notes
split get_elasticsearch_index into get_elasticsearch_subscription_index, and get_elasticsearch_doc_index
using document version to create the actual index
adding new indexes to alias on creation.

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -107,7 +107,7 @@ def _es_search_page(es_query: dict,
     # }
     sort = {"sort": ["_doc"]}
     if _scroll_id is None:
-        page = es_client.search(index=Config.get_es_index_name(ESIndexType.docs, Replica[replica]),
+        page = es_client.search(index=Config.get_es_alias_name(ESIndexType.docs, Replica[replica]),
                                 doc_type=ESDocType.doc.name,
                                 scroll=scroll,
                                 size=per_page,
--- a/dss/config.py
+++ b/dss/config.py
@@ -177,7 +177,16 @@ class Config:
 
     @staticmethod
     def get_es_index_name(index_type: ESIndexType, replica: Replica) -> str:
-        """Returns the index name or the alias for indexes"""
+        """Returns the index name"""
+        deployment_stage = os.environ["DSS_DEPLOYMENT_STAGE"]
+        index = f"dss-{index_type.name}-{replica.name}-{deployment_stage}"
+        if Config._CURRENT_CONFIG == BucketConfig.TEST:
+            index = f"{index}.{IndexSuffix.name}"
+        return index
+
+    @staticmethod
+    def get_es_alias_name(index_type: ESIndexType, replica: Replica) -> str:
+        """Returns the alias for indexes"""
         deployment_stage = os.environ["DSS_DEPLOYMENT_STAGE"]
         index = f"dss-{index_type.name}-{replica.name}-{deployment_stage}"
         if Config._CURRENT_CONFIG == BucketConfig.TEST:
--- a/dss/config.py
+++ b/dss/config.py
@@ -176,10 +176,14 @@ class Config:
         return Config._ALLOWED_EMAILS
 
     @staticmethod
-    def get_es_index_name(index_type: ESIndexType, replica: Replica) -> str:
+    def get_es_index_name(index_type: ESIndexType, replica: Replica,
+                          version: typing.Optional[str] = None
+                          ) -> str:
         """Returns the index name"""
         deployment_stage = os.environ["DSS_DEPLOYMENT_STAGE"]
         index = f"dss-{index_type.name}-{replica.name}-{deployment_stage}"
+        if version:
+            index = f"{index}-{version}"
         if Config._CURRENT_CONFIG == BucketConfig.TEST:
             index = f"{index}.{IndexSuffix.name}"
         return index
--- a/dss/config.py
+++ b/dss/config.py
@@ -177,6 +177,7 @@ class Config:
 
     @staticmethod
     def get_es_index_name(index_type: ESIndexType, replica: Replica) -> str:
+        """Returns the index name or the alias for indexes"""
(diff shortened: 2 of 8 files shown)


tests/test_indexer.py
```diff
@@ -268,6 +268,10 @@ def test_subscription_notification_unsuccessful(self):
                          f"WARNING:.*:Failed notification for subscription {subscription_id}"
                          f" for bundle {bundle_id} with transaction id .+ Code: {error_response_code}")
 
+    @unittest.skip("WIP")
+    def test_index_when_multiple_indexes_using_alias(self):
```
mikebaumann: I think it would be good to add a simple test now that verifies that the alias is being created during indexing.
Then the test for an alias with multiple indexes can be added when that is the case, as you indicated here.

Bento007: Added

dss/events/handlers/index.py
```diff
@@ -130,19 +130,25 @@ def get_bundle_id_from_key(bundle_key: str) -> str:
     raise Exception(f"This is not a key for a bundle: {bundle_key}")
 
 
-def add_index_data_to_elasticsearch(bundle_id: str, index_data: dict, index_name: str, logger) -> None:
-    create_elasticsearch_index(index_name, logger)
+def add_index_data_to_elasticsearch(bundle_id: str, index_data: dict, alias_name: str, logger) -> None:
+    #  TODO (tsmith): get the major version from index data.
+    #  version = index_data['manifest']['format'].split('.')[0]
```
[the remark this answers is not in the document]
Bento007: removed

tests/test_indexer.py
```diff
@@ -268,6 +268,17 @@ def test_subscription_notification_unsuccessful(self):
                          f"WARNING:.*:Failed notification for subscription {subscription_id}"
                          f" for bundle {bundle_id} with transaction id .+ Code: {error_response_code}")
 
+    def test_alias_exists(self):
+        sample_event = self.create_sample_bundle_created_event(self.bundle_key)
+        self.process_new_indexable_object(sample_event, logger)
+        es_client = ElasticsearchClient.get(logger)
+        index_names = [i['i'] for i in es_client.cat.aliases(name=self.dss_alias_name, h=['a', 'i'], format='json')]
+        self.assertNotEqual(len(index_names), 0)
```
mikebaumann: Thanks for adding the test. I think it should include verification that an alias with the expected name exists.
Any reason not to do that?

Bento007: The test does this. If the Alias did no exist then assertNotEqual would fail.

mikebaumann: Great, thanks!

hannes-ucsc:
```diff
@@ -272,8 +272,7 @@
         sample_event = self.create_sample_bundle_created_event(self.bundle_key)
         self.process_new_indexable_object(sample_event, logger)
         es_client = ElasticsearchClient.get(logger)
-        index_names = [i['i'] for i in es_client.cat.aliases(name=self.dss_alias_name, h=['a', 'i'], format='json')]
-        self.assertNotEqual(len(index_names), 0)
+        self.assertTrue(es_client.indices.exists_alias(name=[self.dss_alias_name]))
 
     @unittest.skip("WIP")
     def test_index_when_multiple_indexes_using_alias(self):
```

I think this is simpler, more expressive and stricter, too.

tests/test_indexer.py
```diff
@@ -268,6 +268,17 @@ def test_subscription_notification_unsuccessful(self):
...
+
+    @unittest.skip("WIP")
+    def test_index_when_multiple_indexes_using_alias(self):
+        pass
+
```
hannes-ucsc: I do not like these–for the same reasons as for TODOs.

dss/config.py
```diff
@@ -177,10 +177,11 @@ def get_allowed_email_domains() -> str:
 
     @staticmethod
     def get_es_index_name(index_type: ESIndexType, replica: Replica) -> str:
+        """Returns the index name or the alias for indexes"""
```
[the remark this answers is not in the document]
Bento007: for subscriptions it returns the index name, for doc indexes it returns the alias.

dss/api/search.py
```diff
@@ -49,7 +49,8 @@ def post(json_request_body: dict,
     # }
     sort = {"sort": ["_doc"]}
 
-    # TODO: (tsmith12) determine if a search operation timeout limit is needed
+    # TODO: (tsmith12) determine if a search operation timeout limit is needed.
+    # TODO: (tsmith) allow user to specify index to search.
     try:
```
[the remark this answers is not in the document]
Bento007: The TODOs are outside the scope of the PR. I agree that any PRs with TODOs within the scope of the PR should be resolved.

dss/util/es.py
```diff
@@ -133,16 +133,32 @@ def get(logger):
         return client
 
 
-def get_elasticsearch_index(es_client, index_name, logger, index_mapping=None):
+def get_elasticsearch_doc_index(es_client, index_name, alias_name, logger, index_mapping=None):
```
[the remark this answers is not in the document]
Bento007: renamed to create_elasticsearch_doc_index

dss/config.py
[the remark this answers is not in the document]
Bento007: I split it into get_es_index_name and get_es_alias_name. They are identical currently but will be changing in the near future.

hannes-ucsc:
Looks good. If this were my PR, I would squash all changes down to one commit at this point.

PR #649, opened 2017-11-16
Introduce TimedThread.

ttung:
Copy-paste from the docblock:

This is a "Thread" class that runs a job for a maximum period of time.  The class provides concurrency-safe methods to retrieve and persist a chunk of state.

You can use this primitive to wrap your lambda execution such that you send a consistent chunk of state out of the lambda back to AWS step functions.


ttung:
@amork, @hannes-ucsc this may be of interest to you since you are using step functions.

Issue #598, opened 2017-11-01
Subscribe query according to set of schema major versions


Bento007:
1. How do we keep subscriptions updated as we add new indexes? Add queries to all indexes they are compatible with. Therefore all fields in query must exist in the index.
2. If the field does not exist in the new index what do we do? don't add it. When a new doc is added, check subscriptions to see if their query can be added to the index.
3. If the field had changed names or type in the new schema , how do we add the query to the index for subscription? The user must update their subscriptions.

see epic #289

Issue #539, opened 2017-10-10
Re-indexing support for indexer(s) 


Bento007:
Need to be able to reindex data when schemas change.

ttung:
It is more likely that we would need to reindex for new replicas or bugs. If we have to reindex when schemas change, we have failed.

Bento007:
if we continue to use ES then we can use `scan` to fetch all documents and reindex using `_source`

Bento007:
## use cases
1. index remapping/transformation
1. scheduled reindex/validation

## Sub tasks?
1.  prototype basic reindexing
1. what is the design of reindexing?
1. how is it invoked (CLI, API)?
1. how can lambda coordination support/infrastructure be used?

mikebaumann:
For 2017 Q4, the reindexing must:
1. Perform the current index processing for each bundle in each replica (e.g. so current index document structure is used, current metadata validation/trimming is performed, etc.)
2. Ensure consistency (1:1) between bundles and index documents. This includes ensuring that admin deleted bundles have no index document
3. Provide option to place the new/fresh index document in the same index or in a newly created index (in case the ES mappings in the existing index are incorrect, and index document must be placed in a new index with the correct ES mappings)

Bento007:
bring Hanas up to speed.

hannes-ucsc:
Reindexing is functional (REPAIR and VERIFY) but completely ignores any concurrency issues outlined in the spec RFC. It also does not implement the offline REINDEX into a fresh index generation.

We should make a conscious decision if and how we want to address those issues.

Issue #530, opened 2017-10-10
Identify requirements for index data change management


mikebaumann:

kislyuk:
Update from sprint review on Oct 16:
- Research use of "one index per metadata schema" rule
- Research strategies for handling data outside the schema, such as excluding it from the index altogether
- Consider using other databases
@mikebaumann can you please provide an update and link to working docs?

Bento007:
https://docs.google.com/document/d/1qjb6mm4wyl1dveuRqC8ZQYaMYSwfF66BbFr9Z4OjCAI/edit#heading=h.mqisjbbjxey3

PR #650, opened 2017-11-17
Finish changes for schema version specific subscription/notification

Bento007:
Ensure subscription queries are added to indices, and do so in a way
that allows queries including fields not defined in the index mappings.
This enables subscription of queries that work across metadata schema versions,
including fields that are not present in a schema version or have
changed between versions, such as a field changing from type text to type object.
This also resolves internal constraints of needing to have the mappings defined
in an index before a subscription query for percolation can be added to the index.

This is accomplished by setting `index.percolator.map_unmapped_fields_as_string` to true,
which also works in the case of other field types (e.g. numeric, date).
Note when dynamic templates are used and queries for percolation have been added
to an index before the index contains mappings of fields referenced by those queries,
the queries must be reloaded when the mappings change for the queries to match.
For more information, see: https://github.com/elastic/elasticsearch/issues/5750

### Test plan
- Existing unit tests were updated and new unit tests were added
- Additional unit testing of exception cases is needed to increase code coverage
- Manual interactive testing with deployed system

When updating existing deployments:
The subscription index mapping must be updated. This may be done in a way that
preserves the existing subscriptions using an appropriate script utilizing
Elasticsearch Curator.
The document indices also have a new setting, yet because this change uses a different pattern for index naming, new version-specific indices, no specific additional migration is required for this setting. 

### Release notes
- Subscription registration may now be performed before prior to loading bundles and indexing
- `put subscription` attempts to add the percolate query to every index under the `alias_name`.
- delete subscription removes percolate query from all indices to which it was added

Connects to #598 
Related to #530 and #539

### DIFF
--- a/dss/api/subscriptions.py
+++ b/dss/api/subscriptions.py
@@ -127,7 +127,7 @@ def put(json_request_body: dict, replica: str):
                               f"replica: {replica}, Exception: {ex}")
 
         # Delete percolate query to make sure queries and subscriptions are in sync.
-        _unregister_percolate(es_client, subscribed_indexes, uuid)
+        _unregister_percolate(es_client, alias_name, uuid)
 
         raise DSSException(requests.codes.internal_server_error,
                            "elasticsearch_error",
@@ -157,8 +157,7 @@ def delete(uuid: str, replica: str):
 
     #  get all indexes that use current alias
     alias_name = Config.get_es_alias_name(ESIndexType.docs, Replica[replica])
-    doc_indexes = _get_indexes_by_alias(es_client, alias_name)
-    _unregister_percolate(es_client, doc_indexes, uuid)
+    _unregister_percolate(es_client, alias_name, uuid)
 
     es_client.delete(index=Config.get_es_index_name(ESIndexType.subscriptions, Replica[replica]),
                      doc_type=ESDocType.subscription.name,
@@ -170,8 +169,8 @@ def delete(uuid: str, replica: str):
     return jsonify({'timeDeleted': time_deleted}), requests.codes.okay
 
 
-def _unregister_percolate(es_client: Elasticsearch, subscribed_indexes: List[str], uuid: str):
-    es_client.delete(index=subscribed_indexes,
+def _unregister_percolate(es_client: Elasticsearch, index_alias: str, uuid: str):
+    es_client.delete(index=index_alias,
                      doc_type=ESDocType.query.name,
                      id=uuid,
                      refresh=True)
--- a/dss/api/subscriptions.py
+++ b/dss/api/subscriptions.py
@@ -88,7 +88,7 @@ def put(json_request_body: dict, replica: str):
     get_elasticsearch_subscription_index(es_client, index_name, logger, index_mapping)
 
     #  get all indexes that use current alias
-    alias_name = Config.get_es_index_name(ESIndexType.docs, Replica[replica])
+    alias_name = Config.get_es_alias_name(ESIndexType.docs, Replica[replica])
     doc_indexes = [i['i'] for i in es_client.cat.aliases(name=alias_name, h=['a', 'i'], format='json')]
     #  try to subscribe query to each of the indexes.
     subscribed_indexes = []
--- a/dss/api/subscriptions.py
+++ b/dss/api/subscriptions.py
@@ -172,12 +172,13 @@ def delete(uuid: str, replica: str):
 
 
 def _unregister_percolate(es_client: Elasticsearch, subscribed_indexes: List[str], uuid: str):
-    # TODO Check return value and log errors
-    es_client.delete_by_query(index=subscribed_indexes,
-                              doc_type=ESDocType.query.name,
-                              body={"query": {"ids": {"type": ESDocType.query.name, "values": [uuid]}}},
-                              conflicts="proceed",
-                              refresh=True)
+    response = es_client.delete_by_query(index=subscribed_indexes,
+                                         doc_type=ESDocType.query.name,
+                                         body={"query": {"ids": {"type": ESDocType.query.name, "values": [uuid]}}},
+                                         conflicts="proceed",
+                                         refresh=True)
+    if response['failures']:
+        logger.error("Failed to unregister percolate query for subscription %s: %s", uuid, response)
 
 
 def _register_percolate(es_client: Elasticsearch, index_name: str, uuid: str, es_query: dict, replica: str):
--- a/dss/api/subscriptions.py
+++ b/dss/api/subscriptions.py
@@ -1,23 +1,14 @@
 import datetime
-import io
-import json
-import re
-import typing
 from uuid import uuid4
 
-import iso8601
 import requests
-
-from cloud_blobstore import BlobNotFoundError
 from elasticsearch import Elasticsearch
 from elasticsearch.exceptions import ElasticsearchException, NotFoundError
 from elasticsearch_dsl import Search
-from flask import jsonify, make_response, redirect, request
-from werkzeug.exceptions import BadRequest
+from flask import jsonify, request
 
 from .. import Config, Replica, ESIndexType, ESDocType, get_logger
 from ..error import DSSException, dss_handler
-from ..hcablobstore import FileMetadata, HCABlobStore
 from ..util.es import ElasticsearchClient, get_elasticsearch_subscription_index
 
 logger = get_logger()
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
     # john@example.com would show up because elasticsearch matched example w/ example.
     # By including "index": "not_analyzed", Elasticsearch leaves all owner inputs alone.
     index_name = Config.get_es_index_name(ESIndexType.subscriptions, Replica[replica])
-    #  TODO (tsmith): get all indexes that use current alais
-    #  TODO (tsmith): try to subscribe query to each of the indexes.
-    #  TODO (tsmith): error if no queries are indexed.
     get_elasticsearch_subscription_index(es_client, index_name, logger, index_mapping)
 
-    try:
-        percolate_registration = _register_percolate(es_client, uuid, es_query, replica)
-        logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
-    except ElasticsearchException as ex:
+    #  get all indexes that use current alias
+    alias_name = Config.get_es_index_name(ESIndexType.docs, Replica[replica])
+    doc_indexes = [i['i'] for i in es_client.cat.aliases(name=alias_name, h=['a', 'i'], format='json')]
+    #  try to subscribe query to each of the indexes.
+    subscribed_indexes = []
+    for doc_index in doc_indexes:
+        try:
+
+            percolate_registration = _register_percolate(es_client, doc_index, uuid, es_query, replica)
+            logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
+            subscribed_indexes.append(doc_index)
+        except ElasticsearchException as ex:
+            last_ex = ex
+
+    # error if no queries are successfully indexed.
+    if len(subscribed_indexes) == 0:
         logger.critical("%s", f"Percolate query registration failed: owner: {owner}, uuid: {uuid}, "
-                              f"replica: {replica}, es_query: {es_query}, Exception: {ex}")
+                              f"replica: {replica}, es_query: {es_query}, Exception: {Exception(last_ex)}")
         raise DSSException(requests.codes.internal_server_error,
                            "elasticsearch_error",
-                           "Unable to register elasticsearch percolate query!")
+                           "Unable to register elasticsearch percolate query!") from Exception(last_ex)
 
     json_request_body['owner'] = owner
+    json_request_body['subscribed_indexes'] = subscribed_indexes
 
     try:
         subscription_registration = _register_subscription(es_client, uuid, json_request_body, replica)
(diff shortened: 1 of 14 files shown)


kozbo:
hold before merging  due to missing pre-requisites (reindexing deployed)

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
+
+            percolate_registration = _register_percolate(es_client, doc_index, uuid, es_query, replica)
+            logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
+            subscribed_indexes.append(doc_index)
+        except ElasticsearchException as ex:
+            last_ex = ex
+
+    # error if no queries are successfully indexed.
+    if len(subscribed_indexes) == 0:
         logger.critical("%s", f"Percolate query registration failed: owner: {owner}, uuid: {uuid}, "
```
hannes-ucsc: `logger.critical(f"Percolate…`

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
-                              f"replica: {replica}, es_query: {es_query}, Exception: {ex}")
+                              f"replica: {replica}, es_query: {es_query}, Exception: {Exception(last_ex)}")
```
hannes-ucsc: Q: Why is the wrapping the exception necessary?

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
         raise DSSException(requests.codes.internal_server_error,
                            "elasticsearch_error",
-                           "Unable to register elasticsearch percolate query!")
+                           "Unable to register elasticsearch percolate query!") from Exception(last_ex)
```
hannes-ucsc: This is wrapping it again. Might want to extract a variable as in  `last_ex = Exception(last_ex)` above.

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
+    alias_name = Config.get_es_alias_name(ESIndexType.docs, Replica[replica])
+    doc_indexes = [i['i'] for i in es_client.cat.aliases(name=alias_name, h=['a', 'i'], format='json')]
+    #  try to subscribe query to each of the indexes.
+    subscribed_indexes = []
+    for doc_index in doc_indexes:
+        try:
+
+            percolate_registration = _register_percolate(es_client, doc_index, uuid, es_query, replica)
+            logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
+            subscribed_indexes.append(doc_index)
```
hannes-ucsc: Nit: This line and the one above would be better in an `else:` branch underneath the except.

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
+        except ElasticsearchException as ex:
+            last_ex = ex
```
hannes-ucsc: As I said below, I would `logger.debug()` the exception here.

dss/api/subscriptions.py
```diff
@@ -164,8 +160,15 @@ def delete(uuid: str, replica: str):
     return jsonify({'timeDeleted': time_deleted}), requests.codes.okay
 
 
-def _register_percolate(es_client: Elasticsearch, uuid: str, es_query: dict, replica: str):
-    index_name = Config.get_es_index_name(ESIndexType.docs, Replica[replica])
+def _unregister_percolate(es_client: Elasticsearch, subscribed_indexes: list, uuid: str):
```
hannes-ucsc: Nit: list[str] would be more informative

dss/api/subscriptions.py
```diff
@@ -164,8 +160,15 @@ def delete(uuid: str, replica: str):
...
+    for index in subscribed_indexes:
+        es_client.delete(index=index,
```
hannes-ucsc: According to https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-index.html this should work:

```
    es_client.delete(index=','.join(subscribed_indexes), 
                     doc_type=ESDocType.query.name, 
                     id=uuid, 
                     refresh=True)
```

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
-    #  TODO (tsmith): error if no queries are indexed.
     get_elasticsearch_subscription_index(es_client, index_name, logger, index_mapping)
 
-    try:
-        percolate_registration = _register_percolate(es_client, uuid, es_query, replica)
-        logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
-    except ElasticsearchException as ex:
+    #  get all indexes that use current alias
+    alias_name = Config.get_es_alias_name(ESIndexType.docs, Replica[replica])
+    doc_indexes = [i['i'] for i in es_client.cat.aliases(name=alias_name, h=['a', 'i'], format='json')]
```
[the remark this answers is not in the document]
Bento007: I could not find a simpler way of getting indexes tied to an alias. I'll fix the names

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
+    #  try to subscribe query to each of the indexes.
+    subscribed_indexes = []
+    for doc_index in doc_indexes:
+        try:
+
```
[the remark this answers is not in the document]
Bento007: will removed

dss/api/subscriptions.py
Bento007: Nice

dss/api/subscriptions.py
Bento007: will fix

dss/api/subscriptions.py
Bento007: I agree, minimize the try block.

dss/api/subscriptions.py
Bento007: Will do.

dss/api/subscriptions.py
```diff
@@ -94,22 +85,32 @@ def put(json_request_body: dict, replica: str):
...
+            percolate_registration = _register_percolate(es_client, doc_index, uuid, es_query, replica)
+            logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
+            subscribed_indexes.append(doc_index)
+        except ElasticsearchException as ex:
+            last_ex = ex
+
+    # error if no queries are successfully indexed.
+    if len(subscribed_indexes) == 0:
```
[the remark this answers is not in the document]
Bento007: I added a comment. I could also return the number of indexes subscribed to out of total indexes in the alias? Though I think that might just confuse people.

dss/api/subscriptions.py
Bento007: As strange as it looks we decided this the prefered way. The reason being if one of the variable in the f-string is resolved and has a `%` operator it could cause an exception for an unresolved parameter. This is our work around.

dss/api/subscriptions.py
```diff
@@ -180,3 +192,6 @@ def _register_subscription(es_client: Elasticsearch, uuid: str, json_request_bod
                            id=uuid,
                            body=json_request_body,
                            refresh=True)
+
+def _get_indexes_by_alias(es_client: Elasticsearch, alias_name: str):
+    return [indexes['i'] for indexes in es_client.cat.aliases(name=alias_name, h=['i'], format='json')]
```
hannes-ucsc: `list(es_client.indices.get_alias('foo').keys())`

dss/api/subscriptions.py
```diff
@@ -94,20 +90,32 @@ def put(json_request_body: dict, replica: str):
...
+        except ElasticsearchException as ex:
+            logger.debug(f"Exception occured when registering a document to an index. Exception: {ex}")
+            last_ex = ex
+        else:
+            logger.debug(f"Percolate query registration succeeded:\n{percolate_registration}")
+            subscribed_indexes.append(doc_index)
+
+    # Queries are unlikely to fit in all of the indexes, therefore errors will almost always occur. Only return an error
+    # if no queries are successfully indexed.
+    if len(doc_indexes) > 0 and not subscribed_indexes:
```
hannes-ucsc: `if doc_indexes and not subscribed_indexes` because non-empty lists are `True-ish`

I prefer this whenever it reads better to the human eye, and I think in this case it does.

hannes-ucsc:
@Bento007 One more thing, I actually rebased your branch yesterday and force pushed it. I remember there being conflicts. Since you rebased this branch, too, and must have seen those conflicts, I wonder if you resolved them properly. I marked some of the hunks I think are wrong in my review.

What tool do you use to resolve conflicts?

hannes-ucsc:
@Bento007, never mind, I see what you did there. You cherry-picked 068fc1053b0476da23125de6f0c7d47005e6451a.

dss/api/subscriptions.py
mikebaumann: Yes, that works unless the alias is not defined, in which case either an exception is thrown (default), or an error tuple is returned (if "ignore=404" is also passed in). I also don't like using the 'cat' API. Instead, we may want to wrap the snippet you provided in an easy-to-use convenience function that does what we need it to, instead of using 'cat'.

hannes-ucsc: I'd prefer

```
    try:
        return list(es_client.indices.get_alias(alias_name).keys())
    except NotFoundError:
        return []
```

mikebaumann: Yup, looks good to me. Thanks!

dss/api/subscriptions.py
```diff
@@ -1,23 +1,15 @@
 import datetime
-import io
-import json
-import re
-import typing
+from typing import List
```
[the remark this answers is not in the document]
Bento007: it's used for `_unregister_percolate` list typing.

dss/api/subscriptions.py
Bento007: I agree

dss/api/subscriptions.py
Bento007: fixed

dss/events/handlers/index.py
```diff
@@ -12,7 +12,7 @@
 import requests
 from cloud_blobstore import BlobStore, BlobStoreError
 from collections import defaultdict
-from elasticsearch.helpers import scan
+from elasticsearch.helpers import scan, bulk, streaming_bulk, BulkIndexError
```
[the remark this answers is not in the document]
Bento007: removed

dss/events/handlers/index.py
```diff
@@ -205,6 +201,41 @@ def add_data_to_elasticsearch(bundle_id: str, index_data: dict, index_name: str,
                      json.dumps(index_data, indent=4))
         raise
 
+    try:
```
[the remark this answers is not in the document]
Bento007: Resolved

dss/events/handlers/mapping.json
```diff
@@ -7,6 +7,11 @@
           "filter": ["lowercase"]
         }
       }
+    },
```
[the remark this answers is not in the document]
Bento007: resolved

dss/api/subscriptions.py
```diff
@@ -164,8 +170,14 @@ def delete(uuid: str, replica: str):
     return jsonify({'timeDeleted': time_deleted}), requests.codes.okay
 
 
-def _register_percolate(es_client: Elasticsearch, uuid: str, es_query: dict, replica: str):
-    index_name = Config.get_es_index_name(ESIndexType.docs, Replica[replica])
+def _unregister_percolate(es_client: Elasticsearch, subscribed_indexes: List[str], uuid: str):
+    es_client.delete(index=subscribed_indexes,
```
hannes-ucsc: Use alias here.

dss/config.py
```diff
@@ -216,7 +216,7 @@ def get_es_index_name(index_type: ESIndexType, replica: Replica,
     def get_es_alias_name(index_type: ESIndexType, replica: Replica) -> str:
         """Returns the alias for indexes"""
         deployment_stage = os.environ["DSS_DEPLOYMENT_STAGE"]
-        index = f"dss-{index_type.name}-{deployment_stage}-{replica.name}"
+        index = f"dss-{index_type.name}-{replica.name}-{deployment_stage}-alias"
```
hannes-ucsc: dss-{stage}-{replica}-{type}-{schema} for indices
dss-{stage}-{replica}-{type}-alias for the alias

dss/events/handlers/index.py
```diff
@@ -205,6 +201,41 @@ def add_data_to_elasticsearch(bundle_id: str, index_data: dict, index_name: str,
...
+                             '_id': hit['_id'],
+                             '_source': hit['_source']['es_query']
+                             }
+                            for hit in scan(ElasticsearchClient.get(logger),
+                                            index=subscription_index_name,
+                                            doc_type=ESDocType.subscription.name,
+                                            query={'query': {'match_all': {}}})
+                            ]
+
+    if len(subscription_queries) > 0:
```
hannes-ucsc: `if subscription_queries:`

dss/config.py
Bento007: updated

dss/api/subscriptions.py
Bento007: changed

dss/events/handlers/index.py
Bento007: fixes

Issue #582, opened 2017-10-25
Bug: do not return _source when output_format != raw


Bento007:
### Expected behavior
When output_format != raw don't return the source field from Elasticsearch. This can fixed by add `"_source": false` to the query. Example:
```
{
    "_source": false,
    "query" : {
        "term" : { "user" : "kimchy" }
    }
}
```

This will require inspecting the query for "_source" if it is not present and output_format != raw then add `"_source": false`.

* Test user defined _source is used when defined and output_format is summary
* Test _source is not returned when output_format is summary and no user defined _source in query.
* Test Error is returned when output_format is raw and user defines `"_source": false`

See for more details on [source filtering](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-source-filtering.html)

### Actual behavior
When a query is performed `_sources` is returned even when it's not used. This is a waste of resources.

mweiden:
https://github.com/HumanCellAtlas/data-store/pull/661

PR #661, opened 2017-11-22
DSS-582: only return _source info on 'raw' output_format

mweiden:
Fixes https://github.com/HumanCellAtlas/data-store/issues/582
Connects to https://waffle.io/HumanCellAtlas/data-store/cards/59f0d13a29243f01cc41eb1d

### Test plan
TODO: The absence of the `metadata` field in non-raw output queries is tested for. The presence of `_source` information in the `metadata` field is tested for in raw output queries.

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -20,6 +20,7 @@ pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noq
 sys.path.insert(0, pkg_root)  # noqa
 
 import dss
+from copy import deepcopy
 from dss.config import IndexSuffix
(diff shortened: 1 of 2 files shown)


Bento007:
as a naming convention our branches usually start with your username, for example: `mweiden-582-source`

tests/test_search.py
```diff
@@ -99,6 +110,7 @@ def test_search_returns_no_results_when_no_documents_indexed(self):
     def test_search_returns_no_result_when_query_does_not_match_indexed_documents(self):
         query = \
             {
+                "_source": False,
```
Bento007: why did you add this?

mweiden:
@Bento007 ty

PR #662, opened 2017-11-27
Index documents according to metadata schema major version number.

### LINKED ISSUE #594: Index bundles according to set of schema major versions

mikebaumann:
Place bundle index document in an ES index appropriate for the shape/structure/format 
of the metadata in the index document, so that it may be indexed successfully
without conflict.

The name of the index includes identification of the metadata schema release major number:
For example:
    v3 - Bundle contains metadata in the version 3 format
    v4 - Bundle contains metadata in the version 4 format
    ...

If no metadata version information is contained in the bundle, the bundle index document
is (attempted to be) added to the an index without a version identifier.
Currently this occurs in the case of the empty bundle used for deployment testing,
and perhaps other test data.

If/when bundle schemas are available, indexing should be updated to reflect the
bundle schema type and major version number.

Fixes #594

### DIFF
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -167,7 +167,7 @@ def get_index_shape_identifier(index_document: dict, logger) -> str:
             "The bundle contains mixed schema major version numbers: {}".format(sorted(list(schema_version_map.keys())))
         return "v" + list(schema_version_map.keys())[0]
     else:
-        return ""  # No files with schema identifiers were found
+        return None  # No files with schema identifiers were found
 
 
 def get_bundle_id_from_key(bundle_key: str) -> str:
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -6,10 +6,12 @@ import re
 import uuid
 import ipaddress
 import socket
+import typing
 from urllib.parse import urlparse, unquote
 
 import requests
 from cloud_blobstore import BlobStore, BlobStoreError
+from collections import defaultdict
 from elasticsearch.helpers import scan
 from requests_http_signature import HTTPSignatureAuth
 
@@ -51,10 +53,10 @@ def process_new_indexable_object(bucket_name: str, key: str, replica: str, logge
         bundle_id = get_bundle_id_from_key(key)
         index_data = create_index_data(blobstore, bucket_name, bundle_id, manifest, logger)
         alias_name = Config.get_es_alias_name(ESIndexType.docs, Replica[replica])
-        #  TODO (tsmith): get the major version from index data.
-        version = '1'
-        index_name = Config.get_es_index_name(ESIndexType.docs, Replica[replica], version)
-        add_index_data_to_elasticsearch(bundle_id, index_data, index_name, alias_name, logger)
+        index_shape_identifier = get_index_shape_identifier(index_data, logger)
+        index_name = Config.get_es_index_name(ESIndexType.docs, Replica[replica], index_shape_identifier)
+        get_elasticsearch_index(index_name, alias_name, logger)
+        add_data_to_elasticsearch(bundle_id, index_data, index_name, logger)
         subscriptions = find_matching_subscriptions(index_data, alias_name, logger)
         process_notifications(bundle_id, subscriptions, replica, logger)
         logger.debug(f"Finished index processing of {replica} creation event for bundle: {key}")
@@ -124,6 +126,48 @@ def create_index_data(handle: BlobStore, bucket_name: str, bundle_id: str, manif
             index_files[index_filename] = file_json
     index['files'] = index_files
     return index
+
+
+def get_index_shape_identifier(index_document: dict, logger) -> str:
+    """ Return string identifying the shape/structure/format of the data in the index document,
+    so that it may be indexed appropriately.
+
+    Currently, this returns a string identifying the metadata schema release major number:
+    For example:
+        v3 - Bundle contains metadata in the version 3 format
+        v4 - Bundle contains metadata in the version 4 format
+        ...
+
+    This includes verification that schema major number is the same for all index metadata
+    files in the bundle, as is the working plan for the 2017 Q4 demo.
+    If no metadata version information is contained in the bundle, the empty string is returned.
+    Currently this occurs in the case of the empty bundle used for deployment testing.
+
+    If/when bundle schemas are available, this function should be updated to reflect the
+    bundle schema type and major version number.
+
+    Other projects (non-HCA) may manage their metadata schemas (if any) and schema versions.
+    This should be an extension point that is customizable by other projects according to their metadata.
+    """
+
+    schema_version_map = defaultdict(set)  # type: typing.MutableMapping[str, typing.MutableSet[str]]
+    files = index_document['files']
+    for filename, file_content in files.items():
+        core = file_content.get('core')
+        if core is not None:
+            schema_type = core['type']
+            schema_version = core['schema_version']
+            schema_version_major = schema_version.split(".")[0]
+            schema_version_map[schema_version_major].add(schema_type)
+        else:
+            logger.info("%s", (f"File {filename} does not contain a 'core' section to identify "
+                               "the schema and schema version."))
+    if schema_version_map:
+        assert len(schema_version_map.keys()) == 1, \
+            "The bundle contains mixed schema major version numbers: {}".format(sorted(list(schema_version_map.keys())))
+        return "v" + list(schema_version_map.keys())[0]
+    else:
+        return ""  # No files with schema identifiers were found
 
 
 def get_bundle_id_from_key(bundle_key: str) -> str:
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -288,21 +288,21 @@ class TestIndexerBase(DSSAssertMixin, DSSStorageMixin, DSSUploadMixin):
                 }
             }
         }
-        with self.subtest("Same major version."):
+        with self.subTest("Same major version."):
             self.assertEqual(get_index_shape_identifier(index_document, logger), "v3")
 
         index_document['files']['assay_json']['core']['schema_version'] = "4.0.0"
-        with self.subtest("Mixed/inconsistent metadata schema release versions in the same bundle"):
+        with self.subTest("Mixed/inconsistent metadata schema release versions in the same bundle"):
             with self.assertRaisesRegex(AssertionError,
                                         "The bundle contains mixed schema major version numbers: \['3', '4'\]"):
                 get_index_shape_identifier(index_document, logger)
 
         index_document['files']['sample_json']['core']['schema_version'] = "4.0.0"
-        with self.subtest("Consistent versions, with a different version value"):
+        with self.subTest("Consistent versions, with a different version value"):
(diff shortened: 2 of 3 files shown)


tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
                          f"WARNING:.*:Failed notification for subscription {subscription_id}"
                          f" for bundle {bundle_id} with transaction id .+ Code: {error_response_code}")
 
-    def test_alias_exists(self):
+    def test_get_index_shape_identifier(self):
```
Bento007: Don't test different conditions in the same test. Either create a new test case or use subtest.

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
...
+                    'sample_json': {
+                        'core': {
+                            'schema_url': "http://hgwdev.soe.ucsc.edu/~kent/hca/schema/sample.json",
+                            'schema_version': "3.0.0",
+                            'type': "sample"
+                        }
+                    }
+                }
+            }
+        self.assertEqual(get_index_shape_identifier(index_document, logger), "v3")
```
Bento007: new test or subtest `with self.subtest("postive test"):`

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
...
+
+        # Test with mixed/inconsistent metadata schema release versions in the same bundle
```
Bento007: new test or subtest  `with self.subtest("mixed/inconsistent metadata schema release versions in the same bundle"):`

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
...
+        index_document['files']['assay_json']['core']['schema_version'] = "4.0.0"
+        with self.assertRaisesRegex(AssertionError,
+                                    "The bundle contains mixed schema major version numbers: \['3', '4'\]"):
+            get_index_shape_identifier(index_document, logger)
+
+        # Test with consistent versions, with a different version value
```
Bento007: new test or subtest `with self.subtest("consistent versions, with a different version value"):`

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
...
+        index_document['files']['sample_json']['core']['schema_version'] = "4.0.0"
+        self.assertEqual(get_index_shape_identifier(index_document, logger), "v4")
+
+        # Test with an version file and unversioned file
```
Bento007: new test or subtest `with self.subtest("a version file and unversioned file"):`

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
...
+        index_document['files']['assay_json'].pop('core')
+        with self.assertLogs(logger, level="INFO") as log_monitor:
+            get_index_shape_identifier(index_document, logger)
+        self.assertRegex(log_monitor.output[0], ("INFO:.*File assay_json does not contain a 'core' section "
+                                                 "to identify the schema and schema version."))
+        self.assertEqual(get_index_shape_identifier(index_document, logger), "v4")
+
+        # Test with no versioned file
```
Bento007: new test or subtest `with self.subtest("no versioned file"):`

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
...
+        self.assertEqual(get_index_shape_identifier(index_document, logger), "")
+
+    def test_alias_and_versioned_index_exists(self):
         sample_event = self.create_sample_bundle_created_event(self.bundle_key)
         self.process_new_indexable_object(sample_event, logger)
         es_client = ElasticsearchClient.get(logger)
         self.assertTrue(es_client.indices.exists_alias(name=[self.dss_alias_name]))
+        alias = es_client.indices.get_alias(name=[self.dss_alias_name])
+        doc_index_name = dss.Config.get_es_index_name(dss.ESIndexType.docs, dss.Replica[self.replica], "v3")
+        self.assertTrue(doc_index_name in alias)
```
Bento007: use `assertIn`

dss/events/handlers/index.py
```diff
@@ -126,6 +128,48 @@ def create_index_data(handle: BlobStore, bucket_name: str, bundle_id: str, manif
...
+            schema_type = core['type']
+            schema_version = core['schema_version']
+            schema_version_major = schema_version.split(".")[0]
+            schema_version_map[schema_version_major].add(schema_type)
+        else:
+            logger.info("%s", (f"File {filename} does not contain a 'core' section to identify "
+                               "the schema and schema version."))
+    if schema_version_map:
+        assert len(schema_version_map.keys()) == 1, \
+            "The bundle contains mixed schema major version numbers: {}".format(sorted(list(schema_version_map.keys())))
```
Bento007: you don't need list(). Also is sorting there just to make testing easier?

dss/events/handlers/index.py
```diff
@@ -126,6 +128,48 @@ def create_index_data(handle: BlobStore, bucket_name: str, bundle_id: str, manif
...
+        return "v" + list(schema_version_map.keys())[0]
+    else:
+        return ""  # No files with schema identifiers were found
```
hannes-ucsc: I changed my mind: I think this should be `None`. That is IMO the most natural value for something that "you don't have" or "does not exist". 

get_es_index_name should be changed to explicitly test for None rather than any false-ish value.

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
                          f"WARNING:.*:Failed notification for subscription {subscription_id}"
                          f" for bundle {bundle_id} with transaction id .+ Code: {error_response_code}")
 
-    def test_alias_exists(self):
+    def test_get_index_shape_identifier(self):
+        from dss.events.handlers.index import get_index_shape_identifier
+        index_document = \
```
hannes-ucsc: Not a change-request but dict(…) is more concise than `{…}` when the keys are valid identifiers:

```diff
@@ -270,25 +270,18 @@
 
     def test_get_index_shape_identifier(self):
         from dss.events.handlers.index import get_index_shape_identifier
-        index_document = \
-            {
-                'files': {
-                    'assay_json': {
-                        'core': {
-                            'schema_url': "http://hgwdev.soe.ucsc.edu/~kent/hca/schema/assay.json",
-                            'schema_version': "3.0.0",
-                            'type': "assay"
-                        }
-                    },
-                    'sample_json': {
-                        'core': {
-                            'schema_url': "http://hgwdev.soe.ucsc.edu/~kent/hca/schema/sample.json",
-                            'schema_version': "3.0.0",
-                            'type': "sample"
-                        }
-                    }
-                }
-            }
+        index_document = dict(
+            files=dict(
+                assay_json=dict(
+                    core=dict(
+                        schema_url="http://hgwdev.soe.ucsc.edu/~kent/hca/schema/assay.json",
+                        schema_version="3.0.0",
+                        type="assay")),
+                sample_json=dict(
+                    core=dict(
+                        schema_url="http://hgwdev.soe.ucsc.edu/~kent/hca/schema/sample.json",
+                        schema_version="3.0.0",
+                        type="sample"))))
         self.assertEqual(get_index_shape_identifier(index_document, logger), "v3")
 
         # Test with mixed/inconsistent metadata schema release versions in the same bundle
```

Also, having the closing curly braces on separate lines is unconventional in the Python world, IIRC.

tests/test_indexer.py
```diff
@@ -268,11 +268,60 @@ def test_subscription_notification_unsuccessful(self):
...
+                    },
+                    'sample_json': {
+                        'core': {
+                            'schema_url': "http://hgwdev.soe.ucsc.edu/~kent/hca/schema/sample.json",
+                            'schema_version': "3.0.0",
+                            'type': "sample"
+                        }
+                    }
+                }
+            }
```
hannes-ucsc: Not a change-request but dict(…) is more concise than `{…}` when the keys are valid identifiers:

```diff
@@ -270,25 +270,18 @@
 
     def test_get_index_shape_identifier(self):
         from dss.events.handlers.index import get_index_shape_identifier
-        index_document = \
-            {
-                'files': {
-                    'assay_json': {
-                        'core': {
-                            'schema_url': "http://hgwdev.soe.ucsc.edu/~kent/hca/schema/assay.json",
-                            'schema_version': "3.0.0",
-                            'type': "assay"
-                        }
-                    },
-                    'sample_json': {
-                        'core': {
-                            'schema_url': "http://hgwdev.soe.ucsc.edu/~kent/hca/schema/sample.json",
-                            'schema_version': "3.0.0",
-                            'type': "sample"
-                        }
-                    }
-                }
-            }
+        index_document = dict(
+            files=dict(
+                assay_json=dict(
+                    core=dict(
+                        schema_url="http://hgwdev.soe.ucsc.edu/~kent/hca/schema/assay.json",
+                        schema_version="3.0.0",
+                        type="assay")),
+                sample_json=dict(
+                    core=dict(
+                        schema_url="http://hgwdev.soe.ucsc.edu/~kent/hca/schema/sample.json",
+                        schema_version="3.0.0",
+                        type="sample"))))
         self.assertEqual(get_index_shape_identifier(index_document, logger), "v3")
 
         # Test with mixed/inconsistent metadata schema release versions in the same bundle
```

Also, having the closing curly braces on separate lines is unconventional in the Python world, IIRC.

hannes-ucsc:
Since the original base branch has landed, I changed the base branch to master.

Bento007:
@hannes-ucsc I think each test should only have one set of conditions to produce a specific outcome. Otherwise the intent of the test gets lost. I'd prefer each test be separated into it's own test case but subtest will work just as well in this case.

hannes-ucsc:
@Bento007 I don't think the intent gets lost because the assertion makes that clear in either case.

If I understand correctly, what subtests give you is better failure output and the ability to move past the first failing assertion in a test method.

https://www.caktusgroup.com/blog/2017/05/29/subtests-are-best/

My issue with your proposal in this case is that subtests don't actually achieve the second goal when state is modified between assertions:

```
with subtest():
    a = []
    assert len() == 0
with subtest():
    a.append(1) # buggy
    assert len() == 1
with subtest():
    a.append(2)
    assert len() == 2
```

Now lets say append had a bug that prevented the first call to succeed. Without subtests you'd get just one failure. With subtests you'd get two but the second one does not provide any additional information about the nature of the failure. 

The shared state makes the individual subtests depend on each other. Yet subtests only provide better test output if they *don't* depend on each other.

The alternative of individual test methods would require duplicating a lot of code.

Bento007:
@hannes-ucsc I adjust the subtest to keep each test as individual as possible.

tests/test_indexer.py
Bento007: resolved

tests/test_indexer.py
Bento007: resolved

hannes-ucsc:
Good to go from my side.

PR #664, opened 2017-11-27
Force AWS CLI output to json if it has been configured

kislyuk:
Encountered this issue when setting up DSS on the biohub account.
### Test plan
Tested manually.


environment
```diff
@@ -63,6 +63,10 @@ DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_TEST="human-cell-atlas-travis-test.iam.gserv
 PYTHONWARNINGS=ignore:ResourceWarning,ignore::UserWarning:zipfile:
 set +a
 
+if grep -q output ~/.aws/config 2>/dev/null; then
```
ttung: can you run `aws configure get output` instead?

also, I think you should just error if it mismatches.

ttung: Also, you could set `AWS_DEFAULT_OUTPUT` instead, which overrides the user's `.aws/config`.

kislyuk: Thanks for the pointer. You're right, setting the environment variable is preferable.

(I didn't want to run `aws configure get` because it's slow)

kislyuk: Updated to set the environment variable instead of running `aws configure`.

PR #665, opened 2017-11-27
Grant permission to create state machine to the Travis CI group

amork:
Grant Travis CI group permission to create state machine. This is needed as port of the deployment performed by Travis


ttung:
Does something like `states:*` make sense?

amork:
@ttung I'd rather avoid granting more permissions than needed. Since we now have a workaround for the policy file size limitation I'd keep this as is, if that's ok with you.

PR #673, opened 2017-11-28
Clear index per test

### LINKED ISSUE #644: python elasticsearch connection timeout / retries too high
I'm applying this diff:
```
[czipa1osx186 (.venv)]:~/hca/data-store:master> hg diff dss/api/search.py 
diff -r a8e024e21e60 dss/api/search.py
--- a/dss/api/search.py Mon Nov 13 14:02:44 2017 -0800
+++ b/dss/api/search.py Thu Nov 16 13:51:15 2017 -0800
@@ -52,6 +52,8 @@
     # TODO: (tsmith12) determine if a search operation timeout limit is needed
     try:
         es_client = ElasticsearchClient.get(get_logger())
+        import time
+        print("before time: ", time.time())
         if _scroll_id is None:
             page = es_client.search(index=Config.get_es_index_name(ESIndexType.docs, Replica[replica]),
                                     doc_type=ESDocType.doc.name,
@@ -64,6 +66,7 @@
         else:
             get_logger().debug("%s", f"Retrieve ES results from scroll instance Scroll_id: {_scroll_id}")
             page = es_client.scroll(scroll_id=_scroll_id, scroll=scroll)
+        print("after time: ", time.time())
 
         # TODO: (tsmith12) allow users to retrieve previous search results
         _scroll_id = page['_scroll_id']
```

and when I run `test_search.py`, I get this for the bogus ES endpoint test:

```
before time:  1510868889.667299
[2017-11-16 13:48:49,686] ERROR in search: Elasticsearch Invalid Endpoint. Exception: ConnectionError((<urllib3.connection.HTTPConnection object at 0x10a79a898>, 'Connection to bogus timed out. (connect timeout=10)')) caused by: ConnectTimeoutError((<urllib3.connection.HTTPConnection object at 0x10a79a898>, 'Connection to bogus timed out. (connect timeout=10)'))
.[2017-11-16T13:48:50,239][INFO ][o.e.c.m.MetaDataDeleteIndexService] [Q15ZIzd] [dss-docs-aws-dev-test_search/1-hT2TwtRgOCMRLh9Subww] deleting index
[2017-11-16T13:48:50,309][INFO ][o.e.c.m.MetaDataCreateIndexService] [Q15ZIzd] [dss-docs-gcp-dev-test_search] creating index, cause [api], templates [], shards [5]/[1], mappings [query, doc]
before time:  1510868930.386119
```

~ 40s.

This is longer than the API gateway timeout, so this is not desirable, as we will never trigger the correct exception codepath.  Instead, the global timeout will be triggered and we'll get a global timeout exception.  Furthermore, the connection library thinks the timeout is set to 10s, not 40s.  I'm not sure if this is retries.

### LINKED ISSUE #654: test_indexer is not a reliable test
https://travis-ci.org/HumanCellAtlas/data-store/builds/303896170

Bento007:
Clear the contents of ES indices between tests instead of removing the indices completely.
Index creation is time consuming and this improves the performance of the tests.

Fixes #654 
Fixes #644 

No conflicts

### DIFF
--- a/tests/es/__init__.py
+++ b/tests/es/__init__.py
@@ -1,5 +1,6 @@
 import logging
-
+from typing import List
+from dss.config import IndexSuffix, Config, BucketConfig
 from dss.util.es import ElasticsearchClient
 
 logging.basicConfig(level=logging.INFO)
@@ -8,6 +9,11 @@ logger.setLevel(logging.INFO)
 
 
 def elasticsearch_delete_index(index_name: str):
+    # ensure the indexes are test index.
+    assert Config._CURRENT_CONFIG == BucketConfig.TEST
+    assert IndexSuffix.name
+    assert index_name.endswith(IndexSuffix.name)
+
     try:
         es_client = ElasticsearchClient.get(logger)
         es_client.indices.delete(index=index_name, ignore=[404])
@@ -19,3 +25,22 @@ def elasticsearch_delete_index(index_name: str):
 def close_elasticsearch_connections(es_client):
     for conn in es_client.transport.connection_pool.connections:
         conn.pool.close()
+
+
+def clear_indexes(index_names: List[str], doctypes: List[str]):
+    '''erases all of the documents in indexes with any of the doctypes provided. This can only be used in TEST
+    configuration with IndexSuffix.name set. Only indexes with the same IndexSuffix.name can be erased.'''
+
+    # ensure the indexes are test index.
+    assert Config._CURRENT_CONFIG == BucketConfig.TEST
+    assert IndexSuffix.name
+    for index_name in index_names:
+        assert index_name.endswith(IndexSuffix.name)
+
+    es_client = ElasticsearchClient.get(logger)
+    if es_client.indices.exists(index_names):
+        es_client.delete_by_query(index=index_names,
+                                  body={'query': {'match_all': {}}},
+                                  doc_type=doctypes,
+                                  refresh=True,
+                                  conflicts='proceed')
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -102,7 +102,7 @@ class TestIndexerBase(DSSAssertMixin, DSSStorageMixin, DSSUploadMixin):
         Config.set_config(BucketConfig.TEST)
         _, _, cls.test_bucket = Config.get_cloud_specific_handles(cls.replica)
         cls.dss_alias_name = dss.Config.get_es_alias_name(dss.ESIndexType.docs, dss.Replica[cls.replica])
-        create_elasticsearch_index(f"initial_{IndexSuffix.name}", cls.replica, logger)
+        create_elasticsearch_index(f"initial_{IndexSuffix.name}", dss.Replica[cls.replica], logger)
         cls.subscription_index_name = dss.Config.get_es_index_name(dss.ESIndexType.subscriptions,
                                                                    dss.Replica[cls.replica])
 
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -102,7 +102,6 @@ class TestIndexerBase(DSSAssertMixin, DSSStorageMixin, DSSUploadMixin):
         Config.set_config(BucketConfig.TEST)
         _, _, cls.test_bucket = Config.get_cloud_specific_handles(cls.replica)
         cls.dss_alias_name = dss.Config.get_es_alias_name(dss.ESIndexType.docs, dss.Replica[cls.replica])
-        create_elasticsearch_index(f"initial_{IndexSuffix.name}", dss.Replica[cls.replica], logger)
         cls.subscription_index_name = dss.Config.get_es_index_name(dss.ESIndexType.subscriptions,
                                                                    dss.Replica[cls.replica])
 
@@ -155,6 +154,7 @@ class TestIndexerBase(DSSAssertMixin, DSSStorageMixin, DSSUploadMixin):
                                                          files=smartseq2_paried_ends_indexed_file_list)
 
     def test_key_is_not_indexed_when_processing_an_event_with_a_nonbundle_key(self):
+        elasticsearch_delete_index(f'*{IndexSuffix.name}')
         bundle_uuid = "{}.{}".format(str(uuid.uuid4()), get_version())
         bundle_key = "files/" + bundle_uuid
         sample_event = self.create_sample_bundle_created_event(bundle_key)
@@ -164,14 +164,7 @@ class TestIndexerBase(DSSAssertMixin, DSSStorageMixin, DSSUploadMixin):
             with self.assertLogs(logger, level="DEBUG") as log_monitor:
                 self.process_new_indexable_object(sample_event, logger)
             self.assertRegex(log_monitor.output[0], "DEBUG:.*Not indexing .* creation event for key: .*")
-            indexes = list(ElasticsearchClient.get(logger).indices.get_alias(self.dss_alias_name).keys())
-            for index in indexes:
-                with self.subTest(f"{index}"):
-                    with self.assertRaises(Exception) as ex:
-                        ElasticsearchClient.get(logger).get(index=index,
-                                                            doc_type=dss.ESIndexType.docs,
-                                                            id=bundle_uuid)
-                    self.assertIn('"found":false', ex.exception.error)
+            self.assertFalse(ElasticsearchClient.get(logger).indices.exists_alias(self.dss_alias_name))
         finally:
             logger.setLevel(log_last)
 
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -198,6 +198,7 @@ class TestIndexerBase(DSSAssertMixin, DSSStorageMixin, DSSUploadMixin):
 
     def test_indexed_file_access_error(self):
         inaccesssible_filename = "inaccessible_file.json"
+        elasticsearch_delete_index(f'*{IndexSuffix.name}')
         bundle_key = self.load_test_data_bundle_with_inaccessible_file(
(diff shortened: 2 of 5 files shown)


tests/test_indexer.py
```diff
@@ -382,6 +383,20 @@ def get_search_results(cls, query, expected_hit_count):
...
+            es_client = ElasticsearchClient.get(logger)
+            es_client.delete_by_query(index=self.dss_index_name,
+                                      body={'query': {'match_all': {}}},
+                                      doc_type=[dss.ESDocType.doc.name, dss.ESDocType.query.name],
+                                      refresh=True)
+            es_client.delete_by_query(index=self.subscription_index_name,
+                                      body={'query': {'match_all': {}}},
+                                      doc_type=dss.ESDocType.subscription.name,
+                                      refresh=True)
+        except TransportError:
```
ttung: Under what circumstances would a `TransportError` occur?  If it happens, it is better that we just crash and burn?

tests/test_search.py
```diff
@@ -64,15 +64,16 @@ def search_setup(cls, replica):
         cls.dss_index_name = dss.Config.get_es_index_name(dss.ESIndexType.docs, replica)
         with open(os.path.join(os.path.dirname(__file__), "sample_v3_index_doc.json"), "r") as fh:
             cls.index_document = json.load(fh)
+        dss.Config.set_config(dss.BucketConfig.TEST)
```
ttung: this is redundant with L63, no?

tests/test_indexer.py
Bento007: This occurs when the index has not been created and I try to delete from it. I could narrow the used of it to the specific test cases where the index is not created yet.

tests/test_indexer.py
```diff
@@ -146,6 +148,7 @@ def test_indexed_file_with_invalid_content_type(self):
                                                          files=smartseq2_paried_ends_indexed_file_list)
 
     def test_key_is_not_indexed_when_processing_an_event_with_a_nonbundle_key(self):
+        elasticsearch_delete_index(f"*{IndexSuffix.name}")
```
ttung: Why do we delete the index at the start of a test?

tests/test_indexer.py
```diff
@@ -162,6 +165,7 @@ def test_key_is_not_indexed_when_processing_an_event_with_a_nonbundle_key(self):
             self.assertEqual('index_not_found_exception', ex.exception.error)
         finally:
             logger.setLevel(log_last)
+            ElasticsearchClient.get(logger).indices.create(self.subscription_index_name)
```
ttung: .. and why do we create the index at the end of a test?

tests/test_indexer.py
```diff
@@ -382,6 +383,18 @@ def get_search_results(cls, query, expected_hit_count):
             else:
                 time.sleep(0.5)
 
+    def clear_test_indexes(self):
+        es_client = ElasticsearchClient.get(logger)
+        if es_client.indices.exists(self.dss_index_name):
```
ttung: under what circumstances would the index not exist at the end of a test?

tests/test_search.py
```diff
@@ -338,10 +338,19 @@ def populate_search_index(self, index_document: dict, count: int) -> list:
             bundle_id = f"{bundle_uuid}.{version}"
             bundle_url = (f"https://127.0.0.1:{self.app._port}"
                           f"/v1/bundles/{bundle_uuid}?version={version}&replica={self.replica_name}")
-            es_client.index(index=self.dss_index_name,
-                            doc_type=ESDocType.doc.name,
-                            id=bundle_id,
-                            body=index_document)
+            if i == count - 1:
```
ttung: Why this change?

Bento007: I was running into an issue where the test where complete faster than the shards where updating during the `clear_test_indexes` operation even with refresh=true. This was causing the incorrect number of documents to populate the index. To remedy I make sure the index is refreshed after the last documents is index.

tests/test_indexer.py
Bento007: This is because the test was original designed expecting the indexes to be deleted at the beginning. I'll adjust the expected error message returned to better reflect what is being tested and removed the create and delete of indexes.

tests/test_indexer.py
Bento007: the indexes are created when the first bundle is added. If that process fails, no index will be created.

Bento007:
I made clear_test_indexes generic.

tests/es/__init__.py
```diff
@@ -19,3 +19,14 @@ def elasticsearch_delete_index(index_name: str):
 def close_elasticsearch_connections(es_client):
     for conn in es_client.transport.connection_pool.connections:
         conn.pool.close()
+
+
+def clear_test_indexes(indexes: list, documents: list):
```
kislyuk: Nothing in this method ensures that the indexes deleted are test indexes, so it is more appropriately named `clear_indexes`.

kislyuk:
I'm concerned about the possibility of this code running against production or against a non-local ES instance (in which case it would result in test race conditions). Could you explain how those conditions are prevented, or implement failsafes to protect against them?

tests/es/__init__.py
Bento007: changed

Bento007:
@kislyuk this is handled by setting `refresh=True` for ES index operations. It’s slower but ensures the tests don’t proceed until the index is updated with the latest changes. Also each test module uses it's own unique index so there shouldn't be any other test or operations running that touch the same index.

tests/es/__init__.py
```diff
@@ -19,3 +19,14 @@ def elasticsearch_delete_index(index_name: str):
...
+
+
+def clear_indexes(indexes: list, documents: list):
+    es_client = ElasticsearchClient.get(logger)
+    if es_client.indices.exists(indexes):
+        es_client = ElasticsearchClient.get(logger)
+        es_client.delete_by_query(index=indexes,
+                                  body={'query': {'match_all': {}}},
+                                  doc_type=documents,
+                                  refresh='True',
```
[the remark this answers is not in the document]
Bento007: both are valid, but `True` would be more pythonic. I'll change it.

tests/es/__init__.py
```diff
@@ -19,3 +19,14 @@ def elasticsearch_delete_index(index_name: str):
 def close_elasticsearch_connections(es_client):
     for conn in es_client.transport.connection_pool.connections:
         conn.pool.close()
+
+
+def clear_indexes(indexes: list, documents: list):
```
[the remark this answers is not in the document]
ttung: I don't know why this comment didn't go through. :(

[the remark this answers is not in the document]
Bento007: Reasonable change. I'll fix it

tests/test_search.py
ttung: I don't know if this might be better.  Will leave to your discretion:

```
es_client.index(
  index=self.dss_index_name,
  doc_type=ESDocType.doc.name,
  id=bundle_id,
  body=index_document,
  refresh=(i == count - 1)
)
```

I am pretty weirded out by the fact that you're seeing the behavior you're describing, but I lack the ES expertise to make any judgement on it.

Certainly, you will want to leave a comment explaining this, because otherwise people will be very confused.

Bento007: I like it. Also I'll add in a comment.

mikebaumann:
@kislyuk
Thanks for raising the concern about `clear_indexes` unintentionally running against an important (non-test, e.g. production) index. 
I think the risk of this is low, as this method is located in test code (tests/es/__init__).
Yet, the risk could be further mitigated by having `clear_indexes` verify the index has the test suffix on the index name, which all indexes passed to `clear_indexes`, and no others (e.g. non-test, deployed) should.
Do you think we should add this check?
If we add this check, will it adequately address your concern?

Bento007:
Working on Rebasing

tests/es/__init__.py
```diff
@@ -19,3 +19,13 @@ def elasticsearch_delete_index(index_name: str):
 def close_elasticsearch_connections(es_client):
     for conn in es_client.transport.connection_pool.connections:
         conn.pool.close()
+
+
+def clear_indexes(indexes: list, doctypes: list):
+    es_client = ElasticsearchClient.get(logger)
```
mikebaumann: clear_indexes should ensure it is a test index, checking that IndexSuffix.name is defined and the index name has that as a suffix.

tests/test_indexer.py
```diff
@@ -323,7 +333,7 @@ def test_subscription_query_with_multiple_data_types_indexing_and_notification(s
 
         # Verify the query works correctly as a search
         search_results = self.get_search_results(subscription_query, 1)
-        self.assertEqual(1, len(search_results))
+        self.assertNotEqual(0, len(search_results))
```
mikebaumann: For this test to thoroughly test what it is intended to test, it is necessary for the indices to be removed at the start this specific test. Before this was done in setUp, yet because it is no longer being done there, it must be done explicitly at the start of this specific test.

tests/test_indexer.py
```diff
@@ -101,9 +102,13 @@ def indexer_setup(cls, replica):
         Config.set_config(BucketConfig.TEST)
         _, _, cls.test_bucket = Config.get_cloud_specific_handles(cls.replica)
         cls.dss_alias_name = dss.Config.get_es_alias_name(dss.ESIndexType.docs, dss.Replica[cls.replica])
+        create_elasticsearch_index(f"initial_{IndexSuffix.name}", dss.Replica[cls.replica], logger)
```
mikebaumann: This line (105) should not be necessary, and all the tests pass for me without it. Please remove this line.

PR #679, opened 2017-11-29
Update CI-CD IAM policy template, pull in sfn deploy update

kislyuk:
Unbreaks master.
### Test plan
Deployed locally while impersonating Travis.

### DIFF
--- a/iam/policy-templates/ci-cd.json
+++ b/iam/policy-templates/ci-cd.json
@@ -102,9 +102,7 @@
         "lambda:*",
         "es:*",
         "sns:*",
-        "states:StartExecution",
-        "states:DescribeExecution",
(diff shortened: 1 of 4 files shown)


iam/policy-templates/ci-cd.json
```diff
@@ -102,9 +102,7 @@
         "lambda:*",
         "es:*",
         "sns:*",
-        "states:StartExecution",
-        "states:DescribeExecution",
-        "states:CreateStateMachine"
+        "states:*"
```
amork: you are granting wider set of permissions than needed

kislyuk: These apply only specifically to the dss-* state machines, so I don't think it matters.

Issue #678, opened 2017-11-29
Add ability to maintain requirements{,-dev}.txt incrementally 


hannes-ucsc:
Related to #676 

@ttung, in https://github.com/HumanCellAtlas/data-store/pull/677#issuecomment-347768150:

> I think the particular use case described in #676 can be handled by two sets of Makefile rules:
>
> One builds an entirely new environment from requirements{,-dev}.txt.in. Another builds an environment from the existing requirements{,-dev}.txt and then applies the modified requirements{,-dev}.txt.in. This will constrain the changes to the newly added modules.

PR #680, opened 2017-11-29
refresh_all_requirements target that blows away requirements.

ttung:
1. Modifies the requirements{,-dev}.txt targets such that they are incremental, i.e., they install {requirements,-dev}.txt, then {requirements,-dev}.txt.in, and then freezes.  If you add just one new requirement to {requirements,-dev}.txt.in, it will not update everything.
2. refresh_all_requirements will empty out the existing requirements.

Connects to #678

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,11 +43,19 @@ clean:
 	git clean -df {chalice,daemons/*}/{chalicelib,domovoilib,vendor}
 	git checkout $$(git status --porcelain {chalice,daemons/*}/.chalice/config.json | awk '{print $$2}')
 
+refresh_all_requirements:
+	@echo -n '' >| requirements.txt
+	@echo -n '' >| requirements-dev.txt
+	@if [ $$(uname -s) == "Darwin" ]; then sleep 1; fi  # this is require because Darwin HFS+ only has second-resolution for timestamps.
+	@touch requirements.txt.in requirements-dev.txt.in
+	@$(MAKE) requirements.txt requirements-dev.txt
+
 requirements.txt requirements-dev.txt : %.txt : %.txt.in
 	[ ! -e .requirements-env ] || exit 1
-	echo "# You should not edit this file directly.  Instead, you should edit $<." > $@
 	virtualenv .requirements-env
+	.requirements-env/bin/pip install -r $@
 	.requirements-env/bin/pip install -r $<
+	echo "# You should not edit this file directly.  Instead, you should edit $<." >| $@
 	.requirements-env/bin/pip freeze >> $@
 	rm -rf .requirements-env
 	scripts/find_missing_wheels.py requirements.txt
--- a/README.md
+++ b/README.md
@@ -259,23 +259,22 @@ repeat the `make deploy` step from a privileged account any time you change the 
 The direct runtime dependencies of this project are defined in `requirements.txt.in`. Direct development dependencies
 are defined in `requirements-dev.txt.in`. All dependencies, direct and transitive, are defined in the corresponding
 `requirements.txt` and `requirements-dev.txt` files. The latter two can be generated using `make requirements.txt` or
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -34,12 +34,20 @@ clean:
 	git clean -df {chalice,daemons/*}/{chalicelib,domovoilib,vendor}
 	git checkout $$(git status --porcelain {chalice,daemons/*}/.chalice/config.json | awk '{print $$2}')
 
+refresh_all_requirements:
+	@echo -n '' >| requirements.txt
+	@echo -n '' >| requirements-dev.txt
+	@if [ $$(uname -s) == "Darwin" ]; then sleep 1; fi  # this is require because Darwin HFS+ only has second-resolution for timestamps.
+	@touch requirements.txt.in requirements-dev.txt.in
+	@echo 'You should now run `make requirements.txt requirements-dev.txt`'
```
hannes-ucsc: You could run a submake here instead.

Makefile
```diff
@@ -34,12 +34,20 @@ clean:
...
+
 requirements.txt requirements-dev.txt : %.txt : %.txt.in
 	[ ! -e .requirements-env ] || exit 1
-	echo "# You should not edit this file directly.  Instead, you should edit $<." > $@
 	virtualenv .requirements-env
 	source .requirements-env/bin/activate && \
```
hannes-ucsc: Since you're touching this, you can eliminate sourcing `activate` by invoking the pip binary in the venv. As in `.requirements-env/bin/pip`. The the `&&` chaining and the \ would go away.

ttung: #682 
one thesis per diff. :)

Makefile
ttung: Done.

README.md
```diff
@@ -268,14 +268,13 @@ If you need to modify or add a direct runtime dependency declaration, follow the
 1) Make sure there are no pending changes to `requirements.txt` or `requirements.txt.in`
 2) Run `make requirements.txt`. This will update `requirements.txt` to the most recent releases of direct and transitive dependencies.
 3) Examine the changes to `requirements.txt` and commit them with a message like `Bumping dependencies`
-4) Make the desired change to `requirements.txt.in`
+4) If necessary, make the desired change to `requirements.txt.in`
 5) Run `make requirement.txt` again. This updates `requirements.txt` with direct and transitive entries related to your change.
```
hannes-ucsc: I thought your change eliminated the need for two commits.

README.md
```diff
@@ -268,14 +268,13 @@ If you need to modify or add a direct runtime dependency declaration, follow the
...
 6) Examine the changes to `requirements.txt` and commit them with a message like `Adding dependency on Foobar`
 
 You now have two commits, one that catches up with updates to transitive dependencies, and one that tracks your explict
 change to a direct dependency. This process applies to development dependencies as well, except for
 `requirements-dev.txt` and `requirements-dev.txt.in` respectively.
 
-Avoid being overly restrictive when constraining the allowed version range of direct dependencies in
-`requirements.txt.in` and `requirements-dev.txt.in`. As a rule of thumb, pin only the major version, e.g. `foo > 1.2.3,
-< 2`. Note that `2`` is a shorthand for `2.0` which is equivalent to `2.0.0` and so on.
```
hannes-ucsc: I'm ok with removing the last sentence but the ones before that are useful, no?

hannes-ucsc: This must have leaked in from #683.

README.md
kislyuk: `s/requirement.txt/requirements.txt/`

ttung: I don't think there was ever a requirement for two commits.  I've made changes in a single commit in the past.

ttung:
Updated docs and landing now. If you have any requests for changes, I will be happy to address them in a subsequent diff.

PR #683, opened 2017-11-29
unpin requirements.txt and requirements-dev.txt

ttung:
Since we have an easy way to do small updates to requirements.txt and requirements-dev.txt, we should unpin so we can stay up to date and avoid the accumulation of technical debt.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,11 +43,19 @@ clean:
 	git clean -df {chalice,daemons/*}/{chalicelib,domovoilib,vendor}
 	git checkout $$(git status --porcelain {chalice,daemons/*}/.chalice/config.json | awk '{print $$2}')
 
+refresh_all_requirements:
+	@echo -n '' >| requirements.txt
+	@echo -n '' >| requirements-dev.txt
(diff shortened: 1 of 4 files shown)


kislyuk:
@hannes-ucsc, we discussed this at CZI and we would like to back to not pinning the top level source dependencies unless there is a specific reason to do so. I'd like to revert to the previous state of things until we discuss this more thoroughly, which is what this PR accomplishes.

hannes-ucsc:
Should we keep the lower bounds in place? I've seen pip downgrade dependencies before.

The README should be changed to reflect this change.

Lastly, #680 is a prerequisite for this, correct?

kislyuk:
I think keeping the lower bounds in place is a great idea.

ttung:
> Should we keep the lower bounds in place? I've seen pip downgrade dependencies before.

I don't think pip should/would ever downgrade dependencies.  The only time I've seen that happen is on a network-connectivity compromised situation, where it pulls from cache rather than hit the network.

In any case, I've updated this PR. to preserve the lower bounds.

> Lastly, #680 is a prerequisite for this, correct?

Yes, this is stacked on top of that PR.

PR #685, opened 2017-12-01
unpin requirements.txt and requirements-dev.txt

hannes-ucsc:
Continued from #683 which was merged into the wrong branch by mistake.


hannes-ucsc:
I merged this without review because consensus had already been reached on the same changes as part of #683.

PR #687, opened 2017-12-01
Finish changes for schema version specific subscription/notification

mikebaumann:
Ensure subscription queries are added to indices, and do so in a way
that allows queries including fields not defined in the index mappings.
This enables subscription of queries that work across metadata schema versions,
including fields that are not present in a schema version or have
changed between versions, such as a field changing from type text to type object.
This also resolves internal constraints of needing to have the mappings defined
in an index before a subscription query for percolation can be added to the index.

This is accomplished by setting index.percolator.map_unmapped_fields_as_string to true,
which also works in the case of other field types (e.g. numeric, date).
Note when dynamic templates are used and queries for percolation have been added
to an index before the index contains mappings of fields referenced by those queries,
the queries must be reloaded when the mappings change for the queries to match.
For more information, see: https://github.com/elastic/elasticsearch/issues/5750
### Test plan

- Unit tests updated
- Manual interactive testing with deployed system

When updating existing deployments:
The subscription index mapping must be updated. This may be done in a way that
preserves the existing subscriptions using an appropriate script utilizing
Elasticsearch Curator.
The document indicies also have a new setting, yet because this change is expected
to be merged with the schema specific indexing change, which will create new 
version-specific indicies, no specific additional migration is required for this setting. 

Connects to #598

### DIFF
--- a/dss/api/subscriptions.py
+++ b/dss/api/subscriptions.py
@@ -94,7 +94,7 @@ def put(json_request_body: dict, replica: str):
 
     #  get all indexes that use current alias
     alias_name = Config.get_es_alias_name(ESIndexType.docs, Replica[replica])
-    doc_indexes = [indexes['i'] for indexes in es_client.cat.aliases(name=alias_name, h=['i'], format='json')]
+    doc_indexes = _get_indexes_by_alias(es_client, alias_name)
 
     #  try to subscribe query to each of the indexes.
     subscribed_indexes = []
@@ -157,7 +157,7 @@ def delete(uuid: str, replica: str):
 
     #  get all indexes that use current alias
     alias_name = Config.get_es_alias_name(ESIndexType.docs, Replica[replica])
-    doc_indexes = [indexes['i'] for indexes in es_client.cat.aliases(name=alias_name, h=['i'], format='json')]
+    doc_indexes = _get_indexes_by_alias(es_client, alias_name)
     _unregister_percolate(es_client, doc_indexes, uuid)
 
     es_client.delete(index=Config.get_es_index_name(ESIndexType.subscriptions, Replica[replica]),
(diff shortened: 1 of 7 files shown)


dss/events/handlers/index.py
```diff
@@ -205,6 +201,41 @@ def add_data_to_elasticsearch(bundle_id: str, index_data: dict, index_name: str,
...
+                             }
+                            for hit in scan(ElasticsearchClient.get(logger),
+                                            index=subscription_index_name,
+                                            doc_type=ESDocType.subscription.name,
+                                            query={'query': {'match_all': {}}})
+                            ]
+
+    if len(subscription_queries) > 0:
+        try:
+            bulk(ElasticsearchClient.get(logger), iter(subscription_queries), refresh=True)
```
Bento007: doing a bulk operation with `refresh=True` sounds really expensive. Does it refresh after each operation or at the end of the bulk?

mikebaumann: The refresh is performed after the bulk operation, not for each item. This is described in the Elasticsearch documentation here: https://www.elastic.co/guide/en/elasticsearch/reference/2.2/docs-bulk.html#bulk-refresh
This is not expensive.
The refresh is necessary because the next step is to perform the percolate query of the index document against these queries.

dss/api/subscriptions.py
```diff
@@ -106,15 +110,14 @@ def put(json_request_body: dict, replica: str):
 
     # Queries are unlikely to fit in all of the indexes, therefore errors will almost always occur. Only return an error
     # if no queries are successfully indexed.
-    if not subscribed_indexes:
+    if len(doc_indexes) > 0 and not subscribed_indexes:
```
Bento007: good catch

mikebaumann: This is a valid use case - adding subscriptions after initial deployment yet before documents are indexed. This is something we wanted to do before but couldn't - and now can.

Issue #592, opened 2017-10-30
migrate s3 parallel copy to step functions


ttung:

kozbo:
Can this be merged yet?

PR #701, opened 2017-12-05
Simple method to invoke a step function.

ttung:
Invoking step functions is going to be a common idiom, so let's have a library function to do it.

Connects to #592


xbrianh:
lgtm

PR #704, opened 2017-12-05
Module to simplify the creation and dispatch of parallel branches in state machines

ttung:
This introduces a pre-processor for state machine definitions that manages bindings from state machine parallel states to a separate state machine, which is replicated as parameterized.  The states in the nested state machine are called with the branch id.

PR #707, opened 2017-12-05
Definition for a parallelized s3 copy state machine.

ttung:
Most of the code is lifted from `chained-aws-lambda`.

Connects to #592.
Depends on #704.

### DIFF
--- a/dss/stepfunctions/s3copyclient/implementation.py
+++ b/dss/stepfunctions/s3copyclient/implementation.py
@@ -1,0 +1,257 @@
+import binascii
+import collections
+import concurrent.futures as futures
+import copy
+
+import hashlib
+import threading
+import typing
+
+import boto3
+from cloud_blobstore.s3 import S3BlobStore
+
+from ...stepfunctions import generator
+from ...stepfunctions.lambdaexecutor import TimedThread
+from ...util.aws import get_s3_chunk_size
+
+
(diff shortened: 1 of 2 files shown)


dss/stepfunctions/s3copyclient/implementation.py
```diff
@@ -0,0 +1,257 @@
...
+    event[_Key.SIZE] = source_size
+    event[_Key.PART_SIZE] = part_size
+    event[_Key.PART_COUNT] = part_count
+    event[Key.FINISHED] = False
+
+    return event
+
+
+def copy_worker(event, lambda_context, branch_id):
+    class CopyWorkerTimedThread(TimedThread[dict]):
```
kislyuk: This class does not depend on the parent's args, consider bringing it into the top scope?

ttung: Is there precedence in Python for using a nested class/function for code organization?  The reason I did this was because I thought `CopyWorkerTimedThread` was never going to be used anywhere else except by `copy_worker`.

kislyuk: I guess that's fine, I think there is a negligible performance impact to defining the class at function runtime (unless it makes an I/O call while being defined). Also the fewer indents, the better. But I'm fine keeping it where it is.

PR #708, opened 2017-12-05
Definition for a parallelized s3 copy state machine that also writes file metadata.

ttung:
Most of the code is lifted from dss/events/chunkedtask/s3copyclient.py.

Connects to #592.
Depends on #704, #707.


kislyuk:
We may want to use a unit test to touch this code, or reconfigure codecov.

PR #709, opened 2017-12-05
S3 parallel copy state machine app

ttung:
Connects to #592.
Depends on #704, #707.

### DIFF
--- a/daemons/dss-s3-copy-sfn/app.py
+++ b/daemons/dss-s3-copy-sfn/app.py
@@ -1,0 +1,17 @@
+import os
+import sys
+
+import domovoi
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'domovoilib'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+import dss.stepfunctions.s3copyclient as s3copyclient
+import dss.stepfunctions.generator as generator
+
+app = domovoi.Domovoi()
+
+
+annotation_processor = generator.StateMachineAnnotationProcessor()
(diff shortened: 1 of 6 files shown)


iam/policy-templates/dss-s3-copy-sfn-lambda.json
```diff
@@ -0,0 +1,42 @@
...
+      "Action": [
+        "logs:CreateLogGroup",
+        "logs:CreateLogStream",
+        "logs:PutLogEvents"
+      ],
+      "Resource": "arn:aws:logs:*:*:*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": "s3:*",
```
kislyuk: This lambda shouldn't have delete access to the bucket. That's dangerous. Let's set the permissions on an as-needed basis.

daemons/dss-s3-copy-sfn/app.py
```diff
@@ -0,0 +1,16 @@
...
+sys.path.insert(0, pkg_root)  # noqa
+
+import dss.stepfunctions.s3copyclient as s3copyclient
+import dss.stepfunctions.generator as generator
+
+app = domovoi.Domovoi()
+
+
+transformer = generator.Transformer(app)
+transformer.transform(s3copyclient.sfn)
```
kislyuk: Any chance we could use more descriptive names than generator and transformer?

ttung: Done.

iam/policy-templates/dss-s3-copy-sfn-lambda.json
ttung: Done.

PR #710, opened 2017-12-05
S3 parallel copy + metadata write state machine app

ttung:
Dispatch logic to the s3 parallel copy + metadata write state machine.

Connects to #592.
Depends on #704, #707, #708, #709


ttung:
> This looks OK but why do these (#710 and #709) need to be separate lambdas/state machines? Why not have one copy SFN with an option determining whether or not to also copy metadata?

different permissions.  copy+write writes to the main bucket.  #709 is for the checkout service.

PR #711, opened 2017-12-05
Switch PUT /files to use the new state-machine based parallel s3 copy + metadata write

ttung:
This introduces a pre-processor for state machine definitions that includes:

1. Bindings from state machine states to python methods.
2. Bindings from state machine parallel states to a separate state machine, which is replicated as parameterized.

Connects to #592.
Depends on #704, #707, #708, #710

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -4,19 +4,19 @@ import json
 import re
 import typing
 from enum import Enum, auto
+from uuid import uuid4
 
 import iso8601
 import requests
-import chainedawslambda.aws
 from cloud_blobstore import BlobAlreadyExistsError, BlobNotFoundError, BlobStore
 from flask import jsonify, make_response, redirect, request
 from ..util.version import datetime_to_version_format
 
-from .. import DSSException, dss_handler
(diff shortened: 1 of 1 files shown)


dss/api/files.py
```diff
@@ -176,19 +176,26 @@ class CopyMode(Enum):
...
+            uuid, version,
+            file_metadata_json,
+        )
+        execution_input = json.dumps(state)
+        task_id = str(uuid4())
+
+        sfn = boto3.client('stepfunctions')
+        sts_client = boto3.client("sts")
+        region = boto3.Session().region_name
+        accountid = sts_client.get_caller_identity()['Account']
```
kislyuk: please use `dss.util.aws.ARN.get_region()` and `get_account_id()`

ttung:
Updated to use the API provided in #701 and refactored in #725

ttung:
> This LGTM but I can't quite grok what copy_write_metadata_sfn_event() does from the name alone. Consider renaming it to something more descriptive.

I can't think of a name that makes sense and is not absurdly long, so I just docblocked the function.

PR #718, opened 2017-12-06
[WIP] Indexer refactor

mweiden:
Small refactor to make the flow of the indexer code easier to understand and more modular for reuse in other indexing operations such as delete.

There are two changes here:
1. Change usage of the term `bundle_key` to `bundle_object_name`. This is more consistent with the terms used in the `BlobStore` interface, which uses `(bucket, object_name)` to identify stored entities. By comparison AWS uses the paring `(bucket_name, object_key)` to identify things in s3 while GCP uses the pairing `(bucket, name)`. Using the `BlobStore` terminology should reduce confusion.
2. Decouple the identification of the `index_name` from the creation of the document to be added to the search index. Other indexing operations, like deletions, will need to be performed on the index that do not require the document in the index to be created from scratch. This increases modularity and will reduce network calls.

### DIFF
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -26,9 +26,9 @@ DSS_BUNDLE_KEY_REGEX = r"^bundles/[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-
 def process_new_s3_indexable_object(event, logger) -> None:
     try:
         # This function is only called for S3 creation events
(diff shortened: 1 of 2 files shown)


mweiden:
The second change was not possible. Picking up the first somewhere else.

hannes-ucsc:
My preference would be to avoid field rename refactorings for the time being. There is going to be some contention on the code base if we're trying to tackle Admin Deletes at the same time as reindexing. Renames like that are going to make merges more difficult.

Let's create a ticket for this instead and solve later?

PR #719, opened 2017-12-06
[WIP] Run `make deploy` and smoketest during PR builds [skip ci]

hannes-ucsc:
This enables `make deploy` and `make smoketest` for PR builds in Travis CI. All builds hit the dev stage. To serialize access of CI builds to dev, mutexes are used.

The mutexes use DynamoDB conditional row updates and have a configurable expiration. They are implemented in a separate Python package at https://github.com/hannes-ucsc/lambda-mutex, a fork of https://github.com/chiradeep/lambda-mutex. I added a fix for a race condition when the utility is run for the first time in an AWS region, added blocking mode (with timeout) for mutex.lock() and added a CLI.

Reviewers should pay special attention to the Travis config.

The biggest problem with this is that master and PR builds still share the dev stage. A failed `make deploy` from a PR build could pollute that stage beyond repair, breaking subsequent builds. I think we should create a special sandbox environment for PR builds in order to decouple them from master.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,14 @@ requirements.txt requirements-dev.txt : %.txt : %.txt.in
 
 requirements-dev.txt : requirements.txt.in
 
+lock-stage:
+	# The lock on the mutex expires after 30 min
+	python -m dyndbmutex.cli $$DSS_DEPLOYMENT_STAGE lock --blocking --expiration 1800
+
+unlock-stage:
+	python -m dyndbmutex.cli $$DSS_DEPLOYMENT_STAGE release
+
 .PHONY: lint mypy
 .PHONY: test all_test integration_test standalone_test $(test_srcs) $(integration_test_srcs) $(standalone_test_srcs)
 .PHONY: deploy deploy-chalice deploy-daemons
+.PHONY lock-stage unlock-stage
--- a/Makefile
+++ b/Makefile
@@ -79,14 +79,14 @@ requirements.txt requirements-dev.txt : %.txt : %.txt.in
 
 requirements-dev.txt : requirements.txt.in
 
-lock-stage:
+lock_stage:
 	# The lock on the mutex expires after 30 min
 	python -m dyndbmutex.cli $$DSS_DEPLOYMENT_STAGE lock --blocking --expiration 1800
 
-unlock-stage:
+unlock_stage:
 	python -m dyndbmutex.cli $$DSS_DEPLOYMENT_STAGE release
 
 .PHONY: lint mypy
 .PHONY: test all_test integration_test standalone_test $(test_srcs) $(integration_test_srcs) $(standalone_test_srcs)
 .PHONY: deploy deploy-chalice deploy-daemons
-.PHONY lock-stage unlock-stage
+.PHONY: lock_stage unlock_stage
--- a/common.mk
+++ b/common.mk
@@ -19,7 +19,3 @@ endif
 ifeq ($(findstring Python 3.6, $(shell python --version 2>&1)),)
 $(error Please run make commands from a Python 3.6 virtualenv)
 endif
-
-ifneq ($(shell python -c 'import dyndbmutex' && echo yes), yes)
-$(error The dyndbmutex package is missing. Installing build requirements from requirements-dev.txt should fix that)
-endif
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -32,21 +32,21 @@ install:
...
+    script:
+    - make -j4 -O deploy
+    - set -eo pipefail
+    - make -j4 -O test
+    - make smoketest
     skip_cleanup: true
-    on:
-      branch: staging
+after_deploy:
+- make unlock-stage
```
ttung: Does this run even if the deploy fails?

.travis.yml
```diff
@@ -32,21 +32,21 @@ install:
 script:
 - set -eo pipefail
 - make -j4 -O test
-- if [[ $TRAVIS_BRANCH == staging ]]; then source environment.staging; fi
-- if [[ $TRAVIS_EVENT_TYPE == cron ]]; then make smoketest; fi
 after_success:
 - bash <(curl -s https://codecov.io/bash)
+before_deploy:
+- if [[ $TRAVIS_BRANCH == staging ]]; then source environment.staging; fi
+- make lock-stage
```
kislyuk: please use underscore separated make targets per convention

common.mk
```diff
@@ -19,3 +19,7 @@ endif
 ifeq ($(findstring Python 3.6, $(shell python --version)),)
 $(error Please run make commands from a Python 3.6 virtualenv)
 endif
+
+ifneq ($(shell python -c 'import dyndbmutex' && echo yes), yes)
+$(error The dyndbmutex package is missing. Installing build requirements from requirements-dev.txt should fix that)
+endif
```
kislyuk: please add newline

kislyuk:
I like the distributed dynamodb-based locking facility, it might be useful for other needs too. Let's either publish it on PyPI or bring it in as a submodule of this repo, or just vendor it in this repo.

.travis.yml
hannes-ucsc: This is not documented. I'll probably switch to a shell trap.

.travis.yml
hannes-ucsc: Will do.

common.mk
hannes-ucsc: Will do.

Makefile
```diff
@@ -62,4 +62,11 @@ requirements.txt requirements-dev.txt : %.txt : %.txt.in
...
 
-.PHONY: test lint mypy $(test_srcs) deploy deploy-chalice deploy-daemons
+lock-stage:
+	# The lock on the mutex expires after 30 min
+	python -m dyndbmutex.cli $$DSS_DEPLOYMENT_STAGE lock --blocking --expiration 1800
+
+unlock-stage:
+	python -m dyndbmutex.cli $$DSS_DEPLOYMENT_STAGE release
+
+.PHONY: test lint mypy $(test_srcs) deploy deploy-chalice deploy-daemons lock-stage
```
ttung: Also you want unlock-stage to be a phony target.

common.mk
ttung: Is this necessary?  Will people on the cmd line be running this?

hannes-ucsc: I'd say that if a recipe in the Makefile depends on it, the Makefile should check for it.

As a newcomer to the project I found these checks very helpful.

If the fix required installing something that would *only* be used by the `lock-stage` target, I'd say we should drop it. But since every developer has to install the build requirements, this is helpful.

ttung: But I thought `dyndbmutex` is only used by lock-stage...

hannes-ucsc: Correct, but the fix (installing the dev reqs) is needed by everyone.

Makefile
hannes-ucsc: Yup.

common.mk
hannes-ucsc: Maybe I should list the actual `pip` command line here?

ttung: I still don't get it.  We don't go and validate the presence of the vast majority of packages in requirements.txt in the Makefile.  There's nothing special about this requirement that makes it necessary to validate its presence.

Anyway, not a big deal, if you feel strongly about it, go ahead and do it.

hannes-ucsc: I'll remove it.

kbergin:
Bumping on open PRs that haven't been updated in a while. Anything to do here?

hannes-ucsc:
This is blocked on #856 for which I think two PRs have been opened and closed without a merge. #1120, the most recent one came pretty close to being merged and is in the process of being broken up into more digestible pieces.

PR #723, opened 2017-12-07
Extract BundleDocument class from indexer

hannes-ucsc:
I would like to introduce a little OO into index.py. The benefit of this refactoring is that it makes more obvious how data flows between methods when preparing a document for indexing. The method signatures are simplified because contextual arguments are converted to instance attributes, for example `logger` and `replica`.

This change increases the line count mostly because it adds another way of constructing a BundleDocument from a ES document JSON.

This is a pure refactoring with minimal functional changes.

Reviewers are encouraged to review the diff with a tool that highlights intra-line changes, e.g. PyCharm. The diff will become very small as I paid close attention to not move code around, to the extent that one function became a method of BundleDocument even though it should not be. I will move those methods, deal with the overly long lines and remove the other TODOs once we reach consensus on whether we want this type of refactoring.

### DIFF
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -225,10 +225,10 @@ class BundleDocument(dict):
                               json.dumps(self, indent=4))  # noqa
             initial_mappings = es_client.indices.get_mapping(index_name)[index_name]['mappings']
             es_client.index(index=index_name,
-                      doc_type=ESDocType.doc.name,
-                      id=self.bundle_id,
-                      # FIXME: (hannes) Can this be json.dumps(self, indent=4) ?
-                                                       body=json.dumps(self))  # Don't use refresh here.
+                            doc_type=ESDocType.doc.name,
+                            id=self.bundle_id,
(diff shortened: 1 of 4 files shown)


hannes-ucsc:
A good way to view the diff here on GitHub is to go to split view and append `&w=1` to the URL, e.g.

https://github.com/HumanCellAtlas/data-store/pull/723/files?diff=split&w=1

mikebaumann:
Nice! Thanks for the tip!

kislyuk:
LGTM, except for a formatting nitpick: I see you've added a bunch of noqa's for lines that are too long, but in most/all of those lines, it looks quite easy to wrap them to stay under 120 cols.

hannes-ucsc:
Thanks @kislyuk. For the initial review I wanted to keep the diff minimal. That's why I didn't wrap the lines and instead added noqa's instead.  Now that I've got tentative agreement from @mikebaumann. I will add commits to address the noqa's and method order issues.

hannes-ucsc:
@mikebaumann and I looked over the additional commits and he gave his verbal OK for this merge.

PR #726, opened 2017-12-08
More indexer refactorings

hannes-ucsc:
Starting to reap the benefits of the `BundleDocument` refactoring.

The reasoning for renaming `index shape identifier` to `shape descriptor` is that the former might be misinterpreted as the shape of an index, whereas the latter more accurately describes the shape of a document.

Reviewers should look at each commit separately.

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -198,14 +198,18 @@ class Config:
         return Config._ALLOWED_EMAILS
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


kislyuk:
This needs attention from @mikebaumann or @Bento007 but I just wanted to say good job on the OO refactor, the code looks much neater with it.

PR #727, opened 2017-12-08
Add the 'eventually' convenience method

mweiden:
Eventually is a convenience method for testing against concurrent services. It retries a a function with assertions on an interval until those assertions pass or a timeout is reached.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,5 +1,33 @@
 import datetime
+import functools
+import time
 
 
 def get_version():
     return datetime.datetime.utcnow().strftime("%Y-%m-%dT%H%M%S.%fZ")
+
+
+def eventually(timeout: float, interval: float, errors: set={AssertionError}):
+    """
+    @eventually runs a test until all assertions are satisfied or a timeout is reached.
+    :param timeout: time until the test fails
+    :param interval: time between attempts of the test
+    :param errors: the exceptions to catch and retry on
+    :return: the result of the function or a raised assertion error
+    """
+    def decorate(func):
+        @functools.wraps(func)
+        def call(*args, **kwargs):
+            timeout_time = time.time() + timeout
+            error_tuple = tuple(errors)
+            while True:
+                try:
+                    return func(*args, **kwargs)
+                except error_tuple:
+                    if time.time() >= timeout_time:
+                        raise
+                    time.sleep(interval)
+
+        return call
+
+    return decorate


tests/__init__.py
```diff
@@ -1,5 +1,35 @@
...
+
+
+def eventually(timeout: float, interval: float):
+    """
+    eventually is designed to run a test multiple times until a the assertion is satisfied or a timeout is reached.
+    :param timeout: time until the test fails
+    :param interval: time between attempts of the test
+    :return: the result of the function or a raised assertion error
+    """
+    def decorate(func):
```
ttung: Switch to:

```
@functools.wraps(func)
def call(*args, **kwargs):
```

tests/__init__.py
```diff
@@ -1,5 +1,35 @@
...
+        def call(*args, **kwargs):
+            timeout_time = time.time() + timeout
+            result = None
```
ttung: consider:

```
while True:
    try:
        return func(*args, **kwargs)
    except AssertionError:
        if time.time() >= timeout_time:
            raise
        time.sleep(interval)
```

mweiden:
@ttung changes made 👍

tests/__init__.py
```diff
@@ -1,5 +1,29 @@
...
+import functools
 
 
 def get_version():
     return datetime.datetime.utcnow().strftime("%Y-%m-%dT%H%M%S.%fZ")
+
+
+def eventually(timeout: float, interval: float):
+    """
+    eventually is designed to run a test multiple times until a the assertion is satisfied or a timeout is reached.
```
hannes-ucsc: s/a the/the

hannes-ucsc: Or actually `until all assertions are satisfied`

kislyuk:
Consider making the caught exceptions configurable (by default, `set([AssertionError])`).

tests/__init__.py
mweiden: This change has been made.

tests/__init__.py
mweiden: Good catch. Made this change too.

mweiden:
@kislyuk Good suggestion, I've made this change too.

tests/__init__.py
```diff
@@ -1,5 +1,33 @@
 import datetime
+import time
```
ttung: sort, pls.

tests/__init__.py
```diff
@@ -1,5 +1,33 @@
...
+
+def eventually(timeout: float, interval: float, errors: set={AssertionError}):
+    """
+    @eventually runs a test until all assertions are satisfied or a timeout is reached.
+    :param timeout: time until the test fails
+    :param interval: time between attempts of the test
+    :param errors: the exceptions to catch and retry on
+    :return: the result of the function or a raised assertion error
+    """
+    def decorate(func):
```
ttung: I think you can just remove L18, and `eventually(..)` can return `call`.

ttung: nm, you're right.

PR #714, opened 2017-12-06
A few extra step function utility methods.

xbrianh:
Utility methods needed for visitation batch processor, may
be useful for other step function consumers.

connects to #539

### Test plan
Tested locally through visitation batch processor integration tests. Will be tested through smoketest in the future.

PR #728, opened 2017-12-08
A few extra step function utility methods.

xbrianh:
Utility methods needed for visitation batch processor, may
be useful for other step function consumers.

connects to #539

Continued from #714 which was merged in error.


hannes-ucsc:
These were the same changes as in #714, plus a minor fix for an bogus mypy complaint.

PR #729, opened 2017-12-09
Fix stack overflow when debugging test_indexer.py with PyCharm (alternative 1/3)

hannes-ucsc:
… or any other IDE that uses pydevd.

This change use a more static approach for reexporting the `requests.api` methods on `ThreadedLocalServer`. With the original `__getattr__`-based approach I get

```
Fatal Python error: Cannot recover from stack overflow.
```

and the process crashes with SIGABRT.

The currently recommended workaround of enabling Genvent-compatible debugging avoids the SO but trashed stderr with hundreds of these:

```
  File "/Users/hannes/workspace/hca/data-store/.venv/lib/python3.6/site-packages/werkzeug/urls.py", line 425, in <genexpr>
    if not rest or any(c not in s('0123456789') for c in rest):
SystemError: error return without exception set
Exception ignored in: <generator object url_parse.<locals>.<genexpr> at 0x105169d58>
```

### DIFF
--- a/tests/infra/server.py
+++ b/tests/infra/server.py
@@ -58,18 +58,17 @@ class ThreadedLocalServer(threading.Thread):
             timeout=(1.0, 30.0),
             **kwargs)
 
-    def __getattr__(self, name):
-        """
-        requests.api is a module consisting of all the HTTP request types, defined as methods.  If we're being called
-        with one of these types, then execute the call against the running server.
-        """
-        func = getattr(requests.api, name, None)
-        if func is not None and isinstance(func, types.FunctionType):
-            def result(path, **kwargs):
-                return self._make_call(func, path, **kwargs)
-            return result
+    # requests.api is a module consisting of all the HTTP request types, defined as methods.  If we're being called
+    # with one of these types, then execute the call against the running server.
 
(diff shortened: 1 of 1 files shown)


kislyuk:
I'm not a fan of `exec()`. It doesn't seem necessary here - you seem to be just dynamically declaring methods on a class, right? Can't you just use:
```
def callable(obj, ...):
    ...
setattr(self.__class__, method_name, types.MethodType(callable, self.__class__))
```

hannes-ucsc:
@kislyuk, correct, the code dynamically injects new methods. Because this happens at class definition time (shortly before, actually), there is zero run time overhead (unlike the `__getattr__` approach). Unfortunately, this also means that that there is no `self` or `cls` available. In fact, the entire class has not yet been defined.

If exec() is objectionable (and I don't necessarily agree that it is) this could alternatively be done with a meta class or by injecting the properties after the class is fully defined. I could also see if the `def x(): …` to x = lambda: ...` equivalence helps here.

hannes-ucsc:
See #730 and #731 for alternative approaches.

kislyuk:
Thanks, I prefer #731.

hannes-ucsc:
Closing as per https://github.com/HumanCellAtlas/data-store/pull/729#issuecomment-350767297.

PR #730, opened 2017-12-11
Fix stack overflow when debugging test_indexer.py with PyCharm (alternative 2/3)

hannes-ucsc:
… or any other IDE that uses pydevd.

This change use a more static approach for reexporting the `requests.api` methods on `ThreadedLocalServer`. With the original `__getattr__`-based approach I get

```
Fatal Python error: Cannot recover from stack overflow.
```

and the process crashes with SIGABRT.

The currently recommended workaround of enabling Genvent-compatible debugging avoids the SO but trashed stderr with hundreds of these:

```
  File "/Users/hannes/workspace/hca/data-store/.venv/lib/python3.6/site-packages/werkzeug/urls.py", line 425, in <genexpr>
    if not rest or any(c not in s('0123456789') for c in rest):
SystemError: error return without exception set
Exception ignored in: <generator object url_parse.<locals>.<genexpr> at 0x105169d58>
```

### DIFF
--- a/tests/infra/server.py
+++ b/tests/infra/server.py
@@ -51,25 +51,21 @@ class ThreadedLocalServer(threading.Thread):
         self._server_ready.set()
         self._server.server.serve_forever()
 
-    def _make_call(self, method, path, **kwargs):
-        return method(
-            f"http://127.0.0.1:{self._port}{path}",
-            allow_redirects=False,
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Closing as per https://github.com/HumanCellAtlas/data-store/pull/729#issuecomment-350767297.

Issue #614, opened 2017-11-06
Reindexing: Develop parallel execution infrastructure for reindexing


mikebaumann:
see epic #539

xbrianh:
Finished as thread pool pattern.

PR #735, opened 2017-12-11
Visitation batch processor app core

xbrianh:
Connects to #614 #539 

### Test plan
Unit and integration tests.


dss/stepfunctions/visitation/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+
+def walker_prefix(walker_execution_name: str) -> str:
+    encoded_prefix, sentinel_execution_name = walker_execution_name.split('----')
+    prefix = encoded_prefix.replace('____', '/')
+    return prefix
+
+
+class Timeout:
```
kislyuk: Consider using signals instead. They avoid shutdown issues, and you will still be called if a thread is hogging the GIL.

```
def alarm_handler(signum, frame):
        raise MyTimeout("...")

    signal.signal(signal.SIGALRM, alarm_handler)
    timeout_seconds = (context.get_remaining_time_in_millis() / 1000) - 10
    context.log("Setting timeout to {}".format(timeout_seconds))
    signal.alarm(timeout_seconds)

    try:
        ... do work ...
    except MyTimeout:
        ... shutdown ...
```

dss/stepfunctions/visitation/utils.py
```diff
@@ -0,0 +1,86 @@
...
+        args: typing.List[typing.Any],
+        kwargs: typing.Dict[str, typing.Any]
+    ) -> None:
+        self.thread = threading.Thread(target=target, daemon=True, args=args, kwargs=kwargs)
+        self.thread.start()
+        self.thread.join(self.timeout_seconds)
+
+    def __call__(self, func: typing.Callable[..., typing.Any]) -> typing.Any:
+        def wrapped(*args, **kwargs):
+            self.func = func
```
hannes-ucsc: You could pass `func` to `run` via `start`. `func`, `args` and `kwargs` deserve to be treated uniformly because together they represent a single function invocation.

requirements-dev.txt
```diff
@@ -16,7 +16,7 @@ chalice==1.0.4
 chardet==3.0.4
 click==6.7
 clickclick==1.2.2
-cloud-blobstore==0.0.7
+cloud-blobstore==0.0.8
```
hannes-ucsc: master has this at 1.0.0

dss/stepfunctions/visitation/__init__.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+
+        self.code = StatusCode.RUNNING.name
+
+        for key in blobs:
+            self.process_item(key)
+            self.k_processed += 1
+            self.marker = blobs.start_after_key
+            self.token = blobs.token
+
```
hannes-ucsc: These two could be unified into a single opaque token. The fact that you distinguish between both is an implementation detail.

dss/stepfunctions/visitation/__init__.py
```diff
@@ -0,0 +1,295 @@
...
+        return name
+
+    def sentinel_muster(self) -> typing.List[str]:
+        """
+        Start walkers as needed based on pool of prefixes that are waiting to be executed. When all walkers are
+        finished, sets state function status code to SUCCEEDED
+        """
+
+        if self.waiting is None:  # type: ignore
```
hannes-ucsc: What are these type annotation for?

dss/stepfunctions/visitation/utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def start(
+        self,
+        target: typing.Callable[..., typing.Any],
+        args: typing.List[typing.Any],
+        kwargs: typing.Dict[str, typing.Any]
+    ) -> None:
+        self.thread = threading.Thread(target=target, daemon=True, args=args, kwargs=kwargs)
+        self.thread.start()
+        self.thread.join(self.timeout_seconds)
```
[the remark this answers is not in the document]
xbrianh: This was adapted from @ttung's stuff, but now I understand better why he did what he did. I'll move to either @kislyuk's suggestion, or simple timing.

dss/stepfunctions/visitation/utils.py
xbrianh: Thanks, I'll use this, or check for execution time within the loop.

dss/stepfunctions/visitation/__init__.py
xbrianh: Will think about it... this would require a change to cloud-blobstore.

dss/stepfunctions/visitation/__init__.py
xbrianh: I don't understand how to keep mypy happy without them.

requirements-dev.txt
xbrianh: How can I keep up?

dss/stepfunctions/visitation/__init__.py
hannes-ucsc: I'd suggest doing it in a separate PR, if you decide to do it.

dss/stepfunctions/visitation/__init__.py
hannes-ucsc: Interesting. Let me try.

hannes-ucsc:
Superseded by #792.

PR #741, opened 2017-12-11
annotate tests with standalone vs integration

ttung:
Targets:

tests/test_xyz.py: this will run both the standalone and integration tests in tests/test_xyz.py
integration__tests/test_xyz.py: this will run the integration test in tests/test_xyz.py
standalone__tests/test_xyz.py: this will run the standalone test in tests/test_xyz.py
test: this will run both the standalone and integration tests in tests/*
integration_test: this will run the integration tests in tests/*
standalone_test: this will run the standalone tests in tests/*

The environment DSS_TEST_MODE selects which type of tests to run.  If the word "integration" is in the env var, then it runs the integration tests.  If the word "standalone" is in the env var, then it runs the standalone tests.   DSS_TEST_MODE _can_ contain both words, in which case, both sets of tests are run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,13 +8,29 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test_srcs := $(wildcard tests/test_*.py)
+standalone_test_srcs := $(addprefix standalone__, $(test_srcs))
+integration_test_srcs := $(addprefix integration__, $(test_srcs))
+
+standalone_test: lint mypy $(standalone_test_srcs)
+	coverage combine
+	rm -f .coverage.*
+
+$(standalone_test_srcs): standalone__%.py :
+	DSS_TEST_MODE=standalone coverage run -p --source=dss -m unittest $*.py
+
+integration_test: lint mypy $(integration_test_srcs)
+	coverage combine
+	rm -f .coverage.*
+
+$(integration_test_srcs): integration__%.py :
+	DSS_TEST_MODE=integration coverage run -p --source=dss -m unittest $*.py
 
 test: lint mypy $(test_srcs)
 	coverage combine
 	rm -f .coverage.*
 
 $(test_srcs): %.py :
-	coverage run -p --source=dss -m unittest $@
+	DSS_TEST_MODE="integration standalone" coverage run -p --source=dss -m unittest $@
 
 smoketest:
 	tests/smoketest.py
@@ -63,4 +79,6 @@ requirements.txt requirements-dev.txt : %.txt : %.txt.in
 
 requirements-dev.txt : requirements.txt.in
 
-.PHONY: test lint mypy $(test_srcs) deploy deploy-chalice deploy-daemons
+.PHONY: lint mypy
+.PHONY: test all_test integration_test standalone_test $(test_srcs) $(integration_test_srcs) $(standalone_test_srcs)
+.PHONY: deploy deploy-chalice deploy-daemons
--- a/tests/infra/testmode.py
+++ b/tests/infra/testmode.py
@@ -1,0 +1,14 @@
+import os
+import unittest
+
+
+def standalone(f):
+    return unittest.skipUnless(
+        os.environ.get('DSS_TEST_MODE', "standalone").find("standalone") != -1,
+        "Skipping standalone test")(f)
+
+
+def integration(f):
+    return unittest.skipUnless(
+        os.environ.get('DSS_TEST_MODE', "integration").find("integration") != -1,
+        "Skipping integration test")(f)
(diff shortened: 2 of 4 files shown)


tests/infra/testmode.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def standalone(f):
+    return unittest.skipIf(
+        ("INTEGRATION_TESTS" in os.environ and os.environ['INTEGRATION_TESTS'] != "0"),
+        "Only running integration tests")(f)
+
+
+def integration(f):
+    return unittest.skipUnless("INTEGRATION_TESTS" in os.environ, "Only running integration tests")(f)
```
hannes-ucsc: The double negation is making my head explode. But if I am correct, this will run integration tests if `INTEGRATION_TESTS` is `0`. Is that what you want?

s/integration/standalone in message?

Makefile
```diff
@@ -8,13 +8,30 @@ mypy:
...
-	coverage run -p --source=dss -m unittest $@
+all_test: lint mypy $(standalone_test_srcs) $(integration_test_srcs)
+	coverage combine
+	rm -f .coverage.*
+
+$(standalone_test_srcs): standalone__%.py : %.py
+	INTEGRATION_TESTS=0 coverage run -p --source=dss -m unittest $<
+
+$(standalone_test_srcs): standalone__%.py : %.py
+	INTEGRATION_TESTS=0 coverage run -p --source=dss -m unittest $<
```
hannes-ucsc: Looks like a dupe.

tests/infra/testmode.py
```diff
@@ -0,0 +1,12 @@
+import os
+import unittest
+
+
+def standalone(f):
+    return unittest.skipIf(
```
ttung: This is run if ^(A && B) == ^A || ^B

I think.  Demorgan's Law is so long long ago.

That means we run standalone if `INTEGRATION_TESTS not in os.environ or os.environ['INTEGRATION_TESTS'] == "0"`

tests/infra/testmode.py
ttung: skipUnless = runWhen

ttung:
Fixed dupe and copy-pasta on skip message.

tests/infra/testmode.py
hannes-ucsc: A test marked as `@standalone` will be skipped if condition is true.
<=> A test marked as `@standalone` will be run if condition is false.
<=> A test marked as `@standalone` will be run if `not (IT in env and $IT != "0")`
<=> A test marked as `@standalone` will be run if `IT not in env or $IT == "0"`

That's correct. 

If a test marked as `@standalone` is skipped, the message "Only running standalone tests" is printed. I don't think that's correct.

tests/infra/testmode.py
```diff
@@ -0,0 +1,12 @@
...
+        ("INTEGRATION_TESTS" in os.environ and os.environ['INTEGRATION_TESTS'] != "0"),
+        "Only running standalone tests")(f)
+
+
+def integration(f):
+    return unittest.skipUnless("INTEGRATION_TESTS" in os.environ, "Only running integration tests")(f)
```
hannes-ucsc: A test marked as `@integration` will be skipped unless condition is true.
<=> A test marked as `@integration` will be skipped if condition is not true.
<=> A test marked as `@integration` will be skipped if condition is false.
<=> A test marked as `@integration` will be run if condition is true.
<=> A test marked as `@integration` will be run if `IT in env`
=> A test marked as `@integration` will be run if `$IT=0`

That's not correct.

If a test marked as `@integration` is skipped, the message "Only running integration tests" is printed. I don't think that's correct either.

tests/infra/testmode.py
```diff
@@ -0,0 +1,12 @@
+import os
+import unittest
+
+
+def standalone(f):
+    return unittest.skipIf(
+        ("INTEGRATION_TESTS" in os.environ and os.environ['INTEGRATION_TESTS'] != "0"),
```
hannes-ucsc: A test marked as `@standalone` will be skipped if condition is true.
<=> A test marked as `@standalone` will be run if condition is false.
<=> A test marked as `@standalone` will be run if `not (IT in env and $IT != "0")`
<=> A test marked as `@standalone` will be run if `IT not in env or $IT == "0"`

That's correct. 

If a test marked as `@standalone` is skipped, the message "Only running standalone tests" is printed. I don't think that's correct.

tests/infra/testmode.py
ttung: Ah, I get your concern.  I totally misread your initial comment. :)
Updated.

tests/infra/testmode.py
```diff
@@ -0,0 +1,14 @@
+import os
+import unittest
+
+
+def standalone(f):
```
hannes-ucsc:
```python
for mode in ('standalone', 'integration'):
    globals()[mode] = lambda f: unittest.skipUnless(
        mode not in os.environ.get('DSS_TEST_MODE', mode),
        f"Skipping {mode} test")(f)
```

hannes-ucsc: Of course I am kidding. But am I?

PR #742, opened 2017-12-12
Make a standard version (date) formatting function

mweiden:


mweiden:
@ttung np, thanks to you 🙇

PR #744, opened 2017-12-12
Use enums uniformly for specifying replicas

mweiden:

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -84,18 +84,17 @@ class Config:
         Config._CURRENT_CONFIG = config
 
     @staticmethod
-    def get_cloud_specific_handles(replica: str) -> typing.Tuple[
-            BlobStore, HCABlobStore, str]:
+    def get_cloud_specific_handles(replica: Replica) -> typing.Tuple[BlobStore, HCABlobStore, str]:
 
         handle: BlobStore
-        if replica == 'aws':
+        if replica == Replica.aws:
             handle = S3BlobStore()
             return (
                 handle,
                 S3HCABlobStore(handle),
                 Config.get_s3_bucket()
             )
-        elif replica == 'gcp':
+        elif replica == Replica.gcp:
             credentials = os.environ["GOOGLE_APPLICATION_CREDENTIALS"]
             handle = GSBlobStore(credentials)
             return (
@@ -103,15 +102,16 @@ class Config:
                 GSHCABlobStore(handle),
                 Config.get_gs_bucket()
             )
-        raise ValueError("I don't understand this replica!")
+        raise NotImplementedError(f"Replica `{replica.name}` is not implemented!")
 
     @staticmethod
-    def get_storage_schema(replica: str) -> str:
-        if replica == 'aws':
+    def get_storage_schema(replica: Replica) -> str:
+
+        if replica == Replica.aws:
             return "s3"
-        elif replica == 'gcp':
+        elif replica == Replica.gcp:
             return "gs"
-        raise ValueError("I don't understand this replica!")
+        raise NotImplementedError(f"Replica `{replica.name}` is not implemented!")
 
     @staticmethod
     def get_s3_bucket() -> str:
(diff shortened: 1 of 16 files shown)


hannes-ucsc:
Wow, you are fearless. I started with this a couple of days ago and quickly gave up, restricting myself to index.py.

mweiden:
@hannes-ucsc we'll see. cross your fingers

ttung:
💯 

(once your tests pass :) )

mweiden:
@ttung 🌲 ✅ 🐢 🍏 📗 💚 🥗

dss/config.py
```diff
@@ -92,34 +92,40 @@ def set_config(config: BucketConfig):
...
-        raise ValueError("I don't understand this replica!")
+        raise NotImplementedError(f"Replica `{replica.name}` is not implemented!")
 
     @staticmethod
-    def get_storage_schema(replica: str) -> str:
-        if replica == 'aws':
+    def get_storage_schema(replica: Replica) -> str:
+
+        if type(replica) == str:
+            return Config.get_storage_schema(Replica[str(replica)])  # noqa
```
xbrianh: Why the cast `str(replica)`?

dss/config.py
```diff
@@ -92,34 +92,40 @@ def set_config(config: BucketConfig):
...
                 GSHCABlobStore(handle),
                 Config.get_gs_bucket()
             )
-        raise ValueError("I don't understand this replica!")
+        raise NotImplementedError(f"Replica `{replica.name}` is not implemented!")
 
     @staticmethod
-    def get_storage_schema(replica: str) -> str:
-        if replica == 'aws':
+    def get_storage_schema(replica: Replica) -> str:
```
hannes-ucsc: `replica: typing.Union[Replica,str]`

dss/config.py
mweiden: @xbrianh otherwise mypy complains. It wasn't picking up the `# noqa` comment. I'll try again.

dss/config.py
mweiden: @hannes-ucsc good catch

dss/config.py
ttung: I'd consider doing `instanceof` because you don't require exact type matching.  Generally, `instanceof` is preferable to inspecting the actual type.

dss/config.py
```diff
@@ -92,34 +92,40 @@ def set_config(config: BucketConfig):
         Config._CURRENT_CONFIG = config
 
     @staticmethod
-    def get_cloud_specific_handles(replica: str) -> typing.Tuple[
-            BlobStore, HCABlobStore, str]:
+    def get_cloud_specific_handles(replica: Replica) -> typing.Tuple[BlobStore, HCABlobStore, str]:
```
ttung: should do the union trick here as well.

tests/test_bundle.py
```diff
@@ -2,25 +2,25 @@
...
 import sys
-import time
 import threading
+import time
 import typing
 import unittest
 import urllib.parse
 import uuid
 
+import hashlib
```
ttung: hashlib, io, and os are all stdlib.

tests/test_checkout.py
```diff
@@ -1,21 +1,21 @@
 #!/usr/bin/env python
 # coding: utf-8
 
-import os
 import sys
 import unittest
 
+import os
```
ttung: os is stdlib.

tests/test_file.py
```diff
@@ -1,23 +1,23 @@
...
 import datetime
-import hashlib
-import os
 import sys
-import tempfile
+import typing
 import unittest
 import uuid
 
+import hashlib
```
ttung: same as above.

dss/config.py
mweiden: The `mypy` check was complaining and not picking up the `noqa` tag. I am going to try to remove this statement. It shouldn't be there in the first place.

dss/config.py
mweiden: I'm rolling that back. I think we should force people to use `Replica` or fail fast. This will encourage the usage we want to see.

PR #745, opened 2017-12-12
DSS-721: Standardize terminology

mweiden:
##  Rename `bundle_id` to `bundle_fqid`

The term 'bundle_id' has popped up in the repository, but it is not
always clear what it means.

Here I introduce the term Bundle Fully Qualified Identifier, which is
the combination of the bundle UUID and version string. It is analogous
to FQDN in that it specifies a bundle version exactly and fully.

##  Rename `*_object_name` to `*_key`

The terms 'bundle_key', 'key', and 'file_key' have popped up in the
repository alongside terms like `*_object_name`.

This PR attempts to standardize the terminology.

### DIFF
--- a/dss/util/checkout.py
+++ b/dss/util/checkout.py
@@ -12,7 +12,6 @@ from .aws import get_s3_chunk_size
 from .bundles import get_bundle, get_bundle_from_bucket
 from .. import DSSException, Config
 
-
 log = getLogger()
 blobstore = S3BlobStore()
 
--- a/dss/util/checkout.py
+++ b/dss/util/checkout.py
@@ -48,7 +48,7 @@ def parallel_copy(source_bucket: str, source_key: str, destination_bucket: str, 
     aws.schedule_task(S3ParallelCopySupervisorTask, initial_state)
 
 
-def get_src_object_name(file_metadata: dict):
+def get_src_key(file_metadata: dict):
     return "blobs/" + ".".join((
         file_metadata[BundleFileMeta.SHA256],
         file_metadata[BundleFileMeta.SHA1],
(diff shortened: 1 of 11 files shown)


dss/util/checkout.py
```diff
@@ -53,55 +55,64 @@ def get_src_object_name(file_metadata: dict):
...
-def validate_bundle_exists(replica: str, bucket: str, bundle_id: str, version: str):
+
+def validate_bundle_exists(replica: str, bucket: str, bundle_fqid: str, version: str):
     try:
-        get_bundle_from_bucket(bundle_id, replica, version, bucket)
+        get_bundle_from_bucket(bundle_fqid, replica, version, bucket)
         return ValidationEnum.PASSED, None
     except DSSException:
-        return ValidationEnum.WRONG_BUNDLE_KEY, "Bundle with specified key does not exist"
+        return ValidationEnum.WRONG_BUNDLE_OBJECT_NAME, "Bundle with specified key does not exist"
```
hannes-ucsc: s/key/object name

dss/util/checkout.py
```diff
@@ -112,10 +123,10 @@ def touch_test_file(dst_bucket, replica) -> bool:
 
     try:
         handle.upload_file_handle(
-            dst_bucket,
+            destination_bucket,
```
hannes-ucsc: Why? `dst` and `src` common abbreviations.

hannes-ucsc:
https://github.com/chanzuckerberg/cloud-blobstore/pull/10 was merged but we need a release so we can include it here.

hannes-ucsc:
@mweiden do you want me to file a PR that upgrades this project to the new release of cloud-blobstore or do you want to do the upgrade here?

PR #746, opened 2017-12-12
Mark standalone tests

ttung:
These are the tests I've written or am somewhat responsible for.


hannes-ucsc:
Your decorator can be used at the class-level, too, correct? If so, do you want to use that or do you think it's better to explicitly decorate every method?

ttung:
I don't know if the test classes can be annotated.

PR #750, opened 2017-12-13
Move IndexDocument and its subclasses to their own file

mweiden:
`dss/events/handlers/index.py` was growing quite large, having both classes that handle both storage events and index documents. This pull request moves IndexDocument and it's child classes to their own module.


tests/test_indexer.py
```diff
@@ -12,10 +14,6 @@
 import uuid
 from http.server import BaseHTTPRequestHandler, HTTPServer
 
-import google.auth
```
hannes-ucsc: Nit: I'm totally fine with sneaking in such small fixes into an unrelated PR as long as they are in a separate commit AND the PR lands unsquashed.

dss/storage/index_document.py
```diff
@@ -0,0 +1,350 @@
...
+            self.logger.info(f"Successfully notified for subscription {subscription_id}"
+                             f" for bundle {self.bundle_id} with transaction id {transaction_id} "
+                             f"Code: {response.status_code}")
+        else:
+            self.logger.warning(f"Failed notification for subscription {subscription_id}"
+                                f" for bundle {self.bundle_id} with transaction id {transaction_id} "
+                                f"Code: {response.status_code}")
+
+
+class BundleTombstoneDocument(IndexDocument):
```
hannes-ucsc: As much as I would like to merge this quickly, but this should not be introduced in this PR.

dss/storage/index_document.py
```diff
@@ -0,0 +1,350 @@
...
+        blobstore, _, bucket_name = Config.get_cloud_specific_handles(replica.name)
+        bundle_uuid, bundle_version = DSS_OBJECT_NAME_REGEX.search(key).groups()
+
+        tombstone_data = json.loads(blobstore.get(bucket_name, key))
+
+        doc = cls(replica, bundle_uuid, bundle_version, logger)
+        doc.update(tombstone_data)
+        return doc
+
+    def list_dead_bundles(self):
```
hannes-ucsc: Love this name.

dss/storage/index_document.py
hannes-ucsc: My mistake.

mweiden:
@hannes-ucsc ty hannes!

PR #753, opened 2017-12-13
Daemon to kick off integration tests.

ttung:
1. Added configuration in .travis.yml to run standalone tests normally.
2. Added configuration in .travis.yml to run integration tests when TRAVIS_DSS_INTEGRATION_MODE is non-empty.
3. Added a daemon that kicks off integration tests every hour.

### DIFF
--- a/daemons/integration-test/app.py
+++ b/daemons/integration-test/app.py
@@ -1,0 +1,34 @@
+import os
+
+import domovoi
+import requests
+
+
+app = domovoi.Domovoi()
+
+
+@app.scheduled_function("rate(60 minutes)")
+def integration_test(*args, **kwargs):
+    travis_token = os.environ["TRAVIS_TOKEN"]
+    body = {
+        'request': {
+            'branch': "master",
+            'message': "INTEGRATION TESTS",
+            'config': {
+                'merge_mode': "deep_merge",
+                'env': {
+                    'matrix': ["TRAVIS_DSS_INTEGRATION_MODE=1"]
+                },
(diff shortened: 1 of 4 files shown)


daemons/integration-test/app.py
```diff
@@ -0,0 +1,35 @@
...
+            'message': "INTEGRATION TESTS",
+            'config': {
+                'merge_mode': "deep_merge",
+                'env': {
+                    'matrix': ["TRAVIS_DSS_INTEGRATION_MODE=1"]
+                },
+            },
+        }}
+    headers = {
+        'Content-Type': "application/json",
```
kislyuk: this header comes for free with `request(json=...)`

ttung: Removed.

daemons/integration-test/app.py
```diff
@@ -0,0 +1,35 @@
...
+
+import domovoi
+import requests
+
+
+app = domovoi.Domovoi()
+
+
+@app.scheduled_function("rate(60 minutes)")
+def integration_test(*_):
```
kislyuk: nitpick: prefer a more proper `method(*args, **kwargs)`

ttung: How about `method(*_, **_)`? :)

ttung: I switched it to `*args, **kwargs` for now though.

ttung:
Highest frequency for Travis cron is "daily"

PR #758, opened 2017-12-14
ObjectIdentifier and child classes manage storage strings

mweiden:

### DIFF
--- a/dss/events/handlers/index.py
+++ b/dss/events/handlers/index.py
@@ -1,11 +1,10 @@
 """Lambda function for DSS indexing"""
 import json
-import re
 from urllib.parse import unquote
 
 from dss import Replica, Config
 from dss.storage.index_document import BundleDocument, BundleTombstoneDocument
-from ...storage.bundles import DSS_BUNDLE_TOMBSTONE_REGEX, DSS_BUNDLE_KEY_REGEX
+from ...storage.bundles import ObjectIdentifier, BundleFQID, TombstoneID
 
 
 class IndexHandler:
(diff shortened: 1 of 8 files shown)


mweiden:
Note: if it weren't for the tests added here, this PR would reduce application LOC.

dss/events/handlers/index.py
```diff
@@ -16,48 +15,34 @@ def process_new_indexable_object(cls, event, logger) -> None:
 
     @classmethod
     def _process_new_indexable_object(cls, replica: Replica, key: str, logger):
-        if cls._is_bundle_to_index(key):
-            cls._index_and_notify(replica, key, logger)
-        elif cls._is_deletion(key):
-            cls._delete_from_index(replica, key, logger)
+        identifier = ObjectIdentifier.from_key(key)
+        if type(identifier) == BundleFQID:
```
hannes-ucsc: `isinstance(FQID)` would allow for polymorphism and is a more commonly used type check in Python. With non-distinct subtypes I would order the if branches accordingly from more specific to less specific. Just like the `except:` clause works in Python.

dss/storage/bundles.py
```diff
@@ -22,44 +22,66 @@
...
-    match = DSS_OBJECT_NAME_REGEX.match(bundle_key)
-    uuid, version = match.groups() if match else (None, None)
-    if uuid and version:
-        return format_bundle_fqid(uuid, version)
-    else:
-        raise ValueError(f"Object name does not contain a valid bundle identifier: {bundle_key}")
+class ObjectIdentifier(namedtuple('ObjectIdentifier', 'uuid version')):
 
+    @classmethod
+    def from_key(cls, bundle_key: str):
```
hannes-ucsc: If this a generic `ObjectIdentifier` then `bundle_key` can be any key, not just a bundle key correct?

dss/storage/bundles.py
```diff
@@ -22,44 +22,66 @@
...
+        return f"{self.prefix}/{self}"
 
-def format_bundle_fqid(uuid: str, version: str) -> str:
-    if UUID_REGEX.match(uuid) and VERSION_REGEX.match(version):
-        return f"{uuid}.{version}"
-    else:
-        raise ValueError(f"Not a valid version regex-version pair: {uuid}, {version}")
+    @property
+    def prefix(self):
+        return NotImplementedError("'prefix' not implemented!")
```
hannes-ucsc: https://docs.python.org/3/library/abc.html#abc.abstractproperty

May be overkill

dss/storage/index_document.py
```diff
@@ -15,26 +15,20 @@
...
         super().__init__()
         self.logger = logger
         self.replica = replica
-        self.bundle_uuid = bundle_uuid
-        self.bundle_version = bundle_version
-
-    @property
-    def bundle_fqid(self):
-        return format_bundle_fqid(self.bundle_uuid, self.bundle_version)
+        self.identifier = identifier
```
hannes-ucsc: For the sake of brevity and recognizability, I would prefer this to be called fqid and rename ObjectIdentifier to be called FQID. I would gloss over the fact that a version-unspecific TombstoneID is not fully-qualified.

dss/storage/bundles.py
```diff
@@ -22,44 +22,66 @@
...
-    if uuid and version:
-        return format_bundle_fqid(uuid, version)
-    else:
-        raise ValueError(f"Object name does not contain a valid bundle identifier: {bundle_key}")
+class ObjectIdentifier(namedtuple('ObjectIdentifier', 'uuid version')):
 
+    @classmethod
+    def from_key(cls, bundle_key: str):
+        match = DSS_OBJECT_NAME_REGEX.match(bundle_key)
+        object_type, uuid, version, tombstone_suffix = match.groups() if match else (None, None, None, None)
```
hannes-ucsc: We should, consider—in a different refactoring—to move from `str` to `uuid` for the type of the `uuid` member. UUIDs are not hexadecimal strings by nature. The hex str representation is ambiguous wrt to case. I've seen crazy bugs cause by that.

tests/__init__.py
```diff
@@ -29,8 +30,12 @@
 }
 
 
-def get_bundle_fqid():
-    return f"{uuid.uuid4()}.{get_version()}"
+def get_bundle_fqid() -> BundleFQID:
+    return BundleFQID(uuid=uuid.uuid4(), version=get_version())
```
hannes-ucsc: This looks odd. Don't you wwant `str(uuid.uuid4())` ?

tests/__init__.py
```diff
@@ -29,8 +30,12 @@
...
+
+
+def get_file_fqid() -> FileFQID:
+    return FileFQID(uuid=uuid.uuid4(), version=get_version())
```
hannes-ucsc: Ditto.

dss/storage/bundles.py
mweiden: 💯

dss/storage/bundles.py
mweiden: Says `@abstractproperty` is deprecated, but I'll add `@abstractmethod`

tests/__init__.py
mweiden: ✅

tests/__init__.py
mweiden: ✅

dss/storage/bundles.py
mweiden: ✅

dss/events/handlers/index.py
mweiden: ✅

dss/storage/bundles.py
hannes-ucsc: You'd also have to add `metaclass='abc.AbstractBaseClass'` to the `class` line. I wonder if that works with namedtuples.

tests/test_object_identifier.py
```diff
@@ -0,0 +1,117 @@
...
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+import dss
+from dss.storage.bundles import ObjectIdentifier, TombstoneID, FileFQID, BundleFQID, BUNDLE_PREFIX, FILE_PREFIX
+
+
+class TestObjectIdentifier(unittest.TestCase):
+
+    def test_to_str(self):
```
hannes-ucsc: For these kinds of tests I love using doctests but we don't have agreement on whether they are acceptable.

mweiden: Not familiar yet, I'll read up on that.

PR #759, opened 2017-12-14
Fix validation error: Forbidden -> forbidden

xbrianh:
When user email is unauthorized in bundles/delete, "Forbidden" fails
response validation and a 500 error is returned. Lower case "forbidden"
passes response validation, returning the correct 403 response.

Incorrect 500 error:
{'detail': "'Forbidden' is not one of ['forbidden']\n"
           '\n'
           "Failed validating 'enum' in "
           "schema['allOf'][1]['properties']['code']:\n"
           "    {'description': 'Machine-readable error code.  The types of "
           "return '\n"
           "                    'values should not be changed lightly.',\n"
           "     'enum': ['forbidden'],\n"
           "     'type': 'string'}\n"
           '\n'
           "On instance['code']:\n"
           "    'Forbidden'",
 'status': 500,
 'title': 'Response body does not conform to specification',
 'type': 'about:blank'}

### DIFF
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -143,7 +143,7 @@ def delete(uuid: str, replica: str, json_request_body: dict, version: str=None):
     if email not in ADMIN_USER_EMAILS:
         raise DSSException(
             requests.codes.forbidden,
-            "Forbidden",
+            "forbidden",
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Some subscription routes have this problem, too.

PR #762, opened 2017-12-14
Ensure bundle document is in one index only (fixes #738) 

### LINKED ISSUE #738: Reindexing: Ensure bundle only exists in one ES index
Multi-schema support added one index per document shape (aka schema set version). When re-indexing in REPAIR mode, we should make sure that the document only exists in a single index by deleting it from all other indexes except the one matching the document shape.

Because the bundle immutability constraint is enforced programmatically by us, we should guard against bugs by also ensuring document uniqueness during normal indexing.

ES doesn't let us perform a DELETE against an alias, so we need to query the alias for all occurrences of that document and issues a DELETE of that document to each index it doesn't belong into.

hannes-ucsc:


dss/storage/index_document.py
```diff
@@ -191,6 +191,47 @@ def get_shape_descriptor(self) -> typing.Optional[str]:
...
+            'from': 0,
+            'size': page_size,
+            'query': {
+                'terms': {
+                    '_id': [self.bundle_fqid]
+                }
+            }
+        })
+        hits = response['hits']
+        assert hits['total'] <= page_size, 'Document is in too many indices'
```
mweiden: Could you help me understand how this works a little more? I would have expected that the above query would return just one document per versioned index

tests/test_indexer.py
```diff
@@ -185,6 +186,27 @@ def _deletion_results_test():
 
         _deletion_results_test()
 
+    def test_reindexing_with_changed_shape(self):
+        bundle_key = self.load_test_data_bundle_for_path("fixtures/indexing/bundles/unversioned/smartseq2/paired_ends")
+        sample_event = self.create_bundle_created_event(bundle_key)
+        shape_descriptor = 'v99'
+
+        @eventually(timeout=5.0, interval=0.5)
```
mweiden: `@eventually`! 🎉

dss/storage/index_document.py
hannes-ucsc: I expect one hit per presence of a document with the given id in any index. I expect there to be a handful of indices and things would have to go seriously wrong for an id to be in all of them, so 64 seemed like a safe upper bound. I just didn't want to write code for paging and test it. Maybe the assertion message isn't clear enough.

tests/test_indexer.py
hannes-ucsc: Yeah, very helpful.

dss/storage/index_document.py
mweiden: I see. Better to fail fast in such an event, I suppose.

dss/events/handlers/index.py
```diff
@@ -28,6 +28,10 @@ def _index_and_notify(replica: Replica, bundle_fqid: BundleFQID, logger):
         logger.info(f"Received {replica.name} creation event for bundle which will be indexed: {bundle_fqid}")
         document = BundleDocument.from_replica(replica, bundle_fqid, logger)
         index_name = document.prepare_index()
+        versions = document.get_indexed_versions()
+        # No need to remove the document from the index it belongs into
+        versions.pop(index_name, None)
+        document.remove_versions(versions)
```
ttung: IMO a non-empty set for versions at this point deserves some logging.

hannes-ucsc: Good point.

hannes-ucsc:
I had verbal sign-off from @mbaumann.

PR #764, opened 2017-12-14
Slurp stderr when sanity checking Python version

kislyuk:
Anaconda Python prints its version on stderr instead of stdout.

### Test plan
Ran test suite with anaconda python.


ttung:
Shell is not guaranteed to be bash, is it? 

Consider `python -c 'import sys; print(sys.version)'`

kislyuk:
SHELL=/bin/bash at the top should make it bash, no?

kislyuk:
I'll switch to your suggested way later. 👍

PR #768, opened 2017-12-14
Fault tolerance for checkout service

amork:
Implement 60 minutes timeout for the overall bundle checkout
Implement retries for individual steps in the workflow including

TaskFailed
Timeout
Other
Configure intervals, max attempts and back-off rates

Test plan

Manually injected delays and failures in the lambda using alternative stage configuration. Confirmed complex retry behavior by analyzing execution log in AWS Step Function console

connect to #614


amork:
connect to #614

iam/policy-templates/dss-checkout-sfn-lambda.json
```diff
@@ -63,4 +65,4 @@
     }
 
   ]
-}
+}
```
ttung: please add newline.

PR #771, opened 2017-12-14
Replicas are now enums, not strings.

ttung:
I did a grep of `"gcp"` and it's clean.  Found some `"aws"` but will file that separately as that's an entirely different code path.


mweiden:
@ttung darn, I thought that I'd caught them all. Thanks for this.

Issue #772, opened 2017-12-14
dss/api/search.py uses replicas as strings


ttung:
Fortunately, neither really interact with `Config.get_cloud_specific_handles`, so neither is affected.

ttung:
Not sure if this should go to @mweiden or (@mikebaumann|@Bento007|@hannes-ucsc) (extended regex, of course).

mweiden:
@ttung I'm on it

PR #776, opened 2017-12-15
Search API should use Replica enums instead of strings

mweiden:
Fixes https://github.com/HumanCellAtlas/data-store/issues/772

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -31,24 +31,23 @@ def post(json_request_body: dict,
     es_query = json_request_body['es_query']
     per_page = PerPageBounds.check(per_page)
 
-    if replica is None:
-        replica = "aws"
+    replica_enum = Replica[replica] if replica is not None else Replica.aws
 
     get_logger().debug("Received posted query. Replica: %s Query: %s Per_page: %i Timeout: %s Scroll_id: %s",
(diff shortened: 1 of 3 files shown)


dss/api/search.py
```diff
@@ -31,24 +31,23 @@ def post(json_request_body: dict,
     es_query = json_request_body['es_query']
     per_page = PerPageBounds.check(per_page)
 
-    if replica is None:
-        replica = "aws"
+    replica_enum = Replica[replica] if replica else Replica.aws
```
ttung: Strictly speaking, `if replica` is not the same as `if replica is None`

dss/config.py
```diff
@@ -204,6 +204,9 @@ def get_es_index_name(index_type: ESIndexType,
         Returns the fully qualified name of an Elasticsearch index of documents for a given
         replica of a given type and shape.
         """
+        if not isinstance(replica, Replica):
```
ttung: `assert isinstance(replica, Replica)`

dss/api/search.py
mweiden: ✅

dss/config.py
mweiden: ✅

PR #777, opened 2017-12-15
Changed where test data is retrieved from.

Bento007:
Test data is now retrieved and stored in test/fixtures. This gives us more control over our test data especially as new versions of data bundles are released.

### Release notes
Changed where test data is retrieved from. Test data is now populated from test/fixtures.


ttung:
If we're doing this, is there still a need for `data-bundle-examples` to be a git submodule?

Bento007:
I don't see a reason for it.

mikebaumann:
smoketest.py uses the `data-bundle-examples` submodule.

Bento007:
we can leave `data-bundle-examples` for now and take it out when we're ready.

PR #780, opened 2017-12-16
Refactor smoketest into testcase

kislyuk:


tests/test_smoketest.py
```diff
@@ -0,0 +1,122 @@
...
+    if isinstance(command, str):
+        kwargs["shell"] = True
+    print(GREEN(command))
+    try:
+        return runner(command, **kwargs)
+    except CalledProcessError as e:
+        parser.exit(RED(f'{parser.prog}: Exit status {e.returncode} while running "{command}". Stopping.'))
+
+
+@infra.testmode.standalone
```
Bento007: I didn't know you could the decorator works on the whole class. That's handy. Also is this an integration test?

tests/test_smoketest.py
```diff
@@ -0,0 +1,122 @@
...
+
+                res = run(f"{venv_bin}hca dss put-subscription "
+                          "--callback-url https://example.com/ "
+                          "--es-query '{}' "
+                          f"--replica {replica}",
+                          runner=check_output)
+                sub_id = json.loads(res.decode())["uuid"]
+                run(f"{venv_bin}hca dss get-subscriptions --replica {replica}")
+                run(f"{venv_bin}hca dss delete-subscription --replica {replica} --uuid {sub_id}")
+        finally:
```
Bento007: This should be in tearDownClass

tests/test_smoketest.py
```diff
@@ -0,0 +1,122 @@
...
+    try:
+        return runner(command, **kwargs)
+    except CalledProcessError as e:
+        parser.exit(RED(f'{parser.prog}: Exit status {e.returncode} while running "{command}". Stopping.'))
+
+
+@infra.testmode.standalone
+class Smoketest(unittest.TestCase):
+    def test_smoketest(self):
+        if os.path.exists("dcp-cli"):
```
Bento007: This should be in setUpClass

kislyuk: Thanks, done.

tests/test_smoketest.py
kislyuk: Thanks, done.

tests/test_smoketest.py
kislyuk: D'oh. Corrected.

PR #781, opened 2017-12-16
Compare existing document with new one and log differences (fixes #774)

### LINKED ISSUE #774: Reindexing: Compare existing document with new one and log differences

hannes-ucsc:
fixes #774

### DIFF
--- a/dss/storage/index_document.py
+++ b/dss/storage/index_document.py
@@ -30,6 +30,12 @@ class IndexDocument(dict):
         self.replica = replica
         self.fqid = fqid
 
+    @classmethod
+    def from_index(cls, replica: Replica, bundle_fqid: BundleFQID, index_name, logger, version=None):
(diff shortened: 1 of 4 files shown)


dss/storage/index_document.py
```diff
@@ -40,21 +46,30 @@ def add_to_index(self, index_name: str):
...
                               ex, index_name, json.dumps(self, indent=4))
             raise
         return initial_mappings
 
+    def to_json(self):
+        return json.dumps(self)
+
+    def __eq__(self, other: object) -> bool:
+        return (self is other
+                or super().__eq__(other)
```
kislyuk: Please group these with parens. It's not immediately obvious what the priority/logic is here.

hannes-ucsc: Addressed this.

PR #783, opened 2017-12-18
Fix Makefile target mixup

hannes-ucsc:
The first commit fixes a bug in .travis.yml and removes a stale .PHONY declaration.

The bug causes non-IT builds tests to essentially skip tests and IT-builds to only run standalone tests. Any PR builds of branches based on b8290d240a4c7f20753122415a6df5a6666414c8 should be disregarded and the branches rebased once this is merged. 

The second commit makes the target naming in Makefile more intuitive.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ install:
 - export PATH=$(pwd)/make4/usr/bin:$PATH
 script:
 - set -eo pipefail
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
Backed out second commit. Will file another PR for that.

PR #790, opened 2017-12-18
Make variable naming more consistent with test target names

hannes-ucsc:
This is purely cosmetic. It bugged me `test_srcs` wasn't used for the `test` target. With this change

* `test` uses test_srcs
* `integration_test` uses integration_test_srcs
* `all_test` uses all_test_srcs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,29 +8,29 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test_srcs := $(wildcard tests/test_*.py)
-standalone_test_srcs := $(addprefix standalone__, $(test_srcs))
+all_test_srcs := $(addprefix all__, $(test_srcs))
 integration_test_srcs := $(addprefix integration__, $(test_srcs))
 
-test: lint mypy $(standalone_test_srcs)
+test: lint mypy $(test_srcs)
 	coverage combine
 	rm -f .coverage.*
 
-$(standalone_test_srcs): standalone__%.py :
-	DSS_TEST_MODE=standalone coverage run -p --source=dss -m unittest $*.py
+$(test_srcs): %.py :
+	DSS_TEST_MODE="standalone" coverage run -p --source=dss -m unittest $@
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,29 +8,29 @@ mypy:
...
+$(test_srcs): %.py :
+	DSS_TEST_MODE="standalone" coverage run -p --source=dss -m unittest $@
 
 integration_test: lint mypy $(integration_test_srcs)
 	coverage combine
 	rm -f .coverage.*
 
 $(integration_test_srcs): integration__%.py :
-	DSS_TEST_MODE=integration coverage run -p --source=dss -m unittest $*.py
+	DSS_TEST_MODE="integration" coverage run -p --source=dss -m unittest $*.py
```
hannes-ucsc: Fully aware that these quotes are redundant but they help with readability.

Makefile
```diff
@@ -8,29 +8,29 @@ mypy:
...
 
 all_test: lint mypy $(test_srcs)
 	coverage combine
 	rm -f .coverage.*
 
-$(test_srcs): %.py :
-	DSS_TEST_MODE="integration standalone" coverage run -p --source=dss -m unittest $@
+$(all_test_srcs): all__%.py :
+	coverage run -p --source=dss -m unittest $*.py
```
ttung: I'd like to leave the `DSS_TEST_MODE="integration standalone"` in there because if someone accidentally set the env var in their environment, I still want `make all_test` to behave as expected.

hannes-ucsc: My intention with this PR was not to change any semantics, only to make naming more intuitive. Looks like I failed to do so. Just to be sure, is the following correct?

- `make test` should run tests decorated as `@standalone` and undecorated tests, 
- `make `integration_test` should run tests decorated as `@integration` and undecorated tests, 
- `all_test` should run tests decorated as `@standalone` or `@integration` and undecorated tests.

ttung: Yes, that sounds good.

hannes-ucsc: Fixed

PR #797, opened 2017-12-19
[WIP] Integrating changes towards reindexing (REPAIR and VERIFY)

hannes-ucsc:


dss/storage/index_document.py
```diff
@@ -85,17 +127,81 @@ def files(self):
...
+        ES index. The tombstone data overrides the document's data in the index.
+
+        :param tombstone: The document with which to replace this document in the index.
+        :param dryrun: see :py:meth:`~IndexDocument.index`
+        :return: see :py:meth:`~IndexDocument.index`
+        """
+        elasticsearch_retry.add_context(bundle=self, tombstone=tombstone)
+        self.logger.info(f"Writing tombstone for {self.replica.name} bundle: {self.fqid}")
+        # Preare the index using the original data such that the tombstone can be placed in the correct index.
+        index_name = self._prepare_index(dryrun)
```
mikebaumann: As implemented here, indexing a bundle that has been admin deleted will fail if the bundle data has been deleted from the bucket. This will "work" in the short-term, as admin delete does not actually delete the bundle data. Yet, going forward, it would be best if indexing of an admin deleted bundle did not depend on the bundle data existing in the bucket!

hannes-ucsc: https://github.com/HumanCellAtlas/data-store/issues/821

hannes-ucsc:
I split this up into three separate PRs: #822, #823 and #824.

Issue #609, opened 2017-11-06
Validate and constrain metadata files to schema when indexing


mikebaumann:
As advocated by @briandoconnor and @ttung, to ensure bundle metadata is reliably indexed and searchable, enhance the indexer to validate the metadata files according to the schema it provides, and to remove any portions of the index document that do not conform to the schema before it is indexed.

see epic #289

mweiden:
@mikebaumann @Bento007 I'm starting to look into this and have a clarifying question.

I see that the bundle manifest is uploaded to the store before the indexer even touches it.

1. Manifest upload https://github.com/HumanCellAtlas/data-store/blob/master/dss/api/bundles.py#L179-L182
2. Manifest read https://github.com/HumanCellAtlas/data-store/blob/master/dss/events/handlers/index.py#L50 (both AWS & GCP)

Would it be better to validate the manifest before uploading it to storage and warning the user up front? The benefit here would be preventing the upload of data in the system that we can't use.

Further, if we validated at this point we could warn the user that non-conforming portions of the index document will not be used.

Bento007:
keep dynamic mapping.
strip everything that is not in the official schema.

ttung:
@mweiden Once the purple box is up and running, we should not be rejecting anything that lands here.

Bento007:
Looking for a solution for caching the json schema from git. We should not rely on git to retrieve the schema every time a document is indexed. 

We need to investigate if minor version changes of the schema will affect validation.

Issue #608, opened 2017-11-06
Added subscription notifications to smoke test


Bento007:
continuation of #449

Bento007:
This has been completed

PR #802, opened 2017-12-22
Cache URL content in S3

Bento007:
connects to #609
### Test plan
Will add test cases to check urls are stored and buckets cache is used.

### Release notes
Caching the contents of website urls to an S3 bucket.

### DIFF
--- a/dss/storage/validator.py
+++ b/dss/storage/validator.py
@@ -1,23 +1,121 @@
 """Utilities in this file are used to removed extra fields from index data before adding to the index."""
-import re
+import io
+from io import BytesIO
+
+import boto3
+from cloud_blobstore import BlobNotFoundError
 from jsonschema import validators
 from jsonschema import _utils, _validators
 
 import json
-from urllib.request import urlopen
-from urllib.error import URLError
-import ssl
 from typing import List
 import logging
 
-
-def read_json_url(link):
-    gcontext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
-    with urlopen(link, context=gcontext) as f:
-        return json.load(f)
+from dss import Replica, Config
+from hashlib import sha1
+import requests
 
 
-def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
+DSS_Draft4Validator = validators.create(meta_schema=_utils.load_schema("draft4"),
+                                        validators={'$ref': _validators.ref,
+                                                    'additionalProperties': _validators.additionalProperties,
+                                                    'properties': _validators.properties_draft4,
+                                                    'required': _validators.required_draft4
+                                                    },
+                                        version="draft4"
+                                        )
+
+
+class SizeLimitError(IOError):
+    def __init__(self, url, limit) -> None:
+        super().__init__(f"{url} not cached. The URL's contents have exceeded {limit} bytes.")
+
+
+class S3UrlCache:
+    """Caches content of arbitrary URLs the first time they are requested. Currently only supports content lengths of up
+     to a few megabytes."""
+    _max_size_default = 64 * 1024 * 1024  # The default max_size per URL = 64 MB
+    _chunk_size_default = 1024 * 1024  # The default chunk_size = 1 MB
+
+    def __init__(self, logger: logging.Logger,
+                 max_size: int = _max_size_default,
+                 chunk_size: int = _chunk_size_default):
+        """
+        :param max_size: The maximum number of bytes of content that can be cached. An error is returned if the content
+        is greater than max_size, and the URL is not cached.
+        :param chunk_size: The amount of content retrieved from the URL per request."""
+        self.max_size = max_size if max_size is not None else self._max_size_default
+        self.chunk_size = chunk_size if chunk_size is not None else self._chunk_size_default
+        self.s3_client = boto3.client('s3')
+
+        # A URL's contents are only stored in S3 to keep the data closer to the aws lambas which use them.
+        self.blobstore, _, self.bucket = Config.get_cloud_specific_handles(Replica.aws)
+        self.logger = logger
+
+    def resolve(self, url: str) -> bytearray:
+        """
+        Requests the contents of a URL by first checking for the contents in an S3 bucket. If not present, the contents
+        are retrieved from the URL and stored in S3.
+
+        :param url: The url to retrieve the content from.
+        """
+        key = self._url_to_key(url)
+
+        # if key in S3 bucket return value from there.
+        try:
+            content = bytearray(self.blobstore.get(self.bucket, key))
+        except BlobNotFoundError:
+            self.logger.info("%s", f"{url} not found in cache. Adding it to {self.bucket} with key {key}.")
+            with requests.get(url, stream=True) as resp_obj:
+                content = bytearray()
+                for chunk in resp_obj.iter_content(chunk_size=self.chunk_size):
+                    #  check if max_size exceeded before storing content to avoid storing large chunks
+                    if len(content) + len(chunk) > self.max_size:
+                        raise SizeLimitError(url, self.max_size)
+                    content.extend(chunk)
+            self._upload_content(key, url, content)
+        return content
+
+    def evict(self, url: str) -> bool:
+        '''
+        Removes the cached URL content from S3.
+        :param url: the url for the content to removed from S3'''
+        if self.contains(url):
+            self.logger.info(f"{url} removed from cache in {self.bucket}.")
+            self.blobstore.delete(self.bucket, self._url_to_key(url))
+        else:
+            self.logger.info(f"{url} not found and not removed from cache.")
+
+    def contains(self, url: str) -> bool:
+        key = self._url_to_key(url)
+        try:
+            self.blobstore.get_user_metadata(self.bucket, key)['dss_cached_url']
+        except BlobNotFoundError:
+            return False
+        else:
+            return True
+
+    def _reverse_key_lookup(self, key: str) -> str:
+        return self.blobstore.get_user_metadata(self.bucket, key)['dss_cached_url']
+
+    @staticmethod
+    def _url_to_key(url: str) -> str:
+        return 'cache/' + sha1(url.encode("utf-8")).hexdigest()
+
+    def _upload_content(self, key: str, url: str, content: bytearray) -> None:
+        meta_data = {
+            "Metadata": {'dss_cached_url': url},
+            "ContentType": "application/octet-stream",
+        }
+        self.s3_client.upload_fileobj(
+            Fileobj=io.BytesIO(content),
+            Bucket=self.bucket,
+            Key=key,
+            ExtraArgs=meta_data,
+        )
+
+
+def remove_json_fields(json_data: dict, path: List[str], fields: List[str]) -> None:
     """
     Removes fields from the path in json_data.
 
@@ -32,21 +130,17 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
         current.pop(field)
 
 
-DSS_Draft4Validator = validators.create(
-    meta_schema=_utils.load_schema("draft4"),
-    validators={
-        u"$ref": _validators.ref,
-        u"additionalProperties": _validators.additionalProperties,
-        u"properties": _validators.properties_draft4,
-        u"required": _validators.required_draft4,
-    },
-    version="draft4",
-)
-
-resolver = validators.RefResolver(referrer='', base_uri='')
-
-
 def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -> list:
+    cache = S3UrlCache(logger)
+
+    def request_json(url):
+        return json.loads(cache.resolve(url).decode("utf-8"))
+
+    resolver = validators.RefResolver(referrer='',
+                                      base_uri='',
+                                      handlers={'http': request_json,
+                                                'https': request_json}
+                                      )
     extra_fields = []
     extra_documents = []
     for document in index_data.keys():
@@ -54,7 +148,7 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
         schema_url = None if core is None else core.get('schema_url')
         if schema_url is not None:
             try:
-                schema = read_json_url(schema_url)
+                schema = request_json(schema_url)
             except Exception as ex:
                 extra_documents.append(document)
                 logger.warning("%s", f"Unable to retrieve schema from {document} in {bundle_id} "
--- a/tests/test_s3urlcache.py
+++ b/tests/test_s3urlcache.py
@@ -1,0 +1,202 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import logging
+import os
+import io
+import shutil
+import sys
+import unittest
+import threading
+import requests
+from http.server import BaseHTTPRequestHandler, HTTPServer
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from dss import Config, BucketConfig
+from dss.config import Replica
+from tests.infra import start_verbose_logging, testmode
+from dss.storage.validator import S3UrlCache
+from dss.storage.validator import SizeLimitError
+from dss.util import networking
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+start_verbose_logging()
+
+KiB = 1024
+MB = KiB ** 2
+
+randomdata = 'a random string of data'
+
+class HTTPInfo:
+    address = "127.0.0.1"
+    port = None
+    server = None
+    thread = None
+
+    @classmethod
+    def make_url(cls):
+        cls.url = f"http://{cls.address}:{cls.port}"
+
+
+def setUpModule():
+    HTTPInfo.port = networking.unused_tcp_port()
+    HTTPInfo.server = HTTPServer((HTTPInfo.address, HTTPInfo.port), GetTestHandler)
+    HTTPInfo.make_url()
+    HTTPInfo.thread = threading.Thread(target=HTTPInfo.server.serve_forever)
+    HTTPInfo.thread.start()
+
+    global randomdata
+    randomdata = os.urandom(MB)
+
+
+def tearDownModule():
+    HTTPInfo.server.shutdown()
+
+@testmode.standalone
+class TestS3UrlCache(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        replica = Replica.aws
+        Config.set_config(BucketConfig.TEST_FIXTURE)
+        cls.blobstore, _, cls.test_fixture_bucket = Config.get_cloud_specific_handles(replica)
+        Config.set_config(BucketConfig.TEST)
+        cls.test_bucket = replica.bucket
+
+    def setUp(self):
+        self.urls_to_cleanup = set()
+        self.cache = S3UrlCache(logger)
+
+    def tearDown(self):
+        self._delete_cached_urls()
+
+    def test_store_in_cache(self):
+        """The URL contents are stored in S3 and the contents returned, when requested url is not found in cache."""
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+        url_key = S3UrlCache._url_to_key(url)
+
+        self._delete_cached_urls()
+        with self.assertLogs(logger, "INFO") as log_monitor:
+            url_content = self.cache.resolve(url)
+
+        original_data = randomdata[:KiB]
+        self.assertEqual(len(url_content), KiB)
+        self.assertEqual(url_content, original_data)
+        self.assertTrue(log_monitor.output[0].endswith(f"{url} not found in cache. Adding it to "
+                                                       f"{self.test_bucket} with key {url_key}."))
+
+    def test_retrieve_from_cache(self):
+        """Stored URL contents is retrieved from S3, when a cached url is requested"""
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+        with self.assertLogs(logger, 'INFO') as log_monitor:
+            url_content = self.cache.resolve(url)
+            cached_content = self.cache.resolve(url)
+        self.assertEqual(len(log_monitor.output), 1)
+        self.assertEqual(url_content, cached_content)
+
+    def test_bad_url(self):
+        bad_urls = ['', '//', 'http://?']
+        self.urls_to_cleanup.update(bad_urls)
+
+        for url in bad_urls:
+            with self.subTest(bad_url=url):
+                with self.assertRaises(requests.RequestException):
+                    self.cache.resolve(url)
+
+    def test_url_SizeLimitError(self):
+        """Exception returned when URL content size is greater than max_size."""
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+        self.cache.max_size = 1
+        with self.assertRaises(SizeLimitError) as ex:
+            self.cache.resolve(url)
+        self.assertEqual(ex.exception.args[0], f"{url} not cached. The URL's contents have exceeded "
+                                               f"{self.cache.max_size} bytes.")
+
+    def test_content_size_0(self):
+        url = f"{HTTPInfo.url}/0"
+        self.urls_to_cleanup.add(url)
+        url_content = self.cache.resolve(url)
+        self.assertEqual(len(url_content), 0)
+
+    def test_chunked_content(self):
+        size = KiB * 10
+        url = f"{HTTPInfo.url}/{size}"
+        self.urls_to_cleanup.add(url)
+        self.cache.max_size = 1 * MB
+        self.cache.chunk_size = 1 * KiB
+        url_content = self.cache.resolve(url)
+        original_data = randomdata[:size]
+        self.assertEqual(url_content, original_data)
+
+    def test_stored_url_metadata(self):
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+        url_key = S3UrlCache._url_to_key(url)
+        self.cache.resolve(url)
+
+        with self.subTest("check dss_cached_url"):
+            cached_url = self.cache._reverse_key_lookup(url_key)
+            self.assertEqual(cached_url, url)
+
+        with self.subTest("check content_type"):
+            contentType = self.blobstore.get_content_type(self.test_bucket, url_key)
+            self.assertEqual(contentType, "application/octet-stream")
+
+    def test_evict(self):
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+        url_key = S3UrlCache._url_to_key(url)
+
+        with self.assertLogs(logger, "INFO") as log_monitor:
+            # Verify the URL is cached
+            self.cache.resolve(url)
+            self.assertTrue(self.cache.contains(url))
+            # Remove the URL from cache
+            self.cache.evict(url)
+            self.assertTrue(not self.cache.contains(url))
+            self.cache.evict(url)
+            self.assertTrue(not self.cache.contains(url))
+
+        self.assertTrue(log_monitor.output[0].endswith(f"{url} not found in cache. Adding it to "
+                                                       f"{self.test_bucket} with key {url_key}."))
+        self.assertTrue(log_monitor.output[1].endswith(f"{url} removed from cache in {self.test_bucket}."))
+        self.assertTrue(log_monitor.output[2].endswith(f"{url} not found and not removed from cache."))
+
+    def test_contains(self):
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+
+        self.assertTrue(not self.cache.contains(url))
+        self.cache.resolve(url)
+        self.assertTrue(self.cache.contains(url))
+
+    def _delete_cached_urls(self):
+        for url in self.urls_to_cleanup:
+            self.cache.evict(url)
+
+
+class GetTestHandler(BaseHTTPRequestHandler):
+    _response_code = 200
+
+    def do_GET(self):
+        length = int(self.path.split('/')[1])
+
+        self.send_response(self._response_code)
+        self.send_header("content-length", length)
+        self.end_headers()
+
+        if length:
+            shutil.copyfileobj(io.BytesIO(randomdata[:length]), self.wfile)
+        else:
+            return
+
+if __name__ == "__main__":
+    unittest.main()


dss/storage/validator.py
```diff
@@ -9,15 +13,73 @@
...
+                                                    u"additionalProperties": _validators.additionalProperties,
+                                                    u"properties": _validators.properties_draft4,
+                                                    u"required": _validators.required_draft4
+                                                    },
+                                        version="draft4"
+                                        )
+
+
+class S3UrlCache:
+    """Used by the resolver to retrieve the url contents from either a url or an s3 cloud bucket"""
```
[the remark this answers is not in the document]
Bento007: How do you know only a few MB can be stored?

dss/storage/validator.py
```diff
@@ -9,15 +13,73 @@
...
-        return json.load(f)
+        # if key in s3 bucket return value from there.
+        try:
+            content = self.blobstore.get(self.bucket, key)
+        except BlobNotFoundError:
+            content = None
+            self.logger.info("%s", f"{url} not found in cache. Adding it to {self.bucket} with key {key}.")
 
+        # if key not in s3 bucket, retrieve contents of urls and store in s3 bucket.
+        if content is None:
```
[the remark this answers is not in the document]
Bento007: I wanted to do that, but thought it would be better to limit the amount of code in a try except block.

dss/storage/validator.py
```diff
@@ -9,15 +13,73 @@
...
+            self.upload_file_handle(key, url, content)
+        return content.decode("utf-8")
 
-def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
+    def request_json(self, url: str) -> dict:
+        return json.loads(self.request(url))
+
+    def upload_file_handle(self, key: str, url: str, data: str) -> None:
+        meta_data = {
+            "Metadata": {'SCHEMA_URL': url},
```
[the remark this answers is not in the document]
Bento007: Ill go with `DSS_CACHED_URL`

dss/storage/validator.py
```diff
@@ -9,15 +13,73 @@
...
+                                        version="draft4"
+                                        )
+
+
+class S3UrlCache:
+    """Used by the resolver to retrieve the url contents from either a url or an s3 cloud bucket"""
+
+    def __init__(self, logger: logging.Logger):
+        self.s3_client = boto3.client('s3')
+        self.blobstore, _, self.bucket = Config.get_cloud_specific_handles(Replica.aws)
```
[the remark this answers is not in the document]
Bento007: Added a comment.

dss/storage/validator.py
```diff
@@ -9,15 +13,73 @@
...
+        self.logger = logger
+        self.gcontext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
 
+    def request(self, url: str) -> str:
```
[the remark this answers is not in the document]
Bento007: added docstring

dss/storage/validator.py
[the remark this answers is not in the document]
hannes-ucsc: Storage in S3 is not a problem but the fact that you're reading the entire URL's content into memory might be. If you get a URL pointing to a 1G file you'll be using 1G of memory (or a multiple of that, e.g. when decoding the string from UTF-8). You'll also block for a minute or so downloading the file. Files of that size are more and more impractical to store in memory and are usually streamed rather than loaded into memory, especially in a resource constraint environment like a lambda.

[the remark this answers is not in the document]
hannes-ucsc: https://docs.python.org/3/library/shutil.html#shutil.copyfileobj

is a good way to limit the amount of data read from a file handle. You could use this to prevent the lambda from being DOSed by freakishly large URLs.

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+            content = self.blobstore.get(self.bucket, key)
+        except BlobNotFoundError:
+            self.logger.info("%s", f"{url} not found in cache. Adding it to {self.bucket} with key {key}.")
+            with requests.get(url, stream=True) as f:
+                content = b''
+                for chunk in f.iter_content(chunk_size=None):
+                    content = content + chunk
+                    if len(content) > max_size:
+                        self.logger.warning("{url} not cached. Contents have exceed {max_size} Bytes.")
+                        raise SizeLimitError("{url} not cached. Contents have exceed {max_size} Bytes.")
```
hannes-ucsc: `exceeded` and `bytes`

Consider doing this:

```
class SizeLimitException(Exception):
    def __init__(self, url, limit) -> None:
        super().__init__(f"{url} not cached. Contents have exceeded {limit} bytes.")
…
raise SizeLimitException(url, max_size)
```

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+
+            self.upload_file_handle(key, url, content)
+        return content.decode("utf-8")
+
+    def request_json(self, url: str, **kwargs) -> dict:
+        return json.loads(self.request(url), **kwargs)
+
+    def upload_file_handle(self, key: str, url: str, data: str) -> None:
```
hannes-ucsc: This should be a protected method and the name is wrong: there is no "file handle" involved here.

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+                for chunk in f.iter_content(chunk_size=None):
+                    content = content + chunk
+                    if len(content) > max_size:
+                        self.logger.warning("{url} not cached. Contents have exceed {max_size} Bytes.")
+                        raise SizeLimitError("{url} not cached. Contents have exceed {max_size} Bytes.")
+
+            self.upload_file_handle(key, url, content)
+        return content.decode("utf-8")
+
+    def request_json(self, url: str, **kwargs) -> dict:
```
hannes-ucsc: I don't think this method is worth having. 

`json.loads(cache.request(url))`

is not unbearably longer than

`cache.request_json(url)`

I would consider including it if it added the `Expect: application/json` header. to the request. In that case, leaving the method in would support my point for making `max_size` a constructor argument. As currently implemented, the caller has no way to override the `max_size` argument to `request()`.

dss/storage/validator.py
```diff
@@ -32,34 +100,28 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
...
-        u"required": _validators.required_draft4,
-    },
-    version="draft4",
-)
-
-resolver = validators.RefResolver(referrer='', base_uri='')
-
-
 def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -> list:
+    handler = S3UrlCache(logger)
```
hannes-ucsc: Rename to `cache`. Add `def request_json(…): return json.loads(cache.resolve())`

dss/storage/validator.py
```diff
@@ -32,34 +100,28 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
...
         if core is not None:
             try:
-                schema = read_json_url(core['schema_url'])
-            except URLError as ex:
+                schema = handler.request_json(core['schema_url'])
+            except requests.ConnectionError as ex:
                 extra_documents.append(document)
-                logger.warning("%s", f"Unable to retrieve schema from url {core['schema_url']} due to exception: {ex}.")
+                logger.warning("%s",
+                               f"Unable to retrieve schema from url {core['schema_url']} due to exception: {ex}.")
```
hannes-ucsc: Nit: Extracting `core['schema_url']` into a variable would guarantee that the error message never gets out of sync with reality.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+
+def tearDownModule():
+    IndexSuffix.reset()
+
+
+class TestS3UrlCache(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.replica = Replica.aws
```
hannes-ucsc: This is only used in this method here and doesn't need to be a class attribute.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+        cls.url = "http://www.example.com"
+        cls.url_key = "cache/" + sha1(cls.url.encode("utf-8")).hexdigest()
+        cls.cache = S3UrlCache(logger)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.blobstore.delete(cls.test_bucket, cls.url_key)
+
+    def test_caching(self):
+        self.blobstore.delete(self.test_bucket, self.url_key)
```
hannes-ucsc: Delete in `setUp` and in `tearDown` and assert non-existence in test. Tests are partial proofs of correctness. Asserting the non-existence makes the proof stronger.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+
+class TestS3UrlCache(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.replica = Replica.aws
+        Config.set_config(BucketConfig.TEST_FIXTURE)
+        cls.blobstore, _, cls.test_fixture_bucket = Config.get_cloud_specific_handles(cls.replica)
+        Config.set_config(BucketConfig.TEST)
+        _, _, cls.test_bucket = Config.get_cloud_specific_handles(cls.replica)
```
hannes-ucsc: You could use `replica.bucket` since my PR landed.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+        cls.url = "http://www.example.com"
+        cls.url_key = "cache/" + sha1(cls.url.encode("utf-8")).hexdigest()
```
hannes-ucsc: You don't generally want to duplicate an implementation detail of the class under test. Extract the key computation into a protected method of the S3UrlCache class and use that method here and in the S3UrlCache.

Better yet, consider adding an `invalidate(url)` method to `S3UrlCache` that performs the deletion.

Generally speaking, tests should not be "in bed" with the code under test. They should be limited to testing the public interface of the class and avoid using implementation details as much as possible. I can explain in person why. Sometimes, writing a test actually drives the shaping of the public interface. It never occurred to me that we would need `invalidate()` but now it makes sense.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+        cls.cache = S3UrlCache(logger)
```
hannes-ucsc: This should be in `setUp()`. The more you do in `setUpClass()` the higher the likelihood that tests affect each other. A key criterion for good tests is that they are isolated from each other, especially each other's failures. 

Only immutable state or state that's expensive to create should be set up in `setUpClass()`.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+from dss.storage.validator import S3UrlCache
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+start_verbose_logging()
+
+def setUpModule():
+    IndexSuffix.name = __name__.rsplit('.', 1)[-1]
```
hannes-ucsc: This looks fishy. Temporarily modifying global state like can be dangerous. Is there any way we can limit the scope of this modification (to, say, a single test) or avoid it altogether?

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+class TestS3UrlCache(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.replica = Replica.aws
+        Config.set_config(BucketConfig.TEST_FIXTURE)
+        cls.blobstore, _, cls.test_fixture_bucket = Config.get_cloud_specific_handles(cls.replica)
+        Config.set_config(BucketConfig.TEST)
+        _, _, cls.test_bucket = Config.get_cloud_specific_handles(cls.replica)
+        cls.url = "http://www.example.com"
```
hannes-ucsc: Please use a URL that we control. Hint: If an S3 bucket is public, each object in it has a public http: URL.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_url_size(self):
+        #  Exception returned when url content size is greater than max_size.
+        self.blobstore.delete(self.test_bucket, self.url_key)
+        with self.assertRaises(SizeLimitError):
+            self.cache.request(self.url, 1)
+
+    def test_stored_url_metadata(self):
+        self.cache.request(self.url)
+        with self.subTest("check dss_cached_url"):
+            metadata = self.blobstore.get_user_metadata(self.test_bucket, self.url_key)
```
hannes-ucsc: Add `S3UrlCache.reverseLookup(key) => url` method that hides this detail.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,81 @@
...
+
+def setUpModule():
+    IndexSuffix.name = __name__.rsplit('.', 1)[-1]
+
+
+def tearDownModule():
+    IndexSuffix.reset()
+
+
+class TestS3UrlCache(unittest.TestCase):
```
hannes-ucsc: Add a test for a non-existent URL. If you can stomach it, add a test against a URL of 0-length content.

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+
+        # A URL's contents are only stored in s3 to keep the data closer to the aws lambas which use them.
+        self.blobstore, _, self.bucket = Config.get_cloud_specific_handles(Replica.aws)
+        self.logger = logger
 
-def read_json_url(link):
-    gcontext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
-    with urlopen(link, context=gcontext) as f:
-        return json.load(f)
+    def request(self, url: str, max_size: Optional[int]=64000) -> str:
```
hannes-ucsc: Also, 64000 is much too small. Let's use 1 MiB. As per ancient convention, use powers of two, e.g. 64 * 1024.

https://en.wikipedia.org/wiki/Mebibyte

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
-def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
+        # if key in s3 bucket return value from there.
+        try:
+            content = self.blobstore.get(self.bucket, key)
+        except BlobNotFoundError:
+            self.logger.info("%s", f"{url} not found in cache. Adding it to {self.bucket} with key {key}.")
+            with requests.get(url, stream=True) as f:
+                content = b''
+                for chunk in f.iter_content(chunk_size=None):
+                    content = content + chunk
```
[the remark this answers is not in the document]
hannes-ucsc: In the above link, I'm referring to the first bullet of item 6 under Notes.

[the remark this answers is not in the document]
hannes-ucsc: Sorry, second bullet.

dss/storage/validator.py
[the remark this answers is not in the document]
Bento007: max size changed to 1 Mb and moved to `__init__`

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+        """
+        key = 'cache/' + sha1(url.encode("utf-8")).hexdigest()
 
-def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
+        # if key in s3 bucket return value from there.
+        try:
+            content = self.blobstore.get(self.bucket, key)
+        except BlobNotFoundError:
+            self.logger.info("%s", f"{url} not found in cache. Adding it to {self.bucket} with key {key}.")
+            with requests.get(url, stream=True) as f:
```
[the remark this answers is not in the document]
Bento007: changed to `resp_obj`

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+
+
+DSS_Draft4Validator = validators.create(meta_schema=_utils.load_schema("draft4"),
+                                        validators={u"$ref": _validators.ref,
+                                                    u"additionalProperties": _validators.additionalProperties,
+                                                    u"properties": _validators.properties_draft4,
+                                                    u"required": _validators.required_draft4
+                                                    },
+                                        version="draft4"
+                                        )
```
kislyuk: Why is "draft4" (or any other metadata schema version) hardcoded in our code? It should be parameterized. (I realize this was added earlier - sorry I missed the earlier review)

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+
+
+class SizeLimitError(requests.RequestException):
+    "The max size content size has been exceeded"
```
kislyuk: Please remove the extra "size" and clarify this message. The content size of what content?

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+    pass
+
+class S3UrlCache:
```
kislyuk: I think you should replace this whole class with a function (maybe `caching_get_url(...)`?) with a `functools.lru_cache()` decorator.

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
 
+from requests import RequestException
+
+from dss import Replica, Config, DSSException
+from hashlib import sha1
+import requests
+
+
+DSS_Draft4Validator = validators.create(meta_schema=_utils.load_schema("draft4"),
+                                        validators={u"$ref": _validators.ref,
```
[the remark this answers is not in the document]
Bento007: fixed

dss/storage/validator.py
Bento007: using 1Mb

dss/storage/validator.py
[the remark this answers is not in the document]
Bento007: using bytearray

dss/storage/validator.py
hannes-ucsc: I'm confused by this request. This is a persistent cache, not just an in-memory one.

dss/storage/validator.py
hannes-ucsc: That's draft4 of the JSON schema spec, not the HCA metadata schema.

dss/storage/validator.py
Bento007: "The content of the url requested has exceed the max size allowed."

dss/storage/validator.py
Bento007: changed to `__upload_content`

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+        # if key in s3 bucket return value from there.
+        try:
+            content = self.blobstore.get(self.bucket, key)
+        except BlobNotFoundError:
+            self.logger.info("%s", f"{url} not found in cache. Adding it to {self.bucket} with key {key}.")
+            with requests.get(url, stream=True) as f:
+                content = b''
+                for chunk in f.iter_content(chunk_size=None):
+                    content = content + chunk
+                    if len(content) > max_size:
```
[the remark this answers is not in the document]
Bento007: changed to:
```python
                for chunk in resp_obj.iter_content(chunk_size=None):
                    #  check if max_size exceeded before storing content to avoid storing large chunks
                    if len(content) + len(chunk) > self.max_size:
                        raise SizeLimitError("{url} not cached. Contents have exceed {self.max_size} Bytes.")
                    content.extend(chunk)
```

dss/storage/validator.py
```diff
@@ -1,23 +1,91 @@
...
+                        self.logger.warning("{url} not cached. Contents have exceed {max_size} Bytes.")
```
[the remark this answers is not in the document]
Bento007: will raise

dss/storage/validator.py
Bento007: using your suggestion

tests/test_s3urlcache.py
Bento007: It's required for all test to avoid using any index naming collisions when running test in parallel.

Bento007: it's not used so  removed it.

tests/test_s3urlcache.py
Bento007: fixed

dss/storage/validator.py
kislyuk: My bad, got it.

dss/storage/validator.py
kislyuk: Why not make it both, and simplify the interface?
```
@functools.lru_cache()
def caching_get_url(url):
    try:
        requests.get(... from cache ...)
    except ...:
        requests.get(... from original url ...)
        # save to S3
    return content
```

hannes-ucsc: It would mix two separate concerns. A user of the S3 Cache might not want the memory overhead that `lru_cache` entails. If they do want in-memory caching they can easily write the above function themselves.

Furthermore, In this particular use case of the cache–resolving JSON schema URLs–an in memory cache is already provided by the RefResolver as @Bento007 pointed out to me yesterday.

kislyuk: I still think it would be better to use a single function with an in-memory LRU cache provided by the standard library.

kislyuk: Marking as approved to unblock the PR since it looks OK otherwise.

dss/storage/validator.py
Bento007: how about this
```python
    cache = S3UrlCache(logger)
    request_json = lambda url: json.loads(cache.resolve(url).decode("utf-8"))
```

tests/test_s3urlcache.py
Bento007: added

tests/test_s3urlcache.py
Bento007: using local server

tests/test_s3urlcache.py
Bento007: added

Bento007:
relates to #608

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
+        self.blobstore, _, self.bucket = Config.get_cloud_specific_handles(Replica.aws)
+        self.logger = logger
 
-def read_json_url(link):
-    gcontext = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
-    with urlopen(link, context=gcontext) as f:
-        return json.load(f)
+    def resolve(self, url: str) -> bytearray:
+        """
+        Requests the contents of a URL by first checking for the contents in an s3 bucket. If not present, the contents
```
hannes-ucsc: S3

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
+        are retrieved from the URL and stored in s3.
```
hannes-ucsc: S3

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
 
+        :param url: The JSON data from which to remove fields.
+        """
+        key = self._create_key(url)
 
-def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
+        # if key in s3 bucket return value from there.
```
hannes-ucsc: S3

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
+        except BlobNotFoundError:
+            with requests.get(url, stream=True) as resp_obj:
+                content = bytearray()
+                for chunk in resp_obj.iter_content(chunk_size=self.chunk_size):
+                    #  check if max_size exceeded before storing content to avoid storing large chunks
+                    if len(content) + len(chunk) > self.max_size:
+                        raise SizeLimitError(url, self.max_size)
+                    content.extend(chunk)
+            self._upload_content(key, url, content)
+            self.logger.info("%s", f"{url} not found in cache. Added it to {self.bucket} with key {key}.")
```
hannes-ucsc: I would move this to be the first line in the except and s/Added/Adding. That stack trace resulting from an exception will be preceded by the log message, giving additional context to the stack trace.

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
+        return content
+
+    def invalidate(self, url):
+        self.blobstore.delete(self.bucket, self._create_key(url))
```
hannes-ucsc: This will raise an exception if the URL isn't currently cached. I think it shouldn't, i.e. this method should be idempotent. The important post condition is that the cache does not contain the URL and this function should not fail if that post condition is already satisfied. Don't forget to add a test for idempotence.

And while I asked proposed the name `invalidate` in an earlier review, I think `evict` would be a better name. I got the two confused. Eviction is the term commonly used.

https://www.google.com/search?q=cache+eviction

hannes-ucsc: Consider adding a doc string.

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
+    def reverse_key_lookup(self, key):
+        return self.blobstore.get_user_metadata(self.bucket, key)
+
+    @staticmethod
+    def _create_key(url: str) -> str:
+        return 'cache/' + sha1(url.encode("utf-8")).hexdigest()
+
+    def _upload_content(self, key: str, url: str, content: bytearray) -> None:
+        meta_data = {
+            "Metadata": {'DSS_CACHED_URL': url},
```
hannes-ucsc: Why is this upper case while the test uses lower case for this metadata key?

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
+                    content.extend(chunk)
+            self._upload_content(key, url, content)
+            self.logger.info("%s", f"{url} not found in cache. Added it to {self.bucket} with key {key}.")
+        return content
+
+    def invalidate(self, url):
+        self.blobstore.delete(self.bucket, self._create_key(url))
+        self.logger.info(f"{url} removed from cache in {self.bucket}.")
+
+    def reverse_key_lookup(self, key):
```
hannes-ucsc: Consider making this protected. PyCharm will complain at the call site in the test but that's ok. The reason I'm asking this is that you don't want regular users of this class to be confused by this method.

Last but not least, this function should return the URL only, not the entire metadata dict. That way you avoid leaking that impl detail to the test.

dss/storage/validator.py
```diff
@@ -1,23 +1,106 @@
...
+        return self.blobstore.get_user_metadata(self.bucket, key)
+
+    @staticmethod
+    def _create_key(url: str) -> str:
```
hannes-ucsc: Nit: I'm on the fence about this name. "Create" sounds like something is actually being persisted. I would probably name this `_key_from_url()` or `_url_to_key()`. Up to you.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def setUp(self):
+        self.urls_to_cleanup = []
+        self.cache = S3UrlCache(logger)
+
+    def tearDown(self):
+        self.delete_cached_urls(self.urls_to_cleanup)
+
+    def test_store_in_cache(self):
+        "The url contents are stored in s3 and the contents returned, when requested url is not found in cache."
```
hannes-ucsc: Triple quotes.

S3

hannes-ucsc: URL

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+            url_content = self.cache.resolve(url)
+
+        original_data = RANDOMDATA[:KB]
+        self.assertEqual(len(url_content), KB)
+        self.assertEqual(url_content, original_data)
+        self.assertTrue(log_monitor.output[0].endswith(f"{url} not found in cache. Added it to "
+                                                       f"{self.test_bucket} with key {url_key}."))
+
+    def test_retrieve_from_cache(self):
+        "Stored url contents is retrieved from s3, when a cached url is requested"
```
hannes-ucsc: Triple quotes

S3

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+        cls.blobstore, _, cls.test_fixture_bucket = Config.get_cloud_specific_handles(replica)
+        Config.set_config(BucketConfig.TEST)
+        cls.test_bucket = replica.bucket
+
+    def setUp(self):
+        self.urls_to_cleanup = []
+        self.cache = S3UrlCache(logger)
+
+    def tearDown(self):
+        self.delete_cached_urls(self.urls_to_cleanup)
```
hannes-ucsc: This is always called with `self.urls_to_cleanup` as the argument, so I would actually eliminate the argument.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+    @classmethod
+    def setUpClass(cls):
+        replica = Replica.aws
+        Config.set_config(BucketConfig.TEST_FIXTURE)
+        cls.blobstore, _, cls.test_fixture_bucket = Config.get_cloud_specific_handles(replica)
+        Config.set_config(BucketConfig.TEST)
+        cls.test_bucket = replica.bucket
+
+    def setUp(self):
+        self.urls_to_cleanup = []
```
hannes-ucsc: A set would be a more appropriate data structure for this.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+            self.cache.invalidate(url)
+            # Check that it was added to cache again.
+            self.cache.resolve(url)
+        self.assertTrue(log_monitor.output[0].endswith(f"{url} not found in cache. Added it to "
+                                                       f"{self.test_bucket} with key {url_key}."))
+        self.assertTrue(log_monitor.output[1].endswith(f"{url} removed from cache in {self.test_bucket}."))
+        self.assertTrue(log_monitor.output[2].endswith(f"{url} not found in cache. Added it to "
+                                                       f"{self.test_bucket} with key {url_key}."))
+
+    def delete_cached_urls(self, urls):
```
hannes-ucsc: This can be protected.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+start_verbose_logging()
+
+KB = 1024
+MB = KB**2
+
+RANDOMDATA = os.urandom(MB)
```
hannes-ucsc: This is evaluated when the module is first imported. Typically you want to avoid interacting with the operating system at module load time. So I think this should be initialized in `setupModule()`. And since its not exactly a constant, its name should be lower case.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+    _response_code = 200
+
+    def do_GET(self):
+        _, length = self.path.split('/')
+
+        self.send_response(self._response_code)
+        self.send_header("content-length", length)
+        self.end_headers()
+
+        length = int(length)
```
hannes-ucsc: To be super-correct, this should be converted to int right after it is extracted from the path. Even if that means converting it back to `str` when calling `send_header`.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+        url_content = self.cache.resolve(url)
+        original_data = RANDOMDATA[:size]
+        self.assertEqual(url_content, original_data)
+
+    def test_stored_url_metadata(self):
+        url = f"{HTTPInfo.url}/{KB}"
+        self.urls_to_cleanup.append(url)
+        url_key = S3UrlCache._create_key(url)
+        self.cache.resolve(url)
+        with self.subTest("check dss_cached_url"):
```
hannes-ucsc: Nit: Consider adding empty line above this one. Or removing the one before the next `with`statement.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+        url = f"{HTTPInfo.url}/0"
+        self.urls_to_cleanup.append(url)
+        url_content = self.cache.resolve(url)
+        self.assertEqual(len(url_content), 0)
+
+    def test_chunked_content(self):
+        size = KB * 10
+        url = f"{HTTPInfo.url}/{size}"
+        self.urls_to_cleanup.append(url)
+        self.cache.max_size = MB
```
hannes-ucsc: For readability, I would change this to `1 * MB` and `1 * KB`

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,189 @@
...
+from dss.storage.validator import SizeLimitError
+from dss.util import networking
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+start_verbose_logging()
+
+KB = 1024
```
hannes-ucsc: Nit: To be precise, this is a Kibibyte, not a Kilobyte, so the appropriate unit abbreviation is KiB.

dss/storage/validator.py
Bento007: Changed

dss/storage/validator.py
Bento007: changed

dss/storage/validator.py
Bento007: changed

dss/storage/validator.py
Bento007: I like it

dss/storage/validator.py
Bento007: blobstore returns False if nothing was deleted or something else if something was deleted. No exception should be raised.

Bento007: added a doc string

dss/storage/validator.py
Bento007: changed to only return url and making protected.

dss/storage/validator.py
Bento007: _url_to_key sounds acceptable.

dss/storage/validator.py
Bento007: s3 lowercases it when it stores it. There are other places in the code where we have the uppercase metadata stored in s3. I'll lowercase it for code to test consistency

tests/test_s3urlcache.py
Bento007: I'll make the change.

> Typically you want to avoid interacting with the operating system at module load time. 

Do you have an example on why this bad?

tests/test_s3urlcache.py
Bento007: changed to set()

tests/test_s3urlcache.py
Bento007: will do

tests/test_s3urlcache.py
Bento007: fixed

tests/test_s3urlcache.py
Bento007: fixed

tests/test_s3urlcache.py
Bento007: fixed

tests/test_s3urlcache.py
Bento007: fixed

tests/test_s3urlcache.py
Bento007: do you really need to worry about protecting functions with testcases? Fixed.

tests/test_s3urlcache.py
Bento007: fixed

tests/test_s3urlcache.py
hannes-ucsc: Good question. Consider a module `foo.py` with 

```
import os
os.system('rm -rf /')
def f():
    …
```

Simply importing `foo` will trigger the side-effect. Granted, this is an extreme (and malicious) side-effect, but it illustrates that the importing module has no control over the side-effects in the imported module, and no way to opt out. If all side-effects are wrapped in functions or classes, they lose the "side-" aspect and the importing module is given back control over what to invoke. The damage this anti-pattern can do depends on how drastic the side-effects are. `os.urandom(1MB)` isn't the end of the world. But there used to be a time where Windows didn't have a `os.urandom` implementation, so a user would not have been able to import your module at all on Windows. 

This is the same reason that Python scripts (and modules invoked via python -m) should have `if __name__ == `__main__:` guarding the module-level statements. It allows the script to be run from the command line AND to be imported as a module. When imported, the `if` statement prevents the execution of the module-level statements.

I'm not alone:

https://news.ycombinator.com/item?id=7536246

tests/test_s3urlcache.py
hannes-ucsc: It makes it more obvious that this is a utility method and not a test case. We do have a few abstract test cases (test_indexer) where this is naming convention is also useful.

dss/storage/validator.py
```diff
@@ -1,23 +1,109 @@
...
+        return content
+
+    def evict(self, url: str):
+        '''
+        Removes the cached URL content from S3.
+        :param url: the url for the content to removed from S3'''
+        self.blobstore.delete(self.bucket, self._url_to_key(url))
+        self.logger.info(f"{url} removed from cache in {self.bucket}.")
+
+    def _reverse_key_lookup(self, key: str) -> dict:
```
hannes-ucsc: s/dict/str, right? Why is mypy not compaining about this?

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,196 @@
...
+        self.assertEqual(url_content, original_data)
+
+    def test_stored_url_metadata(self):
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+        url_key = S3UrlCache._url_to_key(url)
+        self.cache.resolve(url)
+
+        with self.subTest("check dss_cached_url"):
+            metadata = self.cache._reverse_key_lookup(url_key)
```
hannes-ucsc: s/metadata/cached_url

The term metadata leaks an implementation detail.

tests/test_s3urlcache.py
```diff
@@ -0,0 +1,196 @@
...
+            contentType = self.blobstore.get_content_type(self.test_bucket, url_key)
+            self.assertEqual(contentType, "application/octet-stream")
+
+    def test_evict(self):
+        url = f"{HTTPInfo.url}/{KiB}"
+        self.urls_to_cleanup.add(url)
+        url_key = S3UrlCache._url_to_key(url)
+
+        with self.assertLogs(logger, "INFO") as log_monitor:
+            # Check when deleting url that is not cached.
```
hannes-ucsc: The proof that the URL is not cached is a bit weak. It would be a stronger proof of `evict`'s idempotence to call it twice in a row, without `resolve` in between. Also, log messages don't prove much, only that the log message was emitted, but not that the action was actually performed.

IOW, I would write this test as follows

```
resolve()
assert present
evict()
assert missing
evict()
assert missing
```

For the asserts I would add a `S3UrlCache.contains(url)` method that does a cheap HEAD request against the key. That method might come in handy outside of tests, too.

dss/storage/validator.py
Bento007: fixed

tests/test_s3urlcache.py
Bento007: fixed

tests/test_s3urlcache.py
Bento007: fixed

PR #806, opened 2017-12-22
Validator KeyError Fix (fixes #804)

### LINKED ISSUE #804: Indexing error: File "...validator.py", line 32, in remove_json_fields current.pop(field) KeyError: '^characteristics_.*$'
Many of the following errors are being logged by humancellatlas `dss-index-dev`, which are preventing successful indexing. This particular one was logged at 2017-12-22 15:02:45 UTC, although there are many other occurrences. data-store master is current at commit e32cb8b7f990422d5347f0cf1cc2153a30bc724a 

```
'^characteristics_.*$': KeyError
File "/var/task/domovoi/app.py", line 127, in __call__
result = handler(event, context)
File "/var/task/app.py", line 36, in dispatch_gs_indexer_event
GCPIndexHandler.process_new_indexable_object(gs_event['data'], app.log)
File "/var/task/domovoilib/dss/events/handlers/index.py", line 88, in process_new_indexable_object
cls._process_new_indexable_object(Replica.gcp, key, logger)
File "/var/task/domovoilib/dss/events/handlers/index.py", line 20, in _process_new_indexable_object
cls._index_and_notify(replica, identifier, logger)
File "/var/task/domovoilib/dss/events/handlers/index.py", line 29, in _index_and_notify
doc = BundleDocument.from_replica(replica, bundle_fqid, logger)
File "/var/task/domovoilib/dss/storage/index_document.py", line 76, in from_replica
self['files'] = self._read_file_infos(blobstore, bucket_name)
File "/var/task/domovoilib/dss/storage/index_document.py", line 147, in _read_file_infos
scrub_index_data(index_files, str(self.fqid), self.logger)
File "/var/task/domovoilib/dss/storage/validator.py", line 75, in scrub_index_data
remove_json_fields(index_data, path, fields)
File "/var/task/domovoilib/dss/storage/validator.py", line 32, in remove_json_fields
current.pop(field)
KeyError: '^characteristics_.*$'
```

Bento007:
Fixes #804 

Fix for Validator KeyError

### Test plan
- Added testing for json schemas with patternProperties.
- Added testing for extra fields in different json files.

- Run populate on test fixtures.

### Release notes
- Added EXTRA_FIELD_ERROR_REGEX and adjust regular express to extract field names from both error messages found in jsonschema._validators.additionalProperties.
- Added characteristics field to ./tests/fixtures/datafiles/indexing/bundles/v3/smartseq2/paired_ends/project.json.
- Fixes #804

### DIFF
--- a/dss/storage/validator.py
+++ b/dss/storage/validator.py
@@ -44,7 +44,6 @@ DSS_Draft4Validator = validators.create(
 )
 
 resolver = validators.RefResolver(referrer='', base_uri='')
-EXTRA_FIELD_ERROR_REGEX = re.compile(r"(?<!regexes: )'([^']+)',?")
 
 
 def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -> list:
@@ -65,7 +64,9 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
                     path = [document, *error.path]
                     #  Example error message: "Additional properties are not allowed ('extra_lst', 'extra_top' were
                     #  unexpected)" or "'extra', does not match any of the regexes: '^characteristics_.*$'"
-                    fields_to_remove = (path, [field for field in EXTRA_FIELD_ERROR_REGEX.findall(error.message)])
+                    fields_to_remove = (path,
+                                        [field for field in re.findall(r"(?<!regexes: )'([^']+)',?", error.message)]
+                                        )
                     extra_fields.append(fields_to_remove)
         else:
             extra_documents.append(document)
--- a/dss/storage/validator.py
+++ b/dss/storage/validator.py
@@ -44,6 +44,7 @@ DSS_Draft4Validator = validators.create(
 )
 
 resolver = validators.RefResolver(referrer='', base_uri='')
+EXTRA_FIELD_ERROR_REGEX = re.compile(r"(?<!regexes: )'([^']+)',?")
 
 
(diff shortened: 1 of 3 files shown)


dss/storage/validator.py
```diff
@@ -44,6 +44,7 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]):
 )
 
 resolver = validators.RefResolver(referrer='', base_uri='')
+EXTRA_FIELD_ERROR_REGEX = re.compile(r"(?<!regexes: )'([^']+)',?")
```
Bento007: This is the regex i'm using to extract the fields `r"(?<!regexes: )'([^']+)',?"`. If only only alphanumeric fields are allowed I could simplify it. Though there is the possibility the pattern defined in `patternProperty` could still match as a field.

hannes-ucsc: Nit: I would not extract the regex into a variable unless it is used in more than one place. Keep regexes as close to the place of use as possible. Especially the negative look behind makes more sense when the comment is right above it.

hannes-ucsc: Can you go over the json schema validator code to look for other types of error message that we haven't seen yet in the wild but that we might want to handle?

Bento007: My original reason for doing it was to avoid the cost of compiling the regex everytime `scrub_index_data` was called. The python `re` library caches the patterns so it actually isn't a problem. I'll move it back into the function.

hannes-ucsc: Good point. As long as it is right before the function it should be ok.

PR #807, opened 2017-12-22
Make Validator More Robust

Bento007:
Here is another way we could get the fields. It might cost a bit more since `jsonschema._utils.find_additional_properties` is called twice per `document`. The first call is by `_validators.additionalProperties`, the second time is by us.
 
### Release notes
Using `jsonschema._utils.find_additional_properties` to find extra fields over `re.findall`. This is less prone to breaking since extra fields are extracted by comparing the `json_data` with the `schema` and not the `error.message`.

### DIFF
--- a/dss/storage/validator.py
+++ b/dss/storage/validator.py
@@ -65,8 +65,8 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
                     #  Example error message: "Additional properties are not allowed ('extra_lst', 'extra_top' were
                     #  unexpected)" or "'extra', does not match any of the regexes: '^characteristics_.*$'"
                     fields_to_remove = (path,
-                                        [field for field in re.findall(r"(?<!regexes: )'([^']+)',?", error.message)]
-                                        )
+                                        [field for field in _utils.find_additional_properties(error.instance,
+                                                                                              error.schema)])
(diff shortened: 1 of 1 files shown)


dss/storage/validator.py
```diff
@@ -65,8 +65,8 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
                     #  Example error message: "Additional properties are not allowed ('extra_lst', 'extra_top' were
                     #  unexpected)" or "'extra', does not match any of the regexes: '^characteristics_.*$'"
                     fields_to_remove = (path,
-                                        [field for field in re.findall(r"(?<!regexes: )'([^']+)',?", error.message)]
-                                        )
+                                        [field for field in _utils.find_additional_properties(error.instance,
```
hannes-ucsc: This would result in the scrubbing of fields that are explicitly allowed by `patternProperties`.

It also calls `find_additional_properties` twice: once here and once in the built-in `additionalProperties` validator. 

If this change passes the test then the tests actually need to be updated to make sure that a field mentioned in `patternProperties` does not get scrubbed! And that needs to be done in another PR, independently of this exploration.

hannes-ucsc: Ignore second paragraph. You already mentioned that.

PR #810, opened 2017-12-23
Timeout mechanism using signals.

xbrianh:
Context manager useful for applying time constraints to lambda executions.

### Test plan
Unit tests.

### DIFF
--- a/dss/stepfunctions/visitation/timeout.py
+++ b/dss/stepfunctions/visitation/timeout.py
@@ -1,0 +1,26 @@
+import signal
+from contextlib import AbstractContextManager
+
+class Timeout(AbstractContextManager):
+    def __init__(self, seconds_remaining: int) -> None:
+        self.did_timeout = False
+        self.seconds_remaining = seconds_remaining
+
+    def __enter__(self):
+        def _timeout_handler(signum, frame):
+            raise TimeoutError("time's up!")
+
+        signal.signal(signal.SIGALRM, _timeout_handler)
+        signal.alarm(self.seconds_remaining)
+
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        signal.signal(signal.SIGALRM, signal.SIG_DFL)
+        signal.alarm(0)
+
+        if exc_type == TimeoutError:
+            self.did_timeout = True
+            return True
+
+        return None is exc_type
--- a/tests/test_visitation.py
+++ b/tests/test_visitation.py
@@ -19,6 +19,7 @@ from dss.stepfunctions import step_functions_describe_execution
 from dss.stepfunctions.visitation import implementation
 from dss.stepfunctions.visitation.integration_test import IntegrationTest
 from dss.stepfunctions.visitation import registered_visitations
+from dss.stepfunctions.visitation.timeout import Timeout
 
 from tests import infra  # noqa
 from tests.infra import get_env, testmode  # noqa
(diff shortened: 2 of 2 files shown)


ttung:
I recommend using the TimedThread abstraction, since threading has fewer edge cases and it also provides a clean way to freeze state cleanly.

hannes-ucsc:
@ttung, we (@xbrianh and I) went back and forth between three mechanisms: 

1) a simple time check in the loop (my proposal)
2) a timed thread (@xbrianh's original approach, similar to your more generalized solution)
3) alarm signal (originally proposed by(@kislyuk)

1) is simple but does not work when indexer gets stuck on socket I/O somewhere causing all progress made in one worker execution to be lost

2) always terminates and does not lose progress on a timeout but the timed thread keeps running while the main thread has determined that it should time out. The state snapshots performed by TimeThread take care of that, but only for internal state. The indexer modifies external state, i.e. Elasticsearch.

3) is pretty much guaranteed to terminate without cooperation of the code to be timed out. We verified that the alarm signal is delivered during a blocking socket I/O call. Raising an exception on timeout is a good way to give that code a chance to clean-up gracefully (using `finally:`) without requiring it to cooperate explicitly, e.g. check for a flag etc.

ttung:
> Solution 3 is pretty much guaranteed to terminate without cooperation of the code to be timed out. We verified that the alarm signal is delivered during a blocking socket I/O call. Raising an exception on timeout is a good way to give that code a chance to clean-up gracefully (using finally:) without requiring it to cooperate explicitly, e.g. check for a flag etc.

I don't know if you want your design to be dependent on the ability to clean up gracefully.  You can't assume you will actually get time to run.  Also, I have a general aversion to signals because:

1. threads and signals interact in hilarious though mostly tragic ways.  you might say, "but I don't use any threads" but it turns out a fair number of python libraries do spawn threads under the covers.
2. other code can insert signal handlers or set alarms which will trigger your code in strange and unexpected ways.

I don't think you guys need to reconsider the approach before the Q1 meeting, but I would suggest doing so afterwards.

hannes-ucsc:
> threads and signals interact in hilarious though mostly tragic ways

Could you give any evidence for that? From the official documentation, the interaction between the two is pretty straight forward. 

https://docs.python.org/3.6/library/signal.html#signals-and-threads

So I won't say "but I don't use threads" but I'll say "I can guarantee that the signal handler is installed in the main thread, which is the same thread that needs to be timed out."

> other code can insert signal handlers or set alarms which will trigger your code in strange and unexpected ways.

That could be a concern but it seems improbable to me. Due to the singleton-nature of signal handlers, any good library author should know not to use them.

> I don't know if you want your design to be dependent on the ability to clean up gracefully

We're not depending on it. We're merely making graceful clean-up very likely to occur and and to succeed. We leave enough of a time cushion for graceful cleanup to happen. As long as it doesn't take too long, it will work. It's a best effort.

ttung:
> Could you give any evidence for that? From the official documentation, the interaction between the two is pretty straight forward.

> https://docs.python.org/3.6/library/signal.html#signals-and-threads

> So I won't say "but I don't use threads" but I'll say "I can guarantee that the signal handler is installed in the main thread, which is the same thread that needs to be timed out."

Oh cool, I did not know this.  Most of my signals programming is in C, and there, signals can be delivered to any thread that doesn't ignore the signal.

ttung:
I'm still wary of this, but a lot less wary now. :)

PR #814, opened 2017-12-27
Aggregate output by visitation workers, not discard it

hannes-ucsc:
This improves on the fix in 10c6be3

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,7 @@
 [report]
 # Regexes for lines to exclude from consideration
 exclude_lines =
+    # 23 is ASCII for pound sign
+    \x23 no coverage
     raise AssertionError
     raise NotImplementedError
--- a/dss/stepfunctions/visitation/__init__.py
+++ b/dss/stepfunctions/visitation/__init__.py
@@ -1,5 +1,6 @@
+from typing import Sequence, Any, Union
+
 import copy
-import typing
 from uuid import uuid4
 from enum import Enum, auto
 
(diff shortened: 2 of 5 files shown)


xbrianh:
This feature was consciously excluded due to size constraints on the state machine. It is innocuous for simple counting, such as in the integration test, but for other purposes it may fail in perplexing ways. For instance, state size may be OK for k workers, but exceed limits for k+1 workers.

IMO, users should be discouraged from collecting results in this manner, especially at the end of a parallel step where state size can quickly become large. For this architecture, if result aggregation is desired, it should be handled through persistent storage services such as S3.

Or, if this feature is necessary, there should be an explicit size limit on `work_result`, and `finalize_walker` should be preserved.

hannes-ucsc:
Let's not have perfect be the enemy of the good. I think 10c6be3 is an easy fix at the cost of usability. This is a really useful feature, especially for reindexing where there is currently no simple way to get the aggregate number of bundles reindexed. The current state size limit is 32k, so each of the 32 workers can return 1k which is plenty for small things such as status and counters, for which external persistence would be complete overkill.

I'll take a stab at limiting the size of work_result.

Can you elaborate on what you mean by "finalize_walker should be preserved".

xbrianh:
What’s this perfect vs good stuff?

Sorry, I see that finalize_walker is still ther, just removed from integration test.

Rather than size limits, perhaps a per-walker count that will be populated after the parallel step. This is essentially making work_result an integer, and the logic in _aggregate can be moved into implementation.py

hannes-ucsc:
We need three integers, or at least two, for reindex:

https://github.com/HumanCellAtlas/data-store/pull/797/commits/bdcdcf230618741f5bdbc309c0bd6d86e6f7342a

Furthermore, there is no reason to be that restrictive, as I mentioned we can afford close to 1k per worker.

I also thought some more about applying an explicit size limit on work_result per worker. I don't think it is necessary. The best it can do is fail at run time, which is exactly what the AWS state machinery does.

We are the authors of the visitation sub classes, and we decide the number of workers. IOW, we are in full control of the two only factors that contribute to the limit being exceeded. And if the limit is exceeded, the failure is deterministic and obvious. I think this is an acceptable compromise as it is.

https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good

hannes-ucsc:
Remaining concerns tracked in #819.

ttung:
What's the behavior of stepfunctions if the size limit is exceeded?  Is it any worse than us terminating the code ourselves?

PR #815, opened 2017-12-28
Decorate remaining undecorated tests with `testmode`

hannes-ucsc:


ttung:
I think some classes were just decorated entirely. Do the method-level ones need to be there still?

hannes-ucsc:
If in doubt, we should stick with method-level decorations. Even if all methods in class are decorated, that could be incidental. If a test class is inherently either integration or standalone, I made a class level decoration.

hannes-ucsc:
I captured the concerns in #820 and would like to go ahead with this as is.

ttung:
I clarified my viewpoint in #820, but I'll stamp this PR and let you decide how you want to do it. :)

tests/test_smoketest.py
```diff
@@ -42,7 +43,7 @@ def run(command, runner=check_call, **kwargs):
         parser.exit(RED(f'{parser.prog}: Exit status {e.returncode} while running "{command}". Stopping.'))
 
 
-@infra.testmode.integration
+@testmode.standalone
```
Bento007: shouldn't this be integration

ttung: great catch. :)

hannes-ucsc: Oops. Fixed.

PR #817, opened 2017-12-28
Fix: test for all-versions tombstone is ineffective

hannes-ucsc:


mweiden:
@hannes-ucsc thanks for this

Issue #816, opened 2017-12-28
Prevent zombie bundles


hannes-ucsc:
Two cases come to mind where a deleted bundle version may become visible, despite an existing tombstone overriding it.

* A bundle version is submitted after a UUID-only (aka unqualified, aka unversioned) tombstone is created. The bundle version will be not be tomstoned since the tombstones for all matching bundles are only created once, during the indexing of the tombstone.

* During index REPAIR some time may pass between the processing of a bundle (which will add the bundle to the index) and the processing of the tombstone. During that time, the original bundle will be visible in the index.

I propose that we modify the indexer to always check for a tombstone before indexing a bundle and then Do The Right Thing™.

PR #818, opened 2017-12-28
Refactor bundle deletion test into reusable methods (connected to #816)

hannes-ucsc:
This splits the two cases (delete of a single version and of all versions) into separate test cases. It also reorders the creation of the tombstone to happen after the bundle is created. The original order would interfere with testing of a fix for #816. Having separate methods will make writing tests for #816 easier.


hannes-ucsc:
I'm retracting this. It should be part of the PR that addresses #816.

hannes-ucsc:
… see #823.

PR #824, opened 2018-01-03
Reindexing 3/3: Integrate visitation code, add admin CLI to trigger REPAIR and VERIFY

hannes-ucsc:
This adds a Reindex visitation which invokes the indexer for every visited bundle and tombstone. It adds an admin lambda and a CLI to invoke it. The admin CLI has sub-commands for triggering index verification (aka VERIFY) and in-place index repair (aka REPAIR).


daemons/dss-admin/app.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    signature = inspect.signature(f)
+    kwargs = {name: kwargs.get(name, param.default) for name, param in signature.parameters.items()}
+    return f(**kwargs)
+
+
+class DSSAdmin(domovoi.Domovoi):
+    def __call__(self, options, context):
+        target_class_name = options['target'].capitalize() + 'Target'
+        target_class = globals()[target_class_name]
```
xbrianh: Probably not much opportunity for hijinks here, but I'd be more comfortable with an explicit list in lieu of `globals()`

hannes-ucsc: Why are you uncomfortable with this approach?

hannes-ucsc: Adding assert.

PR #827, opened 2018-01-03
Prevent test output from being delayed by `make`

hannes-ucsc:
When the `-O` make option is used in conjunction with `-j`, the output of an entire recipe is buffered and printed atomically when the recipe finishes. I have the suspicion that this 1) interferes with Travis' build output handler and 2) causes truncated build logs. This PR

* prefixes every line in the test output with the shell's PID which is going to be unique per line in the recipe. The test recipes only consist of one line, so each test module's output will be prefixed with a different number. A stable, lexical sort of the build output will produce a result resembling that of `-O`.

* passes `-v` to unittest for Travis builds, making the test output more lively. A nice side effect of `-v` is that it also prints which individual test methods were skipped.

* removes `-O` for test-related targets. I originally thought that `-O=line` would sync around every line of output but that's not the case. It syncs around the output produced by a line in the recipe.

* introduces the `DSS_UNITTEST_OPTS` variable which can be set to pass additional argument to Python's unittest module when running the tests (run `python -m unittest -h` to see which options are supported)

I propose trying this out for a few days (on master) and backing it out if it does give us any insights. I'm particularly interested in diagnosing the failed builds that seem to have their logs truncated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,10 @@ install:
 - dpkg -x make*.deb make4
 - export PATH=$(pwd)/make4/usr/bin:$PATH
 script:
+- DSS_UNITTEST_OPTS="$DSS_UNITTEST_OPTS -v"
 - set -eo pipefail
-- if [[ ! $TRAVIS_DSS_INTEGRATION_MODE ]]; then make -j4 -O test; fi
-- if [[ $TRAVIS_DSS_INTEGRATION_MODE ]]; then make -j4 -O integration_test; fi
+- if [[ ! $TRAVIS_DSS_INTEGRATION_MODE ]]; then make -j4 test; fi
+- if [[ $TRAVIS_DSS_INTEGRATION_MODE ]]; then make -j4 integration_test; fi
 - if [[ $TRAVIS_BRANCH == staging ]]; then source environment.staging; fi
 after_success:
 - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


ttung:
Are the truncated build logs for failures only?  Do we get the travis message saying that there was a timeout?

hannes-ucsc:
> Are the truncated build logs for failures only?

I think so.

> Do we get the travis message saying that there was a timeout?

No.

ttung:
If this is a travis-only issue, I would suggest that we _don't_ turn `-v` for console builds.  I don't mind leaving the PID change in there, if you insist, though.

I personally think it's much more likely that we should just have a background process emit output regularly.

hannes-ucsc:
I'll try to enable `-v` only for Travis builds. I personally like `-v` but I understand that some people don't want to see that level of verbosity for their local builds.

kislyuk:
I think we should drop the `| sed -e "s/^/$$$$ /"`.

hannes-ucsc:
If we remove them, it will be very hard to correlate the output with individual tests. Can't we just try this out and see how it feel about in a few days. Consider it an experiment.

hannes-ucsc:
This built successfully several times in a row now. The builds of other branches that included my reindexing changes had a success rate of about 50%. So I am very confident that this will fix the issue with builds being aborted during test_indexer.py.

PR #828, opened 2018-01-04
Add the permissions required for the http endpoint to invoke step functions

ttung:


ttung:
If this is good, go ahead and merge this, since it's just based on master.

PR #830, opened 2018-01-04
Fix smoketest imports

ttung:
importing tests.* should happen after the path hack.

@amork noticed that `make smoketest` fails.


tests/test_smoketest.py
```diff
@@ -6,12 +6,12 @@
...
 from subprocess import check_call, check_output, CalledProcessError
 
-from tests.infra import testmode
-
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
 from dss.api.files import ASYNC_COPY_THRESHOLD
+from tests.infra import testmode
+
```
amork: extra empty line

PR #831, opened 2018-01-04
use dependent target to invoke `make smoketest`

ttung:


hannes-ucsc:
Commit titles should start in capital letter. I fixed that when I merged.

A little explanation would have been helpful. I had to check this out locally to find out what the intent was.

hannes-ucsc:
Didn't see that the base of this PR was set to some other branch, not master. I just cherry-picked the commit to master and pushed manually. If possible, I think we should base our PR branches on master.

Issue #801, opened 2017-12-21
Make the bundle UUID more available for searching (e.g. include in bundle manifest)


mikebaumann:
Searching based on bundle UUID (and version) should be fast and easy. In addition to simply verifying the bundle exists in the DSS, now that output format `raw` is available, it is a convenient way to view all the indexed metadata in the bundle. 

Currently, however, the bundle UUID is not included in the bundle manifest, nor is it explicitly added to the index document. **Is there a reason why the bundle UUID is not included in the bundle manifest, along with the bundle version, which is currently included in the bundle manifest?** It seems having the content of the bundle manifest being self-identifying, with both the bundle UUID and version, would be useful in multiple use cases, not just for search.

It is currently possible to query for a given bundle UUID and version using the Elasticsearch Ids query, but that is crufty, awkward and depends on internal implementation details.

```
{
    "query": {
        "ids" : {
            "values" : ["9f61dcee-13d0-4b9f-8dd4-b80e53e4b895.2017-12-21T174128.247427Z"]
        }
    }
}
```

If there is a good reason for not including the bundle UUID in the bundle manifest, then the bundle  UUID should be explicitly added to the index document content. Yet, I think the former is preferable.

Bento007:
@mikebaumann what is left to be done?

hannes-ucsc:
We should use the `(fixes #123)` or `(resolves #123)` tag in commit (!) titles, so GH closes the ticket as soon as the commit lands on master.

hannes-ucsc:
I think 1c21d736d58d23fa4a2c6fe424cd4ef701bd4956 addresses this.

PR #832, opened 2018-01-05
Include bundle UUID in index document to enable searching by bundle UUID

mikebaumann:
The UUID is added to the index document for both regular bundles and bundle tombstones.

Connects to #801 

### Test plan
Test search by bundle UUID in deployed system.

### DIFF
--- a/dss/storage/index_document.py
+++ b/dss/storage/index_document.py
@@ -111,6 +111,7 @@ class BundleDocument(IndexDocument):
         self['manifest'] = self._read_bundle_manifest(blobstore, bucket_name, bundle_fqid)
         self['files'] = self._read_file_infos(blobstore, bucket_name)
         self['state'] = 'new'
+        self['uuid'] = bundle_fqid.uuid
         return self
 
     @classmethod
@@ -476,7 +477,9 @@ class BundleTombstoneDocument(IndexDocument):
     def from_replica(cls, replica: Replica, tombstone_id: TombstoneID, logger):
         blobstore, _, bucket_name = Config.get_cloud_specific_handles(replica)
         tombstone_data = json.loads(blobstore.get(bucket_name, tombstone_id.to_key()))
-        return cls(replica, tombstone_id, logger, tombstone_data)
+        self = cls(replica, tombstone_id, logger, tombstone_data)
+        self['uuid'] = tombstone_id.uuid
+        return self
 
     def _list_dead_bundles(self) -> typing.Sequence[BundleDocument]:
         blobstore, _, bucket_name = Config.get_cloud_specific_handles(self.replica)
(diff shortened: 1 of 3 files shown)


dss/storage/index_document.py
```diff
@@ -476,6 +477,7 @@ class BundleTombstoneDocument(IndexDocument):
     def from_replica(cls, replica: Replica, tombstone_id: TombstoneID, logger):
         blobstore, _, bucket_name = Config.get_cloud_specific_handles(replica)
         tombstone_data = json.loads(blobstore.get(bucket_name, tombstone_id.to_key()))
+        tombstone_data['uuid'] = tombstone_id.uuid
```
hannes-ucsc: I would

```
self = cls(…)
self['uuid'] = …
```

to keep it symmetric with BundleDocument.

tests/test_indexer.py
```diff
@@ -650,7 +652,8 @@ def test_scrub_index_data(self):
         with self.subTest("with schema version"):
             'Extra fields are removed.'
             manifest = read_bundle_manifest(self.blobstore, self.test_bucket, self.bundle_key)
-            index_data = create_index_data(self.blobstore, self.test_bucket, manifest)
+            index_data = create_index_data(self.blobstore, self.test_bucket, BundleFQID.from_key(self.bundle_key).uuid,
```
hannes-ucsc: To make this less verbose, I would pass the key to `create_index_data()` and derive the uuid in that method. It would also fit naturally because the preceding parameter is the bucket.

tests/test_indexer.py
```diff
@@ -189,6 +189,8 @@ def _create_tombstone(self, tombstone_id):
         self.addCleanup(self._delete_tombstone, tombstone_id)
         sample_event = self.create_bundle_deleted_event(tombstone_id.to_key())
         self.process_new_indexable_object(sample_event, logger)
+        # Return the expected tombstone index document, which includes the bundle uuid.
+        tombstone_data['uuid'] = tombstone_id.uuid
```
hannes-ucsc: This feels a little out of place. Consider changing the assertion. The test should assert that tombstone_data is a subset of what's in the index and that it has the uuid property correctly set. We don't actually mind if there is more data in the index for a tombstone.

dss/storage/index_document.py
mikebaumann: Nice suggestion. Done.

tests/test_indexer.py
mikebaumann: This change was made. Please have a look at how the assert is being performed now.

tests/test_indexer.py
mikebaumann: Done.

mikebaumann:
Thanks for the suggestion. Title was shortened.

PR #833, opened 2018-01-05
Added checking when core exists but schema_url does not.

Bento007:
### Test plan
added a unit test to verify this behavior.

- Run reindex on master after merge.
- Run reindex after deploying to staging

### DIFF
--- a/dss/storage/validator.py
+++ b/dss/storage/validator.py
@@ -55,7 +55,7 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
         if schema_url is not None:
             try:
                 schema = read_json_url(schema_url)
-            except URLError as ex:
+            except Exception as ex:
                 extra_documents.append(document)
                 logger.warning("%s", f"Unable to retrieve schema from {document} in {bundle_id} "
                                      f"because retrieving {schema_url} caused exception: {ex}.")
@@ -70,7 +70,7 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
                     extra_fields.append(fields_to_remove)
         else:
             logger.warning("%s", f"Unable to retrieve schema_url from {document} in {bundle_id} because "
-                                 f"core.schema_url does not exists.")
+                                 f"core.schema_url does not exist.")
             extra_documents.append(document)
     if extra_documents:
         extra_fields.append(([], extra_documents))
--- a/dss/storage/validator.py
+++ b/dss/storage/validator.py
@@ -50,8 +50,8 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
     extra_fields = []
     extra_documents = []
     for document in index_data.keys():
-        core = index_data[document].get('core', {'schema_url': None})
-        schema_url = core.get('schema_url', None)
+        core = index_data[document].get('core')
+        schema_url = None if core is None else core.get('schema_url')
         if schema_url is not None:
             try:
                 schema = read_json_url(schema_url)
--- a/dss/storage/validator.py
+++ b/dss/storage/validator.py
@@ -50,25 +50,27 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
     extra_fields = []
     extra_documents = []
     for document in index_data.keys():
-        core = index_data[document].get('core', None)
-        if core is not None:
+        core = index_data[document].get('core', {'schema_url': None})
+        schema_url = core.get('schema_url', None)
+        if schema_url is not None:
             try:
-                schema = read_json_url(core['schema_url'])
+                schema = read_json_url(schema_url)
             except URLError as ex:
                 extra_documents.append(document)
-                logger.warning("%s", f"Unable to retrieve schema from url {core['schema_url']} due to exception: {ex}.")
-                continue
-
-            for error in DSS_Draft4Validator(schema, resolver=resolver).iter_errors(index_data[document]):
-                if error.validator == 'additionalProperties':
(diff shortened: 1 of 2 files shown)


dss/storage/validator.py
```diff
@@ -50,25 +50,27 @@ def scrub_index_data(index_data: dict, bundle_id: str, logger: logging.Logger) -
...
+                    if error.validator == 'additionalProperties':
+                        path = [document, *error.path]
+                        #  Example error message: "Additional properties are not allowed ('extra_lst', 'extra_top' were
+                        #  unexpected)" or "'extra', does not match any of the regexes: '^characteristics_.*$'"
+                        fields_to_remove = (path, [field for field in _utils.find_additional_properties(error.instance,
+                                                                                                        error.schema)])
                     extra_fields.append(fields_to_remove)
         else:
+            logger.warning("%s", f"Unable to retrieve schema_url from {document} in {bundle_id} because "
+                                 f"core.schema_url does not exists.")
```
kislyuk: s/exists/exist

tests/test_indexer.py
```diff
@@ -675,22 +675,39 @@ def test_scrub_index_data(self):
...
-        with self.subTest("without schema version"):
+        with self.subTest("with missing core.schema_url"):
+            doc = 'assay_json'
+            manifest = read_bundle_manifest(self.blobstore, self.test_bucket, self.bundle_key)
+            index_data = create_index_data(self.blobstore, self.test_bucket, self.bundle_key, manifest)
+            index_data['files'][doc]['core'].pop('schema_url')
+            with self.assertLogs(logger, level="WARNING") as log_monitor:
+                scrub_index_data(index_data['files'], bundle_fqid, logger)
+            self.assertRegex(log_monitor.output[0], f"WARNING:[^:]+:Unable to retrieve schema_url from {doc} in "
+                                                    f"{bundle_fqid} because core.schema_url does not exists.*")
```
kislyuk: same here

tests/test_indexer.py
```diff
@@ -675,22 +675,39 @@ def test_scrub_index_data(self):
...
+                                                    f"{bundle_fqid} because retrieving {url} caused exception: .*")
             self.verify_index_document_structure_and_content(
                 index_data,
                 self.bundle_key,
                 files=smartseq2_paried_ends_indexed_file_list,
-                excluded_files=['assay_json']
+                excluded_files=[doc]
             )
 
-        with self.subTest("without schema version"):
```
hannes-ucsc: Just a note: I don't think I'll ever warm up to the "DAMP test code" approach. Tests are code and should be DRY because they are read by humans and maintained by humans. 

The essence of this new test, the signal so to speak, is lost in the noise of all this duplicated code. I'd have to run a textual diff between the sub-tests to get an idea how this sub-test is different from its siblings and what condition it is supposed to test. The argument to `self.subTest` helps but only declares your intent, it is not immediately evident *how* that is achieved.

I'd hate to end up with a large amount of tests that are 99% identical.

So don't fix this now but consider ways to reduce the duplication in future PRs.

tests/test_indexer.py
```diff
@@ -675,22 +675,39 @@ def test_scrub_index_data(self):
...
+            self.assertRegex(log_monitor.output[0], f"WARNING:[^:]+:Unable to retrieve schema_url from {doc} in "
+                                                    f"{bundle_fqid} because core.schema_url does not exists.*")
+            self.verify_index_document_structure_and_content(
+                index_data,
+                self.bundle_key,
+                files=smartseq2_paried_ends_indexed_file_list,
+                excluded_files=[doc]
+            )
+
+        with self.subTest("without schema version aka core"):
```
hannes-ucsc: "aka" makes no sense in this context. Do you mean "or"?

tests/test_indexer.py
```diff
@@ -675,22 +675,39 @@ def test_scrub_index_data(self):
             )
 
         with self.subTest("with invalid schema_url"):
+            url = "http://invalid_url"
```
hannes-ucsc: This is actually a valid URL. If there were a host named "invalid_url" in the local network and that host would be running an HTTP server, this URL would work.

Try `"://invalid_url"`–valid URL must have a non-valid scheme and I don't think this would pass for a relative URL either.

hannes-ucsc: I *might* be wrong about `invalid_url` being a valid hostname but to be safe if `"://invalid_url"`produces an exception I would go with that.

dss/storage/validator.py
Bento007: fixed

tests/test_indexer.py
Bento007: fixed

tests/test_indexer.py
Bento007: changed to "without core"

tests/test_indexer.py
hannes-ucsc: If you intend to do so, can your create a ticket to track that effort?

hannes-ucsc:
When this is deployed to dev (i.e. when this is merged to master) reindexing needs to be run in dev. When this is deployed to staging (i.e. when master is merged to staging) reindexing needs to be run in staging. How do we intend to track this task?

Bento007:
Created issues #847 to reindex after fix

tests/test_indexer.py
Bento007: I removed duplicate code from the subtests, and improved the subtest description. I broke `test_scrub_index_data` into separate tests but it resulted in a lot of unnecessary setup and teardown. This test should be separate into `class TestScrubIndexData(unittest.testcase)`, but that will take significant refactoring of `test_indexer`.

PR #838, opened 2018-01-09
S3 parallel copy integration test

ttung:
1. Creates a large-ish blob.
2. Copies it.

Connects to #592

### DIFF
--- a/tests/test_s3parallelcopy.py
+++ b/tests/test_s3parallelcopy.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+import os
+import sys
+import unittest
+import uuid
+from concurrent.futures import ThreadPoolExecutor
+
+import boto3
+import botocore
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from dss import stepfunctions
+from dss.stepfunctions import s3copyclient
+from dss.stepfunctions.s3copyclient.implementation import LAMBDA_PARALLELIZATION_FACTOR
+from dss.util.aws import AWS_MIN_CHUNK_SIZE
+from tests import eventually, infra
+from tests.infra import testmode
+
+
+class TestS3ParallelCopy(unittest.TestCase):
+    @staticmethod
+    def upload_part(bucket: str, key: str, upload_id: str, part_id: int):
+        s3_client = boto3.client("s3")
+        etag = s3_client.upload_part(
+            Bucket=bucket,
+            Key=key,
+            UploadId=upload_id,
+            PartNumber=part_id,
+            ContentLength=AWS_MIN_CHUNK_SIZE,
+            Body=chr(part_id % 256) * (AWS_MIN_CHUNK_SIZE))['ETag']
+        return part_id, etag
+
+    @eventually(30 * 60, 5.0, {AssertionError, botocore.exceptions.ClientError})
+    def _check_dst_key_etag(self, bucket: str, key: str, expected_etag: str):
+        s3_client = boto3.client("s3")
+        obj_metadata = s3_client.head_object(
+            Bucket=bucket,
+            Key=key
+        )
+        self.assertEqual(expected_etag, obj_metadata['ETag'].strip("\""))
+
+    @testmode.integration
+    def test_large_copy(self, num_parts=LAMBDA_PARALLELIZATION_FACTOR + 1):
+        test_bucket = infra.get_env("DSS_S3_BUCKET_TEST")
+        test_src_key = infra.generate_test_key()
+        s3_client = boto3.client("s3")
+        mpu = s3_client.create_multipart_upload(Bucket=test_bucket, Key=test_src_key)
+
+        with ThreadPoolExecutor(max_workers=8) as tpe:
+            parts_futures = tpe.map(
+                lambda part_id: TestS3ParallelCopy.upload_part(test_bucket, test_src_key, mpu['UploadId'], part_id),
+                range(1, num_parts + 1))
+
+        parts = [dict(ETag=part_etag, PartNumber=part_id) for part_id, part_etag in parts_futures]
+
+        src_etag = s3_client.complete_multipart_upload(
+            Bucket=test_bucket,
+            Key=test_src_key,
+            MultipartUpload=dict(Parts=parts),
+            UploadId=mpu['UploadId'],
+        )['ETag'].strip('"')
+
+        test_dst_key = infra.generate_test_key()
+        state = s3copyclient.copy_sfn_event(
+            test_bucket, test_src_key,
+            test_bucket, test_dst_key)
+        execution_id = str(uuid.uuid4())
+        stepfunctions.step_functions_invoke("dss-s3-copy-sfn-{stage}", execution_id, state)
+
+        self._check_dst_key_etag(test_bucket, test_dst_key, src_etag)
+
+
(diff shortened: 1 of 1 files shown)


tests/test_s3parallelcopy.py
```diff
@@ -0,0 +1,81 @@
...
+        s3_client = boto3.client("s3")
+        mpu = s3_client.create_multipart_upload(Bucket=test_bucket, Key=test_src_key)
+        part_jobs = list()
+
+        with ThreadPoolExecutor(max_workers=8) as tpe:
+            for part_id in range(1, num_parts + 1):
+                part_jobs.append(
+                    (part_id,
+                     tpe.submit(
+                         TestS3ParallelCopy.upload_part, test_bucket, test_src_key, mpu['UploadId'], part_id)))
```
hannes-ucsc: `self.upload_part` works with class, static and instance methods and should also work in this context which is creating a bound method and pass that as a callable.

tests/test_s3parallelcopy.py
```diff
@@ -0,0 +1,81 @@
...
+from dss import stepfunctions
+from dss.stepfunctions import s3copyclient
+from dss.stepfunctions.s3copyclient.implementation import LAMBDA_PARALLELIZATION_FACTOR
+from dss.util.aws import AWS_MIN_CHUNK_SIZE
+from tests import eventually, infra
+from tests.infra import testmode
+
+
+class TestS3ParallelCopy(unittest.TestCase):
+    @staticmethod
```
hannes-ucsc: Why is this static?

tests/test_s3parallelcopy.py
```diff
@@ -0,0 +1,81 @@
...
+                         TestS3ParallelCopy.upload_part, test_bucket, test_src_key, mpu['UploadId'], part_id)))
+
+            parts = [dict(ETag=part_job.result(), PartNumber=part_id) for part_id, part_job in part_jobs]
+
+        src_etag = s3_client.complete_multipart_upload(
+            Bucket=test_bucket,
+            Key=test_src_key,
+            MultipartUpload=dict(Parts=parts),
+            UploadId=mpu['UploadId'],
+        )['ETag'].strip("\"")
```
hannes-ucsc: `'"'` wouldn't require a backslash.

tests/test_s3parallelcopy.py
```diff
@@ -0,0 +1,81 @@
...
+
+        test_dst_key = infra.generate_test_key()
+        state = s3copyclient.copy_sfn_event(
+            test_bucket, test_src_key,
+            test_bucket, test_dst_key)
+        execution_id = str(uuid.uuid4())
+        stepfunctions.step_functions_invoke("dss-s3-copy-sfn-{stage}", execution_id, state)
+
+        self._check_dst_key_etag(test_bucket, test_dst_key, src_etag)
```
hannes-ucsc: You only get the same etag if you use the same part size for the fixture upload as for the copy. I haven't looked at the code under test so I don't now if there is a chance the part size could diverge.

tests/test_s3parallelcopy.py
```diff
@@ -0,0 +1,81 @@
...
+    @testmode.integration
+    def test_large_copy(self, num_parts=LAMBDA_PARALLELIZATION_FACTOR + 1):
+        test_bucket = infra.get_env("DSS_S3_BUCKET_TEST")
+        test_src_key = infra.generate_test_key()
+        s3_client = boto3.client("s3")
+        mpu = s3_client.create_multipart_upload(Bucket=test_bucket, Key=test_src_key)
+        part_jobs = list()
+
+        with ThreadPoolExecutor(max_workers=8) as tpe:
+            for part_id in range(1, num_parts + 1):
```
hannes-ucsc: Using `tpe.map()` might simplify this a little bit.

tests/test_s3parallelcopy.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @testmode.integration
+    def test_large_copy(self, num_parts=LAMBDA_PARALLELIZATION_FACTOR + 1):
+        test_bucket = infra.get_env("DSS_S3_BUCKET_TEST")
+        test_src_key = infra.generate_test_key()
+        s3_client = boto3.client("s3")
+        mpu = s3_client.create_multipart_upload(Bucket=test_bucket, Key=test_src_key)
+        part_jobs = list()
+
+        with ThreadPoolExecutor(max_workers=8) as tpe:
```
hannes-ucsc: I would use `8 * multiprocessing.cpu_count()` here. Since 90% of what these threads will be doing is wait for I/O, you could use significantly more threads than you have cores.

dss-api.yml
```diff
@@ -588,6 +589,7 @@ paths:
       description: |
         Create a new version of a bundle with a given UUID.  The list of file UUID+versions to be included must be
         provided.
+      x-dss-idempotent: true
```
amork: I was under impression that this is not required when using PUT

ttung: Whoops, need to update the PR.

tests/test_s3parallelcopy.py
ttung: I went with this approach because tpe.map doesn't allow additional parameters, but I think I can work around that.

tests/test_s3parallelcopy.py
ttung: In my experience running this, I saturated the network connection and more threads did not help.  I am more concerned with the memory footprint at this point.

tests/test_s3parallelcopy.py
ttung: It has no dependency on the class or the instance.  It is merely here for code organizational purposes.

tests/test_s3parallelcopy.py
ttung: I prefer this because it's consistent with our string quoting policy.

tests/test_s3parallelcopy.py
ttung: Part size is dependent on the size of the output file.  It is 64MB until it needs to grow to accommodate larger files.  See L35 where we reference same constant (`AWS_MIN_CHUNK_SIZE`).

tests/test_s3parallelcopy.py
```diff
@@ -0,0 +1,81 @@
...
+        return etag
+
+    @eventually(30 * 60, 5.0, {AssertionError, botocore.exceptions.ClientError})
+    def _check_dst_key_etag(self, bucket: str, key: str, expected_etag: str):
+        s3_client = boto3.client("s3")
+        obj_metadata = s3_client.head_object(
+            Bucket=bucket,
+            Key=key
+        )
+        self.assertEqual(expected_etag, obj_metadata['ETag'].strip("\""))
```
Bento007: I suggest you use `'"'`

tests/test_s3parallelcopy.py
hannes-ucsc: I used to religiously mark every method that didn't reference `self` as static but stopped doing that because it turned out to be incidental. Subsequent revisions of the code suddenly introduced references to `self` and the decoration had to be removed. This got especially annoying if the method is called through the class, like you do below.

As far as code organization is concerned, it would send a stronger message to there reader if this method were protected.

tests/test_s3parallelcopy.py
hannes-ucsc: Sounds good.

tests/test_s3parallelcopy.py
ttung: Ok ok! :)

Changed it.

Issue #485, opened 2017-09-26
GET /bundles/<bundle_uuid> often results in HTTP Error 504: Gateway Timeout


caaespin:
Getting a Timeout error when calling GET `<staging_url>/v1/bundles/<bundle_uuid>`: 

`HTTP Error 504: Gateway Timeout`

20 calls were threaded for the blue box, and many timed out. Will include a test script shortly.

Bento007:
any update on this?

ttung:
THERE, UPDATES. :)

PR #842, opened 2018-01-10
Mark PUT /files and PUT /bundles idempotent

ttung:
This informs the CLI that these requests can be retried.

Connects to #485


kislyuk:
Should we not treat all PUTs as idempotent, and switch them to POST otherwise?

ttung:
It's probably makes sense to, but I'm going to have each endpoint's owner make that determination rather than mark it as such on the CLI tool.

ttung:
Actually, it looks like the urllib3 marks `PUT` as retryable.

PR #848, opened 2018-01-11
Remove more chained-aws-lambda cruft from the repo.

ttung:


dss/stepfunctions/checkout/checkout_states.py
```diff
@@ -31,7 +31,7 @@
 ]
 
 state_machine_def = {
-    "Comment": "DSS Checkout service state machine that submits a Job to chained copy client"
+    "Comment": "DSS Checkout service state machine that submits a Job to s3 copy client"
```
hannes-ucsc: `checkout` `job` and `S3`

ttung: done.

PR #850, opened 2018-01-12
Use execution id instead of ARN

amork:
In order to avoid exposing account id and region switch from using ARN to uuid4-based sfn execution id.
Also update format of the checkout S3 location.
### Test plan
Tested locally by initiating a checkout process then using returned checkout job id in the check status invokation.
Integration and also mmoke test in dev cover this.

### DIFF
--- a/dss/api/bundles/checkout.py
+++ b/dss/api/bundles/checkout.py
@@ -6,6 +6,7 @@ from ..bundles import get_bundle
 from ... import Config, dss_handler, stepfunctions, Replica
 from ...util.checkout import get_execution_id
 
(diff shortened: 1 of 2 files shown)


dss/api/bundles/checkout.py
```diff
@@ -6,6 +6,7 @@
 from ... import Config, dss_handler, stepfunctions, Replica
 from ...util.checkout import get_execution_id
 
+state_machine_name_template = "dss-checkout-sfn-{stage}"
```
ttung: I would make constants all-caps.

amork: fixed

PR #851, opened 2018-01-12
Add utility for using Kibana with AWS credentials

hannes-ucsc:
This is useful for running Kibana against an Amazon Elasticsearch instance without having to change the access policy to allow requests from your local IP. It works by running Kibana through a proxy that signs the requests with your AWS credentials.

### DIFF
--- a/scripts/kibana-proxy.py
+++ b/scripts/kibana-proxy.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python
+
+"""
+Runs Kibana and aws-signing-proxy locally. The latter is used to sign requests by the former and forward them to an
+Amazon Elasticsearch instance. The default instance is the main ES instance for the current DSS stage.
+
+To install Kibana and aws-signing-proxy follow these steps (macOS only):
+
+1) Run
+
+   brew install golang
+
+2) As instructed by brew, set GOPATH and GOROOT, then set PATH to include $GOPATH/bin
+
+3) Run
+
+   go get github.com/cllunsford/aws-signing-proxy
+
+4) Install a matching version of Kibana (currently 5.5). Finding the homebrew-core commit for that version is the
+   trickiest part of the setup. It is possible that you don't need to do this as you might already have a matching
+   version of Kibana installed locally. I used the following command to install Kibana 5.5.2
+
+   brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/eb26ed9e7a4b3a35c62a40fa4fec89bf0361781f/Formula/kibana.rb
+
+To use this program, set `AWS_PROFILE` and `source environment`. Then run
+
+   kibana-proxy.py
+
+If multiple versions of Kibana are installed, you may want to select the one to be run by this program by setting
+DSS_KIBANA_BIN (in environment.local) to point at the Kibana executable. For example, I have
+
+   export DSS_KIBANA_BIN=/usr/local/Cellar/kibana/5.5.2/bin/kibana
+"""
+
+import logging
+import os
+import shlex
+import signal
+import sys
+from itertools import chain
+
+import boto3
+
+log = logging.getLogger(__name__)
+
+
+class KibanaProxy:
+
+    def __init__(self, options) -> None:
+        self.options = options
+        self.pids = {}
+
+    def run(self):
+        kibana_port = self.options.kibana_port
+        proxy_port = self.options.proxy_port or kibana_port + 10
+        try:
+            self.spawn('aws-signing-proxy',
+                       '-target', self.dss_end_point,
+                       '-port', str(proxy_port),
+                       AWS_REGION=os.environ['AWS_DEFAULT_REGION'])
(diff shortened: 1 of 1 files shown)


scripts/kibana-proxy.py
```diff
@@ -0,0 +1,113 @@
...
+"""
+
+import logging
+import shlex
+import signal
+import sys
+from itertools import chain
+
+import boto3
+import os
```
ttung: this import goes above.

scripts/kibana-proxy.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @property
+    def dss_end_point(self):
+        log.info('Getting domain endpoint')
+        es = boto3.client('es')
+        domain = es.describe_elasticsearch_domain(DomainName=self.options.domain)
+        return 'https://' + domain['DomainStatus']['Endpoint']
+
+    def spawn(self, *args, **env):
+        command = ' '.join(chain(
```
ttung: Why are you setting the env here and in `os.spawnvpe`?

scripts/kibana-proxy.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def main(argv):
+    import argparse
+    cli = argparse.ArgumentParser(description=__doc__)
+    cli.add_argument('--kibana-port', '-p', metavar='PORT', default=5601, type=int,
+                     help="The port Kibana should be listening on.")
+    cli.add_argument('--proxy-port', '-P', metavar='PORT', type=int,
+                     help="The port the proxy should be listening on. The default is the Kibana port plus 10.")
+    default_domain = os.environ.get('DSS_ES_DOMAIN', 'dss-index-' + os.environ['DSS_DEPLOYMENT_STAGE'])
```
ttung: Why would DSS_ES_DOMAIN not be set?

scripts/kibana-proxy.py
hannes-ucsc: It's really just so that the logged command makes transparent what environment overrides are being used when running the command. I though that the log message could emulate a shell command line so it would look familiar to users. That's also why `shlex.quote` is used here even though its not really used to run the command since the `spawn…()` functions don't involve a shell.

scripts/kibana-proxy.py
hannes-ucsc: You are right, this should not duplicate the logic for composing the name.

scripts/kibana-proxy.py
hannes-ucsc: Will fix.

hannes-ucsc:
Addressed comments and answered question.

scripts/kibana-proxy.py
ttung: Ahh, I see.  I would almost name `command` something else to reflect that it's just for logging, to avoid confusion.

hannes-ucsc: Fixed.

scripts/kibana-proxy.py
```diff
@@ -0,0 +1,114 @@
...
+4) Install a matching version of Kibana (currently 5.5). Finding the homebrew-core commit for that version is the
+   trickiest part of the setup. It is possible that you don't need to do this as you might already have a matching
+   version of Kibana installed locally. I used the following command to install Kibana 5.5.2
+
+   brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/eb26ed9e7a4b3a35c62a40fa4fec89bf0361781f/Formula/kibana.rb
+
+To use this program, set `AWS_PROFILE` and `source environment`. Then run
+
+   kibana-proxy.py
+
```
mikebaumann: It would be helpful to add an instruction to access the Kibana web UI by opening output URL in a browser:
```
Server running at http://localhost:5601
```

mikebaumann: What is the appropriate way to cleanly terminate the Kibana proxy setup? Simply ctrl-C in the terminal window from which it was started, or something else?

scripts/kibana-proxy.py
```diff
@@ -0,0 +1,114 @@
...
+Runs Kibana and aws-signing-proxy locally. The latter is used to sign requests by the former and forward them to an
+Amazon Elasticsearch instance. The default instance is the main ES instance for the current DSS stage.
+
+To install Kibana and aws-signing-proxy follow these steps (macOS only):
+
+1) Run
+
+   brew install golang
+
+2) As instructed by brew, set GOPATH and GOROOT, then set PATH to include $GOPATH/bin
```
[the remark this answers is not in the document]
hannes-ucsc: Will fix

scripts/kibana-proxy.py
hannes-ucsc: Will fix

hannes-ucsc: Ctrl-C does it. Will mention this.

hannes-ucsc:
Post-merge review issues are addressed in #858.

Issue #349, opened 2017-08-14
Switch gcp.copy to use rewrite calls


ttung:

kozbo:
What is the status on this?

ttung:
This is done.

PR #853, opened 2018-01-12
GS copy implemented as step functions

ttung:
Unlike S3 copy, GS copy cannot be parallelized.  However, it has a token that can be used to continue a rewrite call.

Connects to #349

### DIFF
--- a/daemons/dss-gs-copy-sfn/app.py
+++ b/daemons/dss-gs-copy-sfn/app.py
@@ -1,0 +1,12 @@
+import os
+import sys
+
+import domovoi
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'domovoilib'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+import dss.stepfunctions.gscopyclient as gscopyclient
+
+app = domovoi.Domovoi()
+app.register_state_machine(gscopyclient.sfn)
--- a/dss/stepfunctions/gscopyclient/implementation.py
+++ b/dss/stepfunctions/gscopyclient/implementation.py
@@ -1,0 +1,98 @@
+import copy
+import typing
+
+from cloud_blobstore.gs import GSBlobStore
+
+from ... import Config, Replica
+from ...stepfunctions.lambdaexecutor import TimedThread
+
+
+# Public input/output keys for the state object.
+class Key:
+    SOURCE_BUCKET = "srcbucket"
+    SOURCE_KEY = "srckey"
+    DESTINATION_BUCKET = "dstbucket"
+    DESTINATION_KEY = "dstkey"
+    FINISHED = "finished"
+
+
+# Internal key for the state object.
+class _Key:
+    SOURCE_CRC32C = "srccrc32c"
+    SIZE = "size"
+    TOKEN = "token"
+
+
+def setup_copy_task(event, lambda_context):
+    source_bucket = event[Key.SOURCE_BUCKET]
+    source_key = event[Key.SOURCE_KEY]
+
+    gs_blobstore = typing.cast(GSBlobStore, Config.get_cloud_specific_handles(Replica.gcp)[0])
+    blob = gs_blobstore.gcp_client.bucket(source_bucket).get_blob(source_key)
+    source_crc32c = blob.crc32c
+    source_size = blob.size
+
+    event[_Key.SOURCE_CRC32C] = source_crc32c
+    event[_Key.SIZE] = source_size
+    event[Key.FINISHED] = False
+
+    return event
+
+
+def copy_worker(event, lambda_context):
+    class CopyWorkerTimedThread(TimedThread[dict]):
+        def __init__(self, timeout_seconds: float, state: dict) -> None:
+            super().__init__(timeout_seconds, state)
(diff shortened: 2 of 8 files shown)


dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -0,0 +1,101 @@
...
+from cloud_blobstore.gs import GSBlobStore
+
+from ... import Config, Replica
+from ...api import files
+from ...stepfunctions import generator
+from ...stepfunctions.lambdaexecutor import TimedThread
+
+
+# Public input/output keys for the state object.
+class Key:
```
hannes-ucsc: I find this indirection to be overkill. Yes, it allows you to change the actual string value without touching the rest of the code but 

1) this is happens infrequently if at all and 

2) all references to the state are local to this small package and 

3) the fact that property name (`SOURCE_BUCKET`) is different from the actual key in the state dictionary is actually rather confusing if you look at the state in the AWS console.

IMO, it would be Pythonic to exploit its dynamic nature using things like `self.__dict__.extend(state)`

I realize that this is a significant change in style but I thought I'd share my view on this.

dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -0,0 +1,101 @@
...
+            while True:
+                response = dst_blob.rewrite(src_blob, token=state.get(_Key.TOKEN, None))
+                if response[0] is None:
+                    state[Key.FINISHED] = True
+                    return state
+                else:
+                    state[_Key.TOKEN] = response[0]
+                    self.save_state(state)
+
+    result = CopyWorkerTimedThread(lambda_context.get_remaining_time_in_millis() / 1000, event).start()
```
hannes-ucsc: I see no value in extracting a variable here.

daemons/dss-gs-copy-sfn/app.py
```diff
@@ -0,0 +1,17 @@
...
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'domovoilib'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+import dss.stepfunctions.gscopyclient as gscopyclient
+import dss.stepfunctions.generator as generator
+
+app = domovoi.Domovoi()
+
+
+annotation_processor = generator.StateMachineAnnotationProcessor()
```
hannes-ucsc: Take a look at how @xbrianh's Vistitation SFN definition gets by without this. It cuts out a whole lot of machinery with very little additional code. It also is more flexible in the kinds of parameterizations it allows and is self-evident since it only uses standard dictionary and list literals.

tests/test_gscopyclient.py
```diff
@@ -0,0 +1,85 @@
...
+        for ix in range(len(test_src_keys) - 1):
+            src_blob_obj = bucket_obj.get_blob(test_src_keys[ix])
+            blobs = [src_blob_obj for _ in range(16)]
+            dst_blob_obj = bucket_obj.blob(test_src_keys[ix + 1])
+
+            dst_blob_obj.content_type = "application/octet-stream"
+            dst_blob_obj.compose(blobs)
+
+        # set the storage class to nearline.
+        # NOTE: compose(..) does not seem to support setting a storage class.  the canonical way of changing storage
```
hannes-ucsc: `… .  The …`

dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -0,0 +1,101 @@
...
+            self.destination_key = state[Key.DESTINATION_KEY]
+            self.size = state[_Key.SIZE]
+
+        def run(self) -> dict:
+            state = self.get_state_copy()
+            src_blob = self.gs_blobstore.gcp_client.bucket(self.source_bucket).get_blob(self.source_key)
+            dst_blob = self.gs_blobstore.gcp_client.bucket(self.destination_bucket).blob(self.destination_key)
+
+            while True:
+                response = dst_blob.rewrite(src_blob, token=state.get(_Key.TOKEN, None))
```
hannes-ucsc: I've never used the GCS API. Looks like this can't be parallelized? Should it be?

daemons/dss-gs-copy-sfn/app.py
ttung: Actually, we don't need to do the annotation processor for this since it's single threaded.  copy-pasta. :(

dss/stepfunctions/gscopyclient/implementation.py
ttung: It's more for avoiding a typo in a string.

dss/stepfunctions/gscopyclient/implementation.py
ttung: It cannot. :(

dss/stepfunctions/gscopyclient/implementation.py
ttung: willfix

tests/test_gscopyclient.py
ttung: Willfix.

ttung:
Updated, and landing.

Issue #829, opened 2018-01-04
Clean up logging


hannes-ucsc:
- [x] Eliminate passing around of logger instances, revert back to the idiomatic way of using module level loggers
- [x] Eliminate `"%s"` hack
- [ ] ~Always use `f"…"` strings for log messages~
- [ ] ~Shorten `logger` to `log`~
- [x] Eliminate long expressions between curly braces in `f"…"` strings
- [x] Fine-tune per-module log levels, for example, suppress boto's HTTP connection logging

hannes-ucsc:
As @xbrianh pointed out to me, we might not need the "%s" hack. The % interpolation is conditional on a the presence of actual args to be interpolated:

https://github.com/python/cpython/blame/3.6/Lib/logging/__init__.py#L337

```python
>>> import logging
>>> log = logging.getLogger('foo')
# interpolated value with % in it 
>>> x = "b%ss"
# without args, no problem
>>> log.warn(f"x is {x}.")
x is b%ss.
```

When args are passed, they will be used to interpolate the % signs that were interpolated via the f-string. The failure modes are a bit more insidious:

```python
>>> log.warn(f"x is {x}.", 1)
x is b1s.
>>> x = 'b%is'
>>> log.warn(f"x is {x} and foo is %s.",'foo')
--- Logging error ---
Traceback (most recent call last):
…
TypeError: %i format: a number is required, not str
…
```

and 

```python
>>> x = 'b%ss'
>>> log.warn(f"x is {x} and foo is %s.",'foo')
--- Logging error ---
Traceback (most recent call last):
…
TypeError: not enough arguments for format string
…
```

I don't see sufficient utility in mixing f-strings and percent-formatting for logging so I'm inclined to simply disallow it by project convention.

Alternatively, we could override the log record factory to produce log records that explicitly disallow passing args to the logger methods.

PR #859, opened 2018-01-18
2/4: Clean up logging

hannes-ucsc:
See #829 and the section I added to the README for a motivation.

I also recommend reviewing commit by commit.


README.md
```diff
@@ -280,6 +280,53 @@ change to a direct dependency. This process applies to development dependencies
...
+    # but not
+    logger.info(f"Foo is {foo} and bar is %s", bar)
+    # Keyword arguments can be used safely in conjunction with f-strings: 
+    logger.info(f"Foo is {foo}", exc_info=True)
+    ```
+    
+4.  To enable verbose logging by application code, set the envrionment variable `DSS_DEBUG` to `1`. To enable verbose 
+    logging by dependencies set `DSS_DEBUG` to `2`. To disable verbose logging unset `DSS_DEBUG` or set it to `0`.
+
+5.  To assert in tests that certain messages were logged, use the `dss` logger or one of its childreb
```
hannes-ucsc: children

ttung:
Would it make sense to have dss-level logger, and have the individual modules set up their loggers as children of the dss logger?

hannes-ucsc:
That's exactly how this PR works. The 'dss' logger is the top-level logger for the app and every module under the `dss` package is a child of that logger.

https://docs.python.org/3/library/logging.html#logger-objects

ttung:
Oh, cool, I did not know that logging automatically constructs the hierarchy.

README.md
```diff
@@ -280,6 +280,53 @@ change to a direct dependency. This process applies to development dependencies
...
+2.  Do not configure logging at module scope. It should be possible to import any module without side-effects on 
+    logging. The `dss.logging` module contains functions that configure logging for this application, its Lambda 
+    functions and unit tests.
+    
+3.  When logging a message, pass either
+ 
+    * an f-string as the first and only positional argument or
+    
+    * a %-string as the first argument and substitution values as subsequent argumentss. Do not mix the two string 
+      interpolation methods. If you mix them, any percent sign in a substituted value will raise an exception.
```
Bento007: The issues i found with this was if the f-string evaluated to something containing %s, %d, ect.. the logger would complain that it expected more parameters. For example `logger.warning(f"hello {x}", "%s")` would cause an exception expecting more arguments. Has this case been handled?

hannes-ucsc: Yes, it is handled by disallowing this particular case. You shouldn't pass an f"" string as the first argument and then pass additional parameters. More details on that can be found here:

https://github.com/HumanCellAtlas/data-store/issues/829#issuecomment-358455378

Bento007: Sorry bad example. This is what i meant,
```python
import logging
x = '%s'
logging.warning(f"{x}")
```
but i tested it and it works out.

dss/logging.py
```diff
@@ -0,0 +1,90 @@
...
+import sys
+
+from typing import Mapping, Union, Sequence, Tuple
+
+import dss
+from dss.config import Config
+
+log_level_t = Mapping[Union[None, str, logging.Logger], Tuple[int, ...]]
+"""
+Values of this type map a given logger and debug level to the log level for that logger at the given debug level. The
```
ttung: I don't really get this section.  I think I understand how it works from the code, not from the code.

README.md
```diff
@@ -280,6 +280,53 @@ change to a direct dependency. This process applies to development dependencies
...
+    logger = logging.getLogger(__name__)
+    ```
+
+2.  Do not configure logging at module scope. It should be possible to import any module without side-effects on 
+    logging. The `dss.logging` module contains functions that configure logging for this application, its Lambda 
+    functions and unit tests.
+    
+3.  When logging a message, pass either
+ 
+    * an f-string as the first and only positional argument or
```
ttung: I'm inclined to just ban f-strings because the modifying a formatting string to take more arguments will cause unnecessary churn or people might forget to switch to interpolation.

ttung:
Rest lgtm.

dss/logging.py
hannes-ucsc: I rewrite the section.

hannes-ucsc: I will rewrite the section.

README.md
```diff
@@ -280,6 +280,53 @@ change to a direct dependency. This process applies to development dependencies
...
+    
+    * a %-string as the first argument and substitution values as subsequent argumentss. Do not mix the two string
```
kislyuk: typo in "argumentss"

dss/logging.py
```diff
@@ -0,0 +1,90 @@
...
+        log_levels = log_levels.copy()
+        log_levels.update(test_log_levels)
+    for logger, levels in log_levels.items():
+        if isinstance(logger, (str, type(None))):
+            logger = logging.getLogger(logger)
+        level = levels[min(debug, len(levels) - 1)]
+        logger.setLevel(level)
+
+
+def defuse_flask_logging(flask_app):
```
kislyuk: Could it be easier to rename the Flask app? Does that avoid the need for this method?

README.md
hannes-ucsc: I thought I understood this comment but now I'm not so sure.

f-strings *are* a kind of interpolation. Do you mean "switch to % interpolation"? There really is no reason to perform that switch.

dss/logging.py
hannes-ucsc: Let me try that out.

dss/logging.py
hannes-ucsc: Rewritten.

README.md
hannes-ucsc: Fixed

README.md
ttung: My understanding is that if you have f-strings with any string formatting happening, then the output could contain `%`, which would cause logging to freak out.  It seems like it's just risky.

In your example:

```
logger.info(f"Foo is {foo} and bar is {bar}")
```

if `foo` is `%s`, wouldn't logging complain that there aren't enough arguments to format the output?

ttung: So I am suggesting that we just use % interpolation everywhere.

hannes-ucsc: > wouldn't logging complain

Nope: https://github.com/HumanCellAtlas/data-store/issues/829#issuecomment-358455378

> So I am suggesting that we just use % interpolation everywhere.

But why?

hannes-ucsc: Maybe I should add a section explaining why it is safe to use an f"" string as the first argument as long as you don't pass any additional positional arguments.

Bento007: didn't we have this discussion https://github.com/HumanCellAtlas/data-store/pull/859#discussion_r163368764  already in a previous comment string or is this something different?

ttung: Same thing. My bad.

dss/logging.py
hannes-ucsc: There was an even easier way:

https://github.com/HumanCellAtlas/data-store/pull/859/commits/c120f63492132b95c732314051a5df134bcfc720#diff-83c83ff7781521cdb005c5d2180b2522L187

Makefile
```diff
@@ -21,6 +21,10 @@ $$($1_test_srcs): $1__%.py :
 	| sed -e "s/^/$$$$$$$$ /"
 
 .PHONY: $1_test $$($1_test_srcs)
+
+# Serialize the tests that use a local Elasticsearch instance
+$1__tests/test_search.py: $1__tests/test_indexer.py
+$1__tests/test_indexer.py: $1__tests/test_subscriptions.py
```
kislyuk: Not that this needs to be fixed here, but I'm decidedly not a fan of all the magic alphabet soup that has crept into the Makefile test rules. To serialize the tests, just make a target that calls them serially:

```makefile
serial_tests:
	$(MAKE) tests/test_search.py
	$(MAKE) tests/test_indexer.py
```
 and exclude them in the wildcard parallel test target. If I were touching the Makefile today, I would rip out all of this target definition, eval, and octuple dollar sign madness.

dss/events/handlers/sync.py
```diff
@@ -17,6 +18,10 @@
 from dss.util.aws import resources, clients, send_sns_msg, ARN
 from dss.util.streaming import get_pool_manager, S3SigningChunker
 
+
+logger = logging.getLogger(__name__)
+
```
kislyuk: Nitpick: This is one blank line too many

Makefile
hannes-ucsc: This change shouldn't be here. I'll remove it. Well spotted.

However, since I've pushed to this PR many times by now, I can already tell that this change does make the build more stable. So we might consider making those changes anyways, in another PR. I think I have a solution to address @ttung's [concern](https://github.com/HumanCellAtlas/data-store/pull/925#discussion_r164573215) but it would require leaving the [magic](https://github.com/HumanCellAtlas/data-store/pull/925#discussion_r165117509) in.

dss/events/handlers/sync.py
hannes-ucsc: Will fix.

PR #860, opened 2018-01-18
refactor the `test_file_put_large` test to more easily accommodate additional replicas

ttung:
Isolate the upload portion, which is the most cloud-specific aspect of the whole operation.

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -76,8 +76,8 @@ class TestFileApi(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
     @testmode.integration
     def test_file_put_large(self):
 
-        def upload_callable_creator(uploader_class: type):
-            def upload_callable(bucket: str, key: str):
+        def upload_callable_creator(uploader_class: type) -> typing.Callable[[str, str], None]:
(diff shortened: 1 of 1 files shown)


tests/test_file.py
```diff
@@ -75,23 +75,34 @@ def _test_file_put(self, replica: Replica, scheme: str, test_bucket: str, upload
...
+                    fh.flush()
+
+                    uploader.checksum_and_upload_file(fh.name, key, "text/plain")
+            return upload_callable
+
+        self._test_file_put_large(Replica.aws, self.s3_test_bucket, upload_callable_creator(S3Uploader))
         # There's no equivalent for GCP ... yet. :)
 
-    def _test_file_put_large(self, replica: Replica, scheme: str, test_bucket: str, uploader: Uploader):
+    def _test_file_put_large(self, replica: Replica, test_bucket: str, upload_func: typing.Callable[[str, str], None]):
```
Bento007: why do you need the None at the end?

ttung: That's to indicate that the callable should return `None`.  I should fix L80 to reflect that as well...

PR #861, opened 2018-01-18
build the gcp client directly rather than get_cloud_specific_handles

ttung:
The latter API is crufty and needs a refactor.  The issue at hand is that S3BlobStore and GSBlobStore are actually not tied to a specific bucket, yet we have to configure the bucket config.

### DIFF
--- a/dss/stepfunctions/gscopyclient/implementation.py
+++ b/dss/stepfunctions/gscopyclient/implementation.py
@@ -1,9 +1,8 @@
 import copy
-import typing
+import os
 
-from cloud_blobstore.gs import GSBlobStore
+from google.cloud.storage import Client
 
-from ... import Config, Replica
 from ...stepfunctions.lambdaexecutor import TimedThread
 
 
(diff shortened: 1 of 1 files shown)


dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -23,12 +22,18 @@ class _Key:
     TOKEN = "token"
 
 
+def get_gcp_client():
+    # TODO: (ttung) remove this once Config.get_cloud_specific_handles is refactored.
```
hannes-ucsc: Unless you have a PR in the works already, this is bound to fall through the cracks. What I do cases like this create a ticket and refer to the ticket in the TODO.

ttung:
> Unless you have a PR in the works already, this is bound to fall through the cracks. What I do cases like this create a ticket and refer to the ticket in the TODO.

> LGTM so long as the follow-up happens

Man, you guys are no fun. :)

PR #864, opened 2018-01-18
Switch PUT /files to use async copy for GCP

ttung:
Enable async copy test for GCP.

Depends on #860, #861, #863 
Connects to #349


ttung:
I most definitely did not close this.  Wtf, github?

ttung:
This says "closed" but it's not landed, and should not until #914 lands.

PR #871, opened 2018-01-22
[easy] rename get_cloud_specific_handles(…) to *_DEPRECATED

ttung:
Depends on #870

PR #872, opened 2018-01-22
Lazy-initialization of cloud-specific handles.

ttung:
This is to address TODOs and comments in #861.

Depends on #871

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -3,9 +3,11 @@ import typing
 from contextlib import contextmanager
 from enum import Enum, EnumMeta, auto
 
+import boto3
 from cloud_blobstore import BlobStore
 from cloud_blobstore.s3 import S3BlobStore
 from cloud_blobstore.gs import GSBlobStore
+from google.cloud.storage import Client
 
 from .hcablobstore import HCABlobStore
 from .hcablobstore.s3 import S3HCABlobStore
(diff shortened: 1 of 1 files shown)


ttung:
> Lazy initialization can be achieved much more easily with the functools.lru_cache decorator.

Happy to switch to this.  Didn't actually know about this. :)

> I propose eliminating the coupling between the three "handles". Very frequently, only one or two of them are needed. They should all be independent methods.

I disagree.  The three handles are a result of a clean abstraction: stuff that the cloud provider provides, stuff we built that are generally useful functions or just paper over the differences between the cloud providers, and stuff that is specific to the HCA project.  I'm ok if we do a delegate approach, such that we construct a single handle, and any calls that aren't handled in that abstraction layer are sent to the next abstraction layer.  However, this is unfriendly for IDEs.

The alternative is to have the HCA layer inherit from the cloud_blobstore layer, and have the cloud_blobstore layer delegate to the native cloud provider handle.  This gives us IDE usability (the native cloud provider methods are all dynamically generated, and so aren't available to the IDE anyway) and a single handle to work with, but is kind of kludgy in that it's half inheritance and half delegation.

> My other objections make this a moot point but I think what the parameterization with lambdas as factories achieves is more commonly done with plain old inheritance.

It would be possible to replace the lambdas with passing in a class, since all the lambdas are basically the same.  Would you prefer that?

> There never is an instance of Config AFAICT. This is an anti-pattern in Python. If all methods of a class are static, and there are no instances of that class and no subclasses of that class, there is no point for that class. The static methods can just as easily be functions at the module level. This has bothered me before and I thought I should mention this while you are touching that code.

I'm not sure I agree.  The `Config` class serves to wrap up some state, and group things together.  I would be on board with splitting `config.py` into a package and have all of the Config class as top-level methods in its own module, but functionally, that's not really all that different.

hannes-ucsc:
>  The three handles are a result of a clean abstraction

That's true. And I like the abstraction. But returning them as a tuple (or an object as in this PR) is coupling them unnecessarily. There is evidence of that in the code base: just count the number call sites that do
`get_cloud_specific_handles()[0]` or `_,foo,_ = get_cloud_specific_handles()[0]`

> It would be possible to replace the lambdas with passing in a class, since all the lambdas are basically the same. Would you prefer that?

Probably not. Introducing yet another class would just make this even more complicated. I think this can be done with only one class or none at all.

> The Config class serves to wrap up some state, 

AFAICT, it does not. At least not mutable state. It only has constants and static methods.

> and group things together. 

But there is no need for further grouping. Everything in config.py is configuration related and dissolving Config by promoting its members to module-level attributes of config.py would not unduly mix concerns.

hannes-ucsc:
How about I take a shot at this,too, and then we compare?

My *claim* is that I can simplify config.py but I'm not 100% certain I can accomplish that.

mweiden:
@hannes-ucsc @ttung I'll hold off on reviewing here, until we see the alternate

ttung:
Abandoning.

PR #874, opened 2018-01-22
Refactor usages of get_cloud_specific_handles_DEPRECATED

ttung:
Use the new API, which is more specific about what is being requested.

I'd like to request a lot of eyes on this, since it touches a ton of places.

Depends on #872

### DIFF
--- a/daemons/dss-sync/app.py
+++ b/daemons/dss-sync/app.py
@@ -55,7 +55,7 @@ def compose_upload(event, context):
     """
     gs_max_compose_parts = 32
     msg = json.loads(event["Records"][0]["Sns"]["Message"])
-    gs = dss.Config.get_cloud_specific_handles_DEPRECATED(Replica.gcp)[0].gcp_client
+    gs = dss.Config.get_cloud_specific_handles(Replica.gcp).native_cloud_handle
     gs_bucket = gs.get_bucket(msg["dest_bucket"])
     while True:
         try:
@@ -106,10 +106,11 @@ platform_to_replica = dict(s3=Replica.aws, gs=Replica.gcp)
 def copy_parts(event, context):
     topic_arn = event["Records"][0]["Sns"]["TopicArn"]
     msg = json.loads(event["Records"][0]["Sns"]["Message"])
-    blobstore_handle = dss.Config.get_cloud_specific_handles_DEPRECATED(platform_to_replica[msg["source_platform"]])[0]
+    blobstore_handle = dss.Config.get_cloud_specific_handles(
+        platform_to_replica[msg["source_platform"]]).blobstore_handle
     source_url = blobstore_handle.generate_presigned_GET_url(bucket=msg["source_bucket"], key=msg["source_key"])
     futures = []
-    gs = dss.Config.get_cloud_specific_handles_DEPRECATED(Replica.gcp)[0].gcp_client
+    gs = dss.Config.get_cloud_specific_handles(Replica.gcp).native_cloud_handle
     with ThreadPoolExecutor(max_workers=4) as executor:
         for part in msg["parts"]:
             context.log(log_msg.format(part=part, **msg))
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -54,8 +54,7 @@ def put(uuid: str, replica: str, json_request_body: dict, version: str = None):
         timestamp = datetime.datetime.utcnow()
     version = datetime_to_version_format(timestamp)
 
-    handle, hca_handle, bucket = Config.get_cloud_specific_handles_DEPRECATED(Replica[replica])
-
+    cloud_handles = Config.get_cloud_specific_handles(Replica[replica])
     # what's the target object name for the bundle manifest?
     bundle_manifest_key = BundleFQID(uuid=uuid, version=version).to_key()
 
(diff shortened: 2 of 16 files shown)


ttung:
> So apparently the "rename to _DEPRECATED" PR is immediately undone by this one. Your PRs are too fragmented for my taste. I can't see where you are going with your changes and they fragmentation is causing too much churn for me as the reviewer.

They're chained from the top (but not the bottom) in the PR summary.  Unfortunately, github is kind of a crappy review tool, and you can't actually build true diff stacks, but it's what we have.

I argue this is the right way to do things because:

1. The easy rename diffs can be rubberstamped and landed.
2. This produces the cleanest _blame_ (though not the cleanest history) without lumping a tremendous amount of changes in one PR.

hannes-ucsc:
Your initial claim was that organizing PRs like that makes reviewing them easier. I think the opposite is true. Fragmenting PRs like that is not magically reducing the amount of change. The fragmentation just makes the changes and their intent much harder to parse for a reviewer.

Does this approach have precedence outside of our project? Who else is using it?

> This produces the cleanest blame (though not the cleanest history)

I don't understand that staement. Both approaches, fragmented/chained and traditional, can produce the same sequence of commits, no?

ttung:
> Your initial claim was that organizing PRs like that makes reviewing them easier. I think the opposite is true. Fragmenting PRs like that is not magically reducing the amount of change. The fragmentation just makes the changes and their intent much harder to parse for a reviewer.

> Does this approach have precedence outside of our project? Who else is using it?

To better answer this question, can you explain what you mean by fragmentation?  Is the issue that you want to see the entire set of changes collapsed into a single view?  That's super easy.  Here it is: https://github.com/HumanCellAtlas/data-store/compare/tonytung-refactor

A 1:1:1 mapping from thesis:commit:PR is practiced at my previous employer.

> I don't understand that staement. Both approaches, fragmented/chained and traditional, can produce the same sequence of commits, no?

I think I need to better understand your objection to answer this.  I thought you meant you didn't like the fragmented commits, but now I'm not so certain.

hannes-ucsc:
> To better answer this question, can you explain what you mean by fragmentation? Is the issue that you want to see the entire set of changes collapsed into a single view? That's super easy. Here it is: https://github.com/HumanCellAtlas/data-store/compare/tonytung-refactor

That view is not useful for doing an actual review since I cannot leave comments on it. I also have to look at the git log to figure out what the furthest advanced branch is.

My objection is not with the commits (even though in this particular chain of PRs, I would drop the "rename to _DEPRECATED" commit). My objection is with the fact that there is one PR per commit and that those PRs make very little sense in isolation. I have to look at all of them before I can give feedback on any single one of them. This is considerable more work for me with zero benefit for the author, the reviewer or the team. At least *I* can't find a single benefit in this approach.

GH allows a 1:m:n relation between issues, PRs and commits. We need to have a good reason for degenerating that to 1:1:1 and I'm just not seeing one. I try to be a prompt and fair reviewer, but the this approach is making it that frustrating for me.

ttung:
> My objection is not with the commits (even though in this particular chain of PRs, I would drop the "rename to _DEPRECATED" commit). My objection is with the fact that there is one PR per commit and that those PRs make very little sense in isolation. I have to look at all of them before I can give feedback on any single one of them. This is considerable more work for me with zero benefit for the author, the reviewer or the team. At least I can't find a single benefit in this approach.

> GH allows a 1:m:n relation between issues, PRs and commits. We need to have a good reason for degenerating that to 1:1:1 and I'm just not seeing one. I try to be a prompt and fair reviewer, but the this approach is making it that frustrating for me.

Ah I see.  I strongly disagree that there is no benefit.  There is a huge benefit, which is that earlier commits can be landed as soon as they're approved.  Carrying around a huge diff stack has non-trivial costs, both for the reviewer and the reviewee.

For instance, I want to approve the individual commits in your logging PR, but I can't.  I recall some other PRs where we had to hold off on landing things because of potential conflicts, and those are pretty avoidable if the non-controversial things land first.

hannes-ucsc:
> Carrying around a huge diff stack has non-trivial costs, both for the reviewer and the reviewee.

That's not what I am proposing. I am proposing the *right* amount of changes in one PR.

> For instance, I want to approve the individual commits in your logging PR, but I can't.

That implies that you want to reject others. And you can do so, you simply request that those commits be taken out.

> I recall some other PRs where we had to hold off on landing things because of potential conflicts, and those are pretty avoidable if the non-controversial things land first.

Aha! So your chain of PRs (and sequence of commits) looks like this:

1) Swap some code around
2) Touch many lines to deprecate a method
3) Add a new function
4) Touch many lines again to swap in the new function

You can slice and dice it any way you want, but the fourth commit is still exposing you to conflicts unless it is merged quickly. By separating out the non-controversial (but also inconsequential) parts, you have done nothing to avoid merge conflicts. You have only made my life as the reviewer harder. In fact, the second commit exposes you in the same way that the first one does. So that's even worse.

I have been in situations were I was able to do the "many one-liner diffs" first and get them merged quickly but that was rare. And it isn't the case here.

We have a lot of contention on the code base, mostly because the team is too big. That's why I try to 
respond to review requests promptly. And that's why I often approve PRs quickly even if I disagree with some cosmetic aspect of them. In return I would like to review PRs that are the "right" size.

mweiden:
@ttung @hannes-ucsc I think this is veering towards a "ways of working" discussion. I suggest that we discuss this during a group meeting when we're all together.

In the short term, let's discuss the code itself and coordinate landing it or choosing an alternative that drives the project forward. Part of this coordination might be seeing what other PRs are in flight and gauging conflicts.

hannes-ucsc:
Sounds good, @mweiden.

Looking at #909, it appears that this one is redundant now.

ttung:
Abandoning.

PR #908, opened 2018-01-23
Individual endpoints to retrieve storage handles.

ttung:
Instead of retrieving all the handles at once, this method changes it so we retrieve one handle at a time.

Depends on #907


dss/config.py
```diff
@@ -230,21 +253,37 @@ def get_notification_email() -> str:
 
 
 class Replica(Enum):
```
hannes-ucsc: It might be worth exploring the possibility of adding `namedtuple(…, 'bucket_getter storage_schema blobstore_class hcablobstore_class') as a baseclass. It gives you immutable attributes and reduces the boilerplate. I just don't know if namedtuple is compatible with Enum.

PR #909, opened 2018-01-23
Refactor calls to get_cloud_specific_handles()

ttung:
1. Retrieve blobstore and hcablobstore from Config.
2. Retrieve bucket name from replica.


mweiden:
@ttung we approved this, but the tests seem to be failing on a compile-time error

ttung:
apologies.  Pushed fixes for both, and passed the tests locally.

Issue #856, opened 2018-01-17
Streamline deployment process


Bento007:
# User Story (Why?)
1. As DSS operator
I want to easily and reliably deploy DSS and its infrastructure into different AWS accounts
so that we can deploy to production.

2. As user of DSS
I want to easily and reliably deploy DSS and its infrastructure into my AWS account
so that I can have my own deployment

# DoD
- [x] Have an automated deployment process.
- [x] Document the setup process.

## Suggested tools.
* TerraForm
* Script
* Dockerized script
* Installation script

hannes-ucsc:
Brian volunteered ;-)

mweiden:
@xbrianh can you please document the infrastructure setup process?

The goal here would be to come to the repo as a total n00b and be able to follow instructions to deploy.

PR #924, opened 2018-01-25
Use gcloud command to deploy dss-gs-event-relay

### LINKED ISSUE #923: Deployment should not rely on existence of test buckets
The deployment package of `dss-gs-event-relay` is staged into the GCP test fixtures bucket:

https://github.com/HumanCellAtlas/data-store/blob/30c7cf1371a9edf2de16f93a3f1b22ed640b8b31/scripts/deploy_gcf.py#L72

If the user does not wish to run tests, and thus has not created the test buckets, their deployment will fail.

xbrianh:
Deployment of `dss-gs-event-relay` currently relies on the existence of GCP test fixture buckets. This can be avoided with `gcloud beta functions deploy`.

Unfortunately the gcloud method is in beta.

Connects to #856 
FIxes #923

### DIFF
--- a/scripts/deploy_gcf.py
+++ b/scripts/deploy_gcf.py
@@ -3,7 +3,7 @@
 This script manages the deployment of Google Cloud Functions.
 """
 
-import os, sys, time, io, zipfile, random, string, binascii, datetime, argparse, base64
+import os, sys, time, io, zipfile, random, string, binascii, datetime, argparse, base64, subprocess
 
 import boto3
 import google.cloud.storage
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
One general thought: IIRC, there is currently no formal requirement on gcloud. Only the README mentions it. With this change, and possibly your other upcoming improvements of the deployment, it becomes a formal dependency of the Makefile. The dependency should be formalized by adding a check for the existence of the gcloud binary in common.mk.

The same goes for the aws cli, I think.

PR #925, opened 2018-01-25
1/4: Fix flaky indexer tests (fixes #918)

### LINKED ISSUE #918: Abnormal Travis build termination in test_indexer
https://gist.github.com/hannes-ucsc/a3ff655a438373673d2411441a2eb603#file-travis-log-L1235

The end of the log looks truncated. However, there is a failed test in the log. Something seems to go wrong with the retry. The first retry is due to a version conflict in the indexer. That in itself is puzzling because none of the tests should involve a version conflict. From then on, things go haywire, a KeyError is raised in what seems to be the second retry and that seems to be retried again even though it shouldn't.

[Remember to undo #919  after fixing this]

hannes-ucsc:


Makefile
```diff
@@ -8,35 +8,30 @@ mypy:
...
-	coverage combine
-	rm -f .coverage.*
+.PHONY: $1_test $$($1_test_srcs)
 
-$(integration_test_srcs): integration__%.py :
-	set -o pipefail \
-	&& DSS_TEST_MODE="integration" coverage run -p --source=dss -m unittest $(DSS_UNITTEST_OPTS) $*.py 2>&1 \
-	| sed -e "s/^/$$$$ /"
+# Serialize the tests that use a local Elasticsearch instance
+$1__tests/test_search.py: $1__tests/test_indexer.py
```
ttung: Correct me if I'm wrong, but does this not mean that if you run `test_search` from the command line, this will also run test_indexer?

dss/config.py
```diff
@@ -183,26 +206,30 @@ def get_es_index_name(index_type: ESIndexType,
...
+        replica.
+        """
+        return Config._get_es_index_name(index_type, replica, 'alias')
+
+    @classmethod
+    def _get_es_index_name(cls,
+                           index_type: ESIndexType,
+                           replica: 'Replica',
+                           suffix: typing.Optional[str] = None) -> str:
+        assert isinstance(replica, Replica)
```
ttung: I would argue that this assertion is not useful.

tests/infra/server.py
```diff
@@ -27,17 +27,17 @@ class ThreadedLocalServer(threading.Thread):
...
     """
     def __init__(self):
-        super().__init__()
+        super().__init__(daemon=True)
         self._port = networking.unused_tcp_port()
         self._server = None
         self._server_ready = threading.Event()
         self._chalice_app = None
 
-    def start(self, *args, **kwargs):
```
ttung: Good catch.

Makefile
hannes-ucsc: You're right. I should use order-only dependencies here. Will fix.

dss/config.py
hannes-ucsc: Will remove.

Makefile
hannes-ucsc: Order-only dependencies won't fix this.

ttung: We could, I guess, switch back to the system where you are *compelled* to have a running ES instance, and have `.travis.xml` kick that off.

hannes-ucsc:
In response to @ttung's [comment](https://github.com/HumanCellAtlas/data-store/pull/925#discussion_r164573215) on the ES test case serialization, I've backed that change out and reduced the heap size for ES instead, hoping this will make them coexist more peacefully. But this is just a wild guess and we may still see client-side connection timeouts. However, this PR addresses a number of other causes for flakiness. 

@ttung, could you do another review pass?

Makefile
hannes-ucsc: That's worth considering. I would like to propose that we try the heap reduction for a few days and then explore your proposal in a separate PR.

tests/test_search.py
```diff
@@ -418,17 +404,19 @@ def check_count(self, es_query, expected_count, timeout=5):
...
-        super().search_setup(dss.Replica.aws)
+class TestAWSSearch(TestSearchBase):
+    replica = dss.Replica.aws
+
 
+# Prevent unittest's discovery from attempting to discover the base test class. The alterative, not inheriting
+# TestCase in the base class, is too inconvenient because it interferes with auto-complete and generates PEP-8
+# warnings about the camel case methods.
+#
+del TestSearchBase
```
Bento007: an alternative would be to move TestSearchBase into another file called searchBase.py and import it into test_search.

tests/test_subscriptions.py
```diff
@@ -243,16 +222,17 @@ def throw_403(self):
...
-    @classmethod
-    def setUpClass(cls):
-        super().subsciption_setup(dss.Replica.aws)
+    replica = dss.Replica.aws
 
+# Prevent unittest's discovery from attempting to discover the base test class. The alterative, not inheriting
+# TestCase in the base class, is too inconvenient because it interferes with auto-complete and generates PEP-8
+# warnings about the camel case methods.
+#
+del TestSubscriptionsBase
```
Bento007: Alternatively you could create subscriptionBase.py and import it into test_subscriptions.py

dss/config.py
```diff
@@ -58,12 +59,32 @@ class ESDocType(Enum):
...
 
 class IndexSuffix:
-    '''Creates a test specific index when in test config.'''
-    name = ''  # type: typing.Optional[str]
+    """
+    Manage growing and shrinking suffixes to Elasticsearch index names
+    """
+    def __init__(self) -> None:
+        super().__init__()
+        self._stack: typing.Deque[str] = deque()
```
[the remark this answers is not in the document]
hannes-ucsc: Unlike a list, deques are optimized for appending/removing at the ends.

dss/util/es.py
```diff
@@ -62,8 +62,7 @@ def __init__(self, startup_timeout_seconds: int=60) -> None:
                     "-E", f"path.data={self.tempdir.name}",
                     "-E", "logger.org.elasticsearch=warn"
                 ],
-            )
-
+                env=dict(os.environ, ES_JAVA_OPTS="-Xms512m -Xmx512m"))
```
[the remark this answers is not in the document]
hannes-ucsc: Will do.

tests/test_search.py
hannes-ucsc: That would be an alternative, but I am not convinced a desirable one. In either case, I'd rather not mix a code moving change into this PR for obvious reasons. With this particular change, these tests finally become usable in PyCharm. I can right click a failed test an rerun it, something that wasn't possible before.

tests/test_subscriptions.py
hannes-ucsc: See above.

tests/infra/server.py
```diff
@@ -75,6 +75,8 @@ def _inject_api_requests_methods(cls):
     def shutdown(self):
         if self._server is not None:
             self._server.server.shutdown()
+        self.join(timeout=10)
```
kislyuk: Is 10 seconds always enough? I wonder if this should be made 30 or 60.

hannes-ucsc: I'll make it 30.

Makefile
```diff
@@ -8,35 +8,26 @@ mypy:
...
-	| sed -e "s/^/$$$$ /"
+.PHONY: $1_test $$($1_test_srcs)
+endef
 
-all_test: lint mypy $(all_test_srcs)
-	coverage combine
-	rm -f .coverage.*
+$(eval $(call test_rules,standalone,standalone))
+$(eval $(call test_rules,integration,integration))
+$(eval $(call test_rules,all,integration standalone))
```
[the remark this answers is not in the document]
hannes-ucsc: The tests aren't actually run serially anymore. That was in an earlier version of this PR and I backed *that* change out after @ttung pointed at an unintended side-effect. I left *this* change in because while increasing complexity, it reduces duplication. So in a way it is more magic, less alphabet soup.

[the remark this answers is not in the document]
kislyuk: We need to refactor our Makefile to make it readable.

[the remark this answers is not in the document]
hannes-ucsc: No such thing as a readable Makefile. But I'll back this out. I think it's an improvement because it is less alphabet soup, but more delicious. Only half kidding.

PR #935, opened 2018-01-30
3/4: Log retried exceptions only once (fixes #922)

### LINKED ISSUE #922: Outer @retry should not log stack traces already logged by inner @retry
https://gist.github.com/hannes-ucsc/a3ff655a438373673d2411441a2eb603#file-travis-log-L1235

We should also pass `stack_info=True` when logging the exception so we get the entire stack and not just up to `wrapper()`.

hannes-ucsc:
When there are nested @retry transactions as in

```python
@retry
def f():
    g()

@retry
def g():
    assert False
```

an exception in an inner transaction gets logged by the outer transaction, too.

Additionally, when logging an exception, this PR adds logging of the outer stack frames, too. Those are the ones leading up to the call of the function decorated with `@retry`.

### DIFF
--- a/dss/util/retry.py
+++ b/dss/util/retry.py
@@ -88,8 +88,8 @@ class retry:
     Traceback (most recent call last):
       ...
     Exception
-    >>> logger.method_calls[0]
-    call.warning("An exception occurred in '%s' of %r", 'foo', {}, exc_info=True)
+    >>> logger.method_calls[1]
+    call.warning('An exception occurred in %r.', retry.TX(name='foo', {}), exc_info=True, stack_info=True)
 
     If one decorated function calls another decorated function, the inner function will be retried independently and
     each inner TX will commence with an empty context and its own timeout and logger. While this seems desirable it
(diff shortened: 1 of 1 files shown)


dss/util/retry.py
```diff
@@ -202,29 +206,31 @@ def wrapper(*args, **kwargs):
                 try:
                     return f(*args, **kwargs)
                 except BaseException as e:
-                    tx.logger.warning("An exception occurred in '%s' of %r", name, tx, exc_info=True)
+                    tx.log_exception(e)
                     if self.retryable(e):
                         attempt = next(attempts, None)
                         if attempt is None:
-                            tx.logger.debug("Exceeded retry limit for '%s' of %r", name, tx)
+                            tx.logger.info("Exceeded retry limit for %r.", tx)
```
mikebaumann: If a retry limit is exceeded, I am inclined to think it should be logged at `warning` rather than `info`, should it not?

hannes-ucsc: I was going back and forth on this but I think you are right.

PR #936, opened 2018-01-30
Separate s3urlcache into it’s own file.

Bento007:


hannes-ucsc:
Ideally, we should wait with requesting a review until the PR is green.

tests/test_s3urlcache.py
```diff
@@ -11,16 +11,17 @@
 import requests
 from http.server import BaseHTTPRequestHandler, HTTPServer
 
+
+
```
xbrianh: Three blank lines seems excessive.

PR #941, opened 2018-01-31
1a/2: Serialize tests that use local ES instance; increase ES heap

hannes-ucsc:
This change does three things: 

1) It works around 

https://github.com/travis-ci/travis-ci/issues/8408

which is the saving grace that currently prevents builds from failing all the time on master. Without this Travis bug, ES would actually run with only 512MiB of heap (introduced in 3fdc29683764d079a4e7d65743957a4ed107accb) and fail most of the time.

2) It bumps up ES heap to 1.6GiB (initial and max, as recommended by ES docs)

3) Serializes the ES-based tests which makes them more stable.

This change is a variation of one that was backed out of #859 due to @ttung's concern that serializing the tests in the Makefile interferes with running them individually and @kislyuk's concern of being too complex syntactically. I've addressed the former by introducing a fourth set of test rules and the latter by adding comments. 

I consider this still a hacky patch until we find a better way of running the tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,35 +8,59 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test_srcs := $(wildcard tests/test_*.py)
-all_test_srcs := $(addprefix all__, $(test_srcs))
-integration_test_srcs := $(addprefix integration__, $(test_srcs))
 
-test: lint mypy $(test_srcs)
+define test_rules
+
+# Generate a set of rules for running test scripts, either all of them or
+# individually, for a given value of DSS_TEST_MODE. Depending on that value,
+# certain test cases or test methods within the test scripts will skipped.
+#
+# $1 is the prefix for the name of the phony target that runs all test scripts
+# (can be empty)
+# 
+# $2 is the prefix that's added to each test script name to create a phony
+# target for running that script individually (if empty, the actual script name
+# is used)
+# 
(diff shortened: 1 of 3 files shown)


ttung:
I think this is fine, except I would flip the polarity, i.e., make -j test should be fully parallelized and make -j safe_test should serialize the ES tests.  Only travis would ever invoke safe_test.

We should seriously consider a different test strategy, as this is incredibly hard to read, and few will be able to make changes to this.

Makefile
```diff
@@ -8,35 +8,41 @@ mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
 test_srcs := $(wildcard tests/test_*.py)
-all_test_srcs := $(addprefix all__, $(test_srcs))
-integration_test_srcs := $(addprefix integration__, $(test_srcs))
 
-test: lint mypy $(test_srcs)
-	coverage combine
-	rm -f .coverage.*
+define test_rules
```
ttung: Suggestion: Comment here for what $(1), $(2), and $(3) are expected to be.

hannes-ucsc:
Agreed on adding `make safe_test` and explaining `$1`, `$2`, ….

hannes-ucsc:
@kislyuk Yes, that's better. I'll give it a shot.

hannes-ucsc:
There was an unrelated test failure. Tracked in https://github.com/HumanCellAtlas/data-store/issues/956.

hannes-ucsc:
Closing in favor of #955.

PR #955, opened 2018-02-07
1b/2: Serialize tests that use local ES instance; increase ES heap

hannes-ucsc:
A variant of https://github.com/HumanCellAtlas/data-store/pull/941, based on @kislyuk's [preference](https://github.com/HumanCellAtlas/data-store/pull/941#pullrequestreview-94536735) but with the following differences:

* retains @ttung's [requirement](https://github.com/HumanCellAtlas/data-store/pull/941#issuecomment-363554076) that, by default, tests are run in parallel
* keep standalone the default test mode, as currently on master
* serial tests are run in parallel to parallel test (but obviously not in parallel to each other)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,36 +7,44 @@ lint:
 mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
-test_srcs := $(wildcard tests/test_*.py)
-all_test_srcs := $(addprefix all__, $(test_srcs))
-integration_test_srcs := $(addprefix integration__, $(test_srcs))
-
-test: lint mypy $(test_srcs)
+tests:=$(wildcard tests/test_*.py)
+serial_tests:=tests/test_search.py \
+	          tests/test_indexer.py \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,36 +7,44 @@ lint:
...
+safe_test: _serial_test $(parallel_tests)
 	coverage combine
 	rm -f .coverage.*
 
-$(integration_test_srcs): integration__%.py :
-	set -o pipefail \
-	&& DSS_TEST_MODE="integration" coverage run -p --source=dss -m unittest $(DSS_UNITTEST_OPTS) $*.py 2>&1 \
-	| sed -e "s/^/$$$$ /"
+_serial_test:
+	for test in $(serial_tests); do $(MAKE) $$test; done
```
ttung: You can just do:

$(MAKE) -j1 $(serial_tests)

hannes-ucsc: Will do.

PR #963, opened 2018-02-08
callbacks for `Future.add_done_callback` expect a single argument.

### LINKED ISSUE #961: Exception in dss-s3-copy-write-metadata-sfn processing: TypeError: callback() takes 0 positional arguments but 1 was given
### Expected behavior
Put file of large files, requiring async copy, should complete successfully, without exception. 
### Actual behavior
The following exception was observed repeatedly in the CloudWatch logs for `dss-s3-copy-write-metadata-sfn`:
```
[ERROR]	2018-01-30T20:53:53.538Z	acec327c-05ff-11e8-a072-3133a77d17a2	exception calling callback for <Future at 0x7f221a64ca20 state=finished returned NoneType>
File "/var/lang/lib/python3.6/concurrent/futures/_base.py", line 297, in _invoke_callbacks
callback(self)
TypeError: callback() takes 0 positional arguments but 1 was given
```
This prevented the put file operation from completing.
### On which version of the code did the bug occur? (git hash, date, branch, be specific)
This was observed in a deployment of the data-store master branch at commit 30c7cf1371a9edf2de16f93a3f1b22ed640b8b31 in the UCSC AWS platform-dev account.

@hannes-ucsc identified this could be resolved by adding `_` as a parameter as shown in code snippet below from `data-store/dss/stepfunctions/s3copyclient/implementation.py`:
```
            def make_on_complete_callback(part_id: int):
                def callback(_):
                    with state_lock:
                        if part_id >= state[_Key.NEXT_PART]:
                            state[_Key.NEXT_PART] = part_id + 1
                            self.save_state(state)

                return callback
```

ttung:
In this case, we don't use it, so just throw it away.

Fixes #961

### DIFF
--- a/dss/stepfunctions/s3copyclient/implementation.py
+++ b/dss/stepfunctions/s3copyclient/implementation.py
@@ -114,7 +114,7 @@ def copy_worker(event, lambda_context, branch_id):
             state_lock = threading.Lock()
 
             def make_on_complete_callback(part_id: int):
-                def callback():
+                def callback(_):
                     with state_lock:
                         if part_id >= state[_Key.NEXT_PART]:
                             state[_Key.NEXT_PART] = part_id + 1


mikebaumann:
This seems like an issue that should have been caught by an integration test that performs a put file of a large (async copy) file.
Is there such a test? If so, do we know why this was not detected?

ttung:
Yes, there are tests that exercise this codepath.  Not sure why they're not triggering.

hannes-ucsc:
I think the exception in #961 (the one that prevents the execution of the callback) was inconsequential because failing to update NEXT_PART (which is what the callback does) only prevents a minor optimization when querying for the parts yet to be copied. It would be hard to test for that optimization.

PR #970, opened 2018-02-09
[fix-s3-copy 2/2] support small files in s3 copy

ttung:
Currently, the s3 copy step function flips out when it is passed a file that's too small for multipart copy.  It loops indefinitely in the first branch of the parallel section, which is not good.

This takes into account the possibility we might be asked to copy a small object.  If that happens, it copies during `setup_copy_task` and jumps straight to the end.

(Technically, this is an independent diff from #969...)


ttung:
Updated to fix the filesize=0 case.

PR #974, opened 2018-02-11
[package-reorg 2/2] Package reorganization

hannes-ucsc:
Convert from relative to absolute imports where applicable. Specifically, all `..` and `...` imports are absolute now (see below for motivation). Most `.` imports remain relative. Collect ES-related modules in `dss.index.es` and storage-related modules in `dss.storage`.

Only moved files around and updated imports as required. Retained history and authorship.

```
dss                                           dss                                  
├── api                                       ├── api                              
│   ├── ...                                   │   ├── ...                          
│   ...                                       │   ...                              
├── events                                    ├── events                           
│   ├── ...                                   │   ├── ...                          
│   ...                                       │   ...                              
├── hcablobstore ───────────────────────┐     ├── index                            
│   ├── __init__.py                     │     │   ├── es                           
│   ├── gs.py               ┌───────────┼───▶ │   │   ├── __init__.py              
│   └── s3.py               │   ┌───────┼───▶ │   │   ├── document.py              
├── stepfunctions           │   │ ┌─────┼───▶ │   │   ├── manager.py               
│   ├── ...                 │   │ │ ┌───┼───▶ │   │   ├── mapping.json             
│   ...                     │   │ │ │ ┌─┼───▶ │   │   └── validator.py             
├── storage                 │   │ │ │ │ │     │   └── __init__.py                  
│   ├── __init__.py         │   │ │ │ │ │     ├── stepfunctions                    
│   ├── bundles.py ─────────┼─┐ │ │ │ │ │     │   ├── ...                          
│   ├── index.py ───────────┼─┼─┼─┘ │ │ │     │   ...                              
│   ├── index_document.py ──┼─┼─┘   │ │ │     ├── storage                          
│   ├── mapping.json ───────┼─┼─────┘ │ └───▶ │   ├── hcablobstore                 
│   └── validator.py ───────┼─┼───────┘       │   │   ├── __init__.py              
├── util                    │ │               │   │   ├── gs.py                    
│   ├── aws                 │ │               │   │   └── s3.py                    
│   ├── aws                 │ │               │   ├── __init__.py                  
│   │   ├── ...             │ │ ┌───────────▶ │   ├── blobstore.py                 
│   │   ...                 │ │ │ ┌─────────▶ │   ├── bundles.py                   
│   ├── state_machine ────┐ │ │ │ │ ┌───────▶ │   ├── checkout.py                  
│   │   └── __init__.py   │ │ └─┼─┼─┼───────▶ │   └── identifiers.py               
│   ├── __init__.py       │ │   │ │ │         ├── util                             
│   ├── blobstore.py ─────┼─┼───┘ │ │         │   ├── aws                          
│   ├── bundles.py ───────┼─┼─────┘ │         │   │   ├── __init__.py              
│   ├── checkout.py ──────┼─┼───────┘         │   │   ├── _boto3_loader.py         
│   ├── email.py          │ │                 │   │   ├── clients.py               
│   ├── es.py ────────────┼─┘                 │   │   ├── cloudwatch_logging.py    
│   ├── networking.py     │                   │   │   └── resources.py             
│   ├── retry.py          │                   │   ├── __init__.py                  
│   ├── s3urlcache.py     │                   │   ├── email.py                     
│   ├── streaming.py      │                   │   ├── networking.py                
│   └── version.py        │                   │   ├── retry.py                     
├── __init__.py           │                   │   ├── s3urlcache.py                
├── config.py             │                   │   ├── streaming.py                 
├── error.py              │                   │   └── version.py                   
└── logging.py            │                   ├── __init__.py                      
                          │                   ├── config.py                        
                          │                   ├── error.py                         
                          ▼                   └── logging.py                                                                                                          
```

Relative vs. absolute imports
-----------------------------

While . relative imports are useful, the .. and ... imports are hard to read and make sorting imports harder. They also interfere with PyCharm's "Move module" and "Rename module"  refactorings. 

My concern with readability of relative imports is that they don't align on the left, because you need a different number of dots, depending how deep your module is in the package tree. They are harder to parse for a human reader because the location of the importing module matters. Absolute imports are explicit and don't depend on that additional piece of context to be parsed.

Relative imports have the theoretical advantage of making a package more relocatable, but in reality that doesn't pan out because they actually confuse the IDEs, rather than help.

Lastly, some of our modules are packaged by *kind*, not purpose (`events` and `stepfunctions`, for example). There is little cohesion between modules in the same package, leading to an abundance of relative imports that have to `..` all the way up to the root and then descend down into a different branch of the package tree. This PR actually improves on that, opening the door for more relative `.` imports, which I would still favor.


ttung:
I had an argument against absolute imports, but it turns out I was wrong!  The only case where it is relevant is when the high-level structure of the project changes or if this is dropped as a submodule of another project, but other than that, I think this is solid.

hannes-ucsc:
@ttung, @xbrianh if you approve or disapprove could you mark this PR as such? If you suspect this to be contentious enough for a majority of the team or its leadership to dissent, you could ask for more reviewers or request that more than one approval is needed.

PR #977, opened 2018-02-13
[URGENT] Fix: test_json_faker fails stochastically

hannes-ucsc:
Only every 16th build succeeds. The exact error message is

```
Error: Cannot create property '_randexp' on string '.{8}-.{4}-.{4}-.{4}-.{12}' in /properties/hca_ingest/properties/document_id
```

This is caused by a recent release of randexp, a transitive dependency, upgrading it from `0.4.6` to `0.4.7`. The fix is to pin that dependency. In order to pin the dependency, we need to switch to using `package-lock.json` in conjuction with `npm install` which is the prescribed way of installing node.js dependencies reproducibly. Another fix would be to bump `json-schema-faker` to `0.4.7` but unfortunately that's not as easy because `json-schema-faker-cli` depends on it with a `^0.3.6` (essentially `0.3.x`) spec. Bumping to `0.4.7` would conflict with that spec, causing `json-schema-faker-cli` to get its own `node_modules` directory, rendering the bump ineffective.

This is the PR build that shows the test as passing, prior to the PR landing:

https://travis-ci.org/HumanCellAtlas/data-store/builds/338217003

This is the build on master after the PR landed:

https://travis-ci.org/HumanCellAtlas/data-store/builds/340713380

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ this directory.
 
 The HCA DSS prototype requires Python 3.6+ to run. Run `pip install -r requirements.txt` in this directory.
 
+The tests require certain node.js packages. They must be installed using `npm`, the node.js package manager. Install 
+[npm](https://www.npmjs.com/get-npm) (on macOS `brew install npm`) and run `npm install` from the project root.
+
 #### Pull sample data bundles
 
 Tests also use data from the data-bundle-examples subrepository. Run: `git submodule update --init`
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,7 @@ before_install:
 install:
 - pip install -r requirements.txt
 - pip install -r requirements-dev.txt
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -27,6 +27,9 @@ this directory.
 
 The HCA DSS prototype requires Python 3.6+ to run. Run `pip install -r requirements.txt` in this directory.
 
+The tests require certain NodeJS packages. They must be installed using `npm`, the NodeJS package manager. Install
```
kislyuk: s/NodeJS/node.js/

hannes-ucsc: Fixed.

hannes-ucsc:
Modified the fix to not require a fork of json-schema-faker-cli. Thanks to @Bento007 for helping with this.

PR #979, opened 2018-02-14
V5 meta support

Bento007:
### Test plan
- adding test cases for `shape_descriptor` and `scrub_index_data`

### Release notes
v5 of the meta data uses `describedBy` to specify versions and type. Extra fields will be removed from v5 metadata bundle  and v5 bundle will be index by major version using `describedBy`. The `core` field in previous version of the metadate will continue to be supported.

### DIFF
--- a/dss/index/es/document.py
+++ b/dss/index/es/document.py
@@ -18,6 +18,7 @@ from dss.index.bundle import Bundle, Tombstone
 from dss.index.es import ElasticsearchClient, elasticsearch_retry
 from dss.index.es.manager import IndexManager
 from dss.index.es.validator import scrub_index_data
+from dss.index.es.schemainfo import SchemaInfo
 from dss.storage.identifiers import BundleFQID, ObjectIdentifier
 
 logger = logging.getLogger(__name__)
@@ -232,15 +233,14 @@ class BundleDocument(IndexDocument):
         """
         schema_version_map = defaultdict(set)  # type: typing.MutableMapping[str, typing.MutableSet[str]]
         for file_name, file_content in self.files.items():
-            core = file_content.get('core')
-            if core is not None:
-                schema_type = core['type']
-                schema_version = core['schema_version']
-                schema_version_major = schema_version.split(".")[0]
-                schema_version_map[schema_version_major].add(schema_type)
+            schema_info = SchemaInfo.from_json(file_content)
+            if schema_info is not None:
+                schema_version_map[schema_info.version].add(schema_info.type)
             else:
-                logger.info(f"File {file_name} does not contain the 'core' section necessary to identify "
-                            f"the schema and its version.")
+                logger.warning(f"Unable to obtain JSON schema info from file '{file_name}'. The file will be indexed "
+                               f"as is, without sanitization. This may prevent subsequent, valid files from being "
+                               f"indexed correctly.")
+
         if schema_version_map:
             schema_versions = schema_version_map.keys()
             assert len(schema_versions) == 1, \
--- a/dss/index/es/schemainfo.py
+++ b/dss/index/es/schemainfo.py
@@ -7,7 +7,7 @@ logger = logging.getLogger(__name__)
 
 # https://somewhere.anything/1.2.3/anything/type123.json
 # "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/project_bundle.json"
-schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w\d/]*/(?P<type>[\w]+)(.json)?$')
+schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w/]*/(?P<type>[\w]+)(\.json)?$')
 
 
 class SchemaInfo(namedtuple('SchemaInfo', ['url', 'version', 'type'])):
--- a/dss/index/es/schemainfo.py
+++ b/dss/index/es/schemainfo.py
@@ -1,0 +1,30 @@
+import re
+from collections import namedtuple
+from typing import Optional
+import logging
+
+logger = logging.getLogger(__name__)
+
+# https://somewhere.anything/1.2.3/anything/type123.json
+# "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/project_bundle.json"
+schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w\d/]*/(?P<type>[\w]+)(.json)?$')
+
+
+class SchemaInfo(namedtuple('SchemaInfo', ['url', 'version', 'type'])):
+    @classmethod
+    def from_json(cls, data: dict) -> Optional['SchemaInfo']:
+        if data.get('describedBy'):
+            url = data['describedBy']
+            version, doc_type = schema_url_regex.search(url).group('major_version', 'type')
+            info = cls(url, version, doc_type)
+        elif data.get('core'):
+            # TODO: Remove this if block once we stop supporting the `core` property (see issue #1015).
+            doc_type = data['core']['type']
+            version = data['core']['schema_version'].split(".")[0]
+            url = data['core']['schema_url']
+            info = cls(url, version, doc_type)
+        else:
+            # TODO: Remove 'core' from log message once we stop supporting the `core` property (see issue #1015).
+            logger.info("Need either 'describedBy' or 'core' property to determine JSON schema. Both are absent.")
+            info = None
+        return info
--- a/dss/index/es/validator.py
+++ b/dss/index/es/validator.py
@@ -5,8 +5,8 @@ import json
 from typing import List
 import logging
 
+from dss.index.es.schemainfo import SchemaInfo
 from dss.util.s3urlcache import S3UrlCache
-
 logger = logging.getLogger(__name__)
 
 
@@ -34,7 +34,6 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]) -> N
     for field in fields:
         current.pop(field)
 
-
 def scrub_index_data(index_data: dict, bundle_id: str) -> list:
     cache = S3UrlCache()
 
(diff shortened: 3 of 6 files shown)


Bento007:
Waiting for metadata V5 to be released

Bento007:
@hannes-ucsc per your request, please review.

dss/index/es/document.py
```diff
@@ -232,15 +232,15 @@ def get_shape_descriptor(self) -> typing.Optional[str]:
...
         schema_version_map = defaultdict(set)  # type: typing.MutableMapping[str, typing.MutableSet[str]]
         for file_name, file_content in self.files.items():
-            core = file_content.get('core')
-            if core is not None:
-                schema_type = core['type']
-                schema_version = core['schema_version']
-                schema_version_major = schema_version.split(".")[0]
-                schema_version_map[schema_version_major].add(schema_type)
+            try:
+                schema_info = HcaSchema.from_schema(file_content)
```
hannes-ucsc: `HcaSchemaInfo`?

dss/index/es/validator.py
```diff
@@ -34,6 +37,33 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]) -> N
...
 
+schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w\d/]*/(?P<type>[\w\d]+)(.json)?$')
+
+
+class HcaSchema(namedtuple('HcaSchema', ['url', 'version', 'type'])):
+    getters = ['describedBy', 'core']
+
+    @classmethod
+    def from_schema(cls, schema: dict) -> 'HcaSchema':
+        getter = [i for i in schema.keys() if i in cls.getters]
```
hannes-ucsc: `[k for k,v in HcaSchema.__dict__.items() if isinstance(v,staticmethod)]`

and don't call it `getter`

dss/index/es/validator.py
```diff
@@ -34,6 +37,33 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]) -> N
...
+        assert len(getter) == 1, "Exactly one HCA schema describer allowed."
+        url, version, schema_type = getattr(cls, getter[0])(cls, schema)
+        return cls(url=url, version=version, type=schema_type)
+
+    @staticmethod
+    def describedBy(schema) -> Tuple[str, str, str]:
```
hannes-ucsc: -> 'HcaSchema' and like we talked about

dss/index/es/validator.py
```diff
@@ -34,6 +37,33 @@ def remove_json_fields(json_data: dict, path: List[str], fields: List[str]) -> N
...
         current.pop(field)
 
+schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w\d/]*/(?P<type>[\w\d]+)(.json)?$')
+
+
+class HcaSchema(namedtuple('HcaSchema', ['url', 'version', 'type'])):
+    getters = ['describedBy', 'core']
+
+    @classmethod
+    def from_schema(cls, schema: dict) -> 'HcaSchema':
```
hannes-ucsc: not a schema and use dss.util.type.JSON

Bento007: Fixed

dss/index/es/document.py
Bento007: fixed

tests/test_indexer.py
```diff
@@ -520,14 +520,56 @@ def test_get_shape_descriptor(self):
         with self.subTest("An versioned file and an unversioned file"):
             with self.assertLogs(dss.logger, level="INFO") as log_monitor:
                 index_document.get_shape_descriptor()
-            self.assertRegex(log_monitor.output[0], "^INFO:.*File assay_json does not contain the 'core' section "
-                                                    "necessary to identify the schema and its version.", )
+            self.assertRegex(log_monitor.output[0], f"^INFO:.*File assay_json does not contain the 'core' or "
```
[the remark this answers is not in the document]
Bento007: fixed

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,33 @@
...
+from collections import namedtuple
+from typing import Optional
+
+from dss.util.types import JSON
+
+schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w\d/]*/(?P<type>[\w\d]+)(.json)?$')
+
+
+class SchemaInfo(namedtuple('SchemaInfo', ['url', 'version', 'type'])):
+    info_formats = ['describedBy', 'core']
```
hannes-ucsc: remove

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @classmethod
+    def from_json(cls, data: JSON) -> Optional['SchemaInfo']:
+        if data.get('describedBy'):
+            try:
```
hannes-ucsc: this should be strict again

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,33 @@
...
+                url = data['describedBy']  # type: str
+                version, doc_type = schema_url_regex.search(url).group('major_version', 'type')
+                info = cls(url, version, doc_type)
+            except (KeyError, AttributeError):
+                info = None
+        elif data.get('core'):
+            # TODO removed once core is not supported git issue #1015
+            try:
```
hannes-ucsc: should be strict too

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,25 @@
+import re
+from collections import namedtuple
+from typing import Optional
+
+from dss.util.types import JSON
```
hannes-ucsc: not used

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,25 @@
...
+    @classmethod
+    def from_json(cls, data: dict) -> Optional['SchemaInfo']:
+        if data.get('describedBy'):
+            url = data['describedBy']
+            version, doc_type = schema_url_regex.search(url).group('major_version', 'type')
+            info = cls(url, version, doc_type)
+        elif data.get('core'):
+            # TODO remove once core is not supported. Github issue #1015
+            doc_type = data['core']['type']  # type: str
+            version = data['core']['schema_version'].split(".")[0]  # type: str
```
hannes-ucsc: looking at the regexp, I think you don't need the split

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,25 @@
+import re
+from collections import namedtuple
+from typing import Optional
+
+from dss.util.types import JSON
+
+schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w\d/]*/(?P<type>[\w\d]+)(.json)?$')
```
hannes-ucsc: Can you add a comment with an example URL before the regex? It would help interpreting the URL.

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,25 @@
...
+class SchemaInfo(namedtuple('SchemaInfo', ['url', 'version', 'type'])):
+    @classmethod
+    def from_json(cls, data: dict) -> Optional['SchemaInfo']:
+        if data.get('describedBy'):
+            url = data['describedBy']
+            version, doc_type = schema_url_regex.search(url).group('major_version', 'type')
+            info = cls(url, version, doc_type)
+        elif data.get('core'):
+            # TODO remove once core is not supported. Github issue #1015
+            doc_type = data['core']['type']  # type: str
```
hannes-ucsc: We still need those type overrides?

Bento007: removed

dss/index/es/schemainfo.py
Bento007: added

dss/index/es/schemainfo.py
Bento007: but the regex is only used in `describedBy`. `core.schema_version` is expected to have the format 1.1.1.

hannes-ucsc: You're absolutely right.

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,24 @@
...
+
+class SchemaInfo(namedtuple('SchemaInfo', ['url', 'version', 'type'])):
+    @classmethod
+    def from_json(cls, data: dict) -> Optional['SchemaInfo']:
+        if data.get('describedBy'):
+            url = data['describedBy']
+            version, doc_type = schema_url_regex.search(url).group('major_version', 'type')
+            info = cls(url, version, doc_type)
+        elif data.get('core'):
+            # TODO remove once core is not supported. Github issue #1015
```
hannes-ucsc:
```
# TODO: Remove this test once we stop supporting the `core` property (see issue #1015).
```

dss/index/es/schemainfo.py
```diff
@@ -0,0 +1,24 @@
+import re
+from collections import namedtuple
+from typing import Optional
+
+# The string: 'https://somewhere.anything/1.2.3/anything/type123.json' will return major_version=1, type=type123
+schema_url_regex = re.compile(r'/(?P<major_version>[0-9]+)\.[0-9]+\.[0-9]+[\w\d/]*/(?P<type>[\w\d]+)(.json)?$')
```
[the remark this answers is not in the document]
Bento007: `\d` can potentially match other digit like characters not just [0-9]. I wanted to keep as explicit as possible when possible. Do you have another suggestion for the regex?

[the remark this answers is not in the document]
hannes-ucsc: First of all, doesn't `\w` include `\d`?

> other digit like characters not just [0-9]

I see. I keep assuming that URLs are ASCII, which they are not. OTOH, it makes little sense to include non-arabic digits but exclude punctuation like dashes. Generally speaking, the regex should either be "permissive" (agnostic to the exact syntax of `type`) or "strict" (enforce an existing, explicit standard). Correct me if I am wrong, but I don't think there is a standard for allowed characters in HCA metadata entity names. Consequently, I recommend using a permissive regex, like `[^.]+`.

Please make the regex as permissive as possible while still allowing for the extraction of necessary fields.

Lastly, the dot in `.json` needs to be escaped.

PR #980, opened 2018-02-14
Parallelize slow tests again, using Travis' new Build Stages feature

hannes-ucsc:
This squeezes build time back to the 9 or so minutes again. I've seen 6 minutes, too. I also think the Travis config is more readable now as it untangles the various conditionals from the actual scripts.

~~I've also snuck in a change the ES server class. See the specific commit for my reasoning.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,30 @@
 language: python
+
 cache:
   pip: true
   directories:
   - chalice/.chalice/venv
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
test_search_returns_N_results_when_N_documents_match_query seems to be a slow test that could trigger the 10min output timeout.

PR #981, opened 2018-02-15
Fix smoketest

hannes-ucsc:
5f547b5 is the fix. This is a annoying since it seems that this test was never actually run because that assertion could never have succeeded.

I recommend running smoketest locally after making changes to it or at least watching the Travis IT build like a hawk after your PR lands. I know, it's annoying that we don't run ITs on PRs but it is the best we can do right now.


hannes-ucsc:
The diff coverage requirement can't be met since this affected code is not run in a PR build.

PR #983, opened 2018-02-16
Fix: spaces in DSS_NOTIFICATION_SENDER break smoke_test

hannes-ucsc:
Regression from e0d7aafc33e2ae35a2c01b184434ae2ab0a4fbc9.

### DIFF
--- a/tests/test_smoketest.py
+++ b/tests/test_smoketest.py
@@ -2,8 +2,7 @@
 """
 A basic integration test of the DSS. This can also be invoked via `make smoketest`.
 """
-
-import os, sys, argparse, time, uuid, json, shutil, tempfile, unittest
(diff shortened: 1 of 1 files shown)


kislyuk:
a bit overkill, i would just use quotes

hannes-ucsc:
The local part of an email can contain both single quotes and double quotes. shlex comes with the Python standard library so I don't really understand why using a standard library function is objectionable here. Furthermore, there is no guarantee that the user didn't introduce a typo when specifying `DSS_NOTIFICATION_SENDER` and that typo could cause at best a weird error message in smoke_test or at worst data corruption, e.g. when they add a unmatched single quote and put in the `<` etc etc.

When interpolating into shell strings one should always use shlex.quote. That's why it exists.

PR #984, opened 2018-02-16
Minor test improvements

hannes-ucsc:
Assorted fixes I stumbled upon during #980.

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -37,8 +37,10 @@ def post(json_request_body: dict,
 
     replica_enum = Replica[replica] if replica is not None else Replica.aws
 
-    logger.debug("Received posted query. Replica: %s Query: %s Per_page: %i Timeout: %s Scroll_id: %s",
-                 replica_enum.name, json.dumps(es_query, indent=4), per_page, _scroll_id)
+    if logger.isEnabledFor(logging.DEBUG):
+        logger.debug("Received POST for replica=%s, es_query=%s, per_page=%i, _scroll_id: %s",
+                     replica_enum.name, json.dumps(es_query, indent=4), per_page, _scroll_id)
+
     # TODO: (tsmith12) determine if a search operation timeout limit is needed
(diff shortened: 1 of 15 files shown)


tests/test_doctests.py
```diff
@@ -5,15 +5,20 @@
 Run selected doctests as a regular unit test.
 """
 
-import os
+import doctest
 import sys
+import unittest
+
```
kislyuk: one newline too many here

hannes-ucsc: PyCharm put that one there. I'm not sure why os is different from sys.

Will fix.

dss/api/search.py
```diff
@@ -37,8 +37,10 @@ def post(json_request_body: dict,
 
     replica_enum = Replica[replica] if replica is not None else Replica.aws
 
-    logger.debug("Received posted query. Replica: %s Query: %s Per_page: %i Timeout: %s Scroll_id: %s",
-                 replica_enum.name, json.dumps(es_query, indent=4), per_page, _scroll_id)
+    if logger.isEnabledFor(logging.DEBUG):
+        logger.debug("Received POST for replica=%s, es_query=%s, per_page=%i, _scroll_id: %s",
+                     replica_enum.name, json.dumps(es_query, indent=4), per_page, _scroll_id)
```
kislyuk: Wait, sorry, I missed this the first time around. Why is it necessary to gate `logger.debug` with `logger.isEnabledFor`? Shouldn't that logic happen automatically inside `logger`?

hannes-ucsc: Yes, the interpolation is only performed if log level is debug or lower. However, `json.dumps` isn't exactly cheap. This is the recommended way to gate expensive computation of interpolation arguments: https://docs.python.org/3/howto/logging.html#optimization

kislyuk: In general, I would recommend not worrying about the cost of json.dumps.

hannes-ucsc: I'll take it out.

PR #985, opened 2018-02-16
ES test server improvements

hannes-ucsc:
I still suspect some test hangs are caused by the code that spawns an Elasticsearch instance on demand during test setup. There was this outer loop that would essentially retry starting a new instance forever if it couldn't connect to recently started one. There were no log statements made during that loop so it would have been hard to detect in the build output. I took it out. The new approach will raise an exception if the situation occurs that the outer loop tried to mitigate. I think we would want to see this happen in the wild before we attempt to fix it.

The way the free ports are determined is still potentially racy. For example, there are two invocations of that function, one right after the other. There is no guarantee that the second one won't return the same port as the first one.

I've increased the logging verbosity of this setup such that we can see more easily if weird ports are being chosen.

### DIFF
--- a/dss/index/es/__init__.py
+++ b/dss/index/es/__init__.py
@@ -21,6 +21,7 @@ from dss.util import networking
 
 logger = logging.getLogger(__name__)
 
+
 class AWSV4Sign(requests.auth.AuthBase):
     """
     AWS V4 Request Signer for Requests.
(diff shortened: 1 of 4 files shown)


hannes-ucsc:
@ttung, care to chime in? You wrote this originally.

PR #987, opened 2018-02-16
Python Fake JSON generator

Bento007:
# Test plan
- created test cases for json_generator.py
- Added `JsonDataGenerator` to json_faker.py
- Removed JaveScript faker

# Release notes
`JsonDataGenerator` class will generate JSON based off the schema provided. The class uses [Faker](https://pypi.python.org/pypi/Faker) to generate fake data, [JSONschema](https://pypi.python.org/pypi/jsonschema) to dereference `$ref` properties, and S3urlcache to retrieve URL data. Only draft4 of json schema is currently supported.

### Validation Keywords for Any Instance Type
type, enum, const
### Validation Keywords for Numeric Instances (number and integer)
multipleOf for integers, maximum, exclusiveMaximum, minimum, exclusiveMinimum
### Validation Keywords for Strings
maxLength, minLength, format, pattern
### Validation Keywords for Arrays
items, maxItems, minItems, additionalItems, uniqueItems
### Validation Keywords for Objects 
maxProperties, minProperties, required, properties, additionalProperties, patternProperties

### Validation Keywords for Numeric Instances (number and integer)
multipleOf for floats
### Validation Keywords for Arrays
contains
### Validation Keywords for Objects
dependencies, propertyNames
### Keywords for Applying Subschemas Conditionally
if, then, else
### Keywords for Applying Subschemas With Boolean Logic
allOf, anyOf, oneOf, not


Bento007:
@hannes-ucsc I believe I have addressed your initial concerns.

hannes-ucsc:
One commit please.

The commit message for d978ee75c42b0ae2112f4eda7b6785b32b64034f is not formatted right. Look at how SourceTree is showing it as one long line.

File name of test module is wrong. Look at the other tests and the Makefile to see what the naming convention is. Before asking for another round of reviews, please make sure that 1) your test is actually run on Travis and 2) all tests pass.

tests/scalability/json_generator.py
```diff
@@ -0,0 +1,273 @@
...
+                item_count = len(items)
+                i = 0
+                retry = 3  # To prevent infinite loops
+                while i < item_count and retry:
+                    retry -= 1
+                    item = self._gen_json(items[i])
+                    if item not in impostor:
+                        impostor.append(item)
+                        i += 1
+                        retry = 3
```
hannes-ucsc: It's a bit brittle that this line duplicates the initial value for retry above the loop. Is it possible to make retry start at zero, be incremented instead of decremented and place the limit in the loop condition?

hannes-ucsc: Are you ok with getting fewer items than item_count in case the retries are exhausted. I guess you are since this is supposed to be random anyways.

tests/scalability/json_generator.py
```diff
@@ -0,0 +1,273 @@
...
+                if additional_items:
+                    unique_gen(additional_items)
+            else:
+                for item in items:
+                    impostor.append(self._gen_json(item))
+                if additional_items:
+                    simple_gen(additional_items)
+        return impostor
+
+    generators = {
```
hannes-ucsc: Since there seems to be a consistent naming convention, you could eliminate this dictionary and use `getattr()` instead.

tests/scalability/json_generator.py
```diff
@@ -0,0 +1,273 @@
+import random
+import rstr
+from faker import Faker
+from jsonschema import RefResolver
+from typing import Union
+
+
+class JsonDataGenerator(object):
```
hannes-ucsc: JSON is always data. I would call it JsonGenerator

tests/scalability/json_generator.py
```diff
@@ -0,0 +1,273 @@
...
+    """Generate a random json document based on the provided schema."""
```
hannes-ucsc: As usual, s/json/JSON here and in all other prose.

tests/scalability/json_generator.py
```diff
@@ -0,0 +1,273 @@
...
+    def generate_json(self, schema: dict) -> dict:
+        """
+        :param schema: the JSON schema to generate data from.
+        :return: generated JSON date
+        """
+        impostor = self._gen_json(schema)
+        return impostor
+
+    def _gen_json(self, schema: dict):
+        scope = schema.get(u"id")
```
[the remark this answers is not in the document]
Bento007: fixed

tests/scalability/json_generator.py
```diff
@@ -0,0 +1,273 @@
...
+        if scope:
+            self.resolver.push_scope(scope)
+        try:
+            # should be inside an object now other wise there should be a ref.
+            ref = schema.get(u"$ref")
+            json_type = schema.get(u"type", "object")
+            if ref is not None:
+                impostor = self._ref(ref)
+            else:
+                impostor = self.generators.get(json_type)(self, schema)
```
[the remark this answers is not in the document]
Bento007: changed

tests/scalability/json_generator.py
Bento007: The chance of getting a duplicate is pretty low to begin with. However there is a chance that minItems will not be satisfied in which case validation could would fail. Since we are producing random test data I think producing some invalid data is helpful for testing. It's worth revisiting in the future if we continue to use this solution.

tests/scalability/json_generator.py
Bento007: fixed

tests/scalability/json_generator.py
Bento007: fixed

tests/scalability/json_generator.py
Bento007: Thought I got them all. Fixed

tests/scalability/json_generator.py
```diff
@@ -0,0 +1,273 @@
...
+        """
+        impostor = self._gen_json(schema)
+        return impostor
+
+    def _gen_json(self, schema: dict):
+        scope = schema.get(u"id")
+        if scope:
+            self.resolver.push_scope(scope)
+        try:
+            # should be inside an object now other wise there should be a ref.
```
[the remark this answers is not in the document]
Bento007: fixed

Bento007:
@hannes-ucsc, I missed that refactor. I'll change it.

PR #992, opened 2018-02-16
Fix sporadic failures in tests/test_file.py

ttung:
Because S3 only guarantees eventual consistency, some clients may have seen the file land while others may still not see the file has landed.  Therefore, we cannot guarantee that the second request will be synchronous.

Fixes #992

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -96,9 +96,11 @@ class TestFileApi(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
         src_key = generate_test_key()
         upload_func(test_bucket, src_key)
 
-        # should be able to do this twice (i.e., same payload, different UUIDs).  first time should be asynchronous
-        # since it's new data.  second time should be synchronous since the data is present.
-        for expect_async in [True, False]:
+        # We should be able to do this twice (i.e., same payload, different UUIDs).  First time should be asynchronous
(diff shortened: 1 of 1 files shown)


tests/test_file.py
```diff
@@ -96,9 +96,11 @@ def _test_file_put_large(self, replica: Replica, test_bucket: str, upload_func:
         src_key = generate_test_key()
         upload_func(test_bucket, src_key)
 
-        # should be able to do this twice (i.e., same payload, different UUIDs).  first time should be asynchronous
-        # since it's new data.  second time should be synchronous since the data is present.
-        for expect_async in [True, False]:
+        # We should be able to do this twice (i.e., same payload, different UUIDs).  First time should be asynchronous
+        # since it's new data.  Second time should be synchronous since the data is present, but because S3 does not
+        # make consistency guarantees, a second client might not see that the data is already there.  Therefore, we do
```
hannes-ucsc: That's not entirely true. S3 guarantees strong consistency for create-then-read, but eventual consistency for read-create-read, update-read, delete-read and all other combinations.

If both clients do check-if-exists-and-create-if-not then this is the right fix for the test. But maybe the code under test could be changed to try a creation first and then branch depending whether it succeeded or not. But I don't really know, your call.

ttung:
> So, essentially, _test_file_put_large now just tests that the request to upload the file was accepted?

not exactly.  second one can be synchronous (202).

PR #994, opened 2018-02-16
[easy] Add read permissions to daemon domovoilib

xbrianh:
Before deployment, add read permissions to domovoilib with
chmod -R ugo+r.


xbrianh:
@kislyuk Changed to `ugo+rX`

PR #996, opened 2018-02-20
Introduce the concept of index backends

hannes-ucsc:
This PR introduces the concept of an index backend (the abstract class `IndexBackend`), refactors the Elasticsearch-specific code to become a concrete implementation of that class (`ElasticsearchIndexBackend`) and adds a mechanism for invoking multiple backends in parallel (`CompositeIndexBackend`, another implementation of that class, using the Composite design pattern). 

Because concurrently invoking multiple backends necessitates the use of threads, that part of the code became a natural place for enforcing limits on the running time of indexing. This enforcement used to be the responsibility of the reindex visitation machinery (because that's where it mattered most). This PR also modifies the reindex visitation to integrate with the new timeout mechanism.


hannes-ucsc:
Requesting @kislyuk to be the primary reviewer.

dss/stepfunctions/visitation/reindex.py
```diff
@@ -58,49 +59,44 @@ def walker_finalize(self):
...
-            if timeout.did_timeout:
-                logger.warning(f'{self.work_id} timed out during reindex')
-                return
-
-            self.marker = blobs.start_after_key
-            self.token = blobs.token
-        else:
-            self._status = WalkerStatus.finished.name
+    def _walk(self) -> None:
+        backends = [ElasticsearchIndexBackend]
```
mweiden: nit: if this list and the list for the index call or there `CompositeIndexBackend` were centralized somewhere, it would decrease the chance that a future dev would add a new backend to the index app, but not to the reindex step function.

tests/test_indexer.py
```diff
@@ -87,6 +88,8 @@ def tearDownModule():
 class TestIndexerBase(ElasticsearchTestCase, DSSAssertMixin, DSSStorageMixin, DSSUploadMixin, metaclass=ABCMeta):
     bundle_key_by_replica = dict()  # type: typing.MutableMapping[str, str]
 
+    backends = [ElasticsearchIndexBackend]
```
mweiden: nit: here again, might be good to have this part of a global config so everything is synced

tests/test_indexer.py
```diff
@@ -105,7 +105,7 @@ def tearDownClass(cls):
 
     def setUp(self):
         super().setUp()
-        self.indexer = self.indexer_cls()
+        self.indexer = self.indexer_cls(backend.ElasticsearchIndexBackend())
```
mweiden: Part of the benefit of the separation of the indexing logic and the actual storage implementation is that this could be replaced with an in-memory implementation, no? z.B. have a `class InMemoryIndexBackend(IndexBackend)`. This might finally make this test fast!

If possible, I guess this improvement could go in a different commit.

tests/test_indexer.py
mweiden: Part of the benefit of the separation of the indexing logic and the actual storage implementation is that this could be replaced with an in-memory implementation, no? z.B. have a class InMemoryIndexBackend(IndexBackend). This might finally make this test fast!

If possible, I guess this improvement could go in a different commit.

dss/stepfunctions/visitation/reindex.py
hannes-ucsc: Yup. Will fix.

tests/test_indexer.py
hannes-ucsc: Fixed the nit.

Fast yes, but also anemic. The meat of the indexer is still in the ES backend, and is ES-specific. If we cut that out, there wouldn't be much left to test. Another cause of slowness is the fact that the metadata is still read from the buckets everytime for every test. That functionality should be tested once, and mocked for the rest of the tests.

PR #1001, opened 2018-02-21
[easy] Two small fixes; one a follow-up from #984

hannes-ucsc:
acc2353 eliminates lines in the build output like https://travis-ci.org/HumanCellAtlas/data-store/builds/344089603#L786 which I find misleading as it looks like "standalone" test mode is being activated, which is not always the case.

6399d59 is a follow-up from #984.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ lint:
 mypy:
 	mypy --ignore-missing-imports $(MODULES)
 
+export DSS_TEST_MODE?=standalone
+
 tests:=$(wildcard tests/test_*.py)
 serial_tests:=tests/test_search.py \
 	          tests/test_indexer.py \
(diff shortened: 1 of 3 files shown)


hannes-ucsc:
Requesting @kislyuk to be sole reviewer.

Makefile
```diff
@@ -34,21 +36,20 @@ serial_test:
...
 $(tests): %.py :
-	export DSS_TEST_MODE=$${DSS_TEST_MODE:-standalone} \
-	&& coverage run -p --source=dss $*.py $(DSS_UNITTEST_OPTS)
+	coverage run -p --source=dss $*.py $(DSS_UNITTEST_OPTS)
 
 # Run standalone and integration tests
 #
 all_test:
-	DSS_TEST_MODE="standalone integration" $(MAKE) test
+	$(MAKE) test DSS_TEST_MODE="standalone integration"
```
ttung: I would do `$(MAKE) DSS_TEST_MODE="standalone integration" test` but otherwise this change is fine by me.

hannes-ucsc: Fixed.

PR #1006, opened 2018-02-23
Only Supported JSON types are generated by jsongenerator

### LINKED ISSUE #1003: tests/test_hcagenerator.py fails sporadically
https://travis-ci.org/HumanCellAtlas/data-store/jobs/344596146#L835

```
ERROR: test_generation (__main__.TestHCAGenerator) [analysis_bundle.json]
  File "tests/test_hcagenerator.py", line 38, in test_generation
    fake_json = self.faker.generate(name)
  File "/home/travis/build/HumanCellAtlas/data-store/tests/json/hca_generator.py", line 41, in generate
    return json.dumps(generated_json)
  File "/opt/python/3.6.3/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/opt/python/3.6.3/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/python/3.6.3/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/python/3.6.3/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'Decimal' is not JSON serializable
```

I propose that the main JSON generation function/method always validates the generated JSON before returning it. This might make diagnosis of bugs *like* this one easier.

I was able to reproduce this locally with 100 repetitions of that test. Some failures mentioned `Decimal`, others mentioned `datetime`.

Bento007:
Fixes #1003.
Preventing Faker from generating types that are not valid json types.

@hannes-ucsc please review.

### DIFF
--- a/tests/test_hcagenerator.py
+++ b/tests/test_hcagenerator.py
@@ -14,7 +14,8 @@ schema_urls = [
     "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/analysis_bundle.json",
     "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/assay_bundle.json",
     "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/project_bundle.json",
-    "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/sample_bundle.json",
+    # TODO (tsmith) uncomment once boolean logic is supported in jsongenerator.py git issue #1012
+    # "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/sample_bundle.json",
 ]
 
 
@@ -32,12 +33,12 @@ class TestHCAGenerator(unittest.TestCase):
             name = url.split('/')[-1]
             self.assertEqual(self.faker.schemas[name], {'$ref': url, 'id': url})
 
-    @unittest.skip("Test is inconsistant")  # TODO (tsmith) remove once tests run consistently
     def test_generation(self):
-        for name in self.faker.schemas.keys():
-            with self.subTest(name):
-                fake_json = self.faker.generate(name)
-                self.assertIsInstance(json.loads(fake_json), dict)
+        for i in range(100):
+            for name in self.faker.schemas.keys():
+                with self.subTest(name):
+                    fake_json = self.faker.generate(name)
+                    self.assertIsInstance(json.loads(fake_json), dict)
 
 
 if __name__ == "__main__":
--- a/tests/test_jsongenerator.py
+++ b/tests/test_jsongenerator.py
@@ -1,15 +1,17 @@
+import json
 import os
 import sys
 import unittest
 from typing import Callable, Any, Dict, Union, Tuple
 
+from faker import Faker
 from jsonschema import Draft4Validator
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
 from tests.infra import testmode
-from tests.json_gen.generator import JsonGenerator
+from tests.json_gen.generator import JsonGenerator, JsonProvider
 
 type_mapping = {'string': str,
                 'object': dict,
@@ -220,6 +222,7 @@ simple_object = {'type': 'object', 'properties': {'thing_1': simple_integer, 'th
 
 
 class Base(unittest.TestCase):
+    repeat = 25
     def setUp(self):
         self.json_gen = JsonGenerator()
 
@@ -238,10 +241,8 @@ class Base(unittest.TestCase):
                     self.assertIn(func({'type': jtype, 'enum': enums}), enums)
 
 
-@unittest.skip("Test is inconsistant")  # TODO (tsmith) remove once tests run consistently
 @testmode.standalone
 class TestNumber(Base):
-    repeat = 5
     numbers = [10.0, 0.1, 0.0, -0.1, -10.0]
     multiple_ofs = [1.5, 2]  # TODO support multipleOf [0.1, -0.2] for floats
 
@@ -316,10 +317,8 @@ class TestNumber(Base):
                     self.assertEqual(value, 1.0)
 
 
-@unittest.skip("Test is inconsistant")  # TODO (tsmith) remove once tests run consistently
 @testmode.standalone
 class TestInteger(Base):
-    repeat = 10
 
     def test_common(self):
         self._test_common(self.json_gen._integer, 'integer', [1, 2], 999)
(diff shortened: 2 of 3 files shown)


tests/json/generator.py
```diff
@@ -1,10 +1,27 @@
...
 from typing import Union, List
 
 
+class JsonProvider(PythonProvider):
+
+    SUPPORTED_JSON_TYPES = ['str', 'str', 'str', 'str', 'float', 'float',
+                            'int', 'int', 'iso8601', 'uri', 'email']
+
+    def jsondict(self, nb_elements=10, variable_nb_elements=True, *value_types):
+        if not value_types:
```
hannes-ucsc: This introduces the possibility of the wrong types being sneaked in. I would just outright ban this case:

`assert not value_types, "Custom value types are not support for JSON"`

hannes-ucsc: You could do set difference but I just wouldn't go there.

tests/json/generator.py
```diff
@@ -1,10 +1,27 @@
...
 import rstr
 from faker import Faker
+from faker.providers.python import Provider as PythonProvider
 from jsonschema import RefResolver
 from typing import Union, List
 
 
+class JsonProvider(PythonProvider):
+
+    SUPPORTED_JSON_TYPES = ['str', 'str', 'str', 'str', 'float', 'float',
```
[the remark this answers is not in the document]
Bento007: You can affect the probability of producing specific types by specifying it more than once.

tests/json/generator.py
Bento007: I want to leave the option to use this for fuzz testing in the future. The defaults are safe but if the user wants to be more adventurous I don't want to limit them.

Bento007:
1. Is addressed in https://github.com/HumanCellAtlas/data-store/pull/1007

tests/json/generator.py
hannes-ucsc: We don't need this flexibility right now. I am more worried about correctness. Lets keep this simple for now and simply block this case.

tests/json/generator.py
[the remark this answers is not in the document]
hannes-ucsc: This would warrant a little comment in the code. 

What's your reasoning for this particular probability distribution?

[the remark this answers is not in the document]
Bento007: It was the distribution in the original Faker but modified to not include invalid json types.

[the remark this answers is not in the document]
hannes-ucsc: Is there any value in not having a uniform distribution?

tests/json/generator.py
```diff
@@ -8,19 +8,26 @@
 
 class JsonProvider(PythonProvider):
 
-    SUPPORTED_JSON_TYPES = ['str', 'str', 'str', 'str', 'float', 'float',
-                            'int', 'int', 'iso8601', 'uri', 'email']
+    __SUPPORTED_JSON_TYPES = {'str', 'float', 'int', 'iso8601', 'uri', 'email', 'bool'}
```
hannes-ucsc: Technically, both of these attributes are protected (one `_`). Is there a reason one should be private and the other one public?

tests/json/generator.py
```diff
@@ -8,19 +8,26 @@
...
+    # The more times a value appears in the tuple the greater the probability of it being generated.
+    DEFAULT_VALUE_TYPES = ('str', 'str', 'str', 'str', 'float', 'float', 'int', 'int', 'iso8601', 'uri', 'email',
+                           'bool')
 
     def jsondict(self, nb_elements=10, variable_nb_elements=True, *value_types):
-        if not value_types:
-            value_types = self.SUPPORTED_JSON_TYPES
+        value_types = self._check_value_types(*value_types)
```
hannes-ucsc: Drop the `*`

tests/json/generator.py
```diff
@@ -8,19 +8,26 @@
...
         return self.pydict(nb_elements, variable_nb_elements, *value_types)
 
     def jsonlist(self, nb_elements=10, variable_nb_elements=True, *value_types):
-        if not value_types:
-            value_types = self.SUPPORTED_JSON_TYPES
+        value_types = self._check_value_types(*value_types)
```
hannes-ucsc: Drop the `*`

tests/json/generator.py
```diff
@@ -8,19 +8,26 @@
...
         return self.pylist(nb_elements, variable_nb_elements, *value_types)
 
+    def _check_value_types(self, *value_types) -> tuple:
```
hannes-ucsc: Drop the `*`

Bento007:
I need to add support for boolean logic in the json before I can fully support TestHCAGenerator, specifically sample.json because it uses `oneOf` and `allOf`. That will take some time.

hannes-ucsc:
I'm assuming the issue with `sample.json` was uncovered when reintroducing the TestHCAGenerator test case with validation and repetitions as I requested. If that is the case, let's leave the TestHCAGenerator skipped and create a separate issue/PR for addressing the `sample.json` issue. Or unskip TestHCAGenerator and take the `sample.json` schema out of the mix for that test. Your call.

I am doubtful as to whether the test for `JsonProvider` is necessary. There are 90 lines of test code for a 20 line class all of which is already covered by the existing tests. The added tests seem to be predominantly testing argument combinations to JsonProvider methods that are not actually used during the generation of random JSON as needed for this project. I do not want to be burdened with the cost of maintaining and running these tests unless they provide a value, for this project, today.

In the same vein, and correct me if I am wrong, the ability to pass in `value_types` to `JsonProvider` methods is not currently used by this project, and is not on the road map. Why are we speculatively supporting that functionality in this PR?

Bento007:
> I'm assuming the issue with sample.json was uncovered when reintroducing the TestHCAGenerator test case with validation and repetitions as I requested. 

Correct.

>If that is the case, let's leave the TestHCAGenerator skipped and create a separate issue/PR for addressing the sample.json issue.

Will do.

> I am doubtful as to whether the test for JsonProvider is necessary..

They do provide good insight into the actual behavior and will be helpful for future development. I will skip them, but I don't want to remove them.

> In the same vein, and correct me if I am wrong, the ability to pass in value_types to JsonProvider methods is not currently used by this project, and is not on the road map. Why are we speculatively supporting that functionality in this PR?

You are correct that we currently don't pass in any values. I thought the tool would be useful for other areas of testing and I was making it reusable.

tests/test_jsongenerator.py
```diff
@@ -637,20 +629,102 @@ def test_propertyNames(self):
...
-        schema_validator.check_schema(schema_analysis)
-        for i in range(20):
+        for i in range(self.repeat):
             with self.subTest(i):
-                gen_data = generate_json(schema_analysis)
-                schema_validator.validate(gen_data)
+                generate_json(schema_analysis)
+
+
+@unittest.skip("Low Priority")
```
hannes-ucsc: Skipping them is the worst possible way of dealing with this unnecessary test. Now we're dragging along this code without ever knowing if/when it breaks. Please remove this test. You can file a separate PR and assign a different reviewer but I won't approve this PR with this test in place.

tests/test_hcagenerator.py
```diff
@@ -14,7 +14,8 @@
     "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/analysis_bundle.json",
     "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/assay_bundle.json",
     "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/project_bundle.json",
-    "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/4.6.0/json_schema/sample_bundle.json",
+    # TODO (tsmith) uncomment once boolean logic is supported in jsongenerator.py
```
hannes-ucsc: This should mention the ticket were we track this effort.

hannes-ucsc:
I modified the commit message a bit. With bug fixes it often makes sense to mention what the invalid behavior is and then prefix that with `Fix:`. Also note the ticket reference at the end.

PR #1007, opened 2018-02-23
Optional validation of generated json data against schema [Part 2]

Bento007:

### DIFF
--- a/tests/json/generator.py
+++ b/tests/json/generator.py
@@ -2,8 +2,8 @@ import random
 import rstr
 from faker import Faker
 from faker.providers.python import Provider as PythonProvider
-from jsonschema import RefResolver
-from typing import Union, List
+from jsonschema import RefResolver, Draft4Validator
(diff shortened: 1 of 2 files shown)


Bento007:
@hannes-ucsc please review.

tests/json/generator.py
```diff
@@ -53,12 +53,14 @@ class JsonGenerator(object):
         'email': 'email'
     }
 
-    def __init__(self, resolver: RefResolver=None, formats: dict=None) -> None:
+    def __init__(self, resolver: RefResolver=None, formats: dict=None, validate: bool=False) -> None:
```
hannes-ucsc: I don't even want this to be optional. Please make the validation mandatory. View it as an important assertion. Correctness always comes first. There is zero value for our team right now in making this optional.

hannes-ucsc:
Please roll this change into the first PR.

PR #1008, opened 2018-02-24
Bump domovoi dependency version

kislyuk:
This is required to pull in a fix in domovoi that should avoid triggering a Lambda CreateAlias API call error limit when creating aliases for step function states. We see this error sporadically causing failed deployments.

### DIFF
--- a/dss-api
+++ b/dss-api
@@ -14,6 +14,7 @@ from chalice.cli import CLIFactory, run_local_server
 import dss
 
 parser = argparse.ArgumentParser(description=__doc__)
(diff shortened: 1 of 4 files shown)


hannes-ucsc:
https://travis-ci.org/HumanCellAtlas/data-store/jobs/345500540#L855

The local DSS instance is prevented from starting I think.

kislyuk:
Thanks for tracking down that line @hannes-ucsc! I did notice that after getting past the intermittent test error. Will fix.

Issue #1021, opened 2018-03-01
Set-up `integration` instance of DSS and perform initial deployment


hannes-ucsc:
Create buckets, API gateway, ES instance etc. needed for an initial deployment of BlueBox.

PR #1027, opened 2018-03-02
Add `make release_integration`

hannes-ucsc:
- Lacks the actual creation of the cloud resources for the `integration` stage (see #1021 for that)
- Adds the `integration` stage to the extent that `make deploy` would work
- Adds `make release_integration`, including Travis' involvement in that
- Fixes inconsistency and brittleness wrt handling certain environment variables


hannes-ucsc:
Requesting @xbrianh as primary reviewer.

chalice/build_deploy_config.sh
```diff
@@ -2,12 +2,12 @@
 
 set -euo pipefail
 
-if [[ $# != 1 ]]; then
-    echo "Usage: $(basename $0) stage"
+if [[ -z $DSS_DEPLOYMENT_STAGE ]]; then
+    echo 'Please run "source environment" in the data-store repo root directory before running this command'
```
xbrianh: Note: during `make deploy` the argument will be printed to STDOUT, e.g. `./build_deploy_config.sh stage`. IMO this is preferable to jus seeing `./build_deploy_config.sh` float by.

hannes-ucsc: What do you suggest I do?

PR #1035, opened 2018-03-06
Allow multi-version-schema bundles to be indexed

xbrianh:
This is a lightly tested PR to allow integration day efforts (6-March, 2018) to move forward with the most reason meta data bundle example, which involves files of mixed schema versions.

This commit may subsequently be reverted upon completion of integration testing. Alternatively, unit tests may be brought up to date with new ES index names.


kislyuk:
48e421c passed tests, but shouldn't have because master immediately started failing with a linter error. We need to ensure that the tests that run in PRs are the exact same tests that run in master, unless they touch deployed assets. @hannes-ucsc interest

PR #1051, opened 2018-03-13
Probabilistically return Retry-after header in GET /files/<uuid> request

amork:
Probabilistically return "Retry-after" header. The retry-after interval can be relatively short now, but it sets up downstream libraries / users for success when we start integrating this with the checkout service.
### Test plan

- [x] Update all unit tests related to Files Get/Head
- [ ] Test with browsers and other http clients
- [x] Update HCA CLI retry, test with CLI

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -161,6 +161,13 @@ paths:
               description: SHA-256 (in hex format) of the file contents in hex.
               type: string
               pattern: "^[a-z0-9]{64}$"
+        301:
+          description: Handle asynchronously, downstream users are expected to retry later.
+          headers:
+            Retry-After:
+              description: Delay in seconds, downstream users are expected to retry after the delay.
+              type: integer
+              format: int64
     get:
       summary: Retrieve a file given a UUID and optionally a version.
       description: |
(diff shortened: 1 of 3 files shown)


dss-api.yml
```diff
@@ -190,6 +190,13 @@ paths:
           type: string
           format: date-time
       responses:
+        301:
+          description: Handle asynchronously, downstream users are expected to retry later.
+          headers:
+            Retry-After:
+              description: Delay in seconds, downstream users are expected to retry after the delay.
+              type: integer
+              format: int64
```
ttung: oh dear, i hope we never exceed 2B seconds.

dss/api/files.py
```diff
@@ -67,6 +75,17 @@ def get_helper(uuid: str, replica: Replica, version: str=None):
     ))
 
     if request.method == "GET":
+        '''
```
ttung: Why single-quote rather than double-quote?  Everywhere else in the repo uses double-quote.

ttung:
btw, what's the browser support like?  We should know that before we land this.

dss/api/files.py
amork: fixed

dss-api.yml
amork: agreed

PR #1059, opened 2018-03-16
Simplify how test, lint and mypy targets interact [alternative A]

hannes-ucsc:
Follow up on https://github.com/HumanCellAtlas/data-store/pull/1055

I don't like how some test targets depend on mypy and lint while others don't. This is the first of two alternative proposals.

It ensures that mypy and lint are run once before any test is run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,12 @@
 include common.mk
 MODULES=dss tests
 
+# Run all standalone tests in parallel
+#
+test: mypy lint $(tests)
+	coverage combine
+	rm -f .coverage.*
+
 lint:
 	flake8 $(MODULES) chalice/*.py daemons/*/*.py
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,8 @@
 include common.mk
 MODULES=dss tests
 
+all: test
```
kislyuk: Can you just bring the test target up top instead of declaring a new one?

Makefile
```diff
@@ -51,7 +53,7 @@ integration_test:
 smoketest:
 	$(MAKE) DSS_TEST_MODE="integration" tests/test_smoketest.py
 
-scaletest:
+scaletest: mypy lint
```
kislyuk: I don't think scaletest should depend on linter targets

Makefile
hannes-ucsc: Will do.

Makefile
hannes-ucsc: Done.

hannes-ucsc:
I had to walk back the change that moved the test target to the top. That target depends on certain variables being set. Having an `all` target at the top is a convention from the old days: https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html#Standard-Targets.

PR #1063, opened 2018-03-17
Improve handling of mixed schema versions in shape descriptor

hannes-ucsc:


hannes-ucsc:
@Bento007 and @xbrianh, you should both look at this as it touches code you recently modified. I'm not requesting a primary reviewer for that reason.

dss/index/es/document.py
```diff
@@ -208,48 +212,70 @@ def _prepare_index(self, dryrun):
...
-                return 'v-' + '-'.join([e[0] + e[1] for e in schema_version_list])
+                schemas = sorted(schemas_by_file.items())
+
+                def entry(file_name, schema):
+                    if schema.type == file_name:
+                        return file_name, schema.version
+                    else:
+                        return file_name, schema.type, schema.version
+
+                return '.'.join(chain('v', *(starmap(entry, schemas))))
```
xbrianh: This may amount to personal preference but I find `'v.' + '.'.join([entry(*t) for t in schemas])` to be more readable, and it avoids the itertools imports.

hannes-ucsc: I'll try it. It does look more readable.

Generally speaking, though, I don't see why itertools is problematic. It can lead to more readable and more optimal code.

PR #1068, opened 2018-03-20
Only cache if response is 200

Bento007:


hannes-ucsc:
Actually, an even better title would be:

"Fix: S3URLCache caches non-200 responses"

Bento007:
done and done

hannes-ucsc:
Can you confirm that you either cleared the cache or deleted the cached non-200 responses?

PR #1077, opened 2018-03-21
User IAM user creds in gs-event-relay daemon.

xbrianh:
* Instead of bororwing deployment creds, create an IAM user with
an access policy restricted to sns publish.
* User deployed/configured with scripts/deploy_aws_gs_event_relay_user.py
* Credentials stored in Travis environment variables


scripts/deploy_aws_gs_event_relay_user.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
```
kislyuk: I think this script is better named `create_aws_gs_event_relay_user.py` instead of `deploy_aws_gs_event_relay_user.py`. You also need to mention this script in the readme, somewhere alongside authorize_aws_deploy.

environment
```diff
@@ -55,6 +55,8 @@ DSS_GS_CHECKOUT_BUCKET_TEST=org-humancellatlas-dss-checkout-test
 DSS_GS_CHECKOUT_BUCKET_TEST_FIXTURES=org-humancellatlas-dss-checkout-test-fixtures
 DSS_GS_CHECKOUT_BUCKET_STAGING=org-humancellatlas-dss-checkout-staging
 DSS_GS_CHECKOUT_BUCKET_PROD=org-humancellatlas-dss-checkout-prod
+DSS_GS_EVENT_RELAY_AWS_ACCESS_KEY_ID=
+DSS_GS_EVENT_RELAY_AWS_SECRET_ACCESS_KEY=
```
kislyuk: I don't think you should clobber these variables here. Did you set environment variables in the Travis UI? If so, their values won't be available since you clobber them here. You should probably leave a comment here instead, naming these variables and prompting the operator to set them when deploying.

scripts/deploy_aws_gs_event_relay_user.py
xbrianh: will fix

environment
xbrianh: Yes, the variables were set in the Travis UI. Will un-clobber and note.

kislyuk:
OK, this is getting closer. Do you need to update https://github.com/HumanCellAtlas/data-store/blob/master/iam/policy-templates/ci-cd.json?

xbrianh:
@kislyuk Yes, the ci-cd policy needed an update.

Also introduced environment var `DSS_PARAMETER_STORE`
Full parameter names are `$DSS_PARAMETER_STORE/$SOME_PARAMETER_NAME`

README.md
```diff
@@ -262,6 +262,13 @@ credentials once from your workstation. After this is done, Travis CI will be ab
 repeat the `make deploy` step from a privileged account any time you change the IAM policies templates in
 `iam/policy-templates/`.
 
+#### Authorizing the event relay
+
+Environment variables provide the AWS credentials needed to relay events originating from supported cloud platforms
+outside of AWS. Run `scripts/create_aws_event_relay_user.py` to create an AWS IAM user with the appropriate
+restricted access policy. The access key id and secret access key are obtained from the AWS console, and need to be
+populated into AWS SSM parameters by running the script `scripts/set_event_relay_parameters.py`.
```
kislyuk: In `scripts/set_event_relay_parameters.py`, passing secrets as command line parameters is insecure. Instead, please use the IAM method for getting the key ID and secret key.

xbrianh: Fixed

environment
```diff
@@ -74,6 +74,10 @@ DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS=${DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_ARRAY[
 DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_TEST="human-cell-atlas-travis-test.iam.gserviceaccount.com"
 DSS_ES_DOMAIN="dss-index-$DSS_DEPLOYMENT_STAGE"
 PYTHONWARNINGS=ignore:ResourceWarning,ignore::UserWarning:zipfile:
+DSS_PARAMETER_STORE="/dss/parameters"
```
Bento007: please include an explanation of what these new environment variables are for in the readme.

PR #1088, opened 2018-03-23
missing dependencies

amork:
add rstr and Faker module as dependecies for scale test Lambda

### DIFF
--- a/daemons/dss-scalability-test/requirements.txt
+++ b/daemons/dss-scalability-test/requirements.txt
@@ -1,2 +1,4 @@
 hca >= 3.2.0
+Faker >= 0.8.11
+rstr >=2.2.6
 -r ../../requirements.txt


daemons/dss-scalability-test/requirements.txt
```diff
@@ -1,2 +1,4 @@
 hca >= 3.2.0
+Faker >= 0.8.11
+rstr >=2.2.6
```
hannes-ucsc: I think this should be done at the *top-level* `requirements.txt` and/or `.in` files.

When the JSON generator was recently moved from `tests` to `dss` its dependencies should have been moved from `requirements-dev.txt` to `requirements.txt`, too.

kislyuk: We should not include these libraries in the general daemon package.

I think this duplication is fine for now. If you really want to avoid duplication, let's split the json generator requirements into a separate requirements file and include it from requirements-dev and from daemons/dss-scalability-test/requirements.

kislyuk: Alternatively, the dss-scalability-test daemon could simply symlink to requirements-dev.txt instead of requirements.txt (I hope all the stuff we have in requirements-dev is packageable).

hannes-ucsc: I vote for just symlinking. The requirements handling is already unwieldy as it is. The amount of developer time & money spent with managing multiple separate dependency lists by far outweighs the space and bandwidth savings they accomplish.

hannes-ucsc:
Superseded by #1090.

PR #1109, opened 2018-04-03
fixed logging from lambdas by add logger name 'app'

Bento007:
## Testing
Ran smoke test before change and verified logs from dss-sync where not present.
Ran smoke test  after change and verified logs from dss-sync lambda function were present.
```[INFO]	2018-04-03T23:16:56.428Z	1a2a61b5-3795-11e8-b3f5-073ab1df4c00	Copying blobs/190a817c2f89732021d118945cb2f1629bb778957601cca9858da04d4b397609.778911c46874cd319dd1ceb75a27727a546d24d6.16a405357a4b5ff8d85a261294d51e00-2.68ba65c9:{'id': 1, 'start': 0, 'end': 67108864, 'total_parts': 1} from s3://dss-tsmith1 to gs://dss-tsmith1```

PR #1111, opened 2018-04-04
Set default logging level to INFO

mweiden:
## Why?
Logs of requests to the DSS API are not currently showing up in CloudWatch Logs. We need these logs for debugging and analytics.

## What?
The default log level seems to be warning. Given that the order of logging levels is `ALL < DEBUG < INFO < WARN < ERROR < FATAL < OFF`, we will not recover the `INFO`-level API logs in CloudWatch Logs. This PR sets the default to `INFO`.

### DIFF
--- a/dss/logging.py
+++ b/dss/logging.py
@@ -62,7 +62,7 @@ def configure_test_logging():
 
 def _configure_logging(test=False, **kwargs):
     root_logger = logging.getLogger()
-    root_logger.setLevel(logging.WARNING)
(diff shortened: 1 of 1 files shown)


mweiden:
@hannes-ucsc can you be primary reviewer on this one?
/cc @kislyuk

mweiden:
Sorry, this seems to have been fixed in seeing https://github.com/HumanCellAtlas/data-store/pull/1109. Just seeing this now

PR #1114, opened 2018-04-05
Configure chalice apps with lambda logging config

mweiden:
Lambdas deployed with chalice are not currently configured to emit logs. This PR makes the daemon log configuration general to all lambdas and adds the logging configuration to chalice lambdas.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -1,7 +1,6 @@
 import collections
 import datetime
 import functools
-import logging
 import os
 import random
 import re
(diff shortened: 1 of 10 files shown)


chalice/app.py
```diff
@@ -27,6 +28,8 @@
 
 Config.set_config(BucketConfig.NORMAL)
 
+dss.logging.configure_lambda_logging()
```
hannes-ucsc: This should be above the Config.set_config() line.

mweiden: Sure

PR #1117, opened 2018-04-06
Fix: chalice app logs a scary message during tests

hannes-ucsc:
Some tests start an instance of the API web app. At that time, test logging is already configured and the web app is expected to complain about logging already being configured. We shouldn't log a scary backtrace in that case: https://travis-ci.org/HumanCellAtlas/data-store/jobs/362872052#L713

@mweiden Import style of configure_lambda_logging should be consistent. Feel free to change the style, but please do so consistently across all modules.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -17,14 +17,15 @@ import nestedcontext
 import requests
 from flask import json
 
+
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'chalicelib'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
-import dss.logging
 from dss import BucketConfig, Config, DeploymentStage, create_app
+from dss.logging import configure_lambda_logging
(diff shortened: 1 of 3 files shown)


mweiden:
@hannes-ucsc I’m not sure that we’re ready for this. Logging from `DSSChaliceApp` still does not work and before we add more complexity to this module, we should form a cohesive strategy that generalizes to Chalice. The Chalice object that `DSSChaliceApp` inherits from maintains and configures its own logger (see https://github.com/aws/chalice/blob/master/chalice/app.py#L443 and https://github.com/aws/chalice/blob/master/chalice/app.py#L468). How do you propose we reconcile these two modules? /cc @kislyuk

hannes-ucsc:
So it turns out that logging works just fine after @mweiden's fix (#1114). This PR doesn't really add any significant complexity, just one global variable and one conditional. It gets rid of an expected, but scary looking back during trace during test runs.

PR #1120, opened 2018-04-06
Streamlined Deployment/Teardown

xbrianh:
* Streamline the configuration of a DSS deployment.
* Where possible, use Terraform scripting to deploy cloud infrastructure.
* Streamline the teardown of a DSS deployment


configure.py
```diff
@@ -0,0 +1,107 @@
...
+        print(f'\t\t{path}')
+    print('\t2. run `scripts/create_config_gs_service_account.sh`')
+    print('\t3. Visit the google console to aquire `application_secrets.json`')
+    print('\t4. run `source environment`')
+    print('\t5. run `make deploy-infra`')
+    print('\t6. run `make deploy`')
+
+
+if __name__ == "__main__":
+    main()
```
kislyuk: This script needs help or usage information. Running this script with `--help` should print that information.

Also, I think this script needs an overhaul. The logic to get user input is overwrought, and should be either simplified or removed. I think the process of manually typing variable values in non-reproducible configuration dialogs is flawed and ill-advised as the standard method for setting up an environment. The script should instead instruct the user how to alter their configuration by entering it in the appropriate files.

deployment/.gitignore
```diff
@@ -0,0 +1,11 @@
...
+!.gitignore
+!Makefile
+!active
+!*/
+!/dev/**
+!/prod/**
+**/gcp-credentials.json
+**/application_secrets.json
+**/.terraform
+**/local_variables.tf
```
kislyuk: I would caution against structuring the .gitignore file this way - I think it will lead to confusion. Are you modeling this after some Terraform best practice?

deployment/Makefile
```diff
@@ -0,0 +1,51 @@
+COMPONENT=
+STAGEPATH=${shell cd active && pwd -P}
+STAGE=${shell basename $(STAGEPATH)}
+DIRS=${shell find $(STAGE)/* -not -path "*/\.*" -type d}
+COMPONENTS=${shell basename $(DIRS)}
+AWS_PROFILE=${shell cat $(STAGE)/local_variables.tf | jq -r .variable.aws_profile.default}
```
kislyuk: The AWS profile name should not be set from the terraform config. AWS profiles are details of the user's configuration that are orthogonal to terraform and should not be entangled.

PR #1121, opened 2018-04-06
Simplify the workaround for static scoping.

ttung:
This is apparently the preferred idiom for doing this on python, so copying it.


ttung:
Oops, meant to close this.

PR #1122, opened 2018-04-06
[reduce-lambda-usage 1/2] Add a new type of annotation that allows us to transform a branch id to any arbitrary value.

ttung:
This value is then stored in the state machine definition.


hannes-ucsc:
You might want to look at the visitation state machine definition. It is parameterized, too, but in a much simpler yet more powerful way.

ttung:
Deprecated by #1126

PR #1127, opened 2018-04-07
refactor dss scalability test not to use dss.stepfunctions.generator

ttung:


amork:
@ttung I'm curious what are the issues with using dss.stepfunctions.generator?

PR #1132, opened 2018-04-09
Invoke reindexing without reaper/retry

hannes-ucsc:


hannes-ucsc:
This code is covered by integration tests only. Hence the failed coverage test.

PR #1139, opened 2018-04-11
checkout bucket in fixture config

amork:
use DSS_S3_CHECKOUT_BUCKET_TEST for TEST_FIXTURE config


hannes-ucsc:
One silly test is failing.

PR #1148, opened 2018-04-13
[2/2] New PUT /bundle error incorrect_file_bundle_uuid

xbrianh:
depends on: #1147 

Require file `bundle_uuid`'s match their bundle in `PUT /bundle`

Addresses:
https://github.com/HumanCellAtlas/data-store/blob/7007beac418dec45d6f097dabdf131b96d094cf4/dss/api/bundles/__init__.py#L101

### DIFF
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -77,14 +77,13 @@ def put(uuid: str, replica: str, json_request_body: dict, version: str = None):
                     file_metadata = handle.get(bucket, metadata_key)
                 except BlobNotFoundError:
                     continue
-                file_metadata_dict = json.loads(file_metadata)
-                if uuid != file_metadata_dict['bundle_uuid']:
+                file['file_metadata'] = json.loads(file_metadata)
+                if uuid != file['file_metadata']['bundle_uuid']:
                     raise DSSException(
                         requests.codes.conflict,
                         "incorrect_file_bundle_uuid",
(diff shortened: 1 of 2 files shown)


dss/api/bundles/__init__.py
```diff
@@ -77,6 +77,13 @@ def put(uuid: str, replica: str, json_request_body: dict, version: str = None):
                     file_metadata = handle.get(bucket, metadata_key)
                 except BlobNotFoundError:
                     continue
+                file_metadata_dict = json.loads(file_metadata)
```
ttung: why are we parsing this twice?  it's already done on L80.

tests/test_bundle.py
```diff
@@ -210,6 +211,25 @@ def _test_bundle_put(self, replica: Replica, fixtures_bucket: str):
             )
             self.assertEqual(resp_obj.json['code'], "file_missing")
 
+        # should *NOT* be able to upload a bundle containing a file with an incorrect bundle_uuid
+        # but we should get requests.codes.conflict
+        resp_obj = self.upload_file_wait(
```
ttung: alternatively, you can just upload a new bundle with the old file, and that should fail.  that potentially saves you the extra file upload.

dss/api/bundles/__init__.py
xbrianh: Because the json parser becomes stronger with exercise?

fixed.

tests/test_bundle.py
xbrianh: fixed.

Issue #1164, opened 2018-04-17
Reap failed indexer lambda invocations


hannes-ucsc:
We should use the reaper to retry failed indexer invocations.

Connected to #1115, 
Connectedf to #1130

PR #1137, opened 2018-04-11
Retry get file blob if not found else raise error (fixes #1115)

Bento007:
- Added lambda context to indexer.
- Retry finding files in manifest if not found until lambda times out.
- If file cannot be found an exception is raised.
- fixes #1115 

## Test Method
updated test cases for indexer.

## Deployment
redeploy indexing lambdas

Issue #1115, opened 2018-04-05
Out-of-order sync of blobs causes bundles to be partially indexed


hannes-ucsc:
Originally reported by @rexwangcc on Slack. When the indexer tried to index the `gcp` copy of a bundle submitted to the `aws` replica earlier, it reported one of the blobs, the one for `biomaterials.json`, missing:

https://logs.dev.data.humancellatlas.org/_plugin/kibana/app/kibana#/doc/*/cwl-2018-04-05/fromFirehose?id=49583086244926168104730357671478770103613383151955476482.0&_g=()

[You might have to copy that link, then click it and then paste it into the browser's address bar.]

```
{
  "_index": "cwl-2018-04-05",
  "_type": "fromFirehose",
  "_id": "49583086244926168104730357671478770103613383151955476482.0",
  "_score": 1,
  "_source": {
    "@message": "[WARNING]\t2018-04-05T20:05:36.671Z\tb421931b-390c-11e8-b44c-1726d06a9efd\tIn bundle 26518a61-12b2-4ce8-ae03-bafd139b80eb.2018-04-05T200527.642039Z the file 'biomaterial.json' is marked for indexing yet could not be accessed. This file will not be indexed. Exception: BlobNotFoundError, file blob key: blobs/2e898cee332d8352ee55b7cc2031d738e141ff97075807891553e04c746134a0.64f83fddbbff486018bd13be5ee0cf362d88b65a.2fefd90d0e7d237337191e7c6fb8052b.4f2e2543\n",
    "@id": "33963114734376066906980778872193634681259338252515016746",
    "@timestamp": "2018-04-05T20:05:36.671Z",
    "@owner": "861229788715",
    "@log_group": "/aws/lambda/dss-index-integration",
    "@log_stream": "2018/04/05/[$LATEST]3f736ddaa55944418c37efa4d4e5b825"
  },
  "fields": {
    "@timestamp": [
      1522958736671
    ]
  }
}
```

The key part is `… this file will not be indexed. Exception: BlobNotFoundError …`.

When I checked an hour later, the blob was present on the `gcp` replica.

As a result, the indexer didn't index that file which caused a GreenBox subscription not to be notified because its query relies on fields from `biomaterial.json`.

The following shows that the ES index documents differ between the two replicas:

The document for the `aws` replica contains the `biomaterials_json` key: 

```
ca dss post-search --replica aws --es-query='
{
  "query": {
    "term": {
      "_id": {
[...]

Issue #1130, opened 2018-04-09
Use step functions to orchestrate sync'ing of entire bundles


hannes-ucsc:
Ensure that bundle manifest is copied last. Permanent fix for #1115.

Eliminate hack that repeatedly lists the parts in an ongoing multi-part file copy in order to determine if all parts have been copied.

mweiden:
@kislyuk there were recent changes to the `dcplib` code for determining s3 multipart upload.
* The multipart upload chunk size is now determined by [this function](https://github.com/HumanCellAtlas/dcplib/blob/master/dcplib/s3_multipart.py#L14)
* The [`ChecksummingBufferedReader`](https://github.com/HumanCellAtlas/dcplib/commit/574640c8da685f290df7d546163876c1c811e91b) needs to be configured with that chunk size (see [example](https://github.com/HumanCellAtlas/dcp-cli/commit/c6ef9a948a36b2d6cf494e923b58662b448fb46d))

PR #1165, opened 2018-04-17
Consolidate indexer run time tracking in Indexer class (mostly)

hannes-ucsc:
So far we've been neglecting to accurately track the Lambda invocation's remaining run time in the indexer. No one noticed until we started retrying to load metadata blobs in #1137 as a fix for #1115. The retries could potentially take so much time that there wouldn't be anything left for the actual indexing operation, like interacting with ES. We should make reasonably sure that there is sufficient time to complete an indexing operation before we commence it, because if we are interrupted, we'll leave the index inconsistent. This PR fixes that on a best effort basis. I've opened #1164 to retry the entire invocation if that best effort fails.

### DIFF
--- a/daemons/dss-index/app.py
+++ b/daemons/dss-index/app.py
@@ -42,14 +42,12 @@ def dispatch_gs_indexer_event(event, context):
 
 
 def _handle_event(replica, event, context):
-    timeout = context.get_remaining_time_in_millis() / 1000 - 10  # ten seconds of safety for lambda shut down
-    assert timeout >= 10, 'Not enough time left for this lambda to process event'
     executor = ThreadPoolExecutor(len(DEFAULT_BACKENDS))
     # We can't use ecxecutor as context manager because we don't want the shutdown to block
     try:
-        backend = CompositeIndexBackend(executor, DEFAULT_BACKENDS, timeout=timeout, context=context)
+        backend = CompositeIndexBackend(executor, DEFAULT_BACKENDS)
         indexer_cls = Indexer.for_replica(replica)
-        indexer = indexer_cls(backend)
+        indexer = indexer_cls(backend, context)
         indexer.process_new_indexable_object(event)
     finally:
         executor.shutdown(False)
--- a/dss/index/bundle.py
+++ b/dss/index/bundle.py
@@ -1,9 +1,7 @@
 import json
 import logging
-from collections import deque
-from typing import Iterable, Mapping, Optional, Sequence
+from typing import Mapping, Optional, Set
 
-import time
 from cloud_blobstore import BlobNotFoundError, BlobStoreError
 
 from dss import Config, Replica
@@ -31,7 +29,7 @@ class Bundle:
         self.files = files
 
     @classmethod
-    def from_replica(cls, replica: Replica, fqid: BundleFQID):
+    def load(cls, replica: Replica, fqid: BundleFQID):
         manifest = cls._read_bundle_manifest(replica, fqid)
         files = cls._read_file_infos(replica, fqid, manifest)
         self = cls(replica, fqid, manifest=manifest, files=files)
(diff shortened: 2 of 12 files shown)


daemons/dss-index/app.py
```diff
@@ -42,14 +42,12 @@ def dispatch_gs_indexer_event(event, context):
 
 
 def _handle_event(replica, event, context):
-    timeout = context.get_remaining_time_in_millis() / 1000 - 10  # ten seconds of safety for lambda shut down
-    assert timeout >= 10, 'Not enough time left for this lambda to process event'
     executor = ThreadPoolExecutor(len(DEFAULT_BACKENDS))
     # We can't use ecxecutor as context manager because we don't want the shutdown to block
     try:
-        backend = CompositeIndexBackend(executor, DEFAULT_BACKENDS, timeout=timeout, context=context)
+        backend = CompositeIndexBackend(executor, DEFAULT_BACKENDS)
```
xbrianh: Is `executor` only used in `backend._delegate`?  I'm curious why it's passed in here.

hannes-ucsc: `_delegate` is a protected method that should not be called from outside the class. The backend methods that call it are public but their signature is restricted by the abstract class (BTW: this is an instance of the Composite design pattern). The lifetimes of the backend and the executor should be the same and providing the executor at construction time makes that explicit.

dss/index/indexer.py
```diff
@@ -35,9 +40,9 @@ def process_new_indexable_object(self, event: Mapping[str, Any]) -> None:
     def index_object(self, key):
```
xbrianh: There seems to be an (understandable) mixing of indexing logic and timing logic, but I wonder if these concerns could be separated.

For instance,
```
num_ops = indexer.load_data(*args)
check_time(num_ops * indexer.time_for_one_thing)
indexer.do_it()
```

mweiden: To my understanding, @xbrianh is emphasizing where the logic lives. I think separating the concerns is possible by returning a data structure that defers the work once you know the size like the one below.

```python
class Task:

    def __init__(self, size, logic, *args, **kwargs):
        self.size = size
        self.logic = logic
        self.args = args
        self.kwargs = kwargs

    def submit(self):
        result = self.logic(*self.args, **self.kwargs)
        self.size = 0
        return result
```

So you could do something like.

```python
def foo(n):
    for i in range(n):
        time.sleep(1)

def some_insertion_logic():
    ...
    return Task(5, foo, 5)

# in a different (parent) module
print("Unexecuted task size is: " + str(task.size))

if task.size < 10:
    result = task.submit()
```

Just a thought.

dss/index/indexer.py
```diff
@@ -48,25 +53,43 @@ def index_object(self, key):
...
+    @property
+    def remaining_time(self) -> float:
+        """
+        Return the remaining runtime of this Lambda invocation in seconds.
+        """
+        remaining_time = self.context.get_remaining_time_in_millis() / 1000
+        logger.debug("Remaining indexing time is %fs", remaining_time)
+        return remaining_time
+
+    def _is_enough_time(self, num_operations):
```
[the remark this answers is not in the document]
hannes-ucsc: Will do.

dss/index/bundle.py
```diff
@@ -64,14 +67,14 @@ def _read_file_infos(cls, replica: Replica, fqid: BundleFQID, manifest: JSON) ->
                 try:
                     file_string = handle.get(bucket_name, file_blob_key).decode("utf-8")
                 except BlobStoreError as ex:
-                    if attempts < 5:
+                    # Requeue at the end, yielding to other files we may not have tried yet
+                    files_info.append((file_info, attempts + 1))
+                    if wait_time + len(files_info) * load_time < indexer.remaining_time:
```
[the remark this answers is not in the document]
hannes-ucsc: Good point. Indexer depends on Bundle but Bundle should not depend on Indexer. In fact, if I add a `Indexer` type hint to this method, I'd have to import `Indexer` and that would result in a circular import.

However, this code *does* need to know the remaining run time on a continuing basis. One alternative to passing an `Indexer` instance is passing the lambda context directly. I thought that that was uglier because it exposes Bundle to something even further removed from its core concern.

The reason I had made peace with this ugliness is that it is going to go away once we either ensure that blobs are copied in order (#1130) or pursue @xbrianh's idea of retrying the entire indexer lambda using the @amork's DLQ mechanism.

dss/index/indexer.py
hannes-ucsc: To summarize our F2F conversation @mweiden, I don't think that the extra complexity of lifting  the unit of work (that's the design pattern your Task class follows) is warranted if we're moving the logic within the class and just one layer up in the call hierarchy.

The thing to keep in mind is that this PR already consolidates timing logic that was previously spread around more.

hannes-ucsc:
So I ripped out the entire retry logic in `_load_file_infos()`. This eliminates the dependency of Bundle on Indexer, thereby making the coupling between the two unidirectional.

At the time @Bento007 and I implemented the retry logic in `_load_file_infos()` the reaper hadn't been merged yet. I think that the reaper is the perfect mechanism to pickup indexer lambda failures like those caused by the out-of-order syncing of the bundle manifest. Because that issue is so rare, it does not warrant the complexity we added, even if it avoids loading files twice or may lead to a successfully indexed bundle sooner.

That addresses one of your two big issues, @mweiden. The other issue of further separating the timing concern from main indexing concerns I find esoteric. As I laid out above, even *if* we were able to address this by introducing the unit of work pattern, the gain would be questionable. This PR is already a great improvement in that it consolidates the logic in a single class. I look forward to reviewing a PR that consolidates and isolates the timeout logic even further.

PR #1170, opened 2018-04-19
Reliable notifications [1/2]

hannes-ucsc:
All the non-contentions stuff from #1100, in particular:

- extract Notification class
- add PUT for notifications
- add form-encoded POSTs for notifications
- add notification integration tests using signed PUTs and POSTs


dss/api/subscriptions.py
```diff
@@ -56,7 +56,7 @@ def find(replica: str):
         'uuid': hit.meta.id,
         'replica': replica,
         'owner': owner,
-        'callback_url': hit.callback_url,
+        'endpoint': hit.endpoint.to_dict(),
```
kislyuk: Not sure which of these is the more appropriate term here. In cases like this I think it's best to go by established convention. GitHub uses just "url". Are there other examples?

hannes-ucsc: Using "url" instead of "endpoint" would be inappropriate. The `endpoint` object *has* a property called `url` but it also has other properties, like for body encoding, HTTP method and  static fields.

So the closest analogue to our `endpoint` would be GH's `hook` object.

Just googling "webhook endpoint" comes up with plenty matches. I'd welcome suggestions for alternatives, but "url" isn't gonna work, I think.

PR #1178, opened 2018-04-21
[1/3] Reliable notifications

hannes-ucsc:
Same as #1170 but denormalized request body schema for `PUT /subscriptions`


hannes-ucsc:
Do not merge before https://github.com/HumanCellAtlas/dcp-cli/pull/108 lands.

PR #1181, opened 2018-04-23
Move relay secrets to AWS Secrets Manager

xbrianh:


xbrianh:
@mweiden Two thoughts: If we can avoid it, it seems preferable not to pass secrets into shell variables,  and not to tempt operators into setting secrets on the command line.

mweiden:
@xbrianh my opinion here is loosely held. Go ahead

PR #1182, opened 2018-04-23
[3/3] Reliable notifications

hannes-ucsc:
Using signed POST against S3 for endpoint, actually verify notifications during the smoke test.

Other improvements to smoke test.

### DIFF
--- a/tests/test_smoketest.py
+++ b/tests/test_smoketest.py
@@ -5,6 +5,9 @@ A basic integration test of the DSS. This can also be invoked via `make smoketes
 import os, sys, argparse, time, uuid, json, shutil, tempfile, unittest
 import subprocess
 
+import boto3
+import botocore
+
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Requesting @mweiden to be primary reviewer.

tests/test_smoketest.py
```diff
@@ -62,98 +69,143 @@ def setUpClass(cls):
...
+        # Check the notifications
+        #
+        for replica, (subscription_id, notification_key) in notifications_proofs.items():
+            obj = s3.get_object(Bucket=self.notification_bucket, Key=notification_key)
+            notification = json.load(obj['Body'])
+            self.assertEquals(subscription_id, notification['subscription_id'])
+            self.assertEquals(bundle_uuid, notification['match']['bundle_uuid'])
+            self.assertEquals(bundle_version, notification['match']['bundle_version'])
+            s3.delete_object(Bucket=self.notification_bucket, Key=notification_key)
+            run(f"{venv_bin}hca dss delete-subscription --replica {replica} --uuid {subscription_id}")
```
mweiden: If the test fails, the test subscriptions will hang around. I think it would be worth it to put this in a `finally` so that we are more sure that the subscriptions will be mopped up.

hannes-ucsc: Will do.

mweiden: I meant to say the s3 object too, but I think you got the idea.

hannes-ucsc: https://github.com/HumanCellAtlas/data-store/pull/1182/commits/dbd9d079a4ddf31572f11193734edf7f584d2f3e

which will be squashed into its parent.

hannes-ucsc:
I ran smoketest manually one last time, and it passed.

Issue #1101, opened 2018-03-28
Big red button script


Bento007:
[Big Red Button Doc](https://docs.google.com/document/d/1eWknDSh2aoDtdBYmSa4Zvp8mTGJWXnKUhcm0aWZBxmI/edit#) 

**User Story:** As an operator of the blue box, I need to be able to quickly freeze and unfreeze the blue box in emergency situations.

**Acceptance Criteria:**
1) Script that disables the service (without losing data and with a message to users that indicates service being down)
2) Script that enables the service
3) Documentation on how both work
4) Correct people have the permission to run them

PR #1184, opened 2018-04-24
The Big Red Button: Script for an emergency halt of the DSS

xbrianh:
#connects to #1101 

Lambda executions may be throttled by setting the concurrency limit to 0.

See the relevant AWS documentation here:
https://docs.aws.amazon.com/lambda/latest/dg/concurrent-executions.html

### DIFF
--- a/scripts/dss-start-stop--think-before-you-invoke-me.py
+++ b/scripts/dss-start-stop--think-before-you-invoke-me.py
@@ -6,6 +6,10 @@ Use the AWS Lambda concurrency property to disable/enable lambda functions:
 
 As a consequence of this script, previously set Lambda concurrency limits
 will be lost.
+
+Asynchronously triggered lambda functions which are throttled are
+automatically added to a redrive queue, and may be retried when lambdas
+are restarted, depending on downtime.
 """
 import os
 import sys
@@ -23,8 +27,14 @@ parser.add_argument("action", choices=["start", "stop"])
 args = parser.parse_args()
 
 
-action = f"DSS {stage} START" if args.action=="start" else f"DSS {stage} STOP"
-if not click.confirm(f"Are you sure you want to do this ({action})?"):
+if "start" == args.action:
+    msg = f"(DSS {stage} START) Lambdas will be restarted with default concurrency limits. Continue?"
+elif "stop" == args.action:
+    msg = f"(DSS {stage} STOP) Lambdas will be halted by setting concurrency=0. Continue?"
+else:
+    raise Exception(f"Unknown action {args.action}")
+
+if not click.confirm(msg):
     sys.exit(0)
 
 
@@ -57,7 +67,9 @@ for f in functions:
         print(f"{f} not deployed, or does not deploy a Lambda function")
         continue
 
-    if args.start:
+    if "start" == args.action:
         enable_lambda(f)
-    elif args.stop:
+    elif "stop" == args.action:
         disable_lambda(f)
+    else:
+        raise Exception(f"Unknown action {args.action}")
--- a/scripts/dss-start-stop--think-before-you-invoke-me.py
+++ b/scripts/dss-start-stop--think-before-you-invoke-me.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+"""
+Use the AWS Lambda concurrency property to disable/enable lambda functions:
+    - set concurrency to 0 to disable execution
+    - remove concurrency setting to enable execution
+"""
+import os
+import sys
+import boto3
+import click
+import argparse
+
+
+stage = os.environ['DSS_DEPLOYMENT_STAGE']
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument("--stop", action="store_true")
+parser.add_argument("--start", action="store_true")
+args = parser.parse_args()
+assert(
+    (args.start or args.stop) and not (args.start and args.stop)
+)
+
+
+action = f"DSS {stage} START" if args.start else f"DSS {stage} STOP"
+if not click.confirm(f"Are you sure you want to do this ({action})?"):
+    sys.exit(0)
+
+
+LAMBDA = boto3.client('lambda')
+
+
+def disable_lambda(name):
+    LAMBDA.put_function_concurrency(
+        FunctionName=name,
+        ReservedConcurrentExecutions=0
+    )
+    print(f"stopped {name}")
+    
+def enable_lambda(name):
+    LAMBDA.delete_function_concurrency(
+        FunctionName=name,
+    )
+    print(f"started {name}")
+
+
+for root, dirs, files in os.walk(os.path.join(pkg_root, 'daemons')):
+    functions = [f'{name}-{stage}' for name in dirs]
+    break
+functions.append(
+    f"dss-{stage}"
+)
+
+
+for f in functions:
+    try:
+        resp = LAMBDA.get_function(
+            FunctionName=f,
+        )
+    except LAMBDA.exceptions.ResourceNotFoundException:
+        print(f"{f} not deployed")
+        continue
+
+    if args.start:
+        enable_lambda(f)
+    elif args.stop:
+        disable_lambda(f)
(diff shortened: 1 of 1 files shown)


scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+import click
+import argparse
+
+
+stage = os.environ['DSS_DEPLOYMENT_STAGE']
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument("--stop", action="store_true")
```
hannes-ucsc: This would be better modeled as a positional with `choices=['start', 'stop']`

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+parser.add_argument("--start", action="store_true")
+args = parser.parse_args()
+assert(
```
hannes-ucsc: Boolean XOR is `!=` but this becomes obsolete with the above comment.

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+    (args.start or args.stop) and not (args.start and args.stop)
+)
+
+
+action = f"DSS {stage} START" if args.start else f"DSS {stage} STOP"
+if not click.confirm(f"Are you sure you want to do this ({action})?"):
```
hannes-ucsc: Because of this I don't think the script needs to have that scary name.

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+    sys.exit(0)
+
+
+LAMBDA = boto3.client('lambda')
```
hannes-ucsc: All-capsing should be reserved for variables with a constant intializer. It is not evident why this variable is all-caps while `action`, `args`, `stage` etc. are not.

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+    - remove concurrency setting to enable execution
+"""
+import os
+import sys
+import boto3
+import click
+import argparse
+
+
+stage = os.environ['DSS_DEPLOYMENT_STAGE']
```
hannes-ucsc: Shouldn't this be further down in the code.=?

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+
+def disable_lambda(name):
+    LAMBDA.put_function_concurrency(
+        FunctionName=name,
+        ReservedConcurrentExecutions=0
+    )
+    print(f"halted {name}")
+    
+def enable_lambda(name):
+    LAMBDA.delete_function_concurrency(
```
hannes-ucsc: We might want to set concurrency limits for normal operations and this blows those away.

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+        FunctionName=name,
+    )
+    print(f"started {name}")
+
+
+for root, dirs, files in os.walk(os.path.join(pkg_root, 'daemons')):
+    functions = [f'{name}-{stage}' for name in dirs]
+    break
+functions.append(
```
hannes-ucsc: No one else in this project uses this style of wrapping. Can you reconsider it?

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+
+for f in functions:
+    try:
+        resp = LAMBDA.get_function(
+            FunctionName=f,
+        )
+    except LAMBDA.exceptions.ResourceNotFoundException:
+        # Either this daemon does not deploy a lambda, or this daemon is not deployed
```
hannes-ucsc: Log or print this! It is important to alert the user of this fact.

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,69 @@
...
+    
+def enable_lambda(name):
+    LAMBDA.delete_function_concurrency(
+        FunctionName=name,
+    )
+    print(f"started {name}")
+
+
+for root, dirs, files in os.walk(os.path.join(pkg_root, 'daemons')):
+    functions = [f'{name}-{stage}' for name in dirs]
```
hannes-ucsc: Don't you need to concat `dir` and `root`? Make sure this doesn't just work coincidentally because you're running the script from the project root.

You could eliminate the odd loop and break construct with 

```
root, dirs, files = next(os.walk(os.path.join(pkg_root, 'daemons')))
```

That would raise if `./daemons` were empty and I'd actually prefer that.

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: It will fail immediately if this environment variable is not set.

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: fixed

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: This is a dangerous script! It might deserve more than a single check.

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: changed

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: Custom concurrency limits are lost as a consequence of running this script. This is a drawback of using this method.

I've added a warning line in the doc string.

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: The concatenation is not necessary, it seems.

Changed to the `next()` construct.

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: Changed :/

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,63 @@
...
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument("action", choices=["start", "stop"])
+args = parser.parse_args()
+
+
+action = f"DSS {stage} START" if args.action=="start" else f"DSS {stage} STOP"
+if not click.confirm(f"Are you sure you want to do this ({action})?"):
```
hannes-ucsc: How about mentioning the consequences here as in `The concurrency limit of each deployed lambda will be set to 0` for stop and `Any concurrency limits on deployed lambdas will be removed` for start.

scripts/dss-start-stop--think-before-you-invoke-me.py
hannes-ucsc: OK

scripts/dss-start-stop--think-before-you-invoke-me.py
```diff
@@ -0,0 +1,63 @@
...
+functions = [f'{name}-{stage}' for name in dirs]
+functions.append(f"dss-{stage}") 
+
+
+for f in functions:
+    try:
+        resp = lambda_client.get_function(FunctionName=f)
+    except lambda_client.exceptions.ResourceNotFoundException:
+        print(f"{f} not deployed, or does not deploy a Lambda function")
+        continue
```
hannes-ucsc: I default to using the `except: else:` for this.

scripts/dss-start-stop--think-before-you-invoke-me.py
xbrianh: It seems reasonable to warn here.

Concurrency limits are lost during both actions btw.

PR #1187, opened 2018-04-25
Run indexer integration tests for AWS replica only

hannes-ucsc:
The integration tests in test_indexer don't benefit from being run for each replica. I'm also worried that if those tests run in parallel, they might conflict with each other as in this case:

https://travis-ci.org/HumanCellAtlas/data-store/jobs/370849706#L1256


hannes-ucsc:
I'll treat this as a hotfix and merge without review. I would have pushed to master directly but I wanted to ensure that this wasn't interfering with standalone tests.

PR #1188, opened 2018-04-25
Extend Elasticsearch timeout on Travis from 10s to 30s

hannes-ucsc:
Also switch from homegrown ES client pooling to `lru_cache` builtin.

This is motivated by the fact that we are now seeing ES timeout during integration tests. We already have been seeing occasional timeouts on PR builds—and alleviated that by running the tests prone to those timeout in separate, easy to retry Travis jobs—but I'm speculating that this is going to improve things.

Independently, being able to configure the ES client timeout via an environment variable is a useful feature to have, even in a production environment.


hannes-ucsc:
I pushed additional tests and removed an `except:` clause for an exception that the tests prove can't happen. That way I got to meet the coverage requirements.

PR #1189, opened 2018-04-25
Add response metrics and logging

mweiden:
This PR adds logging and metrics for requests and their response codes in the DSS HTTP API.

I've used CloudWatch Metrics, but may change this in the future if we decide on a different metrics system.

The call to the metrics server is done on its own thread to prevent the API from lagging if CloudWatch metrics is slow in responding.

HTTP metric data is grouped into two CloudWatch metrics, `HttpRequests by StatusClass` and `HttpRequests by Path, Method, StatusClass`. `HttpRequests by StatusClass` is for dashboarding and will allow coarse-grained indication of errors on the API. `HttpRequests by Path, Method, StatusClass` will allow us to drill down to the problematic endpoint and method when an issue is spotted in logs or on coarser metric plots. The HTTP return codes are grouped in the metrics into different status classes to allow easy visualization.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -29,6 +29,8 @@ configure_lambda_logging()
 
 Config.set_config(BucketConfig.NORMAL)
 
+cloudwatch = boto3.client('cloudwatch')
+
 
 EXECUTION_TERMINATION_THRESHOLD_SECONDS = 5.0
 """We will terminate execution if we have this many seconds left to process the request."""
@@ -107,6 +109,7 @@ def time_limited(chalice_app: DSSChaliceApp):
 
 def get_chalice_app(flask_app) -> DSSChaliceApp:
     app = DSSChaliceApp(app_name=flask_app.name, configure_logs=False)
+    ex = ThreadPoolExecutor(max_workers=2)
 
     @time_limited(app)
     def dispatch(*args, **kwargs):
@@ -134,6 +137,33 @@ def get_chalice_app(flask_app) -> DSSChaliceApp:
             maybe_fake_504_result = maybe_fake_504()
             if maybe_fake_504_result is not None:
                 return maybe_fake_504_result
+
+        def request_metrics(path_pattern, method, status_code):
+            status_class = str(status_code)[0] + 'xx'
+            try:
+                cloudwatch.put_metric_data(
+                    Namespace='DSS',
+                    MetricData=[
+                        dict(MetricName='HttpRequests by StatusClass',
+                             Dimensions=[
+                                 dict(Name='StatusClass', Value=status_class),
+                             ],
+                             Value=1.0,
+                             Unit='Count',
+                             StorageResolution=60),
+                        dict(MetricName='HttpRequests by Path, Method, StatusCode',
+                             Dimensions=[
+                                 dict(Name='Path', Value=path_pattern),
+                                 dict(Name='Method', Value=method),
+                                 dict(Name='StatusCode', Value=str(status_code))
+                             ],
+                             Value=1.0,
+                             Unit='Count',
+                             StorageResolution=60),
+                    ]
+                )
+            except Exception as e:
+                app.log.exception('Metrics post failed!', e)
 
         status_code = None
         try:
@@ -164,6 +194,9 @@ def get_chalice_app(flask_app) -> DSSChaliceApp:
         # API Gateway/Cloudfront adds a duplicate Content-Length with a different value (not sure why)
         res_headers = dict(flask_res.headers)
         res_headers.pop("Content-Length", None)
+
+        # observability logic is deferred to a different thread so as not to delay response
+        ex.submit(request_metrics, path_pattern, method, status_code)
 
         return chalice.Response(status_code=status_code,
                                 headers=res_headers,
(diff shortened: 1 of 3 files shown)


mweiden:
@hannes-ucsc cool, good suggestion. I will look into that now

chalice/app.py
```diff
@@ -155,6 +199,16 @@ def maybe_fake_504() -> typing.Optional[chalice.Response]:
...
         # API Gateway/Cloudfront adds a duplicate Content-Length with a different value (not sure why)
         res_headers.pop("Content-Length", None)
+
+        # observability logic is sent to a different thread so as not to delay response
+        thread = threading.Thread(
+            name='observables',
+            target=observe_request,
+            args=(path, app.current_request.method, flask_res._status_code)
+        )
+        thread.setDaemon(False)  # not a daemon, function returns, but program waits until metrics are sent
```
ttung: Doesn't this mean that we might _not_ finish sending the logging data?

hannes-ucsc: I don't think (!) it matters. Looking at AWS' bootstrap.py there is just an infinite loop that repeatedly calls the handler for every incoming request. My intuition is that the process will be killed at the 5min deadline irregardless of whether there are any threads, daemons or not.

IOW, merely the fact that this is happening in a thread makes it more likely to not be finished.

https://docs.aws.amazon.com/lambda/latest/dg/running-lambda-code.html

> Background processes or callbacks initiated by your Lambda function that did not complete when the function ended resume if AWS Lambda chooses to reuse the Execution Context. You should make sure any background processes or callbacks (in case of Node.js) in your code are complete before the code exits.

hannes-ucsc: We could have a little heuristic that performs this synchronously if we are within 30s of the 5min deadline and logs a warning if we are within 5s of the deadline because that close to the deadline even a synchronous execution is likely going to be interrupted.

chalice/app.py
```diff
@@ -107,18 +109,14 @@ def wrapper(*args, **kwargs):
 
 def get_chalice_app(flask_app) -> DSSChaliceApp:
     app = DSSChaliceApp(app_name=flask_app.name, configure_logs=False)
+    ex = ThreadPoolExecutor(max_workers=2)
```
hannes-ucsc: I think this is invoked for every request so I would put this at module (global) scope.

hannes-ucsc:
We should continue the thought process from here:

https://github.com/HumanCellAtlas/data-store/pull/1189#pullrequestreview-115673673

iam/policy-templates/ci-cd.json
```diff
@@ -151,6 +151,11 @@
         "arn:aws:events:*:$account_id:rule/*"
       ]
     },
+    {
```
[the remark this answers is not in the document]
mweiden: Made this change

chalice/app.py
```diff
@@ -138,24 +136,83 @@ def maybe_fake_504() -> typing.Optional[chalice.Response]:
...
-                        EXECUTION_TERMINATION_THRESHOLD_SECONDS),
-                    skip_on_conflicts=True):
-                flask_res = flask_app.full_dispatch_request()
-        res_headers = dict(flask_res.headers)
+        def request_metrics(path_pattern, method, status_code):
+            status_class = str(status_code)[0] + 'xx'
+            try:
+                cloudwatch.put_metric_data(
+                    Namespace='DSS',
+                    MetricData=[
```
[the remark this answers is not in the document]
mweiden: ✅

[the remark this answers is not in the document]
hannes-ucsc: With `dict`, I put the first entry on the same line as the `dict`.

chalice/app.py
```diff
@@ -138,24 +139,74 @@ def maybe_fake_504() -> typing.Optional[chalice.Response]:
...
+                ' ' + str(query_params) if query_params is not None else '',
+            )
+
         # API Gateway/Cloudfront adds a duplicate Content-Length with a different value (not sure why)
+        res_headers = dict(flask_res.headers)
         res_headers.pop("Content-Length", None)
-        return chalice.Response(status_code=flask_res._status_code,
+
+        # observability logic is deferred to a different thread so as not to delay response
+        if metrics_executor._work_queue.qsize() >= MAX_EXECUTOR_WORKERS:
```
hannes-ucsc: What do you think about making the threshold 4 and the max pool size 8? And then also log the current size.

Independently, I would log as early as possible *before* adding tasks.

And I would log the size of both pools.

mweiden: Good suggestions, ✅✅✅

chalice/app.py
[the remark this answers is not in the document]
mweiden: ✅

hannes-ucsc:
We should keep a close eye on the logs after this is deployed.

mweiden:
@hannes-ucsc Yes, definitely 👀👀

hannes-ucsc:
I'm under the impression that none of these commits would work or be useful in isolation. I would have squashed them. It's OK for a later commit to depend on an earlier one but if a later commit redoes or tweaks changes from the earlier one, that would be a smell indicating they should be squashed.

mweiden:
@hannes-ucsc OK, will squash next time. We can always revert the merge commit if necessary.

Issue #1190, opened 2018-04-26
TestGSCopy.test_simple_copy fails sporadically


hannes-ucsc:
https://travis-ci.org/HumanCellAtlas/data-store/builds/371635683#L1489:

```
coverage run -p --source=dss tests/test_gscopyclient.py -v
test_simple_copy (__main__.TestGSCopy) ... ERROR
======================================================================
ERROR: test_simple_copy (__main__.TestGSCopy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_gscopyclient.py", line 83, in test_simple_copy
    test_output()
  File "/home/travis/build/HumanCellAtlas/data-store/tests/__init__.py", line 60, in call
    return func(*args, **kwargs)
  File "tests/test_gscopyclient.py", line 80, in test_output
    dst_checksum = self.gs_blobstore.get_cloud_checksum(self.test_bucket, dest_key)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/cloud_blobstore/gs.py", line 176, in get_cloud_checksum
    raise BlobNotFoundError()
cloud_blobstore.BlobNotFoundError
----------------------------------------------------------------------
Ran 1 test in 51.615s
FAILED (errors=1)
Makefile:44: recipe for target 'tests/test_gscopyclient.py' failed
make[1]: *** [tests/test_gscopyclient.py] Error 1
make[1]: Leaving directory '/home/travis/build/HumanCellAtlas/data-store'
Makefile:54: recipe for target 'integration_test' failed
make: *** [integration_test] Error 2
```

There already is a 30s busy loop around the check so I don't think this is eventual consistency.

hannes-ucsc:
We should print/log the missing key when the exceptions occurs so we search the logs. Ideally, the missing key should be part of the exception message but that fix would have to occur in cloud-blobstore.

hannes-ucsc:
https://travis-ci.org/HumanCellAtlas/data-store/builds/371799785#L1483

PR #1191, opened 2018-04-26
cloud-blobstore 2.1.1 for better error messages on BlobNotFoundError

ttung:
Should help track down #1190

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,14 +4,14 @@ azure-common==1.1.8
 azure-nspkg==2.0.0
 azure-storage==0.36.0
 boto3==1.6.0
-botocore==1.10.8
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -4,14 +4,14 @@ azure-common==1.1.8
 azure-nspkg==2.0.0
 azure-storage==0.36.0
 boto3==1.6.0
-botocore==1.10.8
+botocore==1.9.23
```
xbrianh: Does `botocore 1.9.23` have support AWS Secrets Manager?

This may cause deploy to fail.

ttung: #1192 

It turns out pip is pretty bad at dependency management.

PR #1192, opened 2018-04-26
fix botocore version.

ttung:


hannes-ucsc:
We know that 1.9.x won't work so the `.in` file should contain `botocore >= 1.10` If we don't, the same downgrade is likely going to happen again.

Issue #1098, opened 2018-03-28
Duplicate bundles associated with DSS 504 errors


xbrianh:
During bundle/PUT submission to DSS, ingest encountered a 504 error and retried. Both operations successfully uploaded a bundle to the DSS, resulting in two versions.

This appears to have occurred twice on 28-March, 2018. The effected bundles are:
`17a3d288-01a0-464a-9599-7375fda3353d.2018-03-28T150956.976400Z`
`17a3d288-01a0-464a-9599-7375fda3353d.2018-03-28T151023.074974Z`
`59a9ba09-998d-4dde-98db-c6a05b5481ee.2018-03-28T182154.619685Z`
`59a9ba09-998d-4dde-98db-c6a05b5481ee.2018-03-28T182220.776570Z`

A possible cause may be that the dss-prod lambda not responding within the API Gateway timeout. Lambda timeout is 5 minutes, while API Gateway timeout is 29 seconds. Note that the bundle versions are almost exactly 26 seconds apart.

Indeed, we can find dss-prod lambda durations longer than 29 seconds:
>  RequestId: 7f833f8b-32b4-11e8-abb0-477def236f75\tDuration: 36038.25 ms\tBilled Duration: 36100 ms \tMemory Size: 128 MB\tMax Memory Used: 73 MB
>  RequestId: 800f04cb-32b4-11e8-b671-8173331ce5c6\tDuration: 36969.55 ms\tBilled Duration: 37000 ms \tMemory Size: 128 MB\tMax Memory Used: 73 MB
>  RequestId: 7f443b02-32b4-11e8-9c60-e3f5b738ea44\tDuration: 38453.51 ms\tBilled Duration: 38500 ms \tMemory Size: 128 MB\tMax Memory Used: 73 MB
>  RequestId: 7fcb930f-32b4-11e8-8e3f-71c37960d666\tDuration: 39173.49 ms\tBilled Duration: 39200 ms \tMemory Size: 128 MB\tMax Memory Used: 73 MB
>  RequestId: 7f5837de-32b4-11e8-a1fe-918d8506d64d\tDuration: 40014.84 ms\tBilled Duration: 40100 ms \tMemory Size: 128 MB\tMax Memory Used: 73 MB
>  RequestId: 7f54dc13-32b4-11e8-9e30-ede523155a18\tDuration: 40409.30 ms\tBilled Duration: 40500 ms \tMemory Size: 128 MB\tMax Memory Used: 73 MB
[...]

PR #1196, opened 2018-04-27
Fighting the duplicate bundle threat with timeouts

xbrianh:
Connects to #1098 

Gateway timeouts (504 errors) have been probabilistically observed during PUT /bundle.

PUT /bundle should either succeed, or fail with an appropriate error, before the gateway times out. This PR attempts to mitigate against occasionally high latency storage by using a more aggressive timeout strategy.

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -6,10 +6,13 @@ from contextlib import contextmanager
 from enum import Enum, EnumMeta, auto
 
 import boto3
+import botocore.config
 from cloud_blobstore import BlobStore
 from cloud_blobstore.s3 import S3BlobStore
 from cloud_blobstore.gs import GSBlobStore
 from google.cloud.storage import Client
+from google.oauth2 import service_account
+import google.resumable_media.common
 
 from dss.storage.hcablobstore import HCABlobStore
 from dss.storage.hcablobstore.s3 import S3HCABlobStore
@@ -94,6 +97,11 @@ class Config:
     _S3_CHECKOUT_BUCKET = None  # type: typing.Optional[str]
     _GS_CHECKOUT_BUCKET = None  # type: typing.Optional[str]
 
+    BLOBSTORE_CONNECT_TIMEOUT = 5  # type: float
+    BLOBSTORE_READ_TIMEOUT = 5  # type: float
+    BLOBSTORE_BOTO_RETRIES = 2  # type: int
+    BLOBSTORE_GS_MAX_CUMULATIVE_RETRY = 15  # type: float  ## seconds
+
     _ALLOWED_EMAILS = None  # type: typing.Optional[str]
     _CURRENT_CONFIG = BucketConfig.ILLEGAL  # type: BucketConfig
     _NOTIFICATION_SENDER_EMAIL = None  # type: typing.Optional[str]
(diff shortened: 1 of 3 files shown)


dss/config.py
```diff
@@ -94,6 +97,11 @@ class Config:
     _S3_CHECKOUT_BUCKET = None  # type: typing.Optional[str]
```
hannes-ucsc: I think you can actually say `_S3_CHECKOUT_BUCKET: typing.Optional[str] = None`

xbrianh:
@hannes-ucsc 

See timeout tests from `cloud-blobstore`:
https://github.com/chanzuckerberg/cloud-blobstore/blob/afd769116994d1d2534ea971340cbb2af75cdee0/tests/test_gsblobstore.py#L49
https://github.com/chanzuckerberg/cloud-blobstore/blob/afd769116994d1d2534ea971340cbb2af75cdee0/tests/test_s3blobstore.py#L159

hannes-ucsc:
Very cool!

hannes-ucsc:
Correct me if I am wrong: The dupes are caused by 

1) The client requests `PUT /bundles` without a version.
2) S3 occasionally has a high latency that slows down server-side processing of the request. 
3) A timeout occurs and a 504 is returned to the client by either 
   * the API Gateway or 
   * @ttung's timeout logic in `chalice/app.py`
4) Concurrently, the event handling thread continues to run and eventually succeeds.  
5) The client repeats the request, again without a version. Two versions of essentially the same bundle are the result.

This PR makes 3) less likely (but not impossible) by causing 4) to timeout by itself before 3) kicks in. Since there is no guarantee that 3) never occurs, only a reduced probability, we still need to handle that case. Making version mandatory encourages the client to repeat the request with both UUID and version present. If the repeated request was preceded by a 504 that resulted from 3), the client will likely get a 409 and the client can treat that as a success.

hannes-ucsc:
Independently of the fact that we should make version mandatory for the sake of correctness—this PR is an improvement in its own right simply because it reduces the probability of returning 504s while still creating a bundle version in the background. Can we move this forward, @ttung?

ttung:
So this is just to return 503 instead of 504?

I'm rather curious how often things time out because of reading the /files metadata vs the write to /bundles.  I strongly suspect it's the former, in which case, this will almost never get triggered.

Also, #1173 prevented the spin-loop from working correctly....

ttung:
I won't object to this landing, but I'm guessing it'll almost never get triggered.

xbrianh:
I agree @ttung, this will not usually be triggered.. although I've seen some surprisingly long write times for small files on S3 during casual testing.

But if the timeouts are working correctly, including request timeouts from this PR, bundles should never be created during timeouts. What's the point of requiring `version`? Are there other conditions where a bundle is created but the API returns 5xx?

hannes-ucsc:
Unless we can guarantee that the client never receives a 5xx in response to a `PUT /bundles` request, we have to make version mandatory. 

Since @ttung is [ok](https://github.com/HumanCellAtlas/data-store/pull/1196#issuecomment-386410041) with merging this, I think we can do so.

PR #1198, opened 2018-04-29
Support for notification attachments (resolves #1092)

### LINKED ISSUE #1092: Include metadata fields in notifications
As the secondary analysis service, we'd like it to be possible to include bundle metadata fields like project_json.project_shortname in notifications, so that we can more easily search and sort things for troubleshooting purposes.

When registering a subscription, we would like the ability to include an optional list of fields to be returned in each notification, along with the bundle id and version.

The subscription would work the same as before, except that each notification, in addition to including the bundle id and version, would also include the extra requested fields.

Having this information available in the notification would allow us to more easily and effectively search and sort queued up notifications and workflows in the secondary analysis service, which would be very helpful for monitoring and troubleshooting purposes.

hannes-ucsc:
… aka custom notification fields.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -483,6 +483,70 @@ paths:
                 description: |
                   An optional key ID to use with `hmac_secret_key`.
                 type: string
+              attachments:
+                description: |
+                  The set of bundle metadata items to be included in the payload of a notification request to a
+                  subscription endpoint. Each property in this object represents an attachment to the notification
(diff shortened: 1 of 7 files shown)


hannes-ucsc:
The hardest part was specifying and documenting the externally facing API changes. I had most of those already done before we discussed punting this on Friday. The implementation was straight forward.

PR #1209, opened 2018-05-02
Move GCP credentials and application secrets to Secrets Manager

xbrianh:


.travis.yml
```diff
@@ -36,6 +32,7 @@ install:
 
 before_script:
 - export -n _JAVA_OPTIONS # https://github.com/travis-ci/travis-ci/issues/8408
+- source environment
```
hannes-ucsc: What's the reason for adding this here? It's already part of `before_install`. One of the two invocations is redundant, correct?

.travis.yml
```diff
@@ -19,10 +19,6 @@ addons:
     - gettext
 
 before_install:
-- openssl aes-256-cbc -K $encrypted_ead445d7a1e2_key -iv $encrypted_ead445d7a1e2_iv
-  -in gcp-credentials.json.enc -out gcp-credentials.json -d
-- openssl aes-256-cbc -K $encrypted_ead445d7a1e2_key -iv $encrypted_ead445d7a1e2_iv
-  -in application_secrets.json.enc -out application_secrets.json -d
```
hannes-ucsc: Since you're getting rid of both of these, I think you can remove the `secure` lines under `env.global`.

environment
```diff
@@ -90,3 +91,6 @@ set +a
 if [[ -f "${DSS_HOME}/environment.local" ]]; then
     source "${DSS_HOME}/environment.local"
 fi
+
+scripts/get_gcp_credentials.py || echo "Unable to recover gcp credentials from Secrets Manager. Your configuratoin is incorrect/incomplete."
+export ADMIN_USER_EMAILS=`jq ".client_email" ${GOOGLE_APPLICATION_CREDENTIALS} | sed 's/"//g'` || ADMIN_USER_EMAILS=""
```
hannes-ucsc: `jq -r` omits the quotes, I think, and that may allow you to skip the `sed` stage. This is not only an optimization: email addresses can contain double quotes, and they'd be escaped in the JSON string. The `sed` would replace them  indiscriminately.

https://tools.ietf.org/html/rfc5322#page-17

Search for definition of `quoted-string`

I'm aware that this is not your code originally but the fix is straight-forward so I thought I'd mention it.

.travis.yml
xbrianh: The reason is in the PR description.

.travis.yml
xbrianh: Fixed.

environment
xbrianh: Dropped the `sed`

dss/api/bundles/__init__.py
```diff
@@ -23,6 +23,7 @@
 """This is the minimum amount of time remaining on the lambda for us to retry on a PUT /bundles request."""
 
 ADMIN_USER_EMAILS = set(os.environ['ADMIN_USER_EMAILS'].split(','))
+print("frank", ADMIN_USER_EMAILS)
```
hannes-ucsc: remove

environment
```diff
@@ -90,3 +91,6 @@ set +a
 if [[ -f "${DSS_HOME}/environment.local" ]]; then
     source "${DSS_HOME}/environment.local"
 fi
+
+scripts/get_gcp_credentials.py || echo "Unable to recover gcp credentials from Secrets Manager. Your configuratoin is incorrect/incomplete."
```
kislyuk: typo *configuration

PR #1210, opened 2018-05-03
Fix bug in scripts/create_aws_event_relay_user.py

xbrianh:


xbrianh:
The required information is already in the `integration` secret, so it is not required.

hannes-ucsc:
Are you sure? The secrets look structurally different between staging and dev.

hannes-ucsc:
`integration` and `dev`

xbrianh:
The required key paths are `.AccessKey.AccessKeyId` and `.AccessKey.SecretAccessKey`, which are contained in both secrets. The `dev` secret has a bunch of extra key paths contained in the `create_access_key` response.

If I'm wrong smoketest will fail on integration. And every other test involving sync!

PR #1212, opened 2018-05-04
Remove backlink from file to bundle

### LINKED ISSUE #1172: Allow a file to be included in more than one bundle
Hi bluebox

During an export of a result/analysis bundle to DSS, we specify files from the input bundle to be included in the analysis bundle if it has an exact same content like the data files and project.json metadata file. Moreover, it won't be possible to reupload the data files from previous submission since the upload area where those were previously uploaded will be already deleted by the time the result bundle export happens. 

We specify the same uuid and version of the file when we create a the bundle in DSS. 

It seems that this conflicts the bluebox specification that a file cannot be in more than one bundle. 
Would it be possible to allow this behaviour so that we'll be able to reuse same files in the result bundles. This would also avoid having files duplicated in the data store.

Many thanks!

hannes-ucsc:
Remove backlink from file to bundle (fixes #1172)

This makes it possible for a file to belong to more than one bundle. It has implications on the physical deletion of files. We cannot remove a file without enumerating all bundles, loading their metadata and then checking if any other bundles reference that file. However this is no different to the physical deletion of blobs: since one blob can belong to many files, the same enumeration is required there. Ultimately, physical deletion due to lack of consent will always require the deletion of the actual data, i.e. the blob. These changes do not significantly complicate the search for files and bundles that refer to it, or, IOW, the presence of a backlink from file to bundle does not siginifcantly simplify the problem.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -127,9 +127,6 @@ paths:
           description: Returns metadata
           headers:
             # edits to here should probably be reflected in the 302 section immediately below.
-            X-DSS-BUNDLE-UUID:
-              description: A RFC4122-compliant ID for the bundle that contains this file.
-              type: string
             X-DSS-CREATOR-UID:
               description: User ID who created this file.
               type: integer
@@ -201,9 +198,6 @@ paths:
           description: Redirects to a signed URL with the data.
           headers:
             # edits to here should probably be reflected in the 200 section above.
-            X-DSS-BUNDLE-UUID:
-              description: A RFC4122-compliant ID for the bundle that contains this file.
-              type: string
             X-DSS-CREATOR-UID:
               description: User ID who created this file.
               type: integer
@@ -289,16 +283,12 @@ paths:
                 description: Cloud URL for source data.
                 type: string
                 pattern: "^(gs|s3|wasb)://"
-              bundle_uuid:
-                description: A RFC4122-compliant ID for the bundle that contains this file.
-                type: string
               creator_uid:
                 description: User ID who is creating this file.
                 type: integer
                 format: int64
             required:
               - source_url
-              - bundle_uuid
               - creator_uid
       responses:
         200:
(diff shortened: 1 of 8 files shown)


dss-api.yml
```diff
@@ -1111,28 +1098,6 @@ definitions:
           - name
           - content-type
           - indexed
-  file_metadata:
```
xbrianh: Is file metadata associated with the backlink (file -> bundle)?

hannes-ucsc: No, this is just an orphaned section in the Swagger file. I committed the removal as a separate commit.

xbrianh: If this yml section is really orphaned/unneeded than it LGTM.

ttung: I think once upon a time, I wanted this to be the return value of GET /files/<version>.  However, that data ended up being returned in headers, so this is entirely obsolete.

I would inline blob_common_metadata into file_version, since that's the only usage of it.

dss/storage/hcablobstore/__init__.py
```diff
@@ -48,7 +48,6 @@ class FileMetadata:
     FILE_FORMAT_VERSION = "0.0.3"
```
ttung: I would rev the version.

tests/infra/storage_mixin.py
```diff
@@ -55,8 +54,7 @@ def upload_file(self: typing.Any, bundle_file: TestFile, replica: Replica) -> st
         response = self.upload_file_wait(
             bundle_file.url,
             replica,
-            file_uuid=bundle_file.uuid,
-            bundle_uuid=bundle_file.bundle.uuid,
+            file_uuid=bundle_file.uuid
```
ttung: Don't like trailing commas? :)

dss/storage/hcablobstore/__init__.py
hannes-ucsc: Will do in a follow-up PR.

tests/infra/storage_mixin.py
hannes-ucsc: Is the point for them to have minimal diffs when adding removing items?

ttung: Yup. :)

NBD.

PR #1213, opened 2018-05-04
 Use stage-specific CW rule names for notifier

### LINKED ISSUE #1211: CW rule that triggers notifier lambda conflicts between deployments
The stage that was deployed last will win. The notifiers in all other deployments in the same AWS account and region will not be triggered.

This was first seen when I deployed the notification changes to `integration`. After that `dev`notifier went dark causing smoketest to fail. The cron build on master today reverted it back to `dev` likely causing `integration` to fail.

Hotfix is in the works.

Long-term fix is tracked in https://github.com/kislyuk/domovoi/issues/9

hannes-ucsc:
(fixes #1211)

Since I had to touch the requirements files, I also fixed three regressions there

- aegea was added to the `.in` but the `.txt` was never updated
- the explicit `.in` entry for `botocore` was missing (we know we need 1.10 and the first commit proves that the downgrade to 1.9 does indeed happen without this entry, as I suspected)
- the lower bound on `click` was missing (we agreed to provide lower bounds in the `.in`)

I clobbered @ttung's manual one-time exclusion of `clickclick`. It may be unnecessary but we can't expect all devs to forever redo the exclusion. We need a way to declaratively exclude it or just accept its presence.

### DIFF
--- a/daemons/dss-notify/app.py
+++ b/daemons/dss-notify/app.py
@@ -21,11 +21,7 @@ def deploy_notifier():
 app = domovoi.Domovoi(configure_logs=False)
 
 
+@app.scheduled_function("rate(1 minute)", rule_name='run_notifier_' + Config.deployment_stage())
 def run_notifier(event, context):
     notifier = Notifier.from_config()
     notifier.run(context)
-
-
-# FIXME: https://github.com/HumanCellAtlas/data-store/issues/1211
-run_notifier.__name__ += "_" + Config.deployment_stage()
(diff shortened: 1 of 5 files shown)


ttung:
It's super weird that requirements.txt and requirement-dev.txt require different versions of botocore.  This is all the fault of pip not properly resolving dependencies. :(

hannes-ucsc:
@ttung, I fixed the botocore version discrepancy. I agree it's weird. This could happen anytime but less bad than downgrade to a version we know won't work. Pip sucks.

PR #1214, opened 2018-05-05
Collections API

kislyuk:


dss-api.yml
```diff
@@ -660,6 +660,250 @@ paths:
...
+            properties:
+              addContents:
+                type: array
+                items:
+                  $ref: '#/definitions/CollectionItem'
+              removeContents:
+                type: array
+                items:
+                  $ref: '#/definitions/CollectionItem'
+              name:
```
ttung: these fields need more description love.

dss/api/collections.py
```diff
@@ -0,0 +1,185 @@
...
+def patch(uuid: str, json_request_body: dict, replica: str, version: str):
+    # patch one collection with another
+    uuid = uuid.lower()
+    handle = Config.get_blobstore_handle(Replica[replica])
+    try:
+        cur_collection_blob = handle.get(Replica[replica].bucket, "{}/{}.{}".format(COLLECTION_PREFIX, uuid, version))
+    except BlobNotFoundError:
+        raise DSSException(404, "not_found", "Could not find collection for UUID {}".format(uuid))
+    collection = json.loads(cur_collection_blob)
+    for field in "name", "description", "details":
```
ttung: so you can patch someone else's collection?

dss/api/collections.py
```diff
@@ -0,0 +1,185 @@
...
+        if field in json_request_body:
+            collection[field] = json_request_body[field]
+    remove_contents_set = set(map(hashabledict, json_request_body.get("removeContents", [])))
+    collection["contents"] = [i for i in collection["contents"] if hashabledict(i) not in remove_contents_set]
```
ttung: IMO, I think it should be an error if something in `remove_contents_set` is not in the collection you're patching.

dss/api/collections.py
ttung: Also, you're not validating that the collection was tombstoned.  I think this is probably the right model, as it circumvents some nasty race condition issues, but it also creates some weirdness where someone can revive a collection that was tombstoned.

dss-api.yml
```diff
@@ -1285,6 +1529,52 @@ definitions:
...
+    required:
+      - name
+      - description
+      - details
+      - contents
+  CollectionItem:
+    type: object
+    properties:
+      type:
+        type: string
```
ttung: should this be an enum?  i would argue yes.

dss-api.yml
```diff
@@ -1285,6 +1529,52 @@ definitions:
...
+        format: date-time
+        description: RFC3339 timestamp of the file, bundle, collection, or file containing a metadata fragment.
+      path:
+        type: string
+        description: |
+          JSON Pointer indexing into the JSON conents of the file containing a metadata fragment.
+          Required for links of type other than file, bundle, or collection. Ignored otherwise.
+    required:
+      - type
+      - uuid
```
ttung: shouldn't version be required?

dss-api.yml
```diff
@@ -660,6 +660,250 @@ paths:
...
+          description: Unexpected error
+          schema:
+            allOf:
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, not_found, invalid_link]
```
ttung: I wonder if there should be some text description for the error codes.

dss-api.yml
ttung: Also, what's the magic other type?  I don't get it.  Why not just make it explicit?

dss/api/collections.py
```diff
@@ -0,0 +1,185 @@
...
+                item_metadata[FileMetadata.SHA256],
+                item_metadata[FileMetadata.SHA1],
+                item_metadata[FileMetadata.S3_ETAG],
+                item_metadata[FileMetadata.CRC32C],
+            ))
+            # check that item is marked as metadata, is json, and is less than max size
+            item_doc = json.loads(blobstore_handle.get(replica.bucket, blob_path))
+            item_content = jsonpointer.resolve_pointer(item_doc, item["fragment"])
+            return item_content
+    except Exception as e:
```
ttung:
```
except DSSException:
    raise
except Exception as ex:
    raise DSSException(...
```

dss-api.yml
```diff
@@ -660,6 +660,250 @@ paths:
...
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, not_found]
+                required:
+                  - code
+    patch:
```
ttung: I wonder if it would be better if collections.patch requires a destination version.  I think so, but I don't have the argument fully mapped out.

dss-api.yml
```diff
@@ -660,6 +660,250 @@ paths:
...
+          type: string
+          enum: [aws, gcp, azure]
+      responses:
+        200:
+          description: OK
+          schema:
+            type: object
+            properties:
+              collections:
+                type: array
```
hannes-ucsc: Suggested: Looks like this route returns all collections of a user. Typically, directory routes should return a digest or listing of the items underneath. Or at best a list of sparsely populated items. I don't think we want this route to return all collections along with each collection's items.

hannes-ucsc: Requested: Just realized that this route raises NIE. I think we should not declare/document unimplemented API routes so as to not confuse or frustrate consumers. My suggestion above hints at ways this route might change.

dss/api/collections.py
```diff
@@ -0,0 +1,185 @@
...
+    collection = json.loads(cur_collection_blob)
+    for field in "name", "description", "details":
+        if field in json_request_body:
+            collection[field] = json_request_body[field]
+    remove_contents_set = set(map(hashabledict, json_request_body.get("removeContents", [])))
+    collection["contents"] = [i for i in collection["contents"] if hashabledict(i) not in remove_contents_set]
+    verify_collection(json_request_body.get("addContents", []), Replica[replica], handle)
+    collection["contents"].extend(json_request_body.get("addContents", []))
+    timestamp = datetime.datetime.utcnow()
+    new_collection_version = datetime_to_version_format(timestamp)
```
hannes-ucsc: Not idempotent. See @ttung's and my comments above in the Swagger def.

dss-api.yml
hannes-ucsc: PATCH opens a whole new can of worms when it comes to versioning.

Suggested: The destination version should be mandatory to be consistent with PUT.

Suggested: I would also require the source version and return a 409 if it is not the latest version.

The scenario to think about is having multiple concurrent PATCH requests to the same collection.

dss-api.yml
```diff
@@ -660,6 +660,250 @@ paths:
...
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, not_found, invalid_link]
+                required:
+                  - code
+  /collections/{uuid}:
+    get:
+      summary: Retrieve a collection given a UUID.
+      description: |
+        Given a collection UUID, return the associated collection object.
+      security:
+        - googAuth:
```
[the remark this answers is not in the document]
kislyuk: Collections are intended to stay private until we integrate an authorization mechanism that allows sharing them.

After the Collections implementation lands, the intent is to create a Releases API. A release is a collection promoted into a global namespace and made immutable and public.

dss-api.yml
kislyuk: Added descriptions.

dss-api.yml
kislyuk: Added description for the novel `invalid_link` error code. For other codes, I'll address this in a separate PR.

dss-api.yml
```diff
@@ -660,6 +660,250 @@ paths:
...
+          pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
+        - name: replica
+          in: query
+          description: Replica to delete from. Can be one of aws, gcp, or azure.
+          required: true
+          type: string
+          enum: [aws, gcp, azure]
+        - name: version
+          in: query
+          description: Timestamp of file creation in RFC3339.  If this is not provided, the latest version is returned.
```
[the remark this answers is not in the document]
kislyuk: Fixed, sorry about that.

dss-api.yml
kislyuk: Enum does not allow a wildcard. I specifically wanted custom types to be declarable. This allows the collection pointer schema to be extensible from upstream. For example, imagine someone wants to express "species" as a first class collection content type (made up example). They could declare a type "species" and use the JSON Pointer to point to a particular fragment of a metadata file representing a species in their metadata schema. The intent is to allow entities that are not first class in the DSS API to be consistently referenced within DSS.

dss-api.yml
kislyuk: It should be, fixed, thanks.

dss/api/collections.py
kislyuk: Fixed, thanks.

dss-api.yml
kislyuk: I would prefer that it remain optional, as I do believe keeping it optional increases the usability of the API, and that any orphan objects created by incomplete requests are inconsequential. I'm OK with making it mandatory if there is consensus on that.

The source version is already required. Added a test for that.

dss/api/collections.py
```diff
@@ -0,0 +1,185 @@
...
+logger = logging.getLogger(__name__)
+dss_bucket = Config.get_s3_bucket()
+
+def get_impl(uuid: str, replica: str, version: str = None):
+    uuid = uuid.lower()
+    bucket = Replica[replica].bucket
+    handle = Config.get_blobstore_handle(Replica[replica])
+
+    my_collection_prefix = "{}/{}".format(COLLECTION_PREFIX, uuid)
+    tombstone_key = "{}.{}".format(my_collection_prefix, TOMBSTONE_SUFFIX)
```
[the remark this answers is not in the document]
kislyuk: I think the logic in identifiers.py is overly complicated and does not actually result in an improvement compared to directly formatting strings, and so I would like to not use it here.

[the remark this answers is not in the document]
mweiden: OK sounds good

dss-api.yml
kislyuk: Sounds good. The intent here is to return a list of the current user's collections, paginated in the standard manner. However, this requires either making the primary index (S3 key prefix) be (f(user_email), collection_id), or making a secondary index (like `/collections/by-user/f(user_email)/collection_id` where the object contents are empty) - where f() may need to be sha256 for privacy, and the secondary index would provide a lookup hint only, with consistency verified by HEAD or GET on each collection). I'm not ready to implement either of those just yet, so I've removed the method for now and filed #1269 to track the related work.

dss-api.yml
```diff
@@ -660,6 +660,250 @@ paths:
...
+      summary: Create a collection.
+      description: |
+        Create a new collection.
+      security:
+        - googAuth:
+            - https://www.googleapis.com/auth/userinfo.email
+      parameters:
+        - name: replica
+          in: query
+          description: Replica to write to. Can be one of aws, gcp, or azure.
```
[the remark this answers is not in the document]
kislyuk: Thanks. Added some initial documentation here and added new bullet points related to sync in https://github.com/HumanCellAtlas/data-store/issues/1269.

dss/api/collections.py
kislyuk: It's not idempotent in the sense that it creates an orphan object, but as we discussed offline - I *think* that's not a big deal and orphaned collection objects should not matter. If there's a use case where this becomes an obvious issue, I'd be happy to reconsider. But as per the current design, you have to retrieve the newly assigned version to do anything with the result, which I hope means the API user is naturally guided toward the right pattern. I will look for places to add more docs to clarify this (added another checkbox in #1269).

dss/api/collections.py
```diff
@@ -0,0 +1,190 @@
...
+            "invalid_link",
+            'Error while parsing the link "{}": {}: {}'.format(item, type(e).__name__, e)
+        )
+
+def verify_collection(contents: List[dict], replica: Replica, blobstore_handle: BlobStore):
+    """
+    Given user-supplied collection contents that pass schema validation, resolve all entities in the collection and
+    verify they exist.
+    """
+    executor = ThreadPoolExecutor(max_workers=8)
```
hannes-ucsc: Request: Use TPE as a context manager to ensure timely shutdown.

dss/api/collections.py
```diff
@@ -0,0 +1,190 @@
...
+    for result in executor.map(partial(resolve_content_item, replica, blobstore_handle), contents):
```
hannes-ucsc: Question: This will raise and abort the loop on the first validation error leaving the TPE running until it is GCed, right?

Suggestion: I'd be more comfortable with partitioning the collection into batches of 100 items, running each batch separately through the TPE (via .submit()), exhausting all results by collecting the futures and populating the response with all errors in a batch. Or maybe accumulate errors until you have more than a certain threshold. Also timeouts should be handled more gracefully. This implementation seems overly naìve, but I'm probably overlooking something here which is why I've marked this as a suggestion.

dss/api/collections.py
```diff
@@ -0,0 +1,190 @@
...
+    except DSSException:
+        raise
+    except Exception as e:
+        raise DSSException(
+            422,
+            "invalid_link",
+            'Error while parsing the link "{}": {}: {}'.format(item, type(e).__name__, e)
+        )
+
+def verify_collection(contents: List[dict], replica: Replica, blobstore_handle: BlobStore):
```
hannes-ucsc: Suggestion: It seems that PATCH ultimately assumes that a collection is a set of items. I think we should therefore ensure uniqueness of collection items in PUT as well.

dss-api.yml
kislyuk: The point of this design is to allow custom types. I don't see the utility of controlling the namespace. I suppose we could tighten this up so type identifiers use only alphanumeric characters/underscores etc. I suppose we could also reserve a prefix (e.g. no custom types that start with "dss"), but at this stage my design is deliberately loosely defined to see how it can best accommodate metadata contents/patterns in releases.

dss/api/collections.py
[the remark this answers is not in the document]
kislyuk: I'm not sneaking anything in.

I'm going to go ahead and use identifiers.py as requested.

dss/api/collections.py
kislyuk: The implementation is deliberately simple as a first cut. Yes, the cleanup of the threadpool will need to be tightened up in the future to make sure no threads make it past the return from this frame. I don't think batching is needed. I've added a task in #1269.

dss/api/collections.py
kislyuk: I agree, implementing that.

dss/api/collections.py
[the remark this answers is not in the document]
kislyuk: Completed a refactor, using identifiers.py where appropriate now.

dss-api.yml
```diff
@@ -660,6 +660,244 @@ paths:
...
+          required: true
+          type: string
+          format: date-time
+        - name: json_request_body
+          in: body
+          required: true
+          schema:
+            type: object
+            properties:
+              addContents:
```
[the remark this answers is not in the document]
kislyuk: Replaced with underscore separated variable names.

dss-api.yml
```diff
@@ -660,6 +660,244 @@ paths:
...
+                description: New name for the collection.
+              description:
+                type: string
+                description: New description for the collection.
+              details:
+                type: object
+                description: New details for the collection.
+      responses:
+        200:
+          description: OK
```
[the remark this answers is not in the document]
kislyuk: Done.

dss/api/collections.py
kislyuk: That's the wrong thing to do. Using the ThreadPoolExecutor context manager will cause it to wait for all futures before leaving the frame. We deliberately want to return upon the first error before leaving the frame. When we do so, the futures that were running at that point will finish, but no new futures will run.

hannes-ucsc: > When we do so, the futures that were running at that point will finish, but no new futures will run.

I don't think so:

```py
from concurrent.futures import ThreadPoolExecutor
tpe = ThreadPoolExecutor(8)

def fail(x):
    assert False

l = list(range(100000))
for x in tpe.map(fail, l):
    pass
```

Note how it blocks for a long time. If you don't notice the blocking, add a zero to the end of the range.

Even if we changed this to using `tpe.submit()` and managed to bail out early, we'd still have to demonstrate that all worker threads are automatically shut down right away. Otherwise, we'd risk accumulating large amounts of threads over a short period of time if multiple PUT or PATCH requests come in.

So while I agree that bailing out early is the scalable thing to do, I'd be really surprised if fire and forget is the right approach here. That's why I suggested batching below: you have an upper bound on the number of failures but you do report multiple failures at once, which is user-friendly (and, assuming they fix all the issues before retrying the request, resource-friendly.

dss/api/collections.py
hannes-ucsc: I'd be happy with a simplest possible implementation that demonstrably didn't leak threads.

tests/test_collections.py
```diff
@@ -0,0 +1,146 @@
...
+            invalid_ptr = dict(type="foo", uuid=file_uuid, version=file_version, fragment="/xyz")
+            res = self.app.put("/v1/collections",
+                               headers=get_auth_header(authorized=True),
+                               params=dict(uuid=str(uuid4()), version=datetime.now().isoformat(), replica="aws"),
+                               json=dict(name="n", description="d", details={}, contents=[invalid_ptr]))
+            self.assertEqual(res.status_code, requests.codes.unprocessable_entity)
+        with self.subTest("PATCH invalid fragment reference"):
+            pass
+        with self.subTest("Dedup semantics"):
+            pass
```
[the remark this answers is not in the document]
kislyuk: Now implemented.

dss/api/collections.py
kislyuk: You're right, it is idempotent. However, I would prefer to postpone the implementation of throwing the error. Added a task in #1269.

dss/api/collections.py
kislyuk: Done.

dss/api/collections.py
kislyuk: You are right. What I said is true for long-running futures that block on I/O (which ours are), but can't be said in general. I think my original approach is still safe and efficient given the specifics of where and how our threads spend our time, but I'm fine with the batch+wait approach, since it's barely any more complex.

As an immediate solution, I've implemented batching and the context manager.

dss/api/collections.py
kislyuk: Updated as per the other comment.

Added a task in #1269 to make sure meaningful errors are thrown on timeout.

dss-api.yml
```diff
@@ -660,6 +660,258 @@ paths:
...
+        - name: uuid
+          in: query
+          description: A RFC4122-compliant ID for the collection.
+          required: true
+          type: string
+          pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
+        - name: version
+          in: query
+          description: |
+            Timestamp of collection creation in RFC3339 format. If this is not provided, the version is automatically
```
ttung: text needs to be updated.

dss-api.yml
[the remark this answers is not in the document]
ttung: Aren't collections also immutable?

dss-api.yml
```diff
@@ -660,6 +660,258 @@ paths:
...
+          description: Unexpected error
+          schema:
+            allOf:
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, not_found]
```
ttung: out of the scope of this diff, but we should move not_found to a different 404 definition.

dss-api.yml
```diff
@@ -660,6 +660,258 @@ paths:
...
                   - code
+  /collections:
+    put:
+      summary: Create a collection.
+      description: |
+        Create a new collection.
+
+        Collections are sets of links to files, bundles, other collections, or fragments of JSON metadata files. Each
+        entry in the input set of links is checked for referential integrity (the link target must exist in the replica
+        referenced). Up to 1000 items can be referenced in a new collection, or added or removed using
```
ttung: Nit: My reading from the later API specification is that 1000 items can be added and 1000 items can be removed.  My reading of this text here suggests that 1000 can be added or removed.

dss-api.yml
```diff
@@ -660,6 +660,258 @@ paths:
...
+        - name: replica
+          in: query
+          description: Replica to update the collection on. Updates are propagated to other replicas.
+          required: true
+          type: string
+          enum: [aws, gcp, azure]
+        - name: version
+          in: query
+          description: |
+            Timestamp of collection creation in RFC3339 format. Required to identify the collection to update.
```
ttung: Timestamp of the collection to update?

dss-api.yml
```diff
@@ -660,6 +660,258 @@ paths:
...
+          description: OK
+          schema:
+            type: object
+            properties:
+              uuid:
+                type: string
+                description: A RFC4122-compliant ID of this collection.
+              version:
+                type: string
+                description: New version of the updated collection.
```
ttung: format: date-time

dss-api.yml
```diff
@@ -660,6 +660,258 @@ paths:
...
+                      input could not be resolved (either because the file, bundle, or collection does not exist on the
+                      replica specified, or because the JSON Pointer reference could not be resolved within the
+                      specified file).
+                    enum: [unhandled_exception, Forbidden, OAuthResponseProblem, not_found, invalid_link]
+                required:
+                  - code
+    delete:
+      summary: Delete a collection.
+      description: |
+        Delete a collection.
```
ttung: I think I may have already asked this before, but I can't find the comment, but no delete-a-version?

dss/api/collections.py
```diff
@@ -0,0 +1,191 @@
...
+class hashabledict(dict):
+    def __hash__(self):
+        return hash(tuple(sorted(self.items())))
+
+@dss_handler
+def patch(uuid: str, json_request_body: dict, replica: str, version: str):
+    authenticated_user_email = request.token_info['email']
+
+    uuid = uuid.lower()
+    owner = get_impl(uuid=uuid, replica=replica)["owner"]
```
ttung: if patch indeed does respect tombstones, you can have race conditions.

dss-api.yml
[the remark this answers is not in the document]
kislyuk: What I meant was that release name for releases is analogous to (collection uuid, collection version), not just collection uuid. We could also make deletions impossible for releases (or collections pointed to by releases, if we choose to implement releases as pointers), though I'm not set on that.

dss-api.yml
kislyuk: I'm not sure if I follow, can you elaborate?

dss-api.yml
kislyuk: Changed to your wording, thanks.

dss-api.yml
kislyuk: Fixed.

dss-api.yml
kislyuk: Right, I'd prefer to keep this simple. Not sure if I see the use case for delete-a-version on collections yet.

dss/api/collections.py
kislyuk: I'm not sure if I follow. The point of tombstones is to be resilient to race conditions, no? Once an all-versions tombstone is placed, no version of the collection can be retrieved.

(Or are you concerned that the PATCH client will be under the mistaken assumption that since the PATCH request succeeded, its result is available?)

dss-api.yml
ttung: Right now, all these error can be returned for any non-201 code.  we'd ideally only want it for not_found to be returned for 404, and 404 should only allow not_found.

dss-api.yml
kislyuk: I agree there's a bit of ambiguity, but not sure how to re-word this better. Any suggestions?

kislyuk:
Merged but let's continue post-commit discussion here and in #1269

dss/api/collections.py
ttung: I take it back. We're replicating the files blindly, not replicating the operations.

PR #1218, opened 2018-05-07
Run integration tests on travis-ci.com instead of .org

hannes-ucsc:
I already deployed this daemon with this change to `dev`.


hannes-ucsc:
An IT build was triggered at 3pm, so this seems to be working.

PR #1219, opened 2018-05-07
Deploy buckets with terraform

xbrianh:
* Terraform objects defined in `infra/{component}`
* Only one component right now: `buckets`
* Terraform configuration carried out in `infra/build_deploy_config.py`

Testing (with custom bucket names!) may be carried out with:
`make deploy-infra`
`make -C infra apply COMPONENT=buckets`
`make -C infra destroy COMPONENT=buckets`

### DIFF
--- a/environment
+++ b/environment
@@ -77,6 +77,9 @@ EVENT_RELAY_AWS_USERNAME="dss-event-relay-${DSS_DEPLOYMENT_STAGE}"
 EVENT_RELAY_AWS_ACCESS_KEY_SECRETS_NAME="event-relay-user-aws-access-key"
 GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME="gcp-credentials.json"
 GOOGLE_APPLICATION_SECRETS_SECRETS_NAME="application_secrets.json"
+
+# `{account_id}`, if present, will be replaced with the account ID associated with the AWS credentials used for
+# deployment. It can be safely omitted.
 DSS_TERRAFORM_BACKEND_BUCKET_TEMPLATE="org-humancellatlas-dss-config-{account_id}"
 
 # Configure the delays between notification attempts. The first attempt is immediate. The second attempt is one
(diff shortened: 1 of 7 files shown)


environment
```diff
@@ -78,6 +78,7 @@ EVENT_RELAY_AWS_USERNAME="dss-event-relay-${DSS_DEPLOYMENT_STAGE}"
 EVENT_RELAY_AWS_ACCESS_KEY_SECRETS_NAME="event-relay-user-aws-access-key"
 GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME="gcp-credentials.json"
 GOOGLE_APPLICATION_SECRETS_SECRETS_NAME="application_secrets.json"
+DSS_TERRAFORM_BACKEND_BUCKET_TEMPLATE="org-humancellatlas-dss-config-861229788715"
```
hannes-ucsc: I don't think this should be hard-coded, right?

hannes-ucsc: I mean the account ID.

xbrianh: That bucket deployed externally.

This seemed like the simplest way to parametrize with account id, but I'm all ears for alternate suggestions.

hannes-ucsc: Actually, I'm unsure why we need to include the account id here in the first place. Anyone deploying our product in another AWS account needs to customize each and every bucket name to avoid conflicts with the buckets in our account. So if we parameterize one bucket name with the account ID we need to parameterize all of them.

Lets talk tomorrow after stand-up.

kislyuk:
I definitely meant the account ID needs to be parameterized. We
parameterize them in other places in our scripts, such as the ones that
manage IAM access policies.

environment
```diff
@@ -77,6 +77,7 @@ EVENT_RELAY_AWS_USERNAME="dss-event-relay-${DSS_DEPLOYMENT_STAGE}"
 EVENT_RELAY_AWS_ACCESS_KEY_SECRETS_NAME="event-relay-user-aws-access-key"
 GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME="gcp-credentials.json"
 GOOGLE_APPLICATION_SECRETS_SECRETS_NAME="application_secrets.json"
+DSS_TERRAFORM_BACKEND_BUCKET_TEMPLATE="org-humancellatlas-dss-config-{account_id}"
```
hannes-ucsc: Can you add a one-liner comment above explaining when and with what `{account_id}` is interpolated and that it's OK to remove it, too. It took me a while to figure this out. And I had the context in mind.

xbrianh: Added comment.

PR #1224, opened 2018-05-09
Connect/read blobstore timeouts enabled only for API server

xbrianh:
* Use default client timeouts for daemons, which are unconstrained by API Gateway.
* Revert timeout changes to `tests/test_gscopyclient.py`

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -28,6 +28,10 @@ from dss.util import paginate
 configure_lambda_logging()
 
 Config.set_config(BucketConfig.NORMAL)
+Config.BLOBSTORE_CONNECT_TIMEOUT = 5
+Config.BLOBSTORE_READ_TIMEOUT = 5
+Config.BLOBSTORE_BOTO_RETRIES = 2
+Config.BLOBSTORE_GS_MAX_CUMULATIVE_RETRY = 15
 
 
 EXECUTION_TERMINATION_THRESHOLD_SECONDS = 5.0
--- a/dss/config.py
+++ b/dss/config.py
@@ -97,10 +97,10 @@ class Config:
     _S3_CHECKOUT_BUCKET = None  # type: typing.Optional[str]
     _GS_CHECKOUT_BUCKET = None  # type: typing.Optional[str]
 
-    BLOBSTORE_CONNECT_TIMEOUT = 5  # type: float
-    BLOBSTORE_READ_TIMEOUT = 5  # type: float
-    BLOBSTORE_BOTO_RETRIES = 2  # type: int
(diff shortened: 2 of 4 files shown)


dss/config.py
```diff
@@ -116,32 +116,42 @@ def set_config(config: BucketConfig):
...
+    def get_native_handle(replica: "Replica") -> typing.Any:
         if replica == Replica.aws:
+            return Config._get_native_aws_handle()
+        elif replica == Replica.gcp:
+            return Config._get_native_gcp_handle()
+        raise NotImplementedError(f"Replica `{replica.name}` is not implemented!")
+
+    @staticmethod
+    def _get_native_aws_handle() -> typing.Any:
+        if Config.BLOBSTORE_CONNECT_TIMEOUT is None or Config.BLOBSTORE_READ_TIMEOUT is None:
```
ttung: what if `BLOBSTORE_BOTO_RETRIES` is set?

ttung: Also, this is an _or_, which is kind of weird.

dss/config.py
```diff
@@ -116,32 +116,42 @@ def set_config(config: BucketConfig):
...
                 )
             )
-        elif replica == Replica.gcp:
+
+    @staticmethod
+    def _get_native_gcp_handle() -> typing.Any:
+        if Config.BLOBSTORE_GS_MAX_CUMULATIVE_RETRY is not None:
             google.resumable_media.common.MAX_CUMULATIVE_RETRY = Config.BLOBSTORE_GS_MAX_CUMULATIVE_RETRY
 
+        if Config.BLOBSTORE_CONNECT_TIMEOUT is None or Config.BLOBSTORE_READ_TIMEOUT is None:
```
ttung: ditto, also an _or_.

dss/config.py
xbrianh: My idea here was that you should either provide values for all timeouts, or none of them.

Switched to allowing partial configs.

chalice/app.py
```diff
@@ -28,6 +28,10 @@
 configure_lambda_logging()
 
 Config.set_config(BucketConfig.NORMAL)
+Config.BLOBSTORE_CONNECT_TIMEOUT = 5
```
hannes-ucsc: This is why I think the upper-case convention is useless, and actually misleading. Here you're reassigning to a constant. There are no constants in Python and we should pretend it does.

Anyhow, different PR.

ttung: Yeah, I agree that this is kind of weird, but it's unclear to me what the solution is.

xbrianh:
connects to #1098

PR #1229, opened 2018-05-10
Deploy elasticsearch with terraform

xbrianh:

### DIFF
--- a/environment
+++ b/environment
@@ -71,12 +71,16 @@ DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_ARRAY=(
 DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS=${DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_ARRAY[*]}
 DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_TEST="human-cell-atlas-travis-test.iam.gserviceaccount.com"
 DSS_ES_DOMAIN="dss-index-$DSS_DEPLOYMENT_STAGE"
+DSS_ES_INSTANCE_TYPE="t2.small.elasticsearch"
+DSS_ES_VOLUME_SIZE="10"
+DSS_ES_INSTANCE_COUNT="1"
 PYTHONWARNINGS=ignore:ResourceWarning,ignore::UserWarning:zipfile:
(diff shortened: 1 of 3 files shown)


xbrianh:
Infrastructure managed by Terraform should be updated with Terraform, not with the console.

If we don't want ES access IPs to appear in the repo, Secrets Manager seems like a decent solution. The procedure to update access IPs (update the secret and run `make deploy-infra`), is more awkward than before, unfortunately.

hannes-ucsc:
> Infrastructure managed by Terraform should be updated with Terraform, not with the console.

Can you explain why we need to be dogmatic about this? The approach I propose seems to be more practical.

> If we don't want ES access IPs to appear in the repo

I agree that we don't. Not because they are secret, but because they change all the time.

Also, let me ask again: Is there a way to tell Terraform to not touch the policy on redeployments?

xbrianh:
If changes are made outside of Terraform, the state file becomes out-of-date.

IMO if we want to "deploy and forget", or "deploy and manage elsewhere", we should not use Terraform.

hannes-ucsc:
> If changes are made outside of Terraform, the state file becomes out-of-date.

And why is that a problem?

Also, let me ask again: Is there a way to tell Terraform to not touch the policy on redeployments?

hannes-ucsc:
To formulate the question differently: I'v noticed that the policy is optional in the TF module for ES. If you omit the policy from the TF config, will the state file contain a value for the policy? If it doesn't, then the state file would not become stale by users modifying the console through the console or the CLI.

From a birds eye perspective, I would be surprised if TF had this black or white approach to managed resources: either a resource is managed by TF and you can't configure any aspect of it through any other means than TF. That seems improbable but I don't know TF at all.

mweiden:
@hannes-ucsc Just want to add a datapoint in weighing the alternatives here: to my understanding, you don't have to redeploy to change the access policy on a terraform-managed ES cluster. Doing a `terraform apply` will simply change the access policy if that is the only difference in configuration.

xbrianh:
The ES access policy may be deployable externally to TF, but it seems unclear that a hybrid deployment is better than managing a list IPs through secrets manager.

Moving forward for now.

xbrianh:
Connects to #856

PR #1230, opened 2018-05-11
Add test for bundle with multiple files of the same name

hannes-ucsc:
If we allow duplicate file names in a bundle—and I'm not sure we do—we should have a test for it.

Is this a good way to test it?

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -160,54 +160,35 @@ class TestBundleApi(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
         schema = replica.storage_schema
 
         bundle_uuid = str(uuid.uuid4())
-        file_uuid = str(uuid.uuid4())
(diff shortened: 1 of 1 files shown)


kislyuk:
I think we should try to prevent duplicate file names. Is there a technical reason to avoid doing it?

hannes-ucsc:
@kislyuk I'm not aware of a requirement for duplicate file names in a bundle. I suspect that most people would intuitively expect a bundle to act like a archived file system directory similar to a tarball or zip file.

hannes-ucsc:
Note to self: create ticket before closing this.

ttung:
Strongly advocate prohibiting multiple files with the same filename, as downloading a bundle via the cli tools will be extremely confusing otherwise.

hannes-ucsc:
https://github.com/HumanCellAtlas/data-store/issues/1268

PR #1232, opened 2018-05-12
Deploy custom domain with Terraform

xbrianh:
Deploy a regional API Gateway custom domain name through scripting executed by the Terraform workflow.

This is a workaround while waiting for Terraform support for regional domains:
https://github.com/terraform-providers/terraform-provider-aws/issues/2195

#connect to #856


xbrianh:
Reviewers may find documentation for terraform `local-exec` provisioning useful:
https://www.terraform.io/docs/provisioners/local-exec.html

During testing, after successful execution of `local-exec`, terraform did not attempt to re-deploy upon further calls to `terraform apply`.

infra/domain/create_regional_domain_name.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
+
+"""
+This script configures a regional domain name for AWS Apigateway, based on an existing
```
kislyuk: The full human-readable name for the service is "API Gateway" and I would prefer it if we used it consistently throughout.

xbrianh: fixed.

PR #1233, opened 2018-05-13
Rename Reindex to IndexVisitation

hannes-ucsc:
With this we would have a consistent naming scheme:

`admin_cli.py index verify` => `IndexTarget.verify()` => `IndexVisitation.start(dryrun=True)`
`admin_cli.py index repair` => `IndexTarget.repair()` => `IndexVisitation.start(dryrun=False)`
`admin_cli.py storage verify` => `StorageTarget.verify()` => `StorageVisitation.start()`


xbrianh:
Perhaps consider ReindexVisitation?

hannes-ucsc:
> Perhaps consider ReindexVisitation?

I did. But it wouldn't fit the pattern. Targets are nouns (index, storage) and actions are verbs (verify, repair). This is easy to understand and consistent.

PR #1235, opened 2018-05-14
Instrument DSS chalice apps with AWS X-Ray

mweiden:
## Need (User story)
As a developer/operator of the DSS,
I need to having profiling information on calls in the API
to understand elevated latencies and timeouts under load.

## Approach
This pull request integrates [X-Ray](https://aws.amazon.com/xray/) into DSS's chalice applications so that we can get detailed tracing/profiling information on API endpoints that will experience bursts of load from users and batch processing frameworks like Spark.


chalice/app.py
```diff
@@ -17,10 +17,25 @@
 import requests
 from flask import json
 
-
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'chalicelib'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
+XRAY_TRACE = os.environ.get('XRAY_TRACE', 'false') == 'true'  # noqa
+
+if XRAY_TRACE:  # noqa
```
Bento007: If you want to instrument all of the lambdas you'll need to add this to all of the `app.py`s in `daemons/` right after `Domovoi` is called. Maybe we should add x-ray tracing hook to `Domovoi` or something to avoid lots of duplicate code.

chalice/app.py
```diff
@@ -17,10 +17,25 @@
...
+    from aws_xray_sdk.core import xray_recorder, patch
+    from aws_xray_sdk.core.context import Context
+    from aws_xray_sdk.ext.flask.middleware import XRayMiddleware
+    patch(('boto3', 'requests'))
+    xray_recorder.configure(
+        service='DSS',
+        dynamic_naming='*dss.{}data.humancellatlas.org*'.format(
```
Bento007: why not just use `API_DOMAIN_NAME` here?

mweiden: Good idea, I didn't know about that

PR #1241, opened 2018-05-16
Region env vars for some GS buckets

xbrianh:
Workaround for GS test buckets living in a different region than the dev bucket... for some reason.

connects to #856


mweiden:
@xbrianh can you help me understand what brought this about?

xbrianh:
@mweiden The Google `dev` bucket is not in the same region as the Google test and test-fixtures buckets. The Terraform scripts needs extra variables, beyond `GCP_DEFAULT_REGION`, to describe this. There is also a region mismatch in `staging`.

One alternative is migrating the buckets, if there's not a good reason for them to be in different regions.

xbrianh:
@mweiden Do you want to provide any extra input before this is merged?

mweiden:
@xbrianh no, other than you need to resolve the merge conflicts here

PR #1242, opened 2018-05-16
Fix flaky notifier test

hannes-ucsc:


hannes-ucsc:
@ttung, I'm told that you know sockets and TCP/IP. I'd appreciate your feedback on 3499bb07b1c95f38cd7b8e5a1501e539c7153afc. The fact that I need to send 10MiB worth of data over a loopback socket to detect a hang-up surprises me a little.

hannes-ucsc:
Thank you for looking into this. See also my Slack response.

PR #1244, opened 2018-05-17
Fix duplicate notifications

hannes-ucsc:
Another outcome from the F2F at Boston and the collaboration with @rexwangcc there. 

Backend timeouts are currently not handled very well and could cause duplicate notifications. For the remainder of this text, note that we run the actual indexer backend as a task in a thread pool executor.

E = estimated time for writing one document (tombstone or bundle document)
R = remaining time in lambda execution

Currently the main thread times out after waiting E seconds for the backend thread and exits with an exception. The exception will eventually cause the lambda execution to fail and then to be retried. While the main thread is shutting down, the backend thread continues to run in its own pool thread. The pool isn't shut down (by choice, so it doesn't block) and the backend thread keeps running. It is a daemon however and therefore doesn't prevent the termination of the process, in this case the lambda execution context. In other words, it just gets terminated at a random point in its life. Just by chance, the runaway backend thread sometimes succeeds in queueing the notification for the indexed bundle. One of the two built-in retried execution of the lambda will then likely succeed and queue another duplicate notification.

Increasing E from 10s to 60s eliminated this problem but leaves the last 60s of a lambda execution unused. This is ok when indexing but not so much when reindexing where we want to index many bundles per lambda execution. We can do better by decoupling the time the main thread waits for the backend thread from E and use R instead. In other words, we'll leave in place the check that prevents the indexer from comitting to perform the indexing operations alltogether (update the index and queue the notification) if R < E but once it decided to do so, it will be given R instead of E to doing so.

### DIFF
--- a/daemons/dss-index/app.py
+++ b/daemons/dss-index/app.py
@@ -14,7 +14,7 @@ from dss.index import DEFAULT_BACKENDS
 from dss.index.backend import CompositeIndexBackend
 from dss.index.indexer import Indexer
 from dss.logging import configure_lambda_logging
-from dss.util.time import RemainingLambdaContextTime
+from dss.util.time import RemainingLambdaContextTime, AdjustedRemainingTime
 
 app = domovoi.Domovoi(configure_logs=False)
 configure_lambda_logging()
@@ -46,8 +46,9 @@ def _handle_event(replica, event, context):
     executor = ThreadPoolExecutor(len(DEFAULT_BACKENDS))
     # We can't use ecxecutor as context manager because we don't want the shutdown to block
     try:
-        remaining_time = RemainingLambdaContextTime(context)
-        backend = CompositeIndexBackend(executor, DEFAULT_BACKENDS)
+        remaining_time = AdjustedRemainingTime(actual=RemainingLambdaContextTime(context),
+                                               offset=-10)  # leave 10s for shutdown
+        backend = CompositeIndexBackend(executor, remaining_time, DEFAULT_BACKENDS)
         indexer_cls = Indexer.for_replica(replica)
         indexer = indexer_cls(backend, remaining_time)
         indexer.process_new_indexable_object(event)
--- a/daemons/dss-index/app.py
+++ b/daemons/dss-index/app.py
@@ -14,6 +14,7 @@ from dss.index import DEFAULT_BACKENDS
 from dss.index.backend import CompositeIndexBackend
 from dss.index.indexer import Indexer
 from dss.logging import configure_lambda_logging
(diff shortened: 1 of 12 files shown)


daemons/dss-index/app.py
```diff
@@ -45,7 +45,7 @@ def _handle_event(replica, event, context):
     executor = ThreadPoolExecutor(len(DEFAULT_BACKENDS))
     # We can't use ecxecutor as context manager because we don't want the shutdown to block
```
rexwangcc: nit: a tiny typo with `ecxecutor`

daemons/dss-index/app.py
```diff
@@ -45,7 +45,7 @@ def _handle_event(replica, event, context):
...
     try:
-        backend = CompositeIndexBackend(executor, DEFAULT_BACKENDS)
+        backend = CompositeIndexBackend(executor, DEFAULT_BACKENDS, context)
```
rexwangcc: Could you clarify what does this context refer to?

hannes-ucsc: Lambda context. It is a an object that the Amazon Lambda runtime provides. We use it frequently enough in Blue for everyone to know what `context` refers to.

A word of caution: PRs labeled as `WIP` in their title are a work in progress and could change significantly. In fact I just modified this line.

rexwangcc: I see. Sry I didn't notice it's a WIP PR, will be back later :)

daemons/dss-index/app.py
hannes-ucsc: Fixed

dss/util/time.py
```diff
@@ -0,0 +1,68 @@
...
+    def get(self) -> float:
+        return max(0.0, self._deadline - time.time())
+
+
+class SpecificRemainingTime(RemainingTimeUntil):
+    """
+    A specific amount of wall clock time
+    """
+
+    def __init__(self, amount) -> None:
```
Bento007: add typing

dss/util/time.py
```diff
@@ -0,0 +1,68 @@
...
+    def get(self) -> float:
+        return self._context.get_remaining_time_in_millis() / 1000
+
+
+class RemainingTimeUntil(RemainingTime):
+    """
+    The remaining wall clock time up to a given deadline
+    """
+
+    def __init__(self, deadline) -> None:
```
Bento007: add typing

hannes-ucsc: Fixed and squashed

dss/util/time.py
hannes-ucsc: Fixed and squashed

PR #1245, opened 2018-05-17
[Easy] Component dirs must contain terraform files

xbrianh:
Subdirectories of /infa are considered components by /infa/Makefile only when they contain terraform files (files ending with .tf).

### DIFF
--- a/infra/Makefile
+++ b/infra/Makefile
@@ -1,4 +1,4 @@
-DIRS=${shell find . -name *.tf -exec dirname {} \; | sort --unique}
+DIRS=${shell find . -name "*.tf" -exec dirname {} \; | sort --unique}
 COMPONENTS=${shell for d in $(DIRS); do basename $$d; done}
 
 all: init-all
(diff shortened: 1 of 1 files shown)


infra/Makefile
```diff
@@ -1,4 +1,5 @@
-COMPONENTS=${shell find * -not -path "*/.*" -type d}
+DIRS=${shell find . -name *.tf -exec dirname {} \; | sort --unique}
```
hannes-ucsc: Quote the glob or this will fail if the current directory contains a file matching the glob. The shell would expand it before `find` gets it.

xbrianh: Terraform files will never be placed in the current directory. It would actually be better if this command fails otherwise.

hannes-ucsc: This really is a no-brainer. You need to quote globs if you don't want the the shell to interpret them. Relying on some implicit side-effect to enforce an invariant is is just plain insidious. Besides it doesn't work, it is not *guaranteed* to fail in the presence of `*.tf` file in the CWD, it could produce unpredictable results.

If you don't believe me, consult the manual page for find:

> Don't forget to enclose the pattern in quotes in order to protect it from expansion by the shell. 

(https://linux.die.net/man/1/find)

xbrianh: It is not my intention to rely on any side effects. It seemed useful to explain that placing terraform files in the current directory is non-comformant with the infra build system.

Thanks for the doc pointer: quotes added

PR #1247, opened 2018-05-18
Script for setting up Apigateway base path maping

xbrianh:
A script for idempotently mapping the AWS Apigateway custom domain name to the Apigateway stage. It should be executed for a first-time deployment after successfully running `make deploy-infra` and `make deploy`.

#connect to #856

### DIFF
--- a/scripts/set_apigateway_base_path_mapping.py
+++ b/scripts/set_apigateway_base_path_mapping.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+
+"""
+This script idempotently maps the AWS Apigateway custom domain name to the Apigateway
+stage. It should be executed for a first-time deployment after successfully
+running `make deploy-infra` and `make deploy`.
+"""
+
+import os
+import sys
+import json
+import boto3
+
+stage = os.environ['DSS_DEPLOYMENT_STAGE']
+domain_name = os.environ['API_DOMAIN_NAME']
+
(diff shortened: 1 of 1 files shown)


scripts/set_apigateway_base_path_mapping.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
+
+"""
+This script idempotently maps the AWS Apigateway custom domain name to the Apigateway
```
hannes-ucsc: Nit: Apigateway is a really odd spelling. `Amazon API Gateway` is the official spelling.

scripts/set_apigateway_base_path_mapping.py
```diff
@@ -0,0 +1,46 @@
...
+
+import os
+import sys
+import json
+import boto3
+
+stage = os.environ['DSS_DEPLOYMENT_STAGE']
+domain_name = os.environ['API_DOMAIN_NAME']
+
+APIGATEWAY = boto3.client('apigateway')
```
hannes-ucsc: Nit: The convention of upper-casing variables referencing boto3 clients objects has no precedence in this project. I would suggest sticking with the existing convention or filing a PR that establishes this convention consistently across the project after motivating it to the team.

hannes-ucsc: @Bento007 started upper-casing non-constants so we now have zero an annoying degree of inconsistency.

PR #1248, opened 2018-05-21
Added commit message requirements to contributing.md

Bento007:


hannes-ucsc:
I would really like to add the following language:

> Be liberal in accepting PRs

> Provided that a PR satisfies all other requirements with respect to scalability, test coverage and documentation, but you feel that it could be simplified, made more concise or more efficient, suggest that change (but don't request it). If the suggestion is not accepted by the author, and you still feel the project as a whole would benefit from it, make your case in a follow-on PR.

Bento007:
@hannes-ucsc I believe this is covered with the first and second bullet.

PR #1249, opened 2018-05-21
Instrumenting DSS Chalice with X-Ray

Bento007:
Another attempt at #1235


chalice/app.py
```diff
@@ -17,10 +17,23 @@
 import requests
 from flask import json
 
-
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'chalicelib'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
+XRAY_TRACE = os.environ.get('XRAY_TRACE', 'false') == 'true'  # noqa
```
hannes-ucsc: Can we call this `DSS_XRAY_TRACE` to be consistent with the existing naming convention?

hannes-ucsc: Changing the logic to 

`XRAY_TRACE = int(os.environ.get('XRAY_TRACE', '0')) > 0`

would fail if the environment variable was misconfigured and allow us to later define different levels of tracing.

hannes-ucsc: Nit: Let's not pretend that this is a constant by all-capsing it.

environment
```diff
@@ -92,6 +93,7 @@ DSS_TERRAFORM_BACKEND_BUCKET_TEMPLATE="org-humancellatlas-dss-config-{account_id
 # too infrequent. The last attempt is made seven days after the first one, which is easy to remember.
 #
 DSS_NOTIFY_DELAYS="0 60 600 3600 21600 60540 86400 86400 86400 86400 86400 86400"
+XRAY_TRACE=false
```
hannes-ucsc: Add new line above and a short comment.

chalice/app.py
Bento007: > Changing the logic to
> 
> XRAY_TRACE = int(os.environ.get('XRAY_TRACE', '0')) > 0
>
> would fail if the environment variable was misconfigured and allow us to later define different levels of tracing.

Do do think we will actually have multiple levels of profiling?

hannes-ucsc: We might. BTW: `DSS_DEBUG` works the same way: 0 is off, 1 enables debug-level logging for DSS code and 2 does so for all code. The problem I have with `true` and `false` is that I think it is 1) uncommon in the Unix world and 2) ambiguous wrt upper/lower case.

PR #1250, opened 2018-05-22
Fix flaky test of replica consistency check

hannes-ucsc:


hannes-ucsc:
Made delays deterministic and tightened assertion as suggested.

PR #1252, opened 2018-05-23
Add X-Ray profiling for daemons and subsegments for GET /v1/files

Bento007:
## Need (User story)
As a DSS operator
I want to profile uninstrumented (unpatched) DSS application logic (code not in `boto3` or `requests`)
so that I know what parts of an endpoint's logic are slow.

## Approach
Add xray tracing subsegments to the GET /files logic.

Adding xray tracing subsegments to the GET /files logic will allow us to profile unpatched code.

## Deployment Instructions

Must be deployed from a privileged account in order to update AWS Permissions for lambdas

### DIFF
--- a/daemons/dss-index/app.py
+++ b/daemons/dss-index/app.py
@@ -14,9 +14,13 @@ from dss.index import DEFAULT_BACKENDS
 from dss.index.backend import CompositeIndexBackend
 from dss.index.indexer import Indexer
 from dss.logging import configure_lambda_logging
+from dss.util import tracing
 from dss.util.time import RemainingLambdaContextTime, AdjustedRemainingTime
 
+
 app = domovoi.Domovoi(configure_logs=False)
+
+
 configure_lambda_logging()
 
 dss.Config.set_config(dss.BucketConfig.NORMAL)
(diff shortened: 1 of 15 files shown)


dss/xraytrace.py
```diff
@@ -0,0 +1,28 @@
...
+import typing
+
+DSS_XRAY_TRACE = int(os.environ.get('DSS_XRAY_TRACE', '0')) > 0  # noqa
+
+
+def config_xray() -> None:
+    if DSS_XRAY_TRACE:  # noqa
+        from aws_xray_sdk.core import xray_recorder, patch_all
+        patch_all()
+        xray_recorder.configure(context_missing='LOG_ERROR')
```
mweiden: Do you need to set the context here?

daemons/dss-index/app.py
```diff
@@ -16,7 +19,10 @@
 from dss.logging import configure_lambda_logging
 from dss.util.time import RemainingLambdaContextTime, AdjustedRemainingTime
 
+
 app = domovoi.Domovoi(configure_logs=False)
+
+
```
mweiden: nit: added whitespace?

mweiden:
Also you've got linter issues in Travis

dss/xraytrace.py
Bento007: I thought we only needed that when we weren't using the lambda_context within a lambda.

PR #1255, opened 2018-05-25
Fix flaky attachments test

hannes-ucsc:


hannes-ucsc:
Moving the `configure()` call into the loop ensures that the handler is reset for every sub test. Not resetting introduces a rare race:_get_received_notification_request could pick up the notification from the  previous sub test if the notification handler (running in a separate thread) doesn't update its _request field with the new notification quickly enough.

PR #1257, opened 2018-05-25
Fixing subsegment traces in aws Lambdas

Bento007:
Using Xray to profile lambdas we would like to use subsegments to profile specific segments of code. Currently subsegment tracing is not working using `dss.util.tracing.begin_segment` and `dss.util.tracing.end_segment`. This PR aims to fix that.

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -45,20 +45,18 @@ def get(uuid: str, replica: str, version: str=None):
 
 
 def get_helper(uuid: str, replica: Replica, version: str=None):
-    tracing.begin_segment('parameterization')
-    handle = Config.get_blobstore_handle(replica)
-    bucket = replica.bucket
-    tracing.end_segment('parameterization')
+    with tracing.Subsegment('parameterization'):
+        handle = Config.get_blobstore_handle(replica)
+        bucket = replica.bucket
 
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
I would change this to not end a subsegment when an exception occurs. A subsegment should only be closed when the code inside it runs successfully. Otherwise exceptions would distort your timing stats and they won't be obviously visible as open subsegments.

Bento007:
I agree, I'll just removed the context manager code.

mweiden:
@Bento007 Oops, I misread the documentation on `end_segment`. Thanks for catching this.

hannes-ucsc:
I think you should still use a CM because it provides a more concise syntax, but your `__exit__` methods needs to be a little bit smarter. That method can actually detect whether it is being called in an exception handling situation or under normal conditions.

https://www.python.org/dev/peps/pep-0343/#specification-the-with-statement

Note the parameters passed to the __exit__ method.

Bento007:
@hannes-ucsc thanks for the info. I modified the context manager to raise if an exception is found, but the subsegments is still closed. This will allow xray to mark it as an error when profiling to prevent exceptions from polluting our timing metrics.

PR #1259, opened 2018-05-25
Fix DSS_DEBUG not being propagated to lambdas

melissachang:
Before this PR:
- Put `DSS_DEBUG=1` in `environment.local`
- `cd chalice && make deploy`
- Observe that `chalice/.chalice/config.json` does't have `DSS_DEBUG`

After this PR, `chalice/.chalice/config.json` has `DSS_DEBUG`.


melissachang:
If there's interest in this PR, I can fix the checks

Issue #640, opened 2017-11-14
Ability to use unmanaged buckets by checkout service


amork:
## Need (user story)
As a user of the DSS's consumer API
I want to be able to check out a bundle into an AWS or GCP cloud storage bucket that I manage
so that I can compute on the data in my own account.

amork:
Document how users should configure S3 bucket permissions
Sample policy
Set up a bucket in CZI to test

amork:
Put documentation in the swagger file

Bento007:
should be supported but has not been tested

kozbo:
internally tested, closing

PR #1270, opened 2018-05-30
Checkout lifecycle rule deployment

xbrianh:
* Add 30 day lifecycle rule to GS checkout bucket, which matches the existing lifecycle rule set on the AWS checkout bucket: https://github.com/HumanCellAtlas/data-store/blob/387d908f494d8564f17af6b950b13878c3c5f524/infra/buckets/s3_buckets.tf#L27
* Remove unneeded checkout bucket lifecycle deployment script

Tested on my personal environment.


ttung:
Is there a ticket for renewing a checkout?

xbrianh:
@ttung Not that I'm aware of. We do have a ticket for checking out data to unmanaged buckets, which might work for those who want their data to last longer than the lifecycle period:
https://github.com/HumanCellAtlas/data-store/issues/640

PR #1276, opened 2018-05-31
Minor README changes

melissachang:


README.md
```diff
@@ -166,14 +167,16 @@ Assuming the tests have passed above, the next step is to manually deploy. See t
...
 
-* have the value `dss-index-$DSS_DEPLOYMENT_STAGE`
-
-* or, the environment variable `DSS_ES_DOMAIN` must be set to the domain name of the AWS Elasticsearch Service instance
+* The domain name must either:
+  * have the value `dss-index-$DSS_DEPLOYMENT_STAGE`
+  * or, the environment variable `DSS_ES_DOMAIN` must be set to the domain name of the AWS Elasticsearch Service instance
   to be used.
+* For typical development deployments the t2.small.elasticsearch instance type is more than sufficient.
+* Must be Elasticsearch 5.* instead of 6.* because [more than one document type
```
hannes-ucsc: Not sure we should capture the reasons in such a specific way here. For one, I don't think the doctype thing is the only incompatibility. IIRC, we use 5.x for historic reasons. Maybe @mikebaumann can chime in here.

melissachang: Took out reason. Also fixed the `hca` commands.

PR #1285, opened 2018-05-31
Add support for CircleCI

hannes-ucsc:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,8 @@
 version: 2
 
 defaults:
+  # This makes use of YAML's anchors and aliases in order to reduce duplication. For details on the `&`, `*` and `<<`
+  # operators refer to https://circleci.com/blog/circleci-hacks-reuse-yaml-in-your-circleci-config-with-yaml/
   environment: &environment
     DSS_ES_TIMEOUT: 30
     DSS_UNITTEST_OPTS: -v
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,6 @@ version: 2
 
 defaults:
   environment: &environment
-    DOCKER_REPO: datastoreci/data-store-build
     DSS_ES_TIMEOUT: 30
     DSS_UNITTEST_OPTS: -v
   after_checkout: &after_checkout |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,87 @@
+version: 2
+
+defaults:
+  environment: &environment
+    DOCKER_REPO: datastoreci/data-store-build
+    DSS_ES_TIMEOUT: 30
+    DSS_UNITTEST_OPTS: -v
+  after_checkout: &after_checkout |
+    git submodule sync
+    git submodule update --init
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,87 @@
...
+            - build_image
+      - test_search_and_subscriptions:
+          requires:
+            - build_image
+      - test_aws_indexer:
+          requires:
+            - build_image
+      - test_gcp_indexer:
+          requires:
+            - build_image
```
kislyuk: What do the `<<`, `&`, `*` operators mean? Please document them or put a link to the specific location in the CircleCI docs describing the operators in a leading comment.

Dockerfile.build
```diff
@@ -0,0 +1,62 @@
...
+    && echo 'eval "$(pyenv init -)"' >> ${HOME}/.bashrc
+RUN pyenv install ${PYENV_VERSION}
+RUN pip install --upgrade pip==10.0.1
+RUN pip install virtualenv==16.0.0
+
+ENV PATH /home/circle/make4/usr/bin:${PATH}
+RUN curl --silent --remote-name \
+        http://us-east-1.ec2.archive.ubuntu.com/ubuntu/pool/main/m/make-dfsg/make_4.1-6_amd64.deb \
+    && mkdir -p /home/circle/make4 \
+    && dpkg -x make*.deb /home/circle/make4
```
kislyuk: As we discussed, it would be nice (but not urgent) to get rid of this and go to 18.04. Could you file an issue/task to track that?

.circleci/config.yml
```diff
@@ -0,0 +1,87 @@
+version: 2
+
+defaults:
+  environment: &environment
+    DOCKER_REPO: humancellatlas/data-store-build
```
kislyuk: Consider parameterizing this.

.circleci/config.yml
hannes-ucsc: These are YAML primitives (anchors and aliases). I will add a link to the docs.

Dockerfile.build
hannes-ucsc: https://github.com/HumanCellAtlas/data-store/issues/1293

.circleci/config.yml
hannes-ucsc: Fixed

.circleci/config.yml
hannes-ucsc: Fixed

hannes-ucsc:
I confirmed that PRs from forks aren't built on CircleCI. There is a project-specific setting for that in CircleCI. If that setting is such that forked PRs are indeed built, the secret variables would leak, though. To address that we would can use the *contexts* feature:

> To use environment variables set on the Contexts page, the person running the workflow must be a member of the organization for which the context is set and the rule must allow access to all projects in the org.

https://circleci.com/docs/2.0/contexts/#overview

.circleci/config.yml
kislyuk: I see, thanks. YAML is impressive and scary.

PR #1291, opened 2018-05-31
minor optimization to test_sync

Bento007:
Speeding up the sync test by limiting the generation of random data.


Bento007:
@kislyuk good idea

Issue #1275, opened 2018-05-31
Consistency check should verify /files and /blobs, not just /bundles


hannes-ucsc:
Currently, we only compare the listing of `/bundles` in each replica.

See also #1274.

Issue #813, opened 2017-12-26
Visitation has minor generality/abstraction issues


hannes-ucsc:
While the goal of providing a general mechanism for distributed execution is commendable, and the current implementation goes a long way toward that, there are some issues left.

- The general mechanism has actually nothing to do with visitation. It implements what most people refer to as scatter/gather. Despite the generality of that mechanism, terms like walker are being used when, really, there are no walkers at that level of abstraction, only workers. These are mostly naming issues. Renaming `Visitation` to `Parallelization` or whatever, and `walker` to `worker` should do the trick.

- There is some duplication between the two concrete implementations: `IntegrationTest` and `Reindex`. This suggests that there should be an intermediate implementation, e.g. `BucketVisitation` from which the other two inherit.

- I would prefer to eliminate the _term thread_ pool from the terminology. Threads and thread pools are very specific things. I got royally confused when I first saw its mention in this context.

PR #1296, opened 2018-06-01
Consistency check of /files, /blobs and metadata/contents

### LINKED ISSUE #1274: Consistency check should verify object metadata and content for every key
TBD: Is there a way to infer from metadata that the contents are the same? If not, how do we ensure that the contents is the same? How do we do this efficiently for files and bundles. How do we do this efficiently for blobs?

hannes-ucsc:
Fixes #1274 and #1275.

### DIFF
--- a/daemons/dss-admin/app.py
+++ b/daemons/dss-admin/app.py
@@ -13,7 +13,6 @@ from dss import BucketConfig, Config, Replica
 from dss.logging import configure_lambda_logging
 from dss.stepfunctions.visitation.index import IndexVisitation
 from dss.stepfunctions.visitation.storage import StorageVisitation
-from dss.util import tracing
 from dss.util.types import JSON
 
 
@@ -60,15 +59,21 @@ class StorageTarget(Target):
     """
     Admin operations on the storage buckets and the replication between them.
     """
-    def __init__(self, replicas: Mapping[str, str]) -> None:
+    def __init__(self, replicas: Mapping[str, str], prefix: str = None) -> None:
+        super().__init__()
         replicas = ((Replica[replica], bucket) for replica, bucket in replicas.items())
         self.replicas = {replica.name: bucket or replica.bucket for replica, bucket in replicas}
-        super().__init__()
+        self.prefix = prefix or ''
 
-    def verify(self, workers: int) -> JSON:
+    def verify(self, workers: int, folder: str, quick: bool = False) -> JSON:
         assert 1 < workers
         replicas, buckets = zip(*self.replicas.items())
-        return StorageVisitation.start(workers, replicas=replicas, buckets=buckets)
+        return StorageVisitation.start(workers,
+                                       replicas=replicas,
+                                       buckets=buckets,
+                                       prefix=self.prefix,
+                                       folder=folder,
+                                       quick=quick)
 
 
 def _invoke(f: Callable, kwargs: Mapping[str, Any]) -> Any:
--- a/dss/stepfunctions/visitation/storage.py
+++ b/dss/stepfunctions/visitation/storage.py
@@ -1,12 +1,16 @@
+from abc import ABC, abstractmethod
+from collections import defaultdict
 import logging
 import string
-from typing import Mapping, MutableMapping, Sequence, List, Optional, Tuple
+from typing import List, Mapping, MutableMapping, NamedTuple, Optional, Sequence, Tuple, Type, Hashable, Set
 
-from cloud_blobstore import PagedIter
+from cloud_blobstore import BlobStore, PagedIter
+from frozendict import frozendict
 
 from dss.config import Config, Replica
+from dss.util import require
 from dss.util.iterators import zipalign
-from . import Visitation, WalkerStatus
+from dss.stepfunctions.visitation import Visitation, WalkerStatus
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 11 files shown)


dss/stepfunctions/visitation/storage.py
```diff
@@ -28,38 +31,44 @@ class StorageVisitation(Visitation):
...
         n = len(self.replicas)
         self.work_result = {k: [0] * n for k in self.work_result.keys()}
 
-    walker_state_spec = dict(tokens=None,
-                             row=None)
+    walker_state_spec = dict(tokens=None, row=None)
 
     def walker_initialize(self) -> None:
         n = len(self.replicas)
         self.row: Tuple[Optional[str], ...] = (None,) * n
```
xbrianh: I think `self.row` gets turned into a list when it's round-tripped through Step Functions state:
```
In [8]: json.loads(json.dumps((1,2,3)))
Out[8]: [1, 2, 3]
```

hannes-ucsc: That's the intent. Do you see any problems with that?

xbrianh: No, just confused by the initialization to an immutable type.

hannes-ucsc: … other than the type annotation being inaccurate? This is only one of two assignments and while the first one could be changed to a list, the second one is more or less bound to be a tuple because that's what `zip` conceptually is, an iterable of tuples. Tuples are cheaper than lists. I think I tried to fix the type hint this but the problem was that there is no type class in `typing` that reflects the support of the `[]` operator with an integer argument, i.e. `def __get__(i:int)`. I'll take another look.

The visitation code notoriously defeats the type checkers, both PyCharm and mypy, mostly because of the state (de)serialization magic. I should add this as one of my criticisms to https://github.com/HumanCellAtlas/data-store/issues/813.

xbrianh: In that case, perhaps postpone assignment to `self.row` until after the `while` loop.

hannes-ucsc: Maybe I should just explicitly convert the tuple to a list in the second assignment and do the assignment outside of the loop.

dss/stepfunctions/visitation/storage.py
```diff
@@ -81,6 +96,21 @@ def walker_walk(self) -> None:
                     else:
                         logger.debug(f"Replica {replica} contains {key}")
                         self.work_result['present'][i] += 1
+                        if values is not None:
+                            values[folder.inspect_key(i, handles[i], key)].add(replica)
+                if values is not None:
+                    assert len(values) > 0
+                    if len(values) > 1:
+                        logger.warning('Inconsistency detected for key %s: %r', row.min, values)
```
xbrianh: It might be useful to include the Step Functions execution name in the log messages, which would serve as a correlation id.

hannes-ucsc: I agree, but this should be done for all visitations consistently.

requirements.txt
```diff
@@ -21,6 +21,7 @@ elasticsearch-dsl==5.3.0
 fasteners==0.14.1
 Flask==0.12.2
 future==0.16.0
+frozendict==1.2
```
kislyuk: Consider vendoring or bundling this code (https://github.com/slezica/python-frozendict/blob/master/frozendict/__init__.py) instead of declaring a dependency. I think dependencies that haven't been updated in a while and don't have a lot of stars/other evidence of popularity are a potential hazard.

dss/stepfunctions/visitation/storage.py
```diff
@@ -103,3 +133,90 @@ def _aggregate(self, work_results: Sequence[Mapping[str, List[int]]]) -> Mutable
...
+    """
+    bucket: str
+    key: str
+    user_metadata: Mapping[str, str]
+    size: int
+    content_type: str
+    cloud_checksum: str
+
+
+class StorageFolder(ABC):
```
kislyuk: What's the point of using ABCs here? Is it to just auto-register subclasses in `cls_by_name`? I think it would be better accomplished by simply listing the classes in an array or something like that.

hannes-ucsc: I forgot to actually declare inspect_key() as abstract.

The registration stuff is unrelated. I'll take it out and replace it with an explicit list. It would be more LOCs but less magic.

requirements.txt
hannes-ucsc: Will do.

daemons/dss-admin/app.py
```diff
@@ -13,7 +13,6 @@
 from dss.logging import configure_lambda_logging
 from dss.stepfunctions.visitation.index import IndexVisitation
 from dss.stepfunctions.visitation.storage import StorageVisitation
-from dss.util import tracing
```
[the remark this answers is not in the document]
hannes-ucsc: It wasn't that's why I removed it.

dss/stepfunctions/visitation/storage.py
```diff
@@ -28,38 +32,48 @@ class StorageVisitation(Visitation):
...
 
     def job_initialize(self):
+        folders = StorageFolder.cls_by_name.keys()
+        require(self.folder in folders, f"Folder {self.folder} is not one of {folders}")
         if self._number_of_workers <= 16:
-            self.work_ids = self.prefix_chars
+            self.work_ids = [self.prefix + a for a in self.prefix_chars]
         else:
-            self.work_ids = [a + b for b in self.prefix_chars for a in self.prefix_chars]
+            self.work_ids = [self.prefix + b + a for a in self.prefix_chars for b in self.prefix_chars]
```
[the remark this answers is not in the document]
hannes-ucsc: That would not work since product returns tuples.

dss/stepfunctions/visitation/storage.py
hannes-ucsc: Fixed.

requirements.txt
hannes-ucsc: Vendored `frozendict`

PR #1299, opened 2018-06-04
Deploy test buckets only from the dev stage

xbrianh:
* All stages share test buckets
* Test buckets are only deployed from `dev`
* Add Terraform logic to accommodate GS buckets with different storage region settings.

### DIFF
--- a/infra/buckets/gs_buckets.tf
+++ b/infra/buckets/gs_buckets.tf
@@ -1,14 +1,13 @@
 resource google_storage_bucket dss_gs_bucket {
-  count = "${length(var.DSS_GS_BUCKET) > 0 ? 1 : 0}"
   name = "${var.DSS_GS_BUCKET}"
   provider = "google"
   location = "${length(var.DSS_GS_BUCKET_REGION) > 0 ?
     "${var.DSS_GS_BUCKET_REGION}" : "${var.GCP_DEFAULT_REGION}"}"
-  storage_class = "REGIONAL"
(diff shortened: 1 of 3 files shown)


environment.integration
```diff
@@ -7,6 +7,13 @@ DSS_GS_BUCKET=$DSS_GS_BUCKET_INTEGRATION
 DSS_S3_CHECKOUT_BUCKET=$DSS_S3_CHECKOUT_BUCKET_INTEGRATION
 DSS_GS_CHECKOUT_BUCKET=$DSS_GS_CHECKOUT_BUCKET_INTEGRATION
 DSS_ES_DOMAIN="dss-index-$DSS_DEPLOYMENT_STAGE"
+# Blank test bucket names so /infra does not deploy them
```
mweiden: @xbrianh could you explain this a little more?

infra/buckets/gs_buckets.tf
```diff
@@ -4,7 +4,7 @@ resource google_storage_bucket dss_gs_bucket {
   provider = "google"
   location = "${length(var.DSS_GS_BUCKET_REGION) > 0 ?
     "${var.DSS_GS_BUCKET_REGION}" : "${var.GCP_DEFAULT_REGION}"}"
-  storage_class = "REGIONAL"
+  storage_class = "${var.DSS_DEPLOYMENT_STAGE == "integration" ? "STANDARD" : var.DSS_DEPLOYMENT_STAGE == "staging" ? "STANDARD" : "REGIONAL"}"
```
mweiden: `dev` is regional?

xbrianh: Sorry for the convoluted logic, will fix.

PR #1304, opened 2018-06-07
Add xray trace_id to lambda logs if enabled.

Bento007:
Added xray trace_id to logs when enabled and present.

### DIFF
--- a/dss/util/tracing.py
+++ b/dss/util/tracing.py
@@ -1,12 +1,14 @@
 import os
 import time
 
-from aws_xray_sdk.core.exceptions.exceptions import SegmentNotFoundException
 from functools import wraps
 import typing
+from aws_xray_sdk.core.context import MISSING_SEGMENT_MSG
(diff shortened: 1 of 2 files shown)


dss/util/tracing.py
```diff
@@ -19,6 +21,26 @@
...
+    def filter(self, record):
+        if record.msg == MISSING_SEGMENT_MSG:
+            return False
+        try:
+            entity = xray_recorder.get_trace_entity()
+        except RecursionError:
+            return True
+        else:
+            record.xray_trace_id = entity.trace_id if entity else ""
+            record.msg = f"{record.xray_trace_id}\t{record.msg}"
```
mweiden: How about just a space since I'm pretty sure all other column breaks are also spaces?

Bento007: They are actually `tabs`, so i'm just sticking with consistency for now.

mweiden: kk

PR #1314, opened 2018-06-13
Precompile all .py files to .pyc before deploying

mweiden:
Chalice does not currently compile all `py` files to `pyc` files before deploying lambdas. For this reason, lambda cold start will be slower. This PR introduces compilation as part of the package and deployment of chalice apps.

### DIFF
--- a/scripts/dss-chalice
+++ b/scripts/dss-chalice
@@ -4,11 +4,11 @@
 # for https://github.com/aws/chalice/issues/479. TODO: (akislyuk)
 # revert to using the standard entry point when fixed.
 
-import os, sys
+import os, sys, compileall
 
 from chalice.cli import main
 
-import chalice.awsclient, chalice.deploy.deployer, chalice.deploy.packager
+import chalice.awsclient, chalice.deploy.deployer, chalice.deploy.packager, chalice.package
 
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Would be nice if we could contribute this back to upstream Chalice.

mweiden:
@ttung sadly, no. I still think it is a worthwhile change, however.

hannes-ucsc:
Before you abandon this, can you run one more test with

```
import requests
print requests.__cached__
```

in chalice/app.py in order to see if the interpreter actually picks up the bytecode files from the expected location on the Lambda file system? IIRC, there are restrictions in place as to which file system locations a Lambda can write to. Maybe Amazon redirected the location of the bytecode files, preventing them to be picked up from the default location where `compileall` places them.

ttung:
I suspect that if more memory (mostly) fixes this problem, it's not parsing the `.py` files, which should just be CPU-bound.

mweiden:
@ttung I think these two variables are conflated, though. Increasing the memory [proportionally increases the allotted CPU time](https://docs.aws.amazon.com/lambda/latest/dg/resource-model.html).

PR #1317, opened 2018-06-14
Remove get_logs route from dss-api

Bento007:
This code is not longer used or working.


kislyuk:
Please remove the [WIP] label when requesting reviews.

PR #1324, opened 2018-06-15
Upload a test bundle for checking out.

ttung:
This allows us to do a proper checkout.

Depends on #1321


ttung:
This uploads to the test bucket now.

PR #1292, opened 2018-06-01
[WIP] Added extra large file to smoke test.

Bento007:
This is to improve coverage for the dss-sync lambda. 

- Factored out common code in smoketest
- Added test_smoketest_largefile using the value from dss/events/handlers/sync.py `parts_size['gs']`

PR #1332, opened 2018-06-15
Replicating smoke test using GCP

Bento007:
* Moved common setup code to setUpClass
* Added support for testing additional replicas

### DIFF
--- a/tests/test_smoketest.py
+++ b/tests/test_smoketest.py
@@ -68,37 +68,36 @@ class Smoketest(unittest.TestCase):
             # Disable workdir destructor
             cls.workdir._finalizer.detach()  # type: ignore
 
-    def test_smoketest(self):
-
         # Create a virtualenv and install the CLI
         #
-        venv = os.path.join(self.workdir.name, "venv")
+        venv = os.path.join(cls.workdir.name, "venv")
         run(f"virtualenv -p {sys.executable} {venv}")
-        venv_bin = os.path.join(venv, "bin", "")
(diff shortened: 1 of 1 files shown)


Bento007:
Connects to #1292

ttung:
Did you try running this smoke test?

There is still s3-only code in there:

```
        # Wait for the checkout to complete and assert its success
        #
        for i in range(10):
            res = run_for_json(f"{self.venv_bin}hca dss get-bundles-checkout --checkout-job-id {checkout_job_id}")
            status = res['status']
            self.assertGreater(len(status), 0)
            if status == 'RUNNING':
                time.sleep(6)
            else:
                self.assertEqual(status, 'SUCCEEDED')
                blob_handle = S3BlobStore.from_environment()
                object_key = get_dst_bundle_prefix(bundle_uuid, bundle_version)
                print(f"Checking bucket {self.checkout_bucket} object key: {object_key}")
                files = list(blob_handle.list(self.checkout_bucket, object_key))
                self.assertEqual(len(files), file_count)
                break
        else:
            self.fail("Timed out waiting for checkout job to succeed")
```

Bento007:
@ttung I ran the smoke test and it passed for `aws` but not for `gcp`. We need to finish adding support for `gcp` in some areas. S3 is required for this part of the test since we only support checking out to `S3`.

ttung:
The master integration test was consistently broken. Feel free to remerge this if I'm mistaken about the cause.

PR #1334, opened 2018-06-16
[easy] Rename the checkouts path to "bundles"

ttung:
This allows us to have two namespaces -- checked out files and checked out bundles.


mweiden:
@ttung can you add a little more context on the flow here?

ttung:
Yeah, so to service /files, since there's a M:N relationship between files and bundles, we can't do a bundle checkout to get the file. So we need to create the concept of a file checkout to support that endpoint.

mweiden:
@ttung OK, I see. Go ahead and 🚢

PR #1335, opened 2018-06-16
Unify all the places where we compose the blob key.

ttung:
Unify all the places where we compose the blob key.

Self-explanatory, no?

The only problem is that we apparently have a disagreement about s3-etag (what's written to cloud storage) and s3_etag (what's sent to users).  I recall going through and changing them but apparently I missed one place...?

### DIFF
--- a/dss/api/collections.py
+++ b/dss/api/collections.py
@@ -1,4 +1,4 @@
-import os, sys, json, logging, datetime, io, functools
+import json, logging, datetime, io, functools
 from typing import List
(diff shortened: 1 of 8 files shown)


ttung:
> Is there a way that we can fix the discrepancy?

There are no great solutions to this.  We can rev the storage format, migrate all existing files, and remove support for the old version, or we can rev what we send to the end user.  I don't know how long we would need to support the old data format.

PR #1336, opened 2018-06-17
Implement single-file checkout.

ttung:
Added test for extra fun times.

Depends on #1333, #1334, #1335

### DIFF
--- a/dss/storage/checkout.py
+++ b/dss/storage/checkout.py
@@ -67,6 +67,26 @@ def start_bundle_checkout(
     return execution_id
 
 
+def start_file_checkout(
+        blob_key,
+        replica: Replica,
+        dst_bucket: typing.Optional[str]=None,
+) -> str:
+    """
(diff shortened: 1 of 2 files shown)


dss/storage/checkout.py
```diff
@@ -76,6 +76,26 @@ def start_bundle_checkout(
...
+        blob_key,
+        replica: Replica,
+        dst_bucket: typing.Optional[str]=None,
+) -> str:
+    """
+    Starts a file checkout.
+
+    :param blob_key: The key of the blob that contains the file.
+    :param replica: The replica to execute the checkout in.
+    :param dst_bucket: If provided, check out to this bucket.  If not provided, check out to the default checkout bucket
```
mweiden: Could this be a user-provided bucket?

ttung: This API supports a user-provided bucket.  There is no endpoint to accomplish that, however.

PR #1341, opened 2018-06-19
Support for checkout service in GET /files

### LINKED ISSUE #1310: GET /files/<uuid> should return links to checkout bucket
blah blah.

ttung:
Depends on #1336, #1343 
Fixes #1310

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -1,7 +1,7 @@
 import datetime
 import json
-import random
 import re
+import time
 import typing
 from enum import Enum, auto
 from uuid import uuid4
@@ -13,10 +13,11 @@ from flask import jsonify, make_response, redirect, request
 
 from dss import DSSException, dss_handler, stepfunctions
 from dss.config import Config, Replica
+from dss.storage.checkout import CheckoutTokenKeys, get_dst_key, start_file_checkout
 from dss.storage.files import write_file_metadata
 from dss.storage.hcablobstore import FileMetadata, HCABlobStore, compose_blob_key
 from dss.stepfunctions import gscopyclient, s3copyclient
-from dss.util import tracing
+from dss.util import tracing, UrlBuilder
 from dss.util.aws import AWS_MIN_CHUNK_SIZE
 from dss.util.version import datetime_to_version_format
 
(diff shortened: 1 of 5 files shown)


mweiden:
@ttung What happened here? Are you splitting this up?

ttung:
Github mistakenly closed this.

dss/api/files.py
```diff
@@ -108,6 +109,39 @@ def get_helper(uuid: str, replica: Replica, version: str=None):
...
+        decoded_token = {
+            CheckoutTokenKeys.EXECUTION_ID: execution_id,
+            CheckoutTokenKeys.START_TIME: start_time,
+            CheckoutTokenKeys.ATTEMPTS: attempts
+        }
+    else:
+        try:
+            decoded_token = json.loads(token)
+            decoded_token[CheckoutTokenKeys.ATTEMPTS] += 1
+        except Exception as ex:
```
mweiden: All exceptions? Why not just `(KeyError, ValueError)`?

dss/storage/checkout.py
```diff
@@ -23,6 +26,90 @@ class ValidationEnum(Enum):
...
 
+class BundleNotFoundError(Exception):
+    """Raised when we attempt to check out a non-existent bundle."""
+    pass
+
+
+class CheckoutTokenKeys:
+    """
+    When we are executing a request that involves a checkout, the client will periodically check back in to see if the
+    checkout is complete.  To avoid duplicating checkout requests, the client will check back using a token.  These are
```
mweiden: It seems like the deduplication is only done by checking that the token exists, the only key-value pair touched is `attempts`, and that value isn't leveraged. Seems like a simple flag would perform the same function(?) Is there a dev UX element here?

dss/api/files.py
ttung: Good idea, will fix.

dss/storage/checkout.py
ttung: It's for future use.  If time has elapsed very little, but attempts is high, that means we have a client that's not respecting the Retry-After header.  We could also implement exponential backoff or other creative solutions for managing retries.

Issue #835, opened 2018-01-08
Use elasticsearch "search after" when results < 10,000


Bento007:
### Expected behavior
Having multiple scroll context open at the same time can slow down the system and consumes memory. We should use [search after](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html) and remove the scroll feature.

### Actual behavior
Scroll is used for all searches if the results cannot be returned in one request.

PR #1344, opened 2018-06-19
Replaced scroll with search after

Bento007:
Using search after for paged searching. This will reduce the overhead required to search beyond 10,000 results at the cost of search results being dynamically generated.

Relates to #835

## Test Plan
Removed test cases relating to scroll_id

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -52,7 +52,7 @@ def post(json_request_body: dict,
             response = make_response(request_body, requests.codes.ok)
         else:
             response = make_response(request_body, requests.codes.partial)
-            next_url = _build_scroll_url(page['_scroll_id'], per_page, replica_enum, output_format)
+            next_url = _build_scroll_url(page, per_page, replica_enum, output_format)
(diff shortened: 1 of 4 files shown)


dss/api/search.py
```diff
@@ -96,31 +96,27 @@ def _es_search_page(es_query: dict,
...
-    # {
-    #   "sort": [
-    #     "_doc"
-    #   ]
-    # }
-    sort = {"sort": ["_doc"]}
+    # https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-search-after.html
+    sort = [
+        # "manifest.version:desc",
+        "_uid:desc"]
```
hannes-ucsc: Why can't we just sort by bundle UUID and version (FQID)? Then we won't even need to use the opaque `_uid` field which is probably yet another UUID and looks just like a bundle UUID which could be confusing.

I'd prefer the sorting to be meaningful rather than random.

Bento007:
Ready for review

PR #1350, opened 2018-06-21
Move retry logic into assert_mixin

ttung:
Since both files and bundles may need retries, move that logic into assert_mixin so we can reuse it.

### DIFF
--- a/tests/infra/assert_mixin.py
+++ b/tests/infra/assert_mixin.py
@@ -1,12 +1,14 @@
-import json
-import pprint
-import typing
-
 import collections.abc
 import functools
+import json
+import pprint
 import re
+import typing
 
+import requests
 from flask import wrappers
+
+from dss.util import UrlBuilder
 
 
 class ExpectedErrorFields(typing.NamedTuple):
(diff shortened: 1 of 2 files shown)


tests/infra/assert_mixin.py
```diff
@@ -62,7 +67,16 @@ def assertResponse(
...
 
-        response = getattr(self.app, method)(path, **kwargs)
+        for ix in range(redirect_follow_retry + 1):
+            response = getattr(self.app, method)(path, **kwargs)
+            if ix == redirect_follow_retry or response.status_code != requests.codes.moved:
+                break
+            retry_after = int(response.headers['Retry-After'])
+            self.assertGreaterEqual(retry_after, min_retry_interval_header)
+            url = response.headers['Location']
+            builder = UrlBuilder(url).set(scheme="", netloc="")
```
mweiden: Why no scheme?

ttung: URLs passed to the the unit test chalice server only have the path.  /v1/files, for instance.

PR #1352, opened 2018-06-22
Backed out #1332

ttung:


ttung:
Landing now to unbreak the smoketest.

PR #1353, opened 2018-06-22
Backed out #1351

ttung:
This parameter does exist, just that my ~/.config/hca was outdated (smoketest should really just pull the config anew...)


ttung:
Landing to unbreak tests.

PR #1354, opened 2018-06-23
Remove magic constants from checkout code.

ttung:
Also renamed some variables named `bundle_id` or `bundle_fqid` that were actually bundle_uuid.  Also renamed some variables named `version` that were actually bundle_version.

Test plan: Deployed daemon and ran integration+standalone checkout tests.

### DIFF
--- a/dss/stepfunctions/checkout/checkout_states.py
+++ b/dss/stepfunctions/checkout/checkout_states.py
@@ -12,107 +12,133 @@ from dss.storage.checkout import (
     pre_exec_validate,
     validate_file_dst,
 )
+from .constants import EventConstants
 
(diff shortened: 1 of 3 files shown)


ttung:
Did you mean Enum?

mweiden:
@ttung no I meant a class or struct that you could load with the fields in the event and exposed as properties. Might be easier than lots of dict keys. This is a total nit, though

ttung:
Can one json-serialize a namedtuple?

ttung:
answer: nope.

ttung:
python's json library really needs to a way for types to provide serialization and deserialization methods. :(

Issue #1300, opened 2018-06-04
Investigate API Gateway Timeouts during high loads


Bento007:
### AWS Support ticket:
https://console.aws.amazon.com/support/v1#/case/?displayId=5101082101&language=en

### Graph of the Spike
https://metrics.dev.data.humancellatlas.org/d/F9svZvSik/dss-dev-scale-test-spike?orgId=1 

### Xray traces
https://console.aws.amazon.com/xray/home?region=us-east-1#/service-map?timeRange=2018-05-30T14:00:00Z~2018-05-30T14:20:00Z

### Cloudwatch Logs
https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logStream:group=/aws/lambda/dss-dev;prefix=2018/05/30;streamFilter=typeLogStreamPrefix

Bento007:
In some cases the dss-dev lambda takes >30 seconds to complete. I couldn't find any traces that showed a request taking >30. We may need to add additional instrumentation to find out where the time being spent.

Bento007:
This logged event took 50sec from an API-gateway call.
https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logEventViewer:group=/aws/lambda/dss-dev;stream=2018/06/05/%5B$LATEST%5D03f860b929f84d5f84d7550c9bc70643;refid=34080487634069493873104016217204606658262760580274389138;reftime=1528221919522

Bento007:
I suspect this is being cause by a lambda being cold started. The burst load produced by the load test caused additional lambdas to be spawned which need time to spawn. This causes requests to timeout until an additional lambda spawns.

The [graph](https://metrics.dev.data.humancellatlas.org/d/F9svZvSik/dss-dev-scale-test-spike?orgId=1&from=1528248614345&to=1528249199725&var-interval=5m) shows the load gradually increasing, and the latency increasing with it. The latency peaks and requests begin to timeout while another lambda is spawned. Once the additional lambda spawns the latency quickly drop back down.

Bento007:
> We need to also look into the timeout logic of the chalice/app and ensure that it isn't exceeding 30sec (28 seconds to be safe)

The dss chalice app's timeout needs to be lowered down to 28 seconds. This will make the app timeout quicker and more lambdas will be spawned sooner. 

We need to also remove the retry logic from the chalice app.

Bento007:
[...]

PR #1355, opened 2018-06-25
Changing dss-api lambda to use 256 memory

Bento007:


ttung:
We should put a simple time.time() before and after the import block and feed that to analytics.

Bento007:
@mweiden the other daemons are already set to 1536 MB. The DSS-API was the only once set bellow that.

Bento007:
@ttung we did something to that effect, see https://github.com/HumanCellAtlas/data-store/issues/1300#issuecomment-397090068

mweiden:
@Bento007 OK, I see. OK go ahead and 🚢

ttung:
> @ttung we did something to that effect, see #1300 (comment)

that should be landed for every daemon so we can keep track of how performance is.
we should also track how often daemons time out.

mweiden:
@ttung @Bento007 Just an 💡: this _could_ be accomplished by doing the xray imports first and wrapping imports with a timing subsegment.

ttung:
How expensive is the xray import?

ttung:
Also, I noticed the lambda logs reveal how much memory was consumed per execution.  We should aggregate P10,P50,P90 for that statistic per daemon.

PR #1358, opened 2018-06-25
[sts-bucket 1/2] Write the sts bucket into the checkout sfn message

ttung:
This adds, to the checkout sfn message, the bucket that we intend to write the checkout status key to.  Once this is in dev, we can deploy the sfn code that reads this key.


mweiden:
@ttung ty!

PR #1360, opened 2018-06-25
fix the key for RESULT.

ttung:
It's RESULT and not EMAIL_RESULT.


mweiden:
@ttung all the tests are failing?

ttung:
Probably because I did a test deploy to test #1363.

ttung:
Nope, actually it's #1364

PR #1374, opened 2018-06-26
Handle multipart copy that completes out of order.

ttung:
Currently, we're marking the entire range (A, B) of a multipart copy complete if B completes.  This is obviously not correct, as it only means B completes.  (A, B-1) still may not have completed.  Generally, this resolves itself as the parts do copy, but if the parts in (A, B-1) error out, then we don't ever complete the copy.

see dss-s3-copy-sfn-dev:9d5cb5f4-df3c-44ec-95bf-04dd54abebeb for an example of where this fails.

This rewrites the parallel copy management using the library in #1377.

### DIFF
--- a/dss/stepfunctions/s3copyclient/implementation.py
+++ b/dss/stepfunctions/s3copyclient/implementation.py
@@ -3,7 +3,6 @@ import collections
 import concurrent.futures as futures
 
 import hashlib
-import threading
 import typing
 
 import boto3
@@ -11,6 +10,7 @@ from cloud_blobstore.s3 import S3BlobStore
 
 from dss.stepfunctions.lambdaexecutor import TimedThread
 from dss.storage.files import write_file_metadata
+from dss.util import parallel_worker
 from dss.util.aws import get_s3_chunk_size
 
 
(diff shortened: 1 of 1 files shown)


dss/stepfunctions/s3copyclient/implementation.py
```diff
@@ -111,14 +111,28 @@ def run(self) -> dict:
...
+                complete[part_id] = False
 
             with futures.ThreadPoolExecutor(max_workers=CONCURRENT_REQUESTS) as executor:
                 def copy_one_part(part_id):
                     self.copy_one_part(part_id)
                     with state_lock:
-                        if part_id >= state[_Key.NEXT_PART]:
+                        complete[part_id] = True
+                        for scan_part_id in range(part_id, state[_Key.LAST_PART] + 1):
+                            if not complete[part_id]:
```
Bento007: is this ever `False`

dss/stepfunctions/s3copyclient/implementation.py
```diff
@@ -111,14 +111,28 @@ def run(self) -> dict:
...
+                # ... except for the parts that are missing.
+                complete[part_id] = False
 
             with futures.ThreadPoolExecutor(max_workers=CONCURRENT_REQUESTS) as executor:
                 def copy_one_part(part_id):
                     self.copy_one_part(part_id)
                     with state_lock:
-                        if part_id >= state[_Key.NEXT_PART]:
+                        complete[part_id] = True
+                        for scan_part_id in range(part_id, state[_Key.LAST_PART] + 1):
```
Bento007: what is the point of this loop if you aren't using `scan_part_id`?

ttung: Derp, that is broken.

kislyuk: Oops, I missed that.

ttung:
This is ready for review.

Bento007:
Could this be causing collections tests to timeout?

ttung:
Yeah, quite possibly.

PR #1375, opened 2018-06-27
Rename most things to execution_id

ttung:
The only thing I didn't change is the _value_ of EventConstants.EXECUTION_ID.  The rationale was to preserve existing sfn runs, but maybe we don't care and we can blow away backwards compatibility. :)

Test plan: `DSS_TEST_MODE="standalone integration" make -j tests/test_checkout.py`

Addresses comment https://github.com/HumanCellAtlas/data-store/pull/1373#pullrequestreview-132235939


mweiden:
@ttung the real concern would be checkout operations in production. I queried cloudwatch and there have been no requests to checkout in the month of June `=>` I think you're safe to change this.

ttung:
Ok changed the value to execution_id as well.

PR #1376, opened 2018-06-27
Adding GCP replica to smoketest

Bento007:
* Moved common setup code to setUpClass
* Added support for testing additional replicas
* Added GSBlobstore for GS checkout verification.

### Test plan
Ran smoke test locally against the dev deployment

### DIFF
--- a/tests/test_smoketest.py
+++ b/tests/test_smoketest.py
@@ -7,6 +7,7 @@ import subprocess
 
 import boto3
 import botocore
+from cloud_blobstore import BlobStore
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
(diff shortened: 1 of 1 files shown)


tests/test_smoketest.py
```diff
@@ -207,6 +214,19 @@ def test_smoketest(self):
             self.assertEquals(bundle_uuid, notification['match']['bundle_uuid'])
             self.assertEquals(bundle_version, notification['match']['bundle_version'])
 
+    def test_smoketest(self):
+        for replica in self.replicas:
+            with self.subTest(replica):
+                self.smoketest(replica)
+
+    def get_blobstore(self, replica):
```
ttung: I'd type the replica parameter.

tests/test_smoketest.py
```diff
@@ -52,9 +53,12 @@ def run_for_json(command, **kwargs):
...
 class Smoketest(unittest.TestCase):
-    replicas = "aws", "gcp"
+    replicas = {
+        "aws",
+        "gcp"
+    }
     notification_bucket = os.environ['DSS_S3_BUCKET_TEST']
-    checkout_bucket = os.environ['DSS_S3_CHECKOUT_BUCKET']
+    checkout_bucket = {'aws': os.environ['DSS_S3_CHECKOUT_BUCKET'], 'gcp': os.environ['DSS_GS_CHECKOUT_BUCKET']}
     test_bucket = os.environ['DSS_S3_BUCKET_TEST']
```
ttung: I think this needs to be parameterized too. 

Might be worth creating a replica to parameterization values dict.

tests/test_smoketest.py
```diff
@@ -110,89 +113,93 @@ def test_smoketest(self):
...
-        run(f"{venv_bin}hca dss download --replica gcp --bundle-uuid {bundle_uuid}")
+            # Download bundle from other replica
+            #
+            run(f"{self.venv_bin}hca dss download --replica {replica} --bundle-uuid {bundle_uuid}")
 
-        # Run a CLI search against the two replicas
-        #
         for replica in self.replicas:
-            run(f"{venv_bin}hca dss post-search --es-query='{{}}' --replica {replica} > /dev/null")
+            # Run a CLI search against the replicas
```
ttung: As the number of replicas potentially increases, this may be untenable.

Bento007:
Ready for review

tests/test_smoketest.py
```diff
@@ -52,13 +54,23 @@ def run_for_json(command, **kwargs):
 
 @testmode.integration
 class Smoketest(unittest.TestCase):
-    replicas = "aws", "gcp"
+    params = [
```
ttung: I would do this instead:

```
REPLICAS = {
  Replica.AWS: {
    'checkout_bucket_envvar': "DSS_S3_CHECKOUT_BUCKET",
    'test_bucket_envvar': "DSS_S3_BUCKET_TEST",
  },
  Replica.GCP: {
    'checkout_bucket_envvar': "DSS_GS_CHECKOUT_BUCKET",
    'test_bucket_envvar': "DSS_GS_BUCKET_TEST",
  },
}
```

ttung:
My comment about the number of replicas making things untenable is really a question directed at _all_ the code that checks / downloads from the other replicas, not just search.  I don't know if you need to special case search specifically, or even at all now.  It was a general comment for something we should pay attention to as we expand the # of replicas.

I think the right solution might be to continue checking each <cloud-provider> <-> <cloud-provider> combination.

tests/test_smoketest.py
Bento007: Changes it to use Replica, but I kept it as a list.

Issue #1357, opened 2018-06-25
Track down and fix build issues


kozbo:
# Why
We are experiencing too many build failures. As a developer I would like to have confidence that new code submissions would not be delayed by test failures in unrelated areas.

# DoD
 - [x] catalog test faailures

Bento007:
https://travis-ci.com/HumanCellAtlas/data-store/builds/77498998
```
======================================================================
ERROR: test_subscription_query_with_multiple_data_types_indexing_and_notification (__main__.TestAWSIndexer)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_indexer.py", line 568, in test_subscription_query_with_multiple_data_types_indexing_and_notification
    self.process_new_indexable_object(sample_event)
  File "tests/test_indexer.py", line 135, in process_new_indexable_object
    self.indexer.process_new_indexable_object(event)
  File "/home/travis/build/HumanCellAtlas/data-store/dss/index/indexer.py", line 32, in process_new_indexable_object
    self.index_object(key)
  File "/home/travis/build/HumanCellAtlas/data-store/dss/index/indexer.py", line 44, in index_object
    self._index_bundle(identifier)
  File "/home/travis/build/HumanCellAtlas/data-store/dss/index/indexer.py", line 63, in _index_bundle
    self.backend.index_bundle(bundle)
  File "/home/travis/build/HumanCellAtlas/data-store/dss/index/backend.py", line 95, in index_bundle
    self._delegate(self.index_bundle, args, kwargs)
  File "/home/travis/build/HumanCellAtlas/data-store/dss/index/backend.py", line 121, in _delegate
    raise RuntimeError(f"One or more backends failed: {problems}")
RuntimeError: One or more backends failed: ["Backend ElasticsearchIndexBackend(dryrun=False, notify=True) raised an exception: ConnectionTimeout caused by - ReadTimeoutError(HTTPConnectionPool(host='localhost', port=44051): Read timed out. (read timeout=10))"]
----------------------------------------------------------------------
```

Bento007:
```======================================================================
FAIL: test_reindexing_with_changed_shape (__main__.TestGCPIndexer)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_indexer.py", line 286, in test_reindexing_with_changed_shape
    bundle_key = self.load_test_data_bundle_for_path("fixtures/indexing/bundles/v3/smartseq2/paired_ends")
[...]

PR #1379, opened 2018-06-29
Add Retry to GCP client handler

Bento007:
This is a partial fix for #1357

## Test plan
Ran `put` bundle 100 time without fix ands saw error 3 times **Proof:** https://travis-ci.com/HumanCellAtlas/data-store/builds/77756090
Ran `put` bundle with fix 100 times and saw no errors **Proof:** https://travis-ci.com/HumanCellAtlas/data-store/builds/77763401

## Release notes
Adding retry logic to native gcp handle.


xbrianh:
`gcloud` already has a retry mechanism: it will keep retrying until `MAX_CUMULATIVE_RETRY` seconds have elapsed, and then give up. Occasionally, I think, this may result in 0 retries if the first connection took a long time to fail. It may also retry many times if the connections fail quickly.

The http adapters look conceptually cleaner to me, but the gcloud retry mechanism is still operative. Can there be weird interplay effects? We may want to set `MAX_CUMULATIVE_RETRY` to a large value to ensure that the http adapter retry configurations take effect first.

dss/config.py
```diff
@@ -136,11 +136,8 @@ def _get_native_aws_handle() -> typing.Any:
 
     @staticmethod
     def _get_native_gcp_handle() -> typing.Any:
-        if Config.BLOBSTORE_GS_MAX_CUMULATIVE_RETRY is not None:
-            google.resumable_media.common.MAX_CUMULATIVE_RETRY = Config.BLOBSTORE_GS_MAX_CUMULATIVE_RETRY
```
xbrianh: Consider setting `MAX_CUMULATIVE_RETRY` to a large value. This may cause the http adapter retry  configuration to take effect first.

PR #1382, opened 2018-07-02
[EASY] moving badges to the top of readme

Bento007:


Bento007:
It's really annoying scrolling to the bottom to find it.

kislyuk:
@Bento007 can you please fix this by moving it underneath the top header? The header should be the opening line of the readme.

PR #1388, opened 2018-07-03
CI-CD Deployment scripting for GitLab

xbrianh:


.gitlab-ci.yml
```diff
@@ -0,0 +1,48 @@
...
+  DSS_TEST_ES_PATH: ./elasticsearch-${ES_VERSION}/bin/elasticsearch
+  DSS_ES_TIMEOUT: 30
+  DSS_UNITTEST_OPTS: "-v"
+  TERRAFORM_APPLY_ARGS: "-auto-approve"
+
+stages:
+  - deploy_dev
+
+before_script:
+  - apt-get update -q -y
```
hannes-ucsc: Suggested: I'd prefer the long-forms of these switches. For readability.

scripts/should_deploy.py
```diff
@@ -0,0 +1,247 @@
...
+    for link in linktext.split(","):
+        link_dest = link.split(";")[1]
+        if "next" in link_dest:
+            return link.split("<")[1].split(">")[0] 
+
+    return None
+
+
+def max_items(max_items=50):
+    def decorate(func):
```
hannes-ucsc: Suggested: I always use https://docs.python.org/3.6/library/functools.html#functools.wraps here.

scripts/should_deploy.py
```diff
@@ -0,0 +1,247 @@
...
+        for c in self.commits(branch):
+            return c['id']
+
+
+def should_deploy_dev():
+    owner = "HumanCellAtlas"
+    repo = "data-store"
+    branch = "master"
+
+    GH = GitHub(owner, repo)
```
hannes-ucsc: Suggest: Not a constant.

scripts/should_deploy.py
```diff
@@ -0,0 +1,247 @@
...
+
+    integration_build_state = None
+    unit_tests_build_state = None
+    for b in Travis(owner, repo).builds_for_commit(branch, github_latest):
+        if b['commit']['message'].startswith("Integration"):
+            integration_build_state = b['state']
+        else:
+            unit_tests_build_state = b['state']
+
+        if integration_build_state is None and unit_tests_build_state is None:
```
hannes-ucsc: This looks odd. Don't you want to break out if both of them are *not* None?

.gitlab-ci.yml
xbrianh: Done.

scripts/should_deploy.py
xbrianh: But there aren't any constants in Python.. changed

scripts/should_deploy.py
xbrianh: Insufficient coffee

PR #1390, opened 2018-07-03
Version mandatory for PUT Files and Bundles

Bento007:
### Test plan
- Updated test cases to always include version
- Added test cases to test when no version is present

- the HCA-CLI will need to be updated to handles these changes.

### Release notes
- updated the swagger paths for bundles/put to require version
- updated the swagger paths for files/put to require version

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -257,7 +257,7 @@ class TestBundleApi(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
             requests.codes.conflict,
         )
 
-        print('should *NOT* be abke to this without bundle version.')
+        print('should *NOT* be able to do this without bundle version.')
         bundle_version = datetime_to_version_format(datetime.datetime.utcnow())
         self.put_bundle(
             replica,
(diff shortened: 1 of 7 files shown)


tests/test_bundle.py
```diff
@@ -258,20 +257,31 @@ def _test_bundle_put(self, replica: Replica, fixtures_bucket: str):
             requests.codes.conflict,
         )
 
-        # should *NOT* be able to upload a bundle with a missing file, but we should get requests.codes.conflict.
+        print('should *NOT* be abke to this without bundle version.')
```
xbrianh: Spelling.

This test used to be addressing `requests.codes.conflict`. Has that changed?

Bento007:
Smoke test does pass, but I think retries in the CLI can still cause duplicate bundles.

Bento007:
Here is the update for the CLI https://github.com/HumanCellAtlas/dcp-cli/pull/135

Bento007:
I broke it into subtests since many of the tests are independent of each other.

PR #1393, opened 2018-07-04
Refactored Collections Testcases

Bento007:
- Added additional testcases
- Separated subtests into own tests.

### DIFF
--- a/tests/test_collections.py
+++ b/tests/test_collections.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
+import typing
 
+import itertools
 import os, sys, unittest, io, json
 from uuid import uuid4
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


tests/test_collections.py
```diff
@@ -17,11 +19,22 @@
...
+    app.start()
+
+
+def tearDownModule():
+    global app
+    app.shutdown()
+
+
+class CollectionBase(DSSAssertMixin, DSSUploadMixin):
+    app = app
```
hannes-ucsc: I'd rather eliminate this alias. Class and module attributes have the same lifecycle so the only thing this does for us is make the reference more verbose, e.g. `self.app` instead of just `app`.

tests/test_collections.py
```diff
@@ -17,11 +19,22 @@
...
+    global app
+    app.start()
+
+
+def tearDownModule():
+    global app
+    app.shutdown()
+
+
+class CollectionBase(DSSAssertMixin, DSSUploadMixin):
```
hannes-ucsc: What's the reason for extracting this base class?

tests/test_collections.py
```diff
@@ -40,61 +53,111 @@ def upload_file(self, contents):
...
+        res.raise_for_status()
+        return res.json()["uuid"], res.json()["version"]
+
+    def _delete_collection(self, uuid: str, replica: str='aws'):
+        self.app.delete("/v1/collections/{}".format(uuid),
+                        headers=get_auth_header(authorized=True),
+                        params=dict(replica=replica))
+
+
+class Collections(unittest.TestCase, CollectionBase):
```
hannes-ucsc: What's the reason for breaking the naming convention for test cases?

tests/test_collections.py
Bento007: I was going to have to have multiple `unittest.TestCases` that subclasses this so I separated out all of the common components. I'll recombine everything.

tests/test_collections.py
Bento007: I was going to have to have multiple unittest.TestCases that subclasses this so I separated out all of the common components. I'll recombine everything.

tests/test_collections.py
Bento007: It's shorter, but I'll change it back to our general convention.

tests/test_collections.py
Bento007: good point

tests/test_collections.py
```diff
@@ -157,5 +227,57 @@ def test_collections(self):
                                params=dict(replica="aws"))
             self.assertEqual(res.status_code, requests.codes.not_found)
 
+    def upload_file(self, contents):
```
hannes-ucsc: Consider maintaining authorship and keeping the diff small by moving this back to the original place.

PR #1397, opened 2018-07-05
Simplify scripts/should_deploy.py

xbrianh:
* Only request first page of results from APIs
* GitLab API URL referenced from env var


xbrianh:
Moving this to allspark.

Issue #1323, opened 2018-06-15
Document: Users must implement retries when hitting the API directly to avoid gateway timeouts.


Bento007:
Relates to #1300

As a power user I would like to use the DSS api to perform bursts of requests. Burst loads on the DSS will cause timeout. Implementing retries in the API will correct his issues.

## Definition of Donw
Autogenerate docs from the swagger to inform users of the API to implement retries if they will be using the API directly.

PR #1419, opened 2018-07-11
Add DSS HTTP client best practices docs

### LINKED ISSUE #1263: Document best practice when querying DSS for bundles and files (GET /v1/{bundles,files})
## Need (user story)
As a user of the DSS API
I need documentation on the `GET /v1/files` and `GET /v1/bundles` endpoints
to know that I need to handle `HTTP 301` responses in addition to `HTTP 200` and standard network failures.

## Approach
To solve this issue we can add documentation to the swagger doc and [data consumer vignettes](https://github.com/HumanCellAtlas/data-consumer-vignettes) that displays the proper retry logic for dealing with `301s` from the API.

kislyuk:
Fixes #1263, #1323

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1,7 +1,47 @@
 swagger: '2.0'
 info:
   title: HCA DCP DSS API
-  description: Human Cell Atlas Data Coordination Platform Data Storage System API
+  description: |
+    Human Cell Atlas Data Coordination Platform Data Storage System API
+
+    **HTTP semantics**
+
+    The DSS API requires clients to follow certain HTTP protocol semantics that may require extra configuration in your
+    HTTP client. The reference CLI and SDK (https://hca.readthedocs.io/) is pre-configured to do this. If writing your own
+    client, please note the following:
+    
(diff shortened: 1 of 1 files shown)


dss-api.yml
```diff
@@ -1,7 +1,47 @@
...
+
+    The DSS API requires clients to follow certain HTTP protocol semantics that may require extra configuration in your
+    HTTP client. The reference CLI and SDK (https://hca.readthedocs.io/) is pre-configured to do this. If writing your own
+    client, please note the following:
+    
+    **301 redirects**: Some DSS API routes may return one or more HTTP 301 redirects, including potentially redirects to
+    themselves (combined with the **Retry-After** delay described below). The client must follow these redirects to obtain
+    the resource requested.
+
+    **Retry-After header**: Some DSS API routes may use the **Retry-After** header in combination with HTTP 301 or 500
```
ttung: I think we only return Retry-After with HTTP 301.

kislyuk: I would like to reserve the right to return Retry-After with 5xx in the future :)

PR #1420, opened 2018-07-11
[Easy] Include missing infra env var

xbrianh:
env vars to TF need to be included in `infra/build_deploy_config.py`


xbrianh:
Will do!

PR #1425, opened 2018-07-16
Incorporate terraform workflow into daemon deployment

xbrianh:
This runs TF workflow if terraform files are present in `daemons/{daemon_name}`

### DIFF
--- a/daemons/Makefile
+++ b/daemons/Makefile
@@ -29,7 +29,7 @@ deploy-parallel: $(PARALLEL_AWS_DAEMONS) $(PARALLEL_GCP_DAEMONS)
 
 $(SERIAL_AWS_DAEMONS) $(PARALLEL_AWS_DAEMONS) scheduled-ci-build:
 	@for f in $@/*.tf; do \
-		echo $$f; \
+		echo "Terraforming $@"; \
 		if [ -e $$f ]; then \
 			rm -rf $@/.terraform/*.tfstate;\
 			./build_tf_deploy_config.py $@;\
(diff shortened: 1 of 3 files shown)


daemons/Makefile
```diff
@@ -28,6 +28,16 @@ deploy-serial: $(SERIAL_AWS_DAEMONS) $(SERIAL_GCP_DAEMONS)
 deploy-parallel: $(PARALLEL_AWS_DAEMONS) $(PARALLEL_GCP_DAEMONS)
 
 $(SERIAL_AWS_DAEMONS) $(PARALLEL_AWS_DAEMONS) scheduled-ci-build:
+	@for f in $@/*.tf; do \
+		echo $$f; \
```
kislyuk: Probably want something fancier here like `echo "Terraforming $$f"`

daemons/Makefile
```diff
@@ -28,6 +28,16 @@ deploy-serial: $(SERIAL_AWS_DAEMONS) $(SERIAL_GCP_DAEMONS)
...
+		if [ -e $$f ]; then \
+			rm -rf $@/.terraform/*.tfstate;\
+			./build_tf_deploy_config.py $@;\
+			(cd $@ ; terraform init);\
+			(cd $@ ; TF_VAR_daemon=$@ terraform apply -auto-approve);\
+		fi;\
+		break;\
```
kislyuk: This means only the first .tf file will be applied. Did you mean it that way, or is this a leftover debug breakpoint?

daemons/build_tf_deploy_config.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
```
kislyuk: Please add a docstring here.

daemons/Makefile
xbrianh: Fixed

daemons/Makefile
xbrianh: This was intentional: once the first .tf file is detected, we know we need to perform TF workflow.

daemons/build_tf_deploy_config.py
xbrianh: Done

Issue #1386, opened 2018-07-02
Notification failed during smoketest


Bento007:
See https://github.com/HumanCellAtlas/data-store/issues/1357#issuecomment-401939228

https://travis-ci.com/HumanCellAtlas/data-store/builds/77884346

```
======================================================================
ERROR: test_smoketest (__main__.Smoketest) [aws]
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_smoketest.py", line 224, in test_smoketest
    self.smoketest(**param)
  File "tests/test_smoketest.py", line 215, in smoketest
    obj = s3.get_object(Bucket=self.notification_bucket, Key=notification_key)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/botocore/client.py", line 612, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.
----------------------------------------------------------------------
```

## Definition of Done
Find a cause for the issue and create tickets needed to fix.

Bento007:
@hannes-ucsc If you have any clues, I'd appreciate your help

hannes-ucsc:
1) Extracted all relevant uuids (subscription and notification key) and plugged them into ELK:

`(699111fa-ff9b-4c99-8840-55a169ac89c0 OR e8e369d4-8f6b-494b-a8d9-f42f47354203 OR 15063188-23d5-47e9-befe-20d2a76b9e6e OR 86a8cc4c-077c-402f-9b4f-3dd7fd5c1398 OR 26729545-ebcf-4ed2-9810-ba620caf736e OR d6ee698b-50af-4934-8a4e-ec71d2f40b8c OR 1fa6ed93-2788-479b-88c4-6e3b11e4d53c OR d8e58092-0a39-4779-be90-8dc5b08b1f37) AND Successfully`

[...]

PR #1426, opened 2018-07-16
Admin repair will never send notifications

Bento007:
Changing admin reindex to never send notification during a repair operations.

connects to #1386


hannes-ucsc:
It would be helpful to link this to the ticket that led to this PR as "connected to".

PR #1427, opened 2018-07-17
 Notification only sent on bundle modification

Bento007:
### Test plan
- modified indexer test cases to handle notification on modification.
- added test case to indexer when duplicate indexer events are sent.

- None

### Release notes
-  Notification only sent on bundle modification
- connects to #1386

### DIFF
--- a/tests/test_indexer.py
+++ b/tests/test_indexer.py
@@ -412,6 +412,34 @@ class TestIndexerBase(ElasticsearchTestCase, DSSAssertMixin, DSSStorageMixin, DS
                                         "The hostname in URL 'https://127.0.0.1' resolves to a private IP"):
                 _notify("https://127.0.0.1")
 
+    @testmode.always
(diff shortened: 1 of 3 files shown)


tests/test_indexer.py
```diff
@@ -461,6 +497,13 @@ def test_case(verify_payloads, **endpoint):
 
     @testmode.standalone
     def test_subscription_with_attachments(self):
+        remaining_time = SpecificRemainingTime(300)
+        backend = CompositeIndexBackend(executor=self.executor,
+                                        backends=DEFAULT_BACKENDS,
+                                        remaining_time=remaining_time,
+                                        notify_async=testmode.is_integration(),
+                                        notify=True)
```
hannes-ucsc: Can you extract this into a method? That would also make it more obvious what the difference is in the backend confiugration between the individual tests.

PR #1428, opened 2018-07-17
GCP service account should not be managed from Integration, Staging

xbrianh:
The GCP service account is shared across stages `dev`, `integration`, and `staging`. There is currently Terraform scripting in place to manage this account from the `dev` Terraform state file.

Since infrastructure should not be managed by multiple state files:
1. Tweak the TF scripting to not deploy the service account when `DSS_GCP_SERVICE_ACCOUNT_NAME == ""`
2. Set `DSS_GCP_SERVICE_ACCOUNT_NAME=""` for `integration` and `staging`

### DIFF
--- a/infra/gcp_service_account/main.tf
+++ b/infra/gcp_service_account/main.tf
@@ -1,6 +1,7 @@
 data "google_project" "project" {}
 
 resource "google_service_account" "dss" {
+  count = "${length(var.DSS_GCP_SERVICE_ACCOUNT_NAME) > 0 ? 1 : 0}"
   display_name = "${var.DSS_GCP_SERVICE_ACCOUNT_NAME}"
   account_id = "${var.DSS_GCP_SERVICE_ACCOUNT_NAME}"
 }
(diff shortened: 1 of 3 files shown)


infra/gcp_service_account/main.tf
```diff
@@ -1,6 +1,7 @@
 data "google_project" "project" {}
 
 resource "google_service_account" "dss" {
+  count = "${length(var.DSS_GCP_SERVICE_ACCOUNT_NAME) > 0 ? 1 : 0}"
```
hannes-ucsc: In Orange, we are generating all `.tf.json` files from pure Python templates. I found that pattern to be very useful because I can use comprehensions and conditionals and, well, Python as a Turing-complete language.

Here is a template with a list comprehension:

https://github.com/DataBiosphere/azul/blob/develop/terraform/api_gateway_domain_name.tf.json.template.py#L74

xbrianh: Part of the aspirations of TF seems to be to provide a human readable record of deployed infrastructure that is relatively easy to understand. I would argue that this is defeated by auto-generated TF files, extra logic/conditionals, and even TF's weird interpolation syntax. But ymmv.

Maybe the solution is to create separate GCP service accounts for `integration` and `staging`. Then these conditionals will not be necessary, and I'll be less hypocritical.

PR #1430, opened 2018-07-17
Add /version endpoint to DSS API

calvinnhieu:
Connected to https://github.com/HumanCellAtlas/dcp-cli/issues/113 (cross repo)

The new `/version` endpoint on the DSS API will serve the version of the DSS generated by `git describe --tags --apply` at the time of deployment. This change is part of a larger solution to expose service version(s) describing each DCP service, and ultimately, to consolidate these endpoints in the DCP CLI as a single command that lists the versions of all DCP services.

The version value for DSS is set via environment variable and is injected into the Lambda environment. The return value of the endpoint is a named dictionary, `version_info`, and contains relevant <service/component, version> pair(s) in order to support multiple subcomponents/versions (see [Green Box's /version endpoint](https://github.com/HumanCellAtlas/dcp-cli/issues/113) under _Notes_ for an example with subcomponents).

### Test plan
The new `/version` endpoint was tested on a local instance of the DSS API and a unit test was added in `tests/test_api.py` to test configuration. When deployed, this will be tested by hitting the endpoint and verifying that the version matches the output of `$git describe --tags --always` run in the appropriate branch/commit.

### Release notes
- Added /version endpoint to DSS API

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -54,6 +54,11 @@ OVERRIDE_EXECUTION_LIMIT_SECONDS = None
 """
 This is how long we wait for a request, if set.  If the value is None, we try to use the lambda's timeout.
 """
+DSS_VERSION = os.getenv('DSS_VERSION')
+"""
+Tag describing the version of the currently deployed DSS codebase.  Generated during deployment in the form:
+[<latest-release-tag>-<commits-since-tag>-]<SHA>
+"""
 
 
 class DSSChaliceApp(chalice.Chalice):
@@ -204,6 +209,21 @@ def get_chalice_app(flask_app) -> DSSChaliceApp:
         return chalice.Response(status_code=200,
                                 headers={"Content-Type": "text/html"},
                                 body=swagger_ui_html)
+
+    @app.route("/version")
+    @time_limited(app)
+    def version():
+        data = {
+            'version_info': {
+                'version': DSS_VERSION
+            }
+        }
+
+        return chalice.Response(
+            status_code=requests.codes.ok,
+            headers={'Content-Type': "application/json"},
+            body=data
+        )
 
     @app.route("/internal/health")
     @time_limited(app)
--- a/chalice/build_deploy_config.sh
+++ b/chalice/build_deploy_config.sh
@@ -22,8 +22,9 @@ if ! aws es describe-elasticsearch-domain --domain-name $dss_es_domain; then
     echo "Please create AWS elasticsearch domain $dss_es_domain or set DSS_ES_DOMAIN to an existing domain and try again"
     exit 1
 fi
+export DSS_VERSION=$(git describe --tags --always)
 export DSS_ES_ENDPOINT=$(aws es describe-elasticsearch-domain --domain-name "$dss_es_domain" | jq -r .DomainStatus.Endpoint)
-export EXPORT_ENV_VARS_TO_LAMBDA="$EXPORT_ENV_VARS_TO_LAMBDA DSS_ES_ENDPOINT"
+export EXPORT_ENV_VARS_TO_LAMBDA="$EXPORT_ENV_VARS_TO_LAMBDA DSS_ES_ENDPOINT DSS_VERSION"
 
 cat "$config_json" | jq ".stages.$stage.api_gateway_stage=env.stage" | sponge "$config_json"
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -9,6 +9,8 @@ import os
 import sys
 import unittest
 
+import requests
+
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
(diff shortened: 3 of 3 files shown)


chalice/app.py
```diff
@@ -205,6 +205,21 @@ def serve_swagger_ui():
                                 headers={"Content-Type": "text/html"},
                                 body=swagger_ui_html)
 
+    @app.route("/version")
+    @time_limited(app)
+    def version():
+        data = {
+            'version_info': {
+                'dss_version': os.environ['DSS_VERSION']
```
mweiden: I would move the version to a static variable. That would prevent the application from making a system call every time there is a `GET /version`

mweiden: Also, I would check if the version endpoints should follow a similar JSON schema between DCP components. `dss_version` is a key specific to DSS. Perhaps it would be good to have them all similar?

chalice/build_deploy_config.sh
```diff
@@ -22,6 +22,7 @@ if ! aws es describe-elasticsearch-domain --domain-name $dss_es_domain; then
     echo "Please create AWS elasticsearch domain $dss_es_domain or set DSS_ES_DOMAIN to an existing domain and try again"
     exit 1
 fi
+export DSS_VERSION=$(git describe --tags --always)
```
mweiden: I would either move this variable into `environment`, since environment already references it, or move the variable export (`EXPORT_ENV_VARS_TO_LAMBDA`) to the on line 27.

Between these two options. I kinda like the idea of putting it in `environment`.

chalice/app.py
calvinnhieu: After a conversation with @sampierson, we agreed that some DCP components aren't/can't be described by a single component version. For reference, I followed the format of [Green box's `/version` endpoint](http://pipelines.dev.data.humancellatlas.org/version), using specific subcomponent names. However, I see your point and would propose that we:
- maintain `version_info` key
- use `version` as the component's version key
- use `<subcomponent>_version` as subcomponents' version key
- standardize this format across all `/version` endpoints

chalice/build_deploy_config.sh
calvinnhieu: The reason for defining `DSS_VERSION` in this script is to ensure that the version/release tag generated by `git describe --tags --always` is accurate (only and always generated preceding a deploy). If the same can be guaranteed for `source environment`, then I see no problem with moving it into `environment`. @kislyuk thoughts?

What's the second option re: line 27? Not sure I understand

chalice/app.py
mweiden: Ah I see, that is the standard. Go ahead with that, then.

chalice/build_deploy_config.sh
mweiden: Re. DSS_VERSION being in `build_deploy_config.sh`: Ah, I see. You're right.

Re. line 27: This environment variable takes environment variable names to be exported to the lambda environment. The `EXPORT_ENV_VARS_TO_LAMBDA_ARRAY` variable in `environment` serves a similar purpose.

chalice/app.py
ttung: Getting an environment variable doesn't involve a syscall, FYI.

chalice/app.py
```diff
@@ -205,6 +209,21 @@ def serve_swagger_ui():
                                 headers={"Content-Type": "text/html"},
                                 body=swagger_ui_html)
 
+    @app.route("/version")
```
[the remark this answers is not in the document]
kislyuk: No, I think we should not update the Swagger service definition with this route.

1. This route is diagnostic in nature and not part of the main service API. It's intended primarily for the service operators.
2. Including and documenting this route would involve multiple levels of awkwardness (it would have to be moved out of this location and into a connexion routed location, relocated under the /v1/ prefix, and it would show up in the CLI and bindings, mixing it in with the main API, which I think is undesirable).

chalice/build_deploy_config.sh
kislyuk: Sounds good to me. I agree the logic for determining the version should be here in this file.

chalice/app.py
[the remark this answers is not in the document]
Bento007: Fair enough.

chalice/app.py
mweiden: @ttung [true](https://stackoverflow.com/questions/46623018/does-a-program-make-a-system-call-to-get-the-value-of-an-environment-variable-in). Good catch.

chalice/app.py
```diff
@@ -54,6 +54,10 @@
 """
 This is how long we wait for a request, if set.  If the value is None, we try to use the lambda's timeout.
 """
+DSS_VERSION = os.environ['DSS_VERSION']
+"""
+Current DSS version
```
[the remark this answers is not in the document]
calvinnhieu: Will do

PR #1432, opened 2018-07-18
[Easy] Separate GCP service accounts by stage

xbrianh:
* Stages should ideally have completely separated environments.
* Churn on GCP service account permissions should not crash other environments.
* Service accounts will be created when this PR is approved (`dev` and `prod` already exists).


xbrianh:
Renamed accounts for integration and staging: `org-humancellatlas-{stage}`

PR #1437, opened 2018-07-20
Promote to integration immediately given CI status

xbrianh:
* Add script to check combined CI status from GitHub
* Tweak `scripts/release.sh` to allow machine execution
* Promote to integration, at the end of the master build pipeline, when DCP-wide integration tests are passing

Note that the `--force` flag has been removed from `scripts/release.sh` in favor of the env var `DSS_FORCE_RELEASE`. By setting this env var in the GitLab UI, a pipeline can be retried with forced promotion if needed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -17,6 +17,7 @@ variables:
   GITHUB_API: "https://api.github.com"
 
 stages:
+  - build_image
   - test
   - integration_test
   - deploy_dev
@@ -31,6 +32,14 @@ before_script:
   - pip install -r requirements-dev.txt
   - source environment
   - if [[ -f "environment.$CI_COMMIT_REF_NAME" ]]; then source "environment.$CI_COMMIT_REF_NAME"; fi
+
+
+build_image:
+  stage: build_image
+  script:
+    - docker login
+    - docker build -f Dockerfile.allspark -t {docker_username}/{tag_key}:{tag_value} .`
+    - docker push {docker_username}/{tag_key}:{tag_value}`
 
 unit_tests:
   stage: test
@@ -75,6 +84,9 @@ deploy_dev:
     - export GOOGLE_APPLICATION_CREDENTIALS=$(pwd -P)/gcp-credentials.json
     - make deploy-infra
     - make deploy
+  environment:
+    name: dev
+    url: https://dss.dev.data.humancellatlas.org
   only:
     - master
 
@@ -91,5 +103,8 @@ release_integration:
     -   exit 1
     - fi
     - yes Yes | scripts/release.sh master integration
+  environment:
+    name: integration
+    url: https://dss.integration.data.humancellatlas.org
   only:
     - master


scripts/release.sh
```diff
@@ -13,64 +13,70 @@ if [[ $# != 2 ]] && [[ $# != 3 ]]; then
     echo 'then this script runs "make deploy" after sourcing "environment.{DEST}",'
     echo "where DEST is the destination branch."
     echo
-    echo "If the --force flag is given, deployment will proceed even if CI checks fail."
+    echo "If ${DSS_FORCE_RELEASE}=='force', deployment will proceed even if CI checks fail."
```
kislyuk: I am worried about the use of env vars to drive something as critical as this. Env vars are too implicit to toggle top level flags. I think it's preferable to call this script with `--force` in the caller's location.

scripts/release.sh
```diff
@@ -13,64 +13,70 @@ if [[ $# != 2 ]] && [[ $# != 3 ]]; then
...
+    # Skip when this script is executed by the GitLab runner
+    echo "Please review and confirm your active AWS account configuration:"
+    aws configure list
+    aws sts get-caller-identity
+    aws iam list-account-aliases
+    echo "Is this correct?"
+    select result in Yes No; do
+        if [[ $result != Yes ]]; then exit 1; else break; fi
+    done
+fi
```
kislyuk: I don't think this is a good pattern. It intertwines the concerns. What if I sudo to impersonate this user for some reason? Consider just calling this script with `yes |`:
```bash
yes Yes | scripts/release.sh
```

scripts/release.sh
```diff
@@ -13,64 +13,70 @@ if [[ $# != 2 ]] && [[ $# != 3 ]]; then
...
+    
+    # TODO: (akislyuk) some CI builds no longer deploy or run a subset of tests. Find the last build that ran a deployment.
+    if [[ "$STATE" != success ]]; then
+        if [[ "force" == ${DSS_FORCE_RELEASE} ]]; then
+            echo "Status checks failed on branch $PROMOTE_FROM_BRANCH. Forcing promotion and deployment anyway."
+        else
+            echo "Status checks failed on branch $PROMOTE_FROM_BRANCH."
+            echo "Run with DSS_FORCE_RELEASE='force' to promote $PROMOTE_FROM_BRANCH to $PROMOTE_DEST_BRANCH and deploy anyway."
+            exit 1
+        fi
```
kislyuk: Why are status checks skipped here on the CD runner?

scripts/status.sh
```diff
@@ -0,0 +1,21 @@
...
+GITHUB_API="https://api.github.com"
+
+OWNER=$(python -c "import sys, urllib.parse; print(urllib.parse.quote_plus(sys.argv[1]))" "${1}")
+REPO=$(python -c "import sys, urllib.parse; print(urllib.parse.quote_plus(sys.argv[1]))" "${2}")
+BRANCH=$(python -c "import sys, urllib.parse; print(urllib.parse.quote_plus(sys.argv[1]))" "${3}")
+
+STATUS=$(http GET ${GITHUB_API}/repos/${OWNER}/${REPO}/commits/${BRANCH}/status Accept:application/vnd.github.full+json)
+STATE=$(echo "$STATUS" | jq -r .state)
+
+echo ${STATE}
```
kislyuk: Line 21 is unnecessary, you can just unwrap the subshell on line 19 if you don't need STATE for anything else.

.gitlab-ci.yml
```diff
@@ -73,3 +74,18 @@ deploy_dev:
...
+  stage: release_integration
+  script:
+    - while [[ "pending" == $(scripts/status.sh HumanCellAtlas dcp integration) ]]; do
+    -   echo "waiting for DCP Integration test to complete";
+    -   sleep 60;
+    - done
+    - if [[ "success" != $(scripts/status.sh HumanCellAtlas dcp integration) ]]; then
+    -   echo "DCP Integration test did not succeed";
+    -   exit 1
+    - fi
```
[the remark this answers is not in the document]
xbrianh: The  dcp-wide integration test can be N commits ahead of data-store master. Which sha do we poll for?

Capped loop to 20 minutes.

Dockerfile.allspark
```diff
@@ -42,21 +42,21 @@ RUN python3 -m pip install virtualenv==16.0.0
 RUN ln -s /usr/bin/python3.6 /usr/bin/python
 RUN ln -s /usr/bin/pip3 /usr/bin/pip
 
-RUN useradd -d /home/hca -ms /bin/bash -g root -G sudo hca
-RUN mkdir /HumanCellAtlas && chown hca /HumanCellAtlas
-USER hca
-WORKDIR /home/hca
+RUN useradd -d /home/hca_cicd -ms /bin/bash -g root -G sudo hca_cicd
+RUN mkdir /HumanCellAtlas && chown hca_cicd /HumanCellAtlas
+USER hca_cicd_cicd
```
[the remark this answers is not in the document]
xbrianh: You can never have too many IMO.

fixed.

scripts/release.sh
xbrianh: That's reasonable, reverted to `--force`.

However, another mechanism will be needed if we wish to force promotion through the build box,

scripts/release.sh
xbrianh: Fixed.

scripts/release.sh
xbrianh: In order to reach this stage in the GitLab pipeline, testes have necessarily passed on the build box. Do we also wish to require  that Travis tests have passed?

scripts/status.sh
xbrianh: Fixed.

xbrianh:
Perhaps the best place to document DSS ci/cd and release efforts is in a DSS-specific section for https://docs.google.com/document/d/1HbSWuIABaavaHpCwaO0peA_uwTW5wBCdRnYoz7ZVQuQ/edit#heading=h.yw6cvy5vlmbu

We can look into automating Dockerfile construction. In the meantime, added a pointer in `gitlab-ci.yml` to the usage instructions in `Dockerfile.allspark`.

.gitlab-ci.yml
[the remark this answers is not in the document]
kislyuk: Can we poll for the sha of the commit from which this code is running? `git rev-parse HEAD`?

scripts/release.sh
kislyuk: I guess you're right, we don't want to depend on travis here.

kislyuk:
Makes sense to document this in the SOP doc, but would be nice to move the SOP doc to the ops wiki now that discussion in google docs has completed.

.gitlab-ci.yml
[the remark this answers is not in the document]
kislyuk: Never mind... can't see an easy/elegant way to map the integration test repo sha to DSS sha for deployment OK.

PR #1441, opened 2018-07-23
adding uuid to touch.txt to prevent GCP rate limit errors

### LINKED ISSUE #1440: Checkout touch_test_file causes rate limit errors.
## Software version (git hash, date, branch)
https://github.com/HumanCellAtlas/data-store/commit/d4bff8babed9b74189441a48e166f0b5e5404464

## Expected behavior
upload a unique file per checkout write test.

## Actual behavior
Uploading the same file per checkout write test. This causes GCP rate limit errors.
```
dss.storage.checkout.error.DestinationBucketNotWritableError: 410 PUT https://www.googleapis.com/upload/storage/v1/b/org-humancellatlas-dss-checkout-test/o?uploadType=resumable&upload_id=AEnB2UplJovWOwykx1lbnA0ObzqjCGSvXEQwdo0iQX3kV_P10PFULpdAY_9EjZXFNUzhiW8GvgiJ8aBvES6lMJr2VprIiOH17Q: The total number of changes to the object org-humancellatlas-dss-checkout-test/touch.txt exceeds the rate limit. Please reduce the rate of create, update, and delete requests.
```

https://travis-ci.com/HumanCellAtlas/data-store/jobs/135635904#L1006

## Steps to reproduce the behavior
This is an intermittent error and an accurate reproduction has not been create.

Bento007:
Fixes #1440 

### Test plan
Run test over several days to see if any GCP request rate limits errors are triggered.

None

### Release notes
A touch_*.txt file will be created for every checkout process. These files will eventually be deleted based on the buckets data lifecycle rules. We should also suggest that creators of checkout buckets set a shorter data lifecycle rule for files starting with `touch_`.

### DIFF
--- a/dss/storage/checkout/__init__.py
+++ b/dss/storage/checkout/__init__.py
@@ -1,5 +1,6 @@
 import io
 from enum import Enum, auto
+import uuid
 
 from cloud_blobstore import BlobNotFoundError, BlobStoreUnknownError
 from dss.config import Config, Replica
@@ -57,7 +58,7 @@ def touch_test_file(replica: Replica, dst_bucket: str) -> bool:
     :param replica: the replica to execute the checkout in.
     :return: True if able to write, if not raise DestinationBucketNotWritableError.
     """
-    test_object = "touch.txt"
+    test_object = f"touch_{uuid.uuid1()}.txt"
     handle = Config.get_blobstore_handle(replica)
 
     try:
--- a/dss/storage/checkout/__init__.py
+++ b/dss/storage/checkout/__init__.py
@@ -1,6 +1,7 @@
 import io
+import random
 from enum import Enum, auto
-import uuid
+from string import hexdigits
 
 from cloud_blobstore import BlobNotFoundError, BlobStoreUnknownError
 from dss.config import Config, Replica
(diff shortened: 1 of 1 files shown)


dss/storage/checkout/__init__.py
```diff
@@ -57,7 +59,7 @@ def touch_test_file(replica: Replica, dst_bucket: str) -> bool:
     :param replica: the replica to execute the checkout in.
     :return: True if able to write, if not raise DestinationBucketNotWritableError.
     """
-    test_object = "touch.txt"
+    test_object = f"touch_{''.join(random.choices(hexdigits, k=4))}.txt"
```
ttung: pls change to `f"touch/{''.join(random.choices(hexdigits, k=4))}.txt"` as that will keep the bucket listing less messy.

dss/storage/checkout/__init__.py
```diff
@@ -57,7 +59,7 @@ def touch_test_file(replica: Replica, dst_bucket: str) -> bool:
     :param replica: the replica to execute the checkout in.
     :return: True if able to write, if not raise DestinationBucketNotWritableError.
     """
-    test_object = "touch.txt"
+    test_object = f"touch/{''.join(random.choices(hexdigits, k=4))}.txt"
```
xbrianh: This will eventually create over two hundred thousand files:
```
In [11]: len(string.hexdigits)**4
Out[11]: 234256
```

Bento007: the files will be cleaned up by the buckets data life cycle setting before that size is reached. If that size is reached it will only consume a few MB of storage. Also you can set a specific data life cycle with files with the touch* as the prefix to clean them up soon after they are created. Either way I don't think this is an issue.

PR #1442, opened 2018-07-24
Update infra to reflect current state

xbrianh:
* Accommodate, and bring up-to-date, infrastructure for `integration`.
* With this PR, `make deploy-infra` passes without changes on `dev` and `integration`.


ttung:
Consider building a cron job that runs terraform plan and checks that the state of the infrastructure is consistent with what's checked in (and erroring if not).

xbrianh:
@ttung Currently incorporating that into the build scripting on the new Build Box. Deployment will fail if infra is not up-to-date.

PR #1454, opened 2018-07-25
Reject bundles containing duplicated filenames

### LINKED ISSUE #1268: Prohibit duplicate file names in a bundle
Ingest is moving to splitting metadata files into one file per entity. For example, the multiple entities in Ingest is planning to split `biomaterials.json` (donor, cell suspension, ...l) into one file per entity. An example bundle can be found here:

https://github.com/HumanCellAtlas/data-bundle-examples/tree/bundle_proposal/version5/10x_v2/multifile_bundle/ischaemic-sensitivity/type_as_filename

In this new structure the file name denotes the entity type. To avoid conflicts in case when there is more than one entity of any given type, the file name will be suffixed with a numeric index. 

The bundle manifest as currently defined by DSS is an array of dictionaries, each dictionary representing one file whose name is stored `name` in that dictionary. While this structure technically allows for more than one file of the same name, a bundle is conceptually an archived file system directory containing files and allowing multiple files of the same name contradicts that notion.

We should explicitly enforce the uniqueness of file names in a bundle manifest thereby making explicit the need for suffixing metadata files of the same entity type as described above.

There are two ways to do this: 

1) Leave the manifest structure as is and simply add validation to `PUT /bundles` that disallows files of the same name.

2) Convert the manifest from array to dictionary and use the file name as the key.

xbrianh:
resolves #1268

### DIFF
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -133,7 +133,8 @@ def put(uuid: str, replica: str, json_request_body: dict, version: str):
             raise DSSException(
                 requests.codes.bad_request,
                 "duplicate_filename",
-                "A bundle cannot contain duplicate filenames"
+                f"Duplicate file name detected: {name}. This test fails on the first occurance. Please check bundle "
+                "layout to ensure no duplicated file names are present."
             )
         filenames.add(name)
         files.append({'user_supplied_metadata': file})
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -125,7 +125,18 @@ def put(uuid: str, replica: str, json_request_body: dict, version: str):
     bundle_manifest_key = BundleFQID(uuid=uuid, version=version).to_key()
 
     # decode the list of files.
-    files = [{'user_supplied_metadata': file} for file in json_request_body['files']]
+    files = list()
+    filenames: typing.Set[str] = set()
+    for file in json_request_body['files']:
+        name = file['name']
+        if name in filenames:
+            raise DSSException(
+                requests.codes.bad_request,
+                "duplicate_filename",
+                "A bundle cannot contain duplicate filenames"
+            )
+        filenames.add(name)
+        files.append({'user_supplied_metadata': file})
 
     time_left = nestedcontext.inject("time_left")
 
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -265,6 +265,26 @@ class TestBundleApi(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
                 expected_code=requests.codes.bad_request
             )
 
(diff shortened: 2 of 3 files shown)


tests/test_bundle.py
```diff
@@ -265,6 +265,26 @@ def _test_bundle_put(self, replica: Replica, fixtures_bucket: str):
                 expected_code=requests.codes.bad_request
             )
 
+        with self.subTest(f'{replica}: should *NOT* be able to put a bundle containing a duplicate filename'):
```
Bento007: I prefer we use the "`result` when `condition`" format when describing test when possible. I've been changing them whenever I have to modify any tests. I suggest you change
> "should *NOT* be able to put a bundle containing a duplicate filename" 

to

> "put fails when the bundle contains a duplicated file name"

xbrianh: changed.

dss/api/bundles/__init__.py
```diff
@@ -125,7 +125,18 @@ def put(uuid: str, replica: str, json_request_body: dict, version: str):
...
-    files = [{'user_supplied_metadata': file} for file in json_request_body['files']]
+    files = list()
+    filenames: typing.Set[str] = set()
+    for file in json_request_body['files']:
+        name = file['name']
+        if name in filenames:
+            raise DSSException(
+                requests.codes.bad_request,
+                "duplicate_filename",
+                "A bundle cannot contain duplicate filenames"
```
hannes-ucsc: Consider including the offending file name in the error message.

Also consider determining all duplicates before raising the exception and including all duplicates in the error message.

tests/test_bundle.py
```diff
@@ -265,6 +265,26 @@ def _test_bundle_put(self, replica: Replica, fixtures_bucket: str):
...
+                    file_uuid2,
+                    bundle_uuid=bundle_uuid2,
+                )
+                file_version2 = resp_obj2.json['version']
+                self.put_bundle(
+                    replica,
+                    bundle_uuid2,
+                    [(file_uuid, file_version, "LICENSE"), (file_uuid2, file_version2, "LICENSE")],
+                    bundle_version,
+                    expected_code=requests.codes.bad_request
```
hannes-ucsc: Is there a way to assert the `duplicate_filename` error code?

xbrianh: Done

dss/api/bundles/__init__.py
xbrianh: Filename included in error message.

In lieue of extra duplication-detecting logic, a warning has been added indicating further  duplicate filename may be present in the bundle.

Maybe it wouldn't be a lot of extra logic, but, in general, it seems reasonable for our API backend to fail as quickly as possible on validation errors.

PR #1456, opened 2018-07-27
Fixed multipart upload

ttung:
1. Bumped to cloud-blobstore 2.1.5 to pick up https://github.com/chanzuckerberg/cloud-blobstore/commit/749d4ef7d9c7cff1d9de662a5fe5919d446a2ef5.
2. Changed the multipart threshold to 64MB+1 for the test uploader class.  This ensures that we don't switch to multipart uploads for 64MB objects.
3. Refactored the tests so that synchronous requests (i.e., <= 64MB) tests can run standalone.
4. Clarified expectations for the three classes of requests: aws multipart, gs multipart, and synchronous requests.

Test plan: Ran `DSS_TEST_MODE="standalone integration" make tests/test_file.py`

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -15,10 +15,9 @@ pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noq
 sys.path.insert(0, pkg_root)  # noqa
 
 import dss
-from dss.api.files import RETRY_AFTER_INTERVAL
+from dss.api.files import ASYNC_COPY_THRESHOLD, RETRY_AFTER_INTERVAL
 from dss.config import BucketConfig, override_bucket_config, Replica
 from dss.util import UrlBuilder
-from dss.util.aws import AWS_MIN_CHUNK_SIZE
 from dss.util.version import datetime_to_version_format
 from tests import eventually
 from tests.fixtures.cloud_uploader import GSUploader, S3Uploader, Uploader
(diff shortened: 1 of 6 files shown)


tests/test_file.py
```diff
@@ -111,56 +110,72 @@ def _test_file_put(self, replica: Replica, scheme: str, test_bucket: str, upload
...
+                else:
+                    # We should be able to do this twice (i.e., same payload, different UUIDs).
+                    # First time should be asynchronous since it's new data.  Second time should be
+                    # synchronous since the data is present.
+                    expect_async_results = (True, False)
+            else:
+                # We should be able to do this twice (i.e., same payload, different UUIDs).  Neither
+                # time should be asynchronous.
+                expect_async_results = (False, False)
+
```
Bento007: Make it a subtest either starting here or L168.

tests/test_file.py
```diff
@@ -111,56 +110,72 @@ def _test_file_put(self, replica: Replica, scheme: str, test_bucket: str, upload
...
+        tempdir = tempfile.gettempdir()
+        uploader = uploader_class(tempdir, bucket)
+        with tempfile.NamedTemporaryFile(delete=True) as fh:
+            fh.write(data)
+            fh.flush()
+
+            uploader.checksum_and_upload_file(fh.name, key, "text/plain")
+
+    @testmode.standalone
+    def test_file_put_large_sync(self):
```
Bento007: Add a test for ASYNC_COPY_THRESHOLD - 1. Those are the most common areas for areas occur.

Bento007: These tests are run in parallel, so As long as it's not slower then test_indexer.

ttung: Ah ha ha ha ha, ohhh, test_indexer...

PR #1457, opened 2018-07-27
Add manual GitLab action to release to staging

xbrianh:


.gitlab-ci.yml
```diff
@@ -104,3 +105,28 @@ release_integration:
...
+
+release_staging:
+  stage: release_staging
+  script:
+    - git clean -df
+    - git reset --hard
+    - for i in $(seq 1 10); do
+    -   if [[ "pending" != $(scripts/status.sh HumanCellAtlas dcp staging) ]]; then break; fi
+    -   echo "waiting for DCP Integration test to complete";
+    -   sleep 120;  # This loop will check status for 20 minutes and then quit
```
kislyuk: Try testing more often (every 30s?). No harm in doing so and less delay if someone is waiting for the deployment to go through.

.gitlab-ci.yml
```diff
@@ -104,3 +105,28 @@ release_integration:
...
+    - done
+    - if [[ "success" != $(scripts/status.sh HumanCellAtlas dcp staging) ]]; then
+    -   echo "DCP Integration test did not succeed";
+    -   exit 1
+    - fi
+    - aws secretsmanager get-secret-value --secret-id ${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/gcp-credentials.json | jq -r .SecretString > gcp-credentials.json
```
kislyuk: This line looks like it might be repeated in a number of places. Maybe it's worth packaging into its own script under `data-store/scripts`? Like `fetch-secrets.sh gcp-credentials ...` where the args are names of secrets to fetch?

xbrianh: Seems reasonable to me: created `scripts/fetch_secret.sh`.

PR #1458, opened 2018-07-27
Remove the submodule dependency on checksumming_io

ttung:
Pull it in through dcplib.


requirements-dev.txt
```diff
@@ -71,6 +72,7 @@ paramiko==2.4.1
 pbr==3.1.1
 protobuf==3.5.0.post1
 psutil==5.4.1
+puremagic==1.4
```
xbrianh: What's this for?

ttung: I dunno. Probably sucked in by dcplib.

PR #1465, opened 2018-08-01
Switch to the constants defined in dcplib

ttung:
https://github.com/HumanCellAtlas/dcplib/commit/d446e4d79a93f1e4ccc17d9d002c13717a495c04 replicates the constants defined in dss/util/aws/__init__.py, and should be used throughout the HCA DCP.


ttung:
Yes, very much so.

PR #1469, opened 2018-08-01
Support new bundle layout

xbrianh:
Add DSS support of the new bundle layout, per https://docs.google.com/document/d/15bSO0sgo5KRWJIr6VUgZf97uL8397vUZ68AL17EC_cA/edit

DSS will no longer guarantee support older metadata schema (v3, v4, v5, etc.)


dss/index/es/validator.py
```diff
@@ -20,7 +20,7 @@
                                         )
 
 
-def remove_json_fields(json_data: dict, path: List[str], fields: List[str]) -> None:
+def remove_json_fields(json_data: dict, path: List[Any], fields: List[str]) -> None:
```
hannes-ucsc: Suggested: `Union[str|int]` ?

dss/index/es/validator.py
```diff
@@ -48,33 +48,36 @@ def request_json(url):
...
     for document in index_data.keys():
-        schema_info = SchemaInfo.from_json(index_data[document])
-        if schema_info is not None:
-            try:
-                schema = request_json(schema_info.url)
-            except Exception as ex:
-                extra_documents.append(document)
-                logger.warning(f"Unable to retrieve JSON schema information from {document} in bundle {bundle_id} "
-                               f"because retrieving {schema_info.url} caused exception: {ex}.")
+        for doc_list_ind, document_content in enumerate(index_data[document]):
```
hannes-ucsc: Not quite sure I understand the intent of this hunk. BTW: if this is change is orthogonal to the general intent of the PR, it would make sense to commit it separately. Don't know if that is the case here, though.

tests/test_indexer.py
```diff
@@ -349,16 +362,16 @@ def test_indexed_file_unparsable(self):
         self.assertRegex(log_monitor.output[0],
                          "WARNING:.*:In bundle .* the file 'unparseable_json.json' is marked for indexing"
                          " yet could not be parsed. This file will not be indexed. Exception:")
-        search_results = self.get_search_results(self.smartseq2_paired_ends_query, 1)
-        self.assertEqual(1, len(search_results))
-        self.verify_index_document_structure_and_content(search_results[0], bundle_key,
-                                                         files=smartseq2_paried_ends_indexed_file_list)
+#        search_results = self.get_search_results(smartseq2_paired_ends_v2_or_v3_query, 1)
```
hannes-ucsc: Requested: Delete or comment back in. If the former, could you quickly explain why these assertion aren't needed? In person is fine.

tests/test_indexer.py
```diff
@@ -1426,7 +1449,7 @@ def create_index_data(blobstore, bucket_name, bundle_key, manifest,
     index = dict(state="new", manifest=manifest, uuid=BundleFQID.from_key(bundle_key).uuid)
     files_info = manifest['files']
     excluded_file = [file.replace('_', '.') for file in excluded_files]
-    index_files = {}
+    index_files: typing.Dict = {}
```
hannes-ucsc: See above.

tests/test_indexer.py
```diff
@@ -1439,7 +1462,15 @@ def create_index_data(blobstore, bucket_name, bundle_key, manifest,
             except Exception:
                 continue
             index_filename = file_info["name"].replace(".", "_")
-            index_files[index_filename] = file_json
+            if index_filename.endswith('_json'):
```
hannes-ucsc: Suggested: This looks like it duplicates functionality from above. Consider extracting a function.

dss/index/es/document.py
```diff
@@ -103,7 +105,18 @@ def from_bundle(cls, bundle: Bundle):
         #    Therefore, substitute dot for underscore in the key filename portion of the index. As due diligence,
         #    additional investigation should be performed.
         #
-        files = {name.replace('.', '_'): content for name, content in bundle.files.items()}
+        files: typing.Dict = dict()
```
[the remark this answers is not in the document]
xbrianh: mypy requested the annotation.

dss/index/es/validator.py
xbrianh: This logic is required for `scrub_index_data` to work with the new bundle format.

tests/test_indexer.py
xbrianh: ?

dss/index/es/document.py
```diff
@@ -103,7 +105,18 @@ def from_bundle(cls, bundle: Bundle):
...
+        for name, content in bundle.files:
+            name = name.replace('.', '_')
+            if name.endswith('_json'):
+                name = name[:-5].rstrip(string.digits + "_") + '_json'
```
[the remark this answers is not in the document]
xbrianh: switched to `rpartition`

dss/index/es/validator.py
xbrianh: Union

tests/test_indexer.py
xbrianh: deleted

tests/test_indexer.py
xbrianh: This is a small part of a function that reproduces a large chunk of application logic.

tests/test_indexer.py
hannes-ucsc: Never mind. It's a bug in mypy, I guess. This shouldn't be necessary.

hannes-ucsc:
I was able to revive `test_reindexing_with_changed_shape` and switch `test_reindexing_with_changed_content` to the new fixture with minimal effort. Can you help me understand why you found it necessary to disable them?

```diff
Index: tests/test_indexer.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- tests/test_indexer.py	(revision ed05ad969a4b0efb938bdbd0ddcf89f4e6c2a033)
+++ tests/test_indexer.py	(date 1533341014000)
@@ -249,12 +249,12 @@
 
     @testmode.standalone
     def test_reindexing_with_changed_content(self):
-        bundle_key = self.load_test_data_bundle_for_path("fixtures/indexing/bundles/v3/smartseq2/paired_ends")
+        bundle_key = self.load_test_data_bundle_for_path("fixtures/indexing/bundles/vx/smartseq2/paired_ends")
         sample_event = self.create_bundle_created_event(bundle_key)
 
         @eventually(timeout=5.0, interval=0.5)
         def _assert_reindexing_results(expect_extra_field, expected_version):
-            query = {**smartseq2_paired_ends_v2_or_v3_query, 'version': True}
+            query = {**smartseq2_paired_ends_vx_query, 'version': True}
             hits = self.get_raw_search_results(query, 1)['hits']['hits']
             self.assertEqual(1, len(hits))
             self.assertEquals(expected_version, hits[0]['_version'])
@@ -287,18 +287,18 @@
             self.process_new_indexable_object(sample_event)
         self.assertTrue(any('is already up-to-date' in e for e in log.output))
 
-    @skipStaleBundleFormatTest
     @testmode.standalone
     def test_reindexing_with_changed_shape(self):
-        bundle_key = self.load_test_data_bundle_for_path("fixtures/indexing/bundles/v3/smartseq2/paired_ends")
+        bundle_key = self.load_test_data_bundle_for_path("fixtures/indexing/bundles/vx/smartseq2/paired_ends")
         sample_event = self.create_bundle_created_event(bundle_key)
         shape_descriptor = 'v99'
 
         @eventually(timeout=5.0, interval=0.5)
         def _assert_reindexing_results(expect_shape_descriptor):
-            hits = self.get_raw_search_results(smartseq2_paired_ends_v2_or_v3_query, 1)['hits']['hits']
+            hits = self.get_raw_search_results(smartseq2_paired_ends_vx_query, 1)['hits']['hits']
             self.assertEqual(1, len(hits))
-            self.assertEqual(expect_shape_descriptor, shape_descriptor in hits[0]['_index'])
+            hashed_shape_descriptor = hashlib.sha1(shape_descriptor.encode()).hexdigest()
+            self.assertEqual(expect_shape_descriptor, hashed_shape_descriptor in hits[0]['_index'])
 
         # Index document into the "wrong" index by patching the shape descriptor
         with unittest.mock.patch.object(BundleDocument, 'get_shape_descriptor', return_value=shape_descriptor):
```

I thought about coverage some more. Code coverage is one, but only one, metric of test strength. Data or state coverage is another. Even if one test covers 99% of the code also covered by another test, that seemingly negligible difference in code coverage could be driven by a significant change in state. The two tests I mention fall into that category. They have almost identical code coverage but the state (variable values, ES index contents) differs in significant ways. The tests are proof that the code handles those differences in state correctly.

Also @Bento007, would you be able to evaluate the list of tests Brian disabled, and see which ones we can dispose of and which ones should remain?

dss/index/es/document.py
```diff
@@ -229,35 +248,40 @@ def get_shape_descriptor(self) -> typing.Optional[str]:
...
+                                   f"indexed as is, without sanitization. This may prevent subsequent, valid files "
+                                   f"from being indexed correctly.")
         if schemas_by_file:
-            same_version = 1 == len(set(schema.version for schema in schemas_by_file.values()))
-            same_schema_and_file_name = all(file_name == schema.type for file_name, schema in schemas_by_file.items())
+            same_version = 1 == len(set(schema.version for _, schema in schemas_by_file))
+            same_schema_and_file_name = all(file_name == schema.type for file_name, schema in schemas_by_file)
             if same_version and same_schema_and_file_name:
-                return 'v' + next(iter(schemas_by_file.values())).version
+                return 'v' + schemas_by_file.pop()[1].version
```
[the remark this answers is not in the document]
Bento007: I find it confusing to follow when you have more than one return path. Also reducing the number of returns in this case will allow us to add the hashing of the `shape_description` into `get_shape_descriptor` without adding duplicate code. Here is what I suggest:
```python
        if schemas_by_file:
            ...
            if same_version and same_schema_and_file_name:
                shape_descriptor = 'v' + schemas_by_file.pop()[1].version
            else:
                ...     
                shape_descriptor = 'v.' + '.'.join(entry(*schema) for schema in schemas)
            return hashlib.sha1(str(shape_descriptor).encode("utf-8")).hexdigest()
        else:
            return None  # No files with schema references were found
```
This will eliminate the need for L216:L217 and keep `shape_descriptor` generation code out of `_prepare_index`

dss/index/es/document.py
```diff
@@ -196,6 +213,8 @@ def _write_to_index(self, index_name: str, version: typing.Optional[int] = None)
 
     def _prepare_index(self, dryrun):
         shape_descriptor = self.get_shape_descriptor()
+        if shape_descriptor is not None:
```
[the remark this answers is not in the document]
xbrianh: `get_shape_descriptor` should return a human readable value to support @hannes-ucsc idea of inserting it into the index document. The hash of the shape descriptor is not descriptive of the shape of the document.

dss/index/es/document.py
```diff
@@ -103,7 +105,22 @@ def from_bundle(cls, bundle: Bundle):
         #    Therefore, substitute dot for underscore in the key filename portion of the index. As due diligence,
         #    additional investigation should be performed.
         #
-        files = {name.replace('.', '_'): content for name, content in bundle.files.items()}
+        files: typing.Dict = dict()
+        for name, content in bundle.files:
+            name = name.replace('.', '_')
```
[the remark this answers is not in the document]
xbrianh: need to check the return value of `rpartition`:
```
"blah".rpartition("_")
>>> "blah".rpartition("_")
('', '', 'blah')
>>> "blah_34".rpartition("_")
('blah', '_', '34')
```

dss/index/es/document.py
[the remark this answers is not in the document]
Bento007: than can we give it another name to avoid confusion. Maybe `index_name`, `shape_hash`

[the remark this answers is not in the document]
xbrianh: yeah :)

`hashed_shape_descriptor`

tests/test_indexer.py
```diff
@@ -736,6 +745,7 @@ def test_alias_and_multiple_schema_version_index_exists(self):
         self.assertIn(doc_index_name, alias)
         self.assertTrue(es_client.indices.exists(index=doc_index_name))
 
+    @skipStaleBundleFormatTest
```
Bento007: This may be tested in multiple places at this point. I think it's safe to skip.

tests/test_indexer.py
```diff
@@ -553,14 +560,15 @@ def test_subscription_notification_unsuccessful(self):
...
+        subscription_id = self.subscribe_for_notification(es_query=smartseq2_paired_ends_vx_query,
                                                           **endpoint.to_dict())
         sample_event = self.create_bundle_created_event(self.bundle_key)
         self.process_new_indexable_object(sample_event)
         prefix, _, bundle_fqid = self.bundle_key.partition("/")
-        self.verify_notification(subscription_id, self.smartseq2_paired_ends_query, bundle_fqid, endpoint)
+        self.verify_notification(subscription_id, smartseq2_paired_ends_vx_query, bundle_fqid, endpoint)
         self.delete_subscription(subscription_id)
 
+    @skipStaleBundleFormatTest
```
Bento007: This should not be skipped

tests/test_indexer.py
```diff
@@ -807,6 +818,7 @@ def test_multiple_schema_version_subscription_indexing_and_notification(self):
         self.delete_subscription(subscription_id)
 
     # TODO: Remove this test once we stop supporting the `core` property (see issue #1015).
+    @skipStaleBundleFormatTest
```
Bento007: I think we can remove this test.

tests/test_indexer.py
```diff
@@ -880,6 +892,7 @@ def test_scrub_index_data_with_core(self):
                                                                       f"Actual index document: "
                                                                       f"{json.dumps(index_data, indent=4)}")
 
+    @skipStaleBundleFormatTest
```
Bento007: This is the only test for the validator. We shouldn't skip this.

dss/index/es/document.py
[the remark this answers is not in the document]
Bento007: don't forget
```
>>> "_blah".rpartition('_')
('', '_', 'blah')

>>> "blah_".rpartition('_')
('blah', '_', '')
```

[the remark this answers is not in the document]
Bento007: Here:
```python
if name.endswith('.json'):
    beg, _, end = name.rpartition('.')
    if beg != end
        name = beg + '_json' 
name = name.replace('.', '_')
```

Bento007:
Is there a test with files with different versions/schemas in the same list?

tests/test_indexer.py
hannes-ucsc: Requested: I may be mistaken, but these twelve lines are identical to the ones in `BundleDocument.from_bundle()`. Why do we need to duplicate them?

tests/test_indexer.py
hannes-ucsc: image

hannes-ucsc: This is after a full page reload.

PR #1461, opened 2018-07-30
Add Index Design and search limitations documentation

mweiden:
Fixes https://github.com/HumanCellAtlas/data-store/issues/1084

# Need
This documentation addresses the following user story
* As a user of the DSS `POST /search` API
* I want to understand the design and limitations of DSS's search index
* so that I can quickly develop efficient queries against DSS's Elasticsearch index.

PR #1470, opened 2018-08-02
Use folded block YAML style for docs in swagger

mweiden:
Follow up from https://github.com/HumanCellAtlas/data-store/pull/1461

# Need

The swagger doc currently uses [literal block style](http://yaml.org/spec/1.2/spec.html#id2795688). This causes text in the swagger doc to not format properly. See example below

before

This makes the documentation harder to read for users.

# Approach

To fix this, this PR switches the swagger docs to [folded block style](http://yaml.org/spec/1.2/spec.html#id2796251) as suggested by @hannes-ucsc. See an example of the result below.

after

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1,7 +1,7 @@
 swagger: '2.0'
 info:
   title: HCA DCP DSS API
-  description: |
+  description: >
     Human Cell Atlas Data Coordination Platform Data Storage System API
 
     # HTTP semantics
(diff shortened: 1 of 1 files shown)


mweiden:
@hannes-ucsc I'm really proud of my gifs. Just sayin'.

ttung:
amazed at the gifs.

mweiden:
@hannes-ucsc Es tut mir leid! Ich werde mir Mühe geben, an meiner Arbeit zu bleiben. 😆

PR #1472, opened 2018-08-03
Check dcp-int status on GitLab

xbrianh:
The dcp-wide integration test has moved to GitLab. This allows us to check status there.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -88,12 +88,13 @@ release_integration:
     - git clean -df
     - git reset --hard
     - for i in $(seq 1 40); do
-    -   if [[ "pending" != $(scripts/status.sh HumanCellAtlas dcp integration) ]]; then break; fi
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -88,12 +88,13 @@ release_integration:
     - git clean -df
     - git reset --hard
     - for i in $(seq 1 40); do
-    -   if [[ "pending" != $(scripts/status.sh HumanCellAtlas dcp integration) ]]; then break; fi
+    -   status=$(scripts/status.sh HumanCellAtlas dcp integration)
+    -   if [[ "pending" != ${status} && "running" != ${status} ]]; then break; fi
```
kislyuk: Technically, "pending" and "running" don't need to be quoted, but `${status}` could benefit from being quoted.

xbrianh: fixed

PR #1490, opened 2018-08-07
Improve error response documentation for PATCH /collections

ttung:
1. Explode all the various errors so they're tied to their own response codes.
2. Add illegal_arguments to the default error.  This is REQUIRED for every endpoint's default error handler, because when swagger validates the parameters and fails, it returns this code.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1053,6 +1053,57 @@ paths:
                     enum: [illegal_version]
                 required:
                   - code
+        403:
+          description: >
+            Returned when credentials presented do not grant access to this collection.
+          schema:
+            allOf:
+              - $ref: '#/definitions/Error'
+              - type: object
+                properties:
+                  code:
+                    type: string
+                    description: >
+                      Machine-readable error code.  The types of return values should not be changed lightly.
+
+                      The code `Forbidden` indicates that the user does not have access to this collection.
+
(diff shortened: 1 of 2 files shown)


ttung:
> But thinking about it, I guess invalid_link should always return 422. 404 is returned when the main resource requested is not found. In all conditions for invalid_link, the contents in the request contained an invalid reference, not the requested resource itself.

I agree. I made this change but then I reverted it because I was aiming for less controversial. :)

ttung:
Update all invalid references to return 422.

Makefile
```diff
@@ -41,7 +41,7 @@ serial_test:
 
 # A pattern rule that runs a single test script
 #
-$(tests): %.py : mypy lint
+$(tests): %.py :
```
kislyuk: debug leftovers?

daemons/package_daemon.sh
```diff
@@ -16,7 +16,7 @@ daemon=$1
 
 echo "Running pre-packaging steps for $daemon"
 
-git clean -df $daemon/domovoilib $daemon/vendor
+#git clean -df $daemon/domovoilib $daemon/vendor
```
kislyuk: debug leftovers?

kislyuk:
LGTM if you remove the debug bits.

PR #1496, opened 2018-08-13
[easy] Fix method signature in composite backend

hannes-ucsc:
It's confusing to pass in a callable and then only use its name.

From this morning's arch meeting.


mweiden:
@hannes-ucsc these tests are red. Seems like some issue with the schema(?)

PR #1510, opened 2018-08-16
Checkout from GitHub with token for releases

xbrianh:
For release builds, explicitly clone the repo from GitHub using an access token.

This fixes a bug where release pushes were attempted on the cloned repo in the build box, which failed.

Tested by successfully pushing a test tag to GitHub.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -85,8 +85,8 @@ integration_test:
 release_integration:
   stage: release_integration
   script:
-    - git clean -df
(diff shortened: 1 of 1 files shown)


xbrianh:
CI builds on the Build Box checks the repot out from allspark, (NOT GitHub). Thus, `origin` was pointing to allspark.

By cloning from GitHub with a scoped token, we push directly to GitHub, and get around the one-way mirroring.

PR #1512, opened 2018-08-17
[Easy] Don't mark job as failed when not releasing to integration

xbrianh:
This allows pipelines that pass all tests and deploy succesfully to `dev` to be marked as passed even if they fail to release to the `integration` stage.

This:
aaa

Will look more like this (note the orange shade):
bbb

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -106,6 +106,7 @@ release_integration:
     url: https://dss.integration.data.humancellatlas.org
   only:
     - master
+  allow_failure: true
 
 force_release_integration:
   stage: release_integration


mweiden:
@xbrianh so does this mean that we can potentially not be at fault if an `integration` deployment fails? If so, in what case?

xbrianh:
@mweiden One not-our-fault scenario is when the DCP-wide integration test is failing for reasons unrelated to the DSS. In this case, the GitLab release-to-integration job is designed to fail, expectedly.

It may also fail for our-fault reasons, so the orange "this mostly passed" marking seems reasonable.

Issue #1498, opened 2018-08-13
Checkout service should answer requests synchronously if possible


ttung:
GET /bundle/ and GET /files/ can probably answer requests synchronously if it can verify that the checkout is already present, and meets the advertised TTL.

mweiden:
This is a duplicate of https://github.com/HumanCellAtlas/data-store/issues/1347. Closing this.

PR #1515, opened 2018-08-18
Create environment variables for blob object expiration

calvinnhieu:
These changes introduce 2 new environment variables:
- `DSS_BLOB_TTL`: Number of days that "expirable" blobs are allowed to live in cloud storage (time to live)
- `DSS_BLOB_PUBLIC_TTL`: Number of days that we publicly claim that blobs live in cloud storage for

Some examples of "expirable" blobs are blobs in test buckets and/or checkout buckets. It is up for discussion as to whether we should differentiate TTLs for different uses (GS vs S3, different bucket types), however I don't have an argument for this. One potential case raised by @ttung is if we need to be cost conscious between different cloud providers.

Ensure that `source environment` is run before deploying.

Related to #1498


kislyuk:
It's probably better to put the unit into the var name: `DSS_BLOB_PUBLIC_TTL_DAYS`

.python-version
```diff
@@ -0,0 +1,2 @@
+data-store
```
ttung: I don't think you should be checking this in.  You should put this in your private gitignore (i.e., `.git/info/exclude`).

calvinnhieu: Agreed. My mistake!

calvinnhieu: My bad. Thanks for the tip re: private gitignore!

Issue #1347, opened 2018-06-20
optimization: reduce copying for checkout service


ttung:
If the advertised TTL for checked out bundles is X days, but we set the reaper to run at Y days, we can avoid the copy if the file is between 0 and (Y-X) days old.

ttung:
I won't be working on this.

PR #1516, opened 2018-08-18
Optimize get_file to run synchronously when possible

calvinnhieu:
These changes enable `get_file` (file GETs) to immediately return the location of the file if it has already been checked out with the conditions:
- the file has not "expired"
- the checksum of the checkout file matches the original (existing logic)

Expiration is defined by `request_time - advertised_ttl` (i.e. earliest acceptable last modified date) such that we ensure the file we serve is preserved in checkout for the time that we advertise to the user.

### Test plan
Tested:
- manual testing via local DSS API server
- added integration test

Will test:
- Hit get_file unchecked-out files, checked-out files that are either stale or valid, and observe expected behaviors

Must be merged after https://github.com/HumanCellAtlas/data-store/pull/1515
Related to #1347

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -1,4 +1,7 @@
+import datetime
 import json
+import logging
+import os
 import re
 import time
 import typing
@@ -28,6 +31,8 @@ ASYNC_COPY_THRESHOLD = AWS_MIN_CHUNK_SIZE
 """The retry-after interval in seconds. Sets up downstream libraries / users to
 retry request after the specified interval."""
 RETRY_AFTER_INTERVAL = 10
+
+logger = logging.getLogger(__name__)
 
 
 @dss_handler
@@ -74,7 +79,6 @@ def get_helper(uuid: str, replica: Replica, version: str=None, token: str=None):
 
     if request.method == "GET":
         token, ready = _verify_checkout(replica, token, file_metadata, blob_path)
-
         if ready:
             response = redirect(handle.generate_presigned_GET_url(
                 replica.checkout_bucket,
@@ -108,6 +112,25 @@ def get_helper(uuid: str, replica: Replica, version: str=None, token: str=None):
 def _verify_checkout(
         replica: Replica, token: typing.Optional[str], file_metadata: dict, blob_path: str,
 ) -> typing.Tuple[str, bool]:
+    try:
+        cloud_handle = Config.get_blobstore_handle(replica)
+        hca_handle = Config.get_hcablobstore_handle(replica)
+        file_key = get_dst_key(blob_path)
+        last_modified = cloud_handle.get_last_modified_date(replica.checkout_bucket, file_key)
+        stale_before_date = (datetime.datetime.now(datetime.timezone.utc) -
+                             datetime.timedelta(days=int(os.environ['DSS_BLOB_PUBLIC_TTL_DAYS'])))
+
+        if last_modified > stale_before_date:
+            if hca_handle.verify_blob_checksum_from_dss_metadata(replica.checkout_bucket,
+                                                                 file_key,
+                                                                 file_metadata):
+                return "", True
+            else:
+                logger.error(
+                    f"Checksum verification failed for file {replica.checkout_bucket}/{file_key}")
+    except BlobNotFoundError:
+        pass
+
     decoded_token: dict
     if token is None:
         execution_id = start_file_checkout(replica, blob_path)
@@ -126,15 +149,7 @@ def _verify_checkout(
         except (KeyError, ValueError) as ex:
             raise DSSException(requests.codes.bad_request, "illegal_token", "Could not understand token", ex)
 
-    hcablobstore = Config.get_hcablobstore_handle(replica)
     encoded_token = json.dumps(decoded_token)
-    try:
-        if hcablobstore.verify_blob_checksum_from_dss_metadata(
-                replica.checkout_bucket, get_dst_key(blob_path), file_metadata):
-            return encoded_token, True
-    except BlobNotFoundError:
-        pass
-
     return encoded_token, False
 
 
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -9,6 +9,7 @@ import sys
 import tempfile
 import typing
 import unittest
+from unittest import mock
 import uuid
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
@@ -376,16 +377,22 @@ class TestFileApi(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
             )
 
     @testmode.standalone
-    def test_file_checkout(self):
+    def test_file_get_invalid_token(self):
         """
         Verifies that a checkout request with a malformed token returns a 400.
         :return:
         """
         tempdir = tempfile.gettempdir()
-        self._test_file_checkout(Replica.aws, "s3", self.s3_test_bucket, S3Uploader(tempdir, self.s3_test_bucket))
-        self._test_file_checkout(Replica.gcp, "gs", self.gs_test_bucket, GSUploader(tempdir, self.gs_test_bucket))
+        self._test_file_get_invalid_token(Replica.aws,
+                                          "s3",
+                                          self.s3_test_bucket,
+                                          S3Uploader(tempdir, self.s3_test_bucket))
+        self._test_file_get_invalid_token(Replica.gcp,
+                                          "gs",
+                                          self.gs_test_bucket,
+                                          GSUploader(tempdir, self.gs_test_bucket))
 
-    def _test_file_checkout(self, replica: Replica, scheme: str, test_bucket: str, uploader: Uploader):
+    def _test_file_get_invalid_token(self, replica: Replica, scheme: str, test_bucket: str, uploader: Uploader):
         src_key = generate_test_key()
         src_data = os.urandom(1024)
         with tempfile.NamedTemporaryFile(delete=True) as fh:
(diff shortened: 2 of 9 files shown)


tests/test_file.py
```diff
@@ -414,6 +416,62 @@ def try_get():
...
+        # should be able to do this twice (i.e., same payload, different UUIDs)
+        self.upload_file(source_url, file_uuid, bundle_uuid=bundle_uuid, version=version)
+        url = str(UrlBuilder()
+                  .set(path="/v1/files/" + file_uuid)
+                  .add_query("replica", replica.name)
+                  .add_query("version", version))
+
+        now = datetime.datetime.now(pytz.utc)
+        # assert 301 redirect on first GET
+        self.assertGetResponse(url, requests.codes.moved, redirect_follow_retries=0)
```
kislyuk: This is outside the scope of this PR, but I think the proper status code for the redirect loop is actually 303 See Other, not 301 Moved Permanently.

tests/test_file.py
```diff
@@ -414,6 +416,62 @@ def try_get():
...
+        last_modified_fn = ("cloud_blobstore.s3.S3BlobStore.get_last_modified_date"
+                            if replica.name == "aws"
+                            else "cloud_blobstore.gs.GSBlobStore.get_last_modified_date")
+        with mock.patch(last_modified_fn) as mock_last_modified:
+            # assert 301 redirect on stale file
+            mock_last_modified.return_value = now - datetime.timedelta(days=10)
+            self.assertGetResponse(url, requests.codes.moved, redirect_follow_retries=0)
+            # assert 302 found on subsequent GET
+            mock_last_modified.return_value = now - datetime.timedelta(days=0)
+            try_get()
```
kislyuk: I think you should try actually getting the body of the file here and verify that the contents match.

environment
```diff
@@ -60,6 +60,8 @@ DSS_GS_CHECKOUT_BUCKET_TEST=org-humancellatlas-dss-checkout-test
 DSS_GS_CHECKOUT_BUCKET_INTEGRATION=org-humancellatlas-dss-checkout-integration
 DSS_GS_CHECKOUT_BUCKET_STAGING=org-humancellatlas-dss-checkout-staging
 DSS_GS_CHECKOUT_BUCKET_PROD=org-humancellatlas-dss-checkout-prod
+DSS_BLOB_TTL=7
+DSS_BLOB_PUBLIC_TTL=4
```
kislyuk: Please include units in variables like this. `DSS_BLOB_TTL_DAYS`

tests/test_file.py
calvinnhieu: I see. I'll create an issue for this

tests/test_file.py
calvinnhieu: Right, that'll verify that the correct file got checked out :+1:

tests/test_file.py
calvinnhieu: https://app.zenhub.com/workspace/o/humancellatlas/data-store/issues/1517

dss/api/files.py
```diff
@@ -112,6 +114,20 @@ def get_helper(uuid: str, replica: Replica, version: str=None, token: str=None):
 def _verify_checkout(
         replica: Replica, token: typing.Optional[str], file_metadata: dict, blob_path: str,
 ) -> typing.Tuple[str, bool]:
+    try:
+        cloud_handle = Config.get_blobstore_handle(replica)
+        hca_handle = Config.get_hcablobstore_handle(replica)
+        last_modified = cloud_handle.get_last_modified_date(replica.checkout_bucket, "files/{}".format(blob_path))
+        stale_date = datetime.datetime.now(pytz.utc) - datetime.timedelta(days=int(os.environ['DSS_BLOB_PUBLIC_TTL']))
```
ttung: why not just `datetime.now(timezone.utc)`?

ttung: Also, I don't think `stale_date` is an appropriate description for what this variable holds.

dss/api/files.py
```diff
@@ -112,6 +114,20 @@ def get_helper(uuid: str, replica: Replica, version: str=None, token: str=None):
...
+
+        if ((last_modified > stale_date) and
+            (hca_handle.verify_blob_checksum_from_dss_metadata(replica.checkout_bucket,
+                                                               get_dst_key(blob_path),
+                                                               file_metadata))):
+            return "", True
```
ttung: Would consider making the str optional for the return type to make this less awkward.

ttung: Also, if the file is present but the checksum does not match, I think it's worth logging some very serious error.

tests/test_file.py
```diff
@@ -1,14 +1,17 @@
 #!/usr/bin/env python
 # coding: utf-8
 
+from bs4 import BeautifulSoup
```
ttung: 3rd party imports belong with the second set of imports.

tests/test_file.py
```diff
@@ -414,6 +417,68 @@ def try_get():
...
+
+        now = datetime.datetime.now(pytz.utc)
+        # assert 301 redirect on first GET
+        self.assertGetResponse(url, requests.codes.moved, redirect_follow_retries=0)
+
+        @eventually(20, 2)
+        def test_checkout():
+            resp = self.assertGetResponse(
+                url, requests.codes.found, redirect_follow_retries=0)
+            html = BeautifulSoup(resp.body, 'html.parser')
```
ttung: I think this can be done by inspecting the location header in the response.

calvinnhieu: Got it. I didn't realize `assertGetResponse` returns a wrapper around the original response. Thanks!

dss/api/files.py
calvinnhieu: Oh didn't know about that. Agree that we shouldn't be adding dependencies if there are native ways to do it. And you're right - changing to `stale_before_date`

dss/api/files.py
calvinnhieu: How would one make the return type Tuple(str, bool) or bool?

dss/api/files.py
ttung: Alternatively, you can calculate the time it's going to be stale (last modified time + offset), and test that against the current date.

dss/api/files.py
```diff
@@ -112,6 +116,25 @@ def get_helper(uuid: str, replica: Replica, version: str=None, token: str=None):
...
+                             datetime.timedelta(days=int(os.environ['DSS_BLOB_PUBLIC_TTL_DAYS'])))
+
+        if last_modified > stale_before_date:
+            if hca_handle.verify_blob_checksum_from_dss_metadata(replica.checkout_bucket,
+                                                                 checkout_file_key,
+                                                                 file_metadata):
+                return "", True
+            else:
+                logger.error(
+                    f"Failed to verify file checksum during checkout: {replica.checkout_bucket}/{checkout_file_key}")
```
ttung: Prefer: `Checksum verification failed.`

Failed to verify checksum could mean that you tried to verify the checksum and couldn't actually perform the operation _or_ the checksum verification failed.

tests/test_file.py
```diff
@@ -376,7 +377,7 @@ def test_file_get_no_replica(self):
             )
 
     @testmode.standalone
-    def test_file_checkout(self):
+    def test_file_get_invalid_token(self):
```
ttung: thanks for fixing this. :)

tests/test_file.py
```diff
@@ -414,6 +415,66 @@ def try_get():
                 url, requests.codes.bad_request)
         try_get()
 
+    @testmode.standalone
```
ttung: This is not a standalone test, since it depends on a step function being deployed.

Consider manually checking out, i.e., manually copying to the expected destination (dss/storage/checkout/file.py:get_dst_key), and then execute the second request if you want this to be a standalone test.

There are certainly some significant advantages to this approach, as you can test the checksum-not-matching code path.

tests/test_file.py
calvinnhieu: Gladly!

dss/api/files.py
```diff
@@ -112,6 +116,25 @@ def get_helper(uuid: str, replica: Replica, version: str=None, token: str=None):
 def _verify_checkout(
         replica: Replica, token: typing.Optional[str], file_metadata: dict, blob_path: str,
 ) -> typing.Tuple[str, bool]:
+    try:
+        cloud_handle = Config.get_blobstore_handle(replica)
+        hca_handle = Config.get_hcablobstore_handle(replica)
+        checkout_file_key = get_dst_key(blob_path)
+        last_modified = cloud_handle.get_last_modified_date(replica.checkout_bucket, "files/{}".format(blob_path))
```
mweiden: You define `checkout_file_key` which results in `f"files/{blob_path}"` and then submit `"files/{}".format(blob_path)` to `get_last_modified_date` on the next line?

calvinnhieu: Good catch!

calvinnhieu:
rebased

tests/test_file.py
```diff
@@ -414,6 +421,67 @@ def try_get():
                 url, requests.codes.bad_request)
         try_get()
 
+    @testmode.integration
+    def test_file_get_checkout(self):
+        """
+        Verifies checkout occurs on first get and not on second.
+        :return:
```
ttung: remove this line.

tests/test_file.py
```diff
@@ -414,6 +421,67 @@ def try_get():
...
+
+            uploader.checksum_and_upload_file(fh.name, src_key, "text/plain")
+
+        source_url = f"{scheme}://{test_bucket}/{src_key}"
+
+        file_uuid = str(uuid.uuid4())
+        bundle_uuid = str(uuid.uuid4())
+        version = datetime_to_version_format(datetime.datetime.utcnow())
+
+        # should be able to do this twice (i.e., same payload, different UUIDs)
```
ttung: remove this comment, it's not relevant.

tests/test_file.py
```diff
@@ -414,6 +421,67 @@ def try_get():
...
+        # assert 301 redirect on first GET
+        self.assertGetResponse(url, requests.codes.moved, redirect_follow_retries=0)
+        test_checkout()
+
+        last_modified_fn = ("cloud_blobstore.s3.S3BlobStore.get_last_modified_date"
+                            if replica.name == "aws"
+                            else "cloud_blobstore.gs.GSBlobStore.get_last_modified_date")
+        with mock.patch(last_modified_fn) as mock_last_modified:
+            # assert 301 redirect on stale file
+            blob_ttl_days = int(os.environ['DSS_BLOB_PUBLIC_TTL_DAYS'])
```
ttung: rather than mock the `get_last_modified date` call, you could make the TTL a second-based variable, override that to some very low value, and then do your second get (which should be "stale")

calvinnhieu: In other words, modify `DSS_BLOB_PUBLIC_TTL_DAYS` to be expressed in seconds such that it can be mocked in terms of seconds?

ttung: For ease, you may still want the environment variable to be in units of days, but when read in, you could store it in seconds, and then tweak your behavior based on that.

calvinnhieu: Sounds like another layer of complexity that would exist to modify the implementation of the test. I'd argue to keep this as it is

PR #1522, opened 2018-08-23
Optimize get_bundle to only perform checkout when necessary

### LINKED ISSUE #1347: optimization: reduce copying for checkout service

calvinnhieu:
These changes update bundle GETs (`get_bundle`) to only perform a checkout in the following cases:
- if any file in the bundle has become [stale](https://github.com/HumanCellAtlas/data-store/pull/1516)
- if any file in the checkout bundle is not listed in the bundle manifest
- if the number of files in the checkout bundle does not match the number of files in the bundle manifest

### Test plan
- Manual testing on dev DSS
- Added integration test

### Release notes
Upgraded dependency: [`cloud-blobstore` >= 3.0.0](https://github.com/chanzuckerberg/cloud-blobstore/pull/21)

Should be merged after https://github.com/HumanCellAtlas/data-store/pull/1516
Fixes #1347

### DIFF
--- a/dss/storage/bundles.py
+++ b/dss/storage/bundles.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 import json
 import typing
 
@@ -8,6 +9,7 @@ from dss.storage.identifiers import DSS_BUNDLE_KEY_REGEX, DSS_BUNDLE_TOMBSTONE_R
 from dss.storage.blobstore import test_object_exists
 
 
+@lru_cache(maxsize=32)
 def get_bundle_manifest(
         uuid: str,
         replica: Replica,
--- a/dss/storage/checkout/bundle.py
+++ b/dss/storage/checkout/bundle.py
@@ -1,11 +1,14 @@
 """
 This is the module for bundle checkouts.
 """
+import datetime
 import io
 import json
+import os
 import time
 import typing
 
+from cloud_blobstore import BlobMetadataField
 from dss import stepfunctions
 from dss.config import Replica, Config
 from dss.stepfunctions.checkout.constants import EventConstants, STATE_MACHINE_NAME_TEMPLATE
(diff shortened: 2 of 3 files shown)


calvinnhieu:
TODO: update requirements.txt

dss/api/bundles/__init__.py
```diff
@@ -52,7 +52,7 @@ def get(
 
     if directurls or presignedurls:
         try:
-            token, ready = verify_checkout(_replica, uuid, version, token)
+            token, ready = verify_checkout(_replica, uuid, version, token, bundle_metadata)
```
calvinnhieu: Does this code get copied over to the step function/daemon code during deployment?

Context: Passing in `bundle_metadata` here to reduce calls to `get_bundle_manifest`. Potentially a better solution pointed out by @kislyuk would be to use the `@functools.lru_cache` decorator on `get_bundle_manifest`.

dss/api/files.py
```diff
@@ -112,6 +116,25 @@ def get_helper(uuid: str, replica: Replica, version: str=None, token: str=None):
 def _verify_checkout(
         replica: Replica, token: typing.Optional[str], file_metadata: dict, blob_path: str,
 ) -> typing.Tuple[str, bool]:
+    try:
```
ttung: I am confused as this is not consistent with the PR title.  Is it possible you mixed commits and PRs?

calvinnhieu: Yea I can see how this is confusing... Essentially I had my branches set up as: `master` -> `create-blob-ttl-env-vars` -> `optimize-file-get` -> `optimize-bundle-get` while keeping the downstream branches up-to-date. In hindsight, `optimize-bundle-get` should have branched off of `create-blob-ttl-env-vars`. Regardless, the file-get related commits will disappear from this PR when https://github.com/HumanCellAtlas/data-store/pull/1516 is merged. Sorry for the confusion!

tests/test_file.py
```diff
@@ -414,6 +421,66 @@ def try_get():
...
+
+        last_modified_fn = ("cloud_blobstore.s3.S3BlobStore.get_last_modified_date"
+                            if replica.name == "aws"
+                            else "cloud_blobstore.gs.GSBlobStore.get_last_modified_date")
+        with mock.patch(last_modified_fn) as mock_last_modified:
+            # assert 301 redirect on stale file
+            mock_last_modified.return_value = now - datetime.timedelta(days=10)
+            self.assertGetResponse(url, requests.codes.moved, redirect_follow_retries=0)
+            # assert 302 found on subsequent GET
+            mock_last_modified.return_value = now - datetime.timedelta(days=0)
```
mweiden: nit/idea: you might change `days` to be at the boundary to make sure that your comparison logic is correct

PR #1523, opened 2018-08-23
Add missing quotes to fix tf scripts

calvinnhieu:
Fix for broken deploy:
https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/4991b

Testing:
- ran `make infra-plan` locally
- should we add a test that runs .tf scripts in the travis PR builds?


calvinnhieu:
Cool, can you point me to where that code is?

Issue #1486, opened 2018-08-06
Any logged in user with a valid email address should be able to subscribe


mweiden:
# User Story

* As a user of the DSS consumer API
* I want to submit subscription requests with just my email address
* so that I don't have to be manually approved.

# Definition of Done

Any logged in user with a valid email address can register subscriptions with DSS.

Currently, only email domains whitelisted in `DSS_SUBSCRIPTION_AUTHORIZED_DOMAINS_ARRAY` can subscribe.

ttung:
This is highly problematic.  We need a proper confirmation process so we don't become a spam vector.

Bento007:
@ttung can you open a issue for this?

> This is highly problematic. We need a proper confirmation process so we don't become a spam vector.

@ttung can you open a new issue for this

ttung:
Actually, I _think_ we are okay.  This requires that the user auth in with google, so presumably that means it's not just anyone who can do it.

mweiden:
@ttung 🙊 https://www.fiverr.com/facebook4tune/give-you-100-usa-gmail-account--2

PR #1524, opened 2018-08-24
remove subscription restriction

Bento007:


mweiden:
@Bento007 You need a rebase here.

Bento007:
fixes #1486

PR #1529, opened 2018-08-24
Put the shape descriptor in the ES index doc

### LINKED ISSUE #1453: Write shape descriptor to each document in the index
If we hash the shape descriptor, we should keep track of it in each document.

xbrianh:
Closes #1453

### DIFF
--- a/dss/index/es/document.py
+++ b/dss/index/es/document.py
@@ -114,6 +114,7 @@ class BundleDocument(IndexDocument):
         scrub_index_data(files, str(self.fqid))
         self['files'] = files
         self['uuid'] = self.fqid.uuid
+        self['shape_descriptor'] = str(self.get_shape_descriptor())
         return self
 
     @classmethod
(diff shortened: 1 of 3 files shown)


xbrianh:
@Bento007 I agree that the shape descriptor should only be computed once. However, I don't see the value in computing it lazily since it is part of the index document. To formalize these sentiments, I've  turned it into a static method that is only called during initialization from bundle.

dss/index/es/document.py
```diff
@@ -213,7 +213,8 @@ def _prepare_index(self, dryrun):
             IndexManager.create_index(es_client, self.replica, index_name)
         return index_name
 
-    def get_shape_descriptor(self) -> typing.Optional[str]:
+    @staticmethod
+    def _get_shape_descriptor(document) -> typing.Optional[str]:
```
hannes-ucsc: A static/class method of a class whose sole argument is an instance of that class should definitely be an instance method. Why is conversion necessary? I surely isn't intuitive.

xbrianh: My intention was to discourage a possible bug that could arise from the possibility that `doc._get_shape_descriptor() != doc['shape_descriptor']`, since `doc` is mutable.

This has been backed out.

PR #1534, opened 2018-08-28
Prevent object modification rate limit for touch test files.

Bento007:
Store touch test files across a range of names to prevent the error, "The total number of changes to the object org-humancellatlas-dss-checkout-test/touch.txt exceeds the rate limit. Please reduce the rate of create, update, and delete requests."


ttung:
So what happened to #1441?  Did it not get merged or something?

Bento007:
Honestly I have no idea. It looks like it was never merged into master according to the branch history.

PR #1535, opened 2018-08-29
Consolodate bundle fixture

xbrianh:
* The example bundle fixture is shared and has been moved out of the indexer fixtures
* Update README.md example query

This has been tested against custom fixture buckets. Fixtures will be repopulated when this lands.


mweiden:
@xbrianh how is versioning handled in this new directory scheme? I don't see a directory that specifies that...

xbrianh:
@mweiden There is currently only a single example bundle, so version doesn't carry much meaning internally to the DSS. Do you feel it should nevertheless be specified?

mweiden:
@xbrianh no. Parsimony is best until we need it.

PR #1536, opened 2018-08-29
Remove unused dependency (moto) and its hazardous subdeps

kislyuk:
Also disabled find_missing_wheels since it tries to use code from chalice before it can be installed


ttung:
Ummmm this didn't land in master?

PR #1540, opened 2018-08-30
Refactor gitlab jobs

xbrianh:
Split apart release and deploy.

This should make re-deploying, or rolling back, possible for `integration` and `staging`

#depends on #1537

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -102,6 +102,10 @@ integration_test:
   stage: integration_test
   script:
     - make -j1 integration_test
+  only:
+    - master
+    - integration
+    - staging
 
 release_integration:
   stage: release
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -18,10 +18,9 @@ variables:
 
 stages:
   - test
-  - deploy_dev
+  - deploy
   - integration_test
-  - release_integration
-  - release_staging
+  - release
```
mweiden: nit: order these chronologically

xbrianh: In my mind they are: `test` first, if pass `deploy`, if deployed `integration_test`

mweiden: Ah, good point. OK then, go ahead.

PR #1541, opened 2018-08-31
[EASY] Files are checked out to bucket/blobs

Bento007:


Bento007:
@ttung https://github.com/HumanCellAtlas/data-store/commit/bc6efb90c37b107066a1c78c045f5fc692deaac2#commitcomment-30352347

blob_key has the format `blobs/hash.hash.hash` so all of the files will end up in `bucket/blobs`. The alternative option is to revert this, and modify https://github.com/HumanCellAtlas/data-store/blob/2d11f5144315cf1df9e941af8183478827c01c51/dss/api/bundles/__init__.py#L86 to 
```python
path='files/' + compose_blob_key(file),
```
This will ensure the _directurl_ points to where the file actually is.

Bento007:
@xbrianh Done https://github.com/HumanCellAtlas/data-store/pull/1543

ttung:
Very unclear what problem this solves...

Bento007:
The directurls generated were not actually pointing to anything in checkout bucket. This moves the files checked out to be be in the location expected by the directurls we generate.

ttung:
> The directurls generated were not actually pointing to anything in checkout bucket. This moves the files checked out to be be in the location expected by the directurls we generate.

That doesn't make sense.  /files is not being used to generate directurls at all.  It generates presigned URLs, but never directurls.

PR #1545, opened 2018-08-31
directaccess and presignedurl should go to the checkout bucket

ttung:
1. `dss/api/bundles/__init__.py`: changes to which bucket we point to after the checkout succeeds.  Note that the keys are different.
2. Make the dst replica explicit.  Right now, the state machine is not aware of overrides done in tests.
3. Small change #1 that I'm sneaking in: `bundle_version` is actually possibly None.
4. Small change #2 that I'm sneaking in: fix styling around default parameter values.

Test plan: `DSS_TEST_MODE="standalone integration" make -j tests/test_checkout.py tests/test_file.py tests/test_bundle.py tests/test_api.py`

### DIFF
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -15,9 +15,9 @@ from dss.config import Config, Replica
 from dss.storage.blobstore import test_object_exists, ObjectTest
 from dss.storage.bundles import get_bundle_manifest
 from dss.storage.checkout import CheckoutError, TokenError
-from dss.storage.checkout.bundle import verify_checkout
+from dss.storage.checkout.bundle import get_dst_bundle_prefix, verify_checkout
 from dss.storage.identifiers import BundleTombstoneID, BundleFQID, FileFQID
-from dss.storage.hcablobstore import BundleFileMetadata, BundleMetadata, FileMetadata, compose_blob_key
+from dss.storage.hcablobstore import BundleFileMetadata, BundleMetadata, FileMetadata
 from dss.util import UrlBuilder
 from dss.util.version import datetime_to_version_format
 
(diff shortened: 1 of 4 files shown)


dss/api/bundles/__init__.py
```diff
@@ -83,12 +83,20 @@ def get(
             file_version['url'] = str(UrlBuilder().set(
                 scheme=_replica.storage_schema,
                 netloc=_replica.checkout_bucket,
-                path=compose_blob_key(file),
+                path="{}/{}".format(
+                    get_dst_bundle_prefix(uuid, version),
```
Bento007: This will fail when version is None.

ttung: Fixed.  Added tests to cover this scenario.

dss/api/bundles/__init__.py
```diff
@@ -83,12 +83,20 @@ def get(
             file_version['url'] = str(UrlBuilder().set(
                 scheme=_replica.storage_schema,
                 netloc=_replica.checkout_bucket,
-                path=compose_blob_key(file),
+                path="{}/{}".format(
```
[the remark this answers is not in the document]
ttung: No, they are not.

tests/test_bundle.py
```diff
@@ -85,24 +85,26 @@ def _test_bundle_get(self, replica: Replica):
 
     @testmode.integration
     def test_bundle_get_directaccess(self):
```
xbrianh: This test parses out they key from the direct url and verifies contents.

Is there a test somewhere that confirms the direct url points to the file we expect?

ttung: Contents should be good enough.  It does verify that the bucket is the checkout bucket.

ttung: https://github.com/HumanCellAtlas/data-store/pull/1545/files#diff-307f004e5f5806557cc669e52b0cdd1eR126

xbrianh: Ok, one more question and then I'll hold my peace:  Does it verify that `aws s3 cp {direct_url} .` will work?

ttung: No, it does not.

PR #1551, opened 2018-09-05
Further .gitlab-ci.yml simplifications and fixes

xbrianh:
* Use the scripts/release.sh from source branch, not dest branch
* Don't source environment from scripts/release.sh, which overwrites source env with `dev` env
* Use `--` to make sure git can differentiate flags from branches
* Remove unused gitlab cache
* Remove unused Terraform env var

`.gitlab-ci.yml` has undergone some churn. However, a full test of the release pipeline should now be possible.


mweiden:
@xbrianh why wasn't the cache used?

xbrianh:
It's an artifact from before we created a custom Docker image for the build.

PR #1552, opened 2018-09-05
Timeout fixes for dss tests

ttung:
1. Remove the time_left hack for the missing file test.  This does cut about 30s from test time, but is not consistent with how it works in production.
2. Reduce the server time allowed by chalice to 30 sec to match production.
3. Change the time the client waits to 35 seconds (it must be longer than the server execution time).
4. Remove OVERRIDE_EXECUTION_LIMIT_SECONDS, which is not really doing much.

### DIFF
--- a/tests/infra/server.py
+++ b/tests/infra/server.py
@@ -44,7 +44,7 @@ class ThreadedLocalServer(threading.Thread):
         project_dir = os.path.join(os.path.dirname(__file__), "..", "..", "chalice")
         factory = CLIFactory(project_dir=project_dir)
         self._chalice_app = factory.load_chalice_app()
-        self._chalice_app._override_exptime_seconds = 86400  # something large.  sys.maxsize causes chalice to flip.
+        self._chalice_app._override_exptime_seconds = 30  # something large.  sys.maxsize causes chalice to flip.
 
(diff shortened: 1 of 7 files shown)


tests/infra/server.py
```diff
@@ -44,7 +44,7 @@ def run(self):
         project_dir = os.path.join(os.path.dirname(__file__), "..", "..", "chalice")
         factory = CLIFactory(project_dir=project_dir)
         self._chalice_app = factory.load_chalice_app()
-        self._chalice_app._override_exptime_seconds = 86400  # something large.  sys.maxsize causes chalice to flip.
+        self._chalice_app._override_exptime_seconds = 30  # something large.  sys.maxsize causes chalice to flip.
```
Bento007: Is this needed. I find having this value large makes debugging easier. Otherwise you only have 30 seconds to step though the server code before a timeout error is triggered.

ttung: If it's 86400, then the timeout code for PUT /bundles with a missing file is not really exercised (i.e., it immediately times out).

Bento007: I suppose when debugging we can override this value with something larger. Will you add a note in data_store.tests.readme that this value can be adjusted when debugging.

Bento007:
@ttung will you be making any more changes to this PR?

ttung:
I won't do this for now.

PR #1553, opened 2018-09-06
Parameterize GCP checkout bucker viewers

xbrianh:
These GCP accounts are granted permissions to use direct urls in the checkout bucket.

#closes 1550


mweiden:
@xbrianh should we also add the ones for dev?

xbrianh:
@mweiden Probably, pending arch discussion. The `dev` checkout bucket currently lets all users access direct urls.

PR #1554, opened 2018-09-06
Sync daemon: use SQS and state functions, ensure referential consistency

### LINKED ISSUE #1130: Use step functions to orchestrate sync'ing of entire bundles

kislyuk:
Fixes #1130.

- Sync daemon uses SQS to ensure S3 event queue elasticity/at-least-once semantics under burst loads.
- Rewrite GS event relay to use GCF Python 3.7 runtime.
- Sync daemon uses SFNs to orchestrate copying of data:
  - For metadata manifests (file, bundle and collection manifests), the step function waits for all objects referenced by the manifest to appear in the destination replica.
  - Metadata manifests are always copied in one shot.
  - Blobs below the part size are copied in one shot.
  - Blobs above the part size use the step function threadpool pattern.
- Sync daemon SFNs use standard DSS SFN retry parameters.

### Test plan
Expanded unit tests. Added integration tests. Ran manual tests against test buckets.


daemons/dss-gs-event-relay/main.py
```diff
@@ -0,0 +1,75 @@
...
+        # Metageneration is updated on metadata changes and starts at 1
+        print("Ignoring object metadata update event")
+    elif re.match(r".+\.part\d+$", data["name"]):
+        print("Ignoring multipart object upload event")
+    else:
+        grtc_client = GRTCClient()
+        config_ns = f"projects/{grtc_client.get_project()}/configs"
+        var_ns = f"{config_ns}/{os.environ['ENTRY_POINT']}/variables"
+        for config_var in "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_DEFAULT_REGION", "sqs_queue_url":
+            os.environ[config_var] = base64.b64decode(grtc_client.get(f"{var_ns}/{config_var}")["value"]).decode()
```
xbrianh: These are what look like REST API calls to fetch parameters from GCP Runtime Config. But in order to configure Runtime Config during deployment the parameters are (in part) fetched from AWS Secretsmanager.

Why not dispense with Runtime Config and hit Secretsmanager here? Would that be slower and/or less reliable?

kislyuk: Chicken-and-egg problem.

To fetch secrets from Secretsmanager, you need credentials to access Secretsmanager. On AWS, those are given to you through instance metadata. On Google, you don't get those. You get Google credentials instead though, and those are good for hitting Runtime Config.

If we don't do this dance, then we have to inline/hardcode secrets into the GCF function code or environment variables.

tests/test_sync.py
```diff
@@ -137,17 +157,123 @@ def test_copy_part_gs_to_s3(self):
...
+            FileMetadata.S3_ETAG: "sync_test",
+            FileMetadata.CRC32C: str(uuid.uuid4())
+        }
+
+        collection_key = "{}/{}".format(COLLECTION_PREFIX, get_collection_fqid())
+        collection_blob = self.s3_bucket.Object(collection_key)
+        collection_blob.put(Body=json.dumps(collection_data).encode())
+        self.assertFalse(sync.dependencies_exist(Replica.aws, Replica.aws, collection_key))
+
+        bundle_key = "{}/{}".format(BUNDLE_PREFIX, BundleFQID(uuid=bundle_uuid, version=bundle_version))
```
Bento007: make into subtest

tests/test_sync.py
```diff
@@ -137,17 +157,123 @@ def test_copy_part_gs_to_s3(self):
...
+        bundle_blob = self.s3_bucket.Object(bundle_key)
+        bundle_blob.put(Body=json.dumps(bundle_data).encode())
+
+        self.assertFalse(sync.dependencies_exist(Replica.aws, Replica.aws, collection_key))
+        self.assertFalse(sync.dependencies_exist(Replica.aws, Replica.aws, bundle_key))
+
+        file_key = "{}/{}".format(FILE_PREFIX, FileFQID(uuid=file_uuid, version=file_version))
```
Bento007: make into subtest

tests/test_sync.py
```diff
@@ -137,17 +157,123 @@ def test_copy_part_gs_to_s3(self):
...
+        file_blob.put(Body=json.dumps(file_data).encode())
+
+        @eventually(timeout=8, interval=1, errors={Exception})
+        def check_file_revdeps():
+            self.assertTrue(sync.dependencies_exist(Replica.aws, Replica.aws, collection_key))
+            self.assertTrue(sync.dependencies_exist(Replica.aws, Replica.aws, bundle_key))
+            self.assertFalse(sync.dependencies_exist(Replica.aws, Replica.aws, file_key))
+        check_file_revdeps()
+
+        blob_key = compose_blob_key(file_data)
```
Bento007: make into subtest

tests/test_sync.py
```diff
@@ -137,17 +157,123 @@ def test_copy_part_gs_to_s3(self):
...
+                         "Skipping large file sync test. Set DSS_TEST_LARGE_FILE_SYNC=1 to run it")
+    def test_sync_large_blob(self):
+        "Tests multipart blob syncing."
+        self._sync_blob(payload_size=max(sync.part_size.values()) + 1)
+
+    def _sync_blob(self, payload_size=8):
+        self.cleanup_sync_test_objects()
+        payload = self.get_payload(payload_size)
+        test_metadata = {"metadata-sync-test": str(uuid.uuid4())}
+        test_key = "{}/s3-to-gcs/{}".format(self.test_blob_prefix, uuid.uuid4())
```
Bento007: make into separate test or subtest

tests/test_sync.py
```diff
@@ -137,17 +157,123 @@ def test_copy_part_gs_to_s3(self):
...
+        src_blob = self.s3_bucket.Object(test_key)
+        gs_dest_blob = self.gs_bucket.blob(test_key)
+        src_blob.put(Body=payload, Metadata=test_metadata)
+
+        @eventually(timeout=60, interval=1, errors={Exception})
+        def check_gs_dest():
+            self.assertEqual(gs_dest_blob.download_as_string(), payload)
+        check_gs_dest()
+
+        test_key = "{}/gcs-to-s3/{}".format(self.test_blob_prefix, uuid.uuid4())
```
Bento007: make into subtest

tests/test_sync.py
```diff
@@ -137,17 +157,123 @@ def test_copy_part_gs_to_s3(self):
...
+        src_blob.upload_from_string(payload)
+
+        @eventually(timeout=60, interval=1, errors={Exception})
+        def check_s3_dest(dest_blob):
+            self.assertEqual(dest_blob.get()["Body"].read(), payload)
+            self.assertEqual(dest_blob.metadata, test_metadata)
+        check_s3_dest(dest_blob)
+
+        # GS metadata seems to take a while to propagate, so we wait until the above test completes to read it back
+        gs_dest_blob.reload()
```
Bento007: same thing as with `TestSyncUtils.test_sync_blob`

tests/test_sync.py
```diff
@@ -137,17 +157,123 @@ def test_copy_part_gs_to_s3(self):
...
+
+    def test_sync_small_blob(self):
+        "Tests oneshot blob syncing."
+        self._sync_blob()
+
+    @unittest.skipUnless(json.loads(os.environ.get("DSS_TEST_LARGE_FILE_SYNC", "false")),
+                         "Skipping large file sync test. Set DSS_TEST_LARGE_FILE_SYNC=1 to run it")
+    def test_sync_large_blob(self):
+        "Tests multipart blob syncing."
+        self._sync_blob(payload_size=max(sync.part_size.values()) + 1)
```
Bento007: add tests around the multipart copy threshold. -1, 0, +1

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+                                      do_oneshot_copy, sync_sfn_dep_wait_sleep_seconds, sync_sfn_num_threads)
+
+configure_lambda_logging()
+logger = logging.getLogger(__name__)
+dss.Config.set_config(dss.BucketConfig.NORMAL)
+
+app = domovoi.Domovoi()
+
+# This entry point is for S3 native events forwarded through SQS.
+@app.s3_event_handler(bucket="akislyuk-experiments", events=["s3:ObjectCreated:*"], use_sqs=True)
```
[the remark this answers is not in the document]
kislyuk: Fixed, thanks.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+    else:
+        for event_record in event["Records"]:
+            bucket = resources.s3.Bucket(event_record["s3"]["bucket"]["name"])
+            obj = bucket.Object(unquote(event_record["s3"]["object"]["key"]))
+            if bucket.name != source_replica.bucket:
+                logger.error("Received S3 event for bucket %s with no configured replica", bucket.name)
+                continue
+
+            for dest_replica in Config.get_replication_destinations(source_replica):
+                if exists(dest_replica, obj.key):
```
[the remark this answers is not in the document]
kislyuk: I don't think that should be done in the scope of this PR.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+def launch_from_forwarded_event(event, context):
+    executions = {}
+    for event_record in event["Records"]:
+        message = json.loads(event_record["body"])
+        if message["selfLink"].startswith("https://www.googleapis.com/storage"):
+            source_replica = Replica.gcp
+            source_key = message["name"]
+            bucket = source_replica.bucket
+            for dest_replica in Config.get_replication_destinations(source_replica):
+                if exists(dest_replica, source_key):
```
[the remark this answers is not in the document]
kislyuk: It is difficult. More than half of those lines are parameterized differently. I don't consider them worth refactoring.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+        - Otherwise, configures the state machine to run the threadpool (copy blob parts) and closer (compose the copy).
+    - Otherwise, check if all referenced entities are already copied to the destination:
+        - For a file manifest, check that the blob is there.
+        - For a bundle manifest, check that file manifests for all files in the bundle are there.
+        - For a collection manifest, check that all collection contents are there.
+        If the checks fail, cause the state machine to sleep for 8 seconds, then try again.
+        If the checks succeed, do a one-shot copy of the manifest.
+    """
+    if event["source_key"].startswith("blobs"):
+        if int(event["source_obj_metadata"]["size"]) > AWS_MIN_CHUNK_SIZE:
```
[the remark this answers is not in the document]
kislyuk: The AWS oneshot copy threshold is lower. It would be a logic fault to not trigger the multipart copy code above the AWS oneshot copy threshold, but below the GCP oneshot copy threshold.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+def oneshot_copy(event, context):
+    source_replica = Replica[event["source_replica"]]
+    dest_replica = Replica[event["dest_replica"]]
+    logger.info(f"Begin transfer of {event['source_key']} from {source_replica} to {dest_replica}")
+    do_oneshot_copy(source_replica=source_replica, dest_replica=dest_replica, source_key=event["source_key"])
+    logger.info(f"Finished transfer of {event['source_key']} from {source_replica} to {dest_replica}")
+
+@app.step_function_task(state_name="ComposeParts", state_machine_definition=sfn)
+def compose_parts(events, context):
+    # Since this is a SFN task handler that sits downstream of a Parallel task, it receives an array of outputs.
```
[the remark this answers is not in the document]
kislyuk: I can see how this can be confusing, but neither replacement is quite right. I've replaced "output" with "task output" to hopefully clarify.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+                continue
+            if part_index <= event.get("last_completed_part", -1):
+                continue
+
+            part = dict(id=part_index + 1, start=part_start, end=min(object_size - 1, part_start + part_size - 1))
+
+            logger.info(log_msg.format(part=part, **event))
+            if dest_replica.storage_schema == "s3":
+                # TODO: (s3) Check if the part was completed but not marked as last copied.
+                upload_url = "{host}/{bucket}/{key}?partNumber={part_num}&uploadId={mpu_id}".format(
```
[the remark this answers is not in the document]
kislyuk: I don't think that should be done in this PR.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+                    part_num=part["id"],
+                    mpu_id=event["mpu_id"]
+                )
+            elif dest_replica.storage_schema == "gs":
+                dest_blob_name = "{}.part{}".format(event["source_key"], part_index + 1)
+                dest_blob = gs.get_bucket(dest_replica.bucket).blob(dest_blob_name)
+                if dest_blob.exists():
+                    continue
+
+                upload_url = dest_blob.create_resumable_upload_session(size=part["end"] - part["start"] + 1)
```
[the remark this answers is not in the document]
kislyuk: ditto

dss/config.py
```diff
@@ -429,6 +429,17 @@ def get_elasticsearch_port(cls):
     def get_elasticsearch_timeout(cls):
         return int(os.environ.get('DSS_ES_TIMEOUT', "10"))
 
+    @classmethod
+    def get_replication_graph(cls):
```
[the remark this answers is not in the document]
kislyuk: Because in the future we will want more complex logic for selecting the destinations to live here.

dss/events/handlers/sync.py
```diff
@@ -263,12 +165,105 @@ def compose_upload(msg):
...
-
-def range_request(url, start, end):
+def range_request(url: str, start: int, end: int):
     return http.request("GET", url, preload_content=False, headers=dict(Range=f"bytes={start}-{end}"))
+
+def get_part_size(object_size, dest_replica):
+    if dest_replica.storage_schema == "s3":
+        part_size = get_s3_multipart_chunk_size(object_size)
+    else:
+        part_size = 640 * 1024 * 1024
```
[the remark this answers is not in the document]
kislyuk: No, it's not. Good point, replaced.

tests/test_sync.py
```diff
@@ -11,75 +11,95 @@
...
+from dss.storage.identifiers import FILE_PREFIX, BUNDLE_PREFIX, COLLECTION_PREFIX, FileFQID, BundleFQID, CollectionFQID
+from dss.storage.hcablobstore import BundleFileMetadata, BundleMetadata, FileMetadata, compose_blob_key
+from tests import eventually, get_version, get_collection_fqid
 from tests.infra import testmode
 
-
 def setUpModule():
     configure_test_logging()
 
+class TestSync:
```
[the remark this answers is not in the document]
kislyuk: Replaced.

tests/test_sync.py
kislyuk: Added.

tests/test_sync.py
kislyuk: Added subtests where requested.

tests/test_sync.py
kislyuk: In both these cases, I have moved the metadata check up to immediately follow the previous test clauses and not use the other operations as a delay. The tests passed as a result. So at this point I consider this workaround unnecessary.

I think the practice of using event sequencing this way for testing is not bad, as long as it's properly documented (as it is here) and is (exceedingly) unlikely to cause race conditions (as it is here). For me it certainly works better than splitting tests into multiple files (increasing the code size, splitting topically related code, and making it harder to understand).

kislyuk:
@Bento007, thanks for your review. The refactor of sync daemon code into blobstore was not the focus of this PR. There are several abstraction level mismatches between what the sync daemon does and what the blobstore code does that make it unsuitable for integration without major modification to both codebases. So I consider that to be outside the scope of this PR as well.

kislyuk:
I had to restart the CI test suite 3 times on the most recent commit due to a repeated failure of the `tests/test_indexer.py DSS_UNITTEST_OPTS="-v TestGCPIndexer"` stage. At this point we have seen this stage fail unrelated changes too often, and taken several shots at fixing it to no avail. I would suggest that we remove `tests/test_indexer.py` from Travis tests, and run it manually or separately on a schedule, so it doesn't cause us to lose confidence in our tests.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+            obj = bucket.Object(unquote(event_record["s3"]["object"]["key"]))
+            if bucket.name != source_replica.bucket:
+                logger.error("Received S3 event for bucket %s with no configured replica", bucket.name)
+                continue
+
+            for dest_replica in Config.get_replication_destinations(source_replica):
+                if exists(dest_replica, obj.key):
+                    logger.error("Key %s already exists in %s, skipping sync", obj.key, dest_replica)
+                    continue
+                exec_name = bucket.name + "/" + obj.key + ":" + source_replica.name + ":" + dest_replica.name
```
xbrianh: It may be useful to include the object type (`blob`, `file`, `bundle`, `collection`) in the `sfn` execution name.

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+        message = json.loads(event_record["body"])
+        if message["selfLink"].startswith("https://www.googleapis.com/storage"):
+            source_replica = Replica.gcp
+            source_key = message["name"]
+            bucket = source_replica.bucket
+            for dest_replica in Config.get_replication_destinations(source_replica):
+                if exists(dest_replica, source_key):
+                    logger.info("Key %s already exists in %s, skipping sync", source_key, dest_replica)
+                    continue
+                exec_name = bucket + "/" + message["name"] + ":" + source_replica.name + ":" + dest_replica.name
```
xbrianh: and here :)

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+        "MaxAttempts": 3,
+        "BackoffRate": 1.5
+    },
+    {
+        "ErrorEquals": ["States.Permissions"],
+        "MaxAttempts": 0
+    }
+]
+
+sfn = {
```
xbrianh: The chart (gif) is extremely helpful for understanding `sfn` logic. So hooray for including it :)

daemons/dss-sync-sfn/app.py
```diff
@@ -0,0 +1,306 @@
...
+                mpu_id = initiate_multipart_upload(source_replica=Replica[event["source_replica"]],
+                                                   dest_replica=Replica[event["dest_replica"]],
+                                                   source_key=event["source_key"])
+                new_event.update(mpu_id=mpu_id)
+            return new_event
+        else:
+            return dict(event, sleep=False, do_oneshot_copy=True, do_multipart_copy=False, done=False)
+    return dict(event, sleep=False, do_oneshot_copy=False, do_multipart_copy=False, done=False)
+
+def copy_parts(event, context):
```
xbrianh: There is logical and conceptual ovrerlap here with `dss.stepfunctions.s3copyclient` and `dss.stepfunctions.gscopyclient`. Multipart copy is a fundamental operation that I feel should be defined in one place.

This is just echoing @Bento007's comments, which have already received responses.

kislyuk: The copy clients in dss.stepfunctions copy data by reference, while the sync daemon copies parts by value. Because of this fundamental design requirement, the codebases were engineered differently. While there is otherwise some conceptual overlap, the sync daemon uses low level streaming APIs to copy the data, and the organizational differences in the codebases are unfortunately quite hard to reconcile.

As I mentioned in my responses to @Bento007, I don't believe refactoring this codebase to conform to other portions of the DSS code should be the focus of this PR. This PR does the work necessary to port sync daemon to step functions. We can separately address these functional commonality concerns between `dss.stepfunctions.*copyclient`, `blobstore` and this codebase later.

daemons/dss-sync-sfn/app.py
kislyuk: Will fix, thanks.

daemons/dss-sync-sfn/app.py
kislyuk: Will fix, thanks.

kislyuk:
SQS queues are deployed idempotently by domovoi in the course of the deployment command.

In the future we may want to integrate that step more closely with Terraform or otherwise make it more externally managed (for example via SAM templates that are now supported by Chalice), but that will require some careful conceptual thinking and probably modifications to Domovoi (e.g. to support the aforementioned SAM templates).

daemons/dss-sync-sfn/app.py
kislyuk: Thanks. I wanted to include an svg (in the AWS Console, this is rendered as an svg), but Github broke inline svg rendering 😢

daemons/dss-sync-sfn/app.py
kislyuk: The key name starts with `blobs/`, `files/`, `bundles/` or `collections/` already, so the execution name already includes this information.

The execution name only gets printed in the trampoline lambda log output though, it doesn't get passed into the state functions API as the execution name. To pass it in, I'll need to make a change to Domovoi. Filed that for myself as a task. I'd prefer to do that separately from this PR though.

kislyuk: Created https://github.com/kislyuk/domovoi/issues/16

kislyuk:
I think `scripts/create_config_aws_event_relay_user.py` may need to be run for every environment.

PR #1562, opened 2018-09-07
Retrieving user email from custom JWT claim

Bento007:
### Test plan
updated how test credential are generated

The hca-cli will also need to be updated


dss/config.py
```diff
@@ -371,6 +371,10 @@ def get_audience():
     def get_OIDC_group_claim():
         return Config._get_required_envvar("OIDC_GROUP_CLAIM")
 
+    @staticmethod
+    def get_OIDC_email_claim():
+        return Config._get_required_envvar("OIDC_EMAIL_CLAIM")
```
kislyuk: You're requiring the env var to be present here, but you have an "or" clause in all the places where this function is used. What are the circumstances in which the env var is present but evaluates to false? (Empty string?) Is it more appropriate to have a failsafe in case the env var is not set?

PR #1566, opened 2018-09-11
adding 422 when metadata checksums are missing for put_file

Bento007:
### Test plan
- need to add a test case.

### Release notes
- A 422 is returned when the file metadata is missing checksums.

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -205,11 +205,17 @@ def put(uuid: str, json_request_body: dict, version: str):
     size = handle.get_size(src_bucket, src_key)
     content_type = handle.get_content_type(src_bucket, src_key)
 
-    # format all the checksums so they're lower-case.
-    for metadata_spec in HCABlobStore.MANDATORY_STAGING_METADATA.values():
-        if metadata_spec['downcase']:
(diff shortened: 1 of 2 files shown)


dss/api/files.py
```diff
@@ -205,11 +205,17 @@ class CopyMode(Enum):
...
-    for metadata_spec in HCABlobStore.MANDATORY_STAGING_METADATA.values():
-        if metadata_spec['downcase']:
-            keyname = typing.cast(str, metadata_spec['keyname'])
-            metadata[keyname] = metadata[keyname].lower()
+    try:
+        # format all the checksums so they're lower-case.
+        for metadata_spec in HCABlobStore.MANDATORY_STAGING_METADATA.values():
+            if metadata_spec['downcase']:
+                keyname = typing.cast(str, metadata_spec['keyname'])
+                metadata[keyname] = metadata[keyname].lower()
```
xbrianh: It may be preferable to narrow the scope of the `try: except` clause to just this line.

mweiden:
Merging with @Bento007's permission

dss/api/files.py
Bento007: sure. I'll fix it in another PR.

PR #1569, opened 2018-09-12
Custom email claim fix and tests

Bento007:
* Fixes an error that prevent users from making authorized request with user credentials. Does not affect service credentials.
* Moved common logic for retrieving user email to a function in security. 
* Added tests to confirm the behavior of that function. 

### Test plan
added specific test cases for custom email claims.

None

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -180,8 +180,10 @@ class TestSecurity(unittest.TestCase):
                 self.assertEqual(security.get_token_email(param), result)
 
         with self.subTest("missing claim"):
-            with self.assertRaises(DSSException):
+            with self.assertRaises(DSSException) as ex:
                 security.get_token_email({})
+            self.assertEqual(ex.exception.status, 401)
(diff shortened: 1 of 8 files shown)


tests/test_utils.py
```diff
@@ -160,6 +160,33 @@ def test_negative_verify_jwt(self):
...
+            with self.subTest(f"no custom claim {param}"):
+                self.assertEqual(security.get_token_email(param), result)
+
+        os.environ['OIDC_EMAIL_CLAIM'] = 'TEST_CLAIM'
+        for param, result in tests:
+            with self.subTest(f"custom claim {param}"):
+                self.assertEqual(security.get_token_email(param), result)
+
+        with self.subTest("missing claim"):
+            with self.assertRaises(DSSException):
```
kislyuk: Can you test more specifically here that it's a 401 exception?

Bento007: Will do

PR #1572, opened 2018-09-14
Optimize get bundle to serve existing checkout bundles whenever possible

### LINKED ISSUE #1367: Checkout should not overwrite duplicate objects, rather refresh their TTLs
## User Story 1
* As a user of the DCP consumer API
* I want the DSS to refresh TTLs on duplicate objects in check out buckets rather than overwriting them
* so that time to export is reduced for "cached" objects.

## ~User Story 2~
* ~As an operator of the DSS~
* ~I want to refresh TTLs on duplicate objects in check out buckets rather than overwriting them~
* ~so that the cost of exporting objects is reduced.~

## Definition of Done
- [ ] Check out requests citing duplicate bundles, files, and bucket combinations does not result in a new copy operation
- [ ] The TTLs on the duplicate objects stored in the bucket are refreshed

calvinnhieu:
These changes follow https://github.com/HumanCellAtlas/data-store/pull/1522. With these changes, if a [stale](https://github.com/HumanCellAtlas/data-store/pull/1516) bundle is accessed, a checkout job will be kicked off but the existing checkout bundle will be returned immediately such that the client does not have to wait for the checkout job to complete, and the bundle's TTL will be refreshed.

### Test plan
- Added integration tests for new stale bundle behavior
- Manual testing against local DSS API server

- Upgraded `cloud-blobstore` to 3.1.0

### Release notes
- Optimized GET bundle to return existing checkout bundles when possible

Fixed #1367

### DIFF
--- a/dss/storage/checkout/bundle.py
+++ b/dss/storage/checkout/bundle.py
@@ -74,19 +74,23 @@ def verify_checkout(
         bundle_version: typing.Optional[str],
         token: typing.Optional[str]
 ) -> typing.Tuple[str, bool]:
+    """
+    Ensures that for a specified bundle either:
+    a) a checkout exists
+    b) a job processing a bundle checkout exists
 
-    handle = Config.get_blobstore_handle(replica)
-    prefix = get_dst_bundle_prefix(bundle_uuid, bundle_version)
-    bundle_metadata = get_bundle_manifest(bundle_uuid, replica, bundle_version)
+    This function will first check if a valid checkout exists. Else, it will
+    create or use a supplied checkout job token to process a bundle checkout job.
 
-    expected_files = [prefix + '/' + file['name'] for file in bundle_metadata['files']]
-    blobs_in_checkout = list(handle.list_v2(replica.checkout_bucket, prefix))
-    stale_before_date = (datetime.datetime.now(datetime.timezone.utc) -
-                         datetime.timedelta(days=int(os.environ['DSS_BLOB_PUBLIC_TTL_DAYS'])))
-
-    if (all((key in expected_files) and
-            (blob[BlobMetadataField.LAST_MODIFIED] > stale_before_date) for key, blob in blobs_in_checkout) and
-            len(blobs_in_checkout) == len(expected_files)):
+    :param replica: Cloud replica
+    :param bundle_uuid: Bundle UUID
+    :param bundle_version: Bundle version
+    :param token: Token used to track the status of a checkout job.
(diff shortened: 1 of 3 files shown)


tests/test_bundle.py
```diff
@@ -266,35 +269,64 @@ def _test_bundle_get_checkout(self, replica: Replica, test_fixtures_bucket: str)
                   .add_query("version", bundle_version)
                   .add_query("presignedurls", "true"))
 
+        @eventually(10, 2)
+        def test_creation_dates_updated():
```
kislyuk: Is this a standalone or integration test? If I don't put a decorator on it, does it get run as standalone by default?

calvinnhieu: Neither - it's a helper function defined inside `_test_bundle_get_checkout`

calvinnhieu:
@kislyuk I see. I'll convert my integration test to a standalone test then

tests/test_bundle.py
kislyuk: Oh! sorry didn't notice that.

Usually I try to call such helpers `assert_blah` instead of `test_blah`.

dss/storage/checkout/bundle.py
```diff
@@ -116,6 +120,70 @@ def verify_checkout(
...
+        bundle_uuid: str,
+        bundle_version: typing.Optional[str],
+) -> bool:
+    """
+    Validates the contents of a checkout bundle against its bundle manifest.
+    Ensures the checkout bundle is not near expiration to avoid serving a missing bundle.
+    :param replica: Cloud replica
+    :param bundle_uuid: Bundle UUID
+    :param bundle_version: Bundle version
+    :return: Is checkout bundle valid
```
kislyuk: Consider more precise wording: "True if checkout bundle is valid, False otherwise".

dss/storage/checkout/bundle.py
```diff
@@ -116,6 +120,70 @@ def verify_checkout(
...
+        bundle_uuid: str,
+        bundle_version: typing.Optional[str],
+) -> bool:
+    """
+    Checks if any objects in a checkout bundle are stale and should be refreshed.
+    A bundle is stale if it is past half of its Bucket Lifecycle Policy TTL.
+    :param replica: Cloud replica
+    :param bundle_uuid: Bundle UUID
+    :param bundle_version: Bundle version
+    :return: Is checkout bundle stale
```
kislyuk: Same here.

PR #1574, opened 2018-09-14
Add logging and retries to gs relay daemon

kislyuk:

### DIFF
--- a/daemons/Makefile
+++ b/daemons/Makefile
@@ -59,5 +59,6 @@ import-test:
             ./package_daemon.sh $$daemon; \
             python $$daemon/app.py; \
         done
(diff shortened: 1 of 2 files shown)


daemons/Makefile
```diff
@@ -59,5 +59,6 @@ import-test:
             ./package_daemon.sh $$daemon; \
             python $$daemon/app.py; \
         done
+	python dss-gs-event-relay/main.py
```
xbrianh: What happens here?

kislyuk: This is a lint test. It's just a sanity check to see if the code in the module compiles and imports without triggering any errors.

PR #1578, opened 2018-09-17
Replace all storage buckets

### LINKED ISSUE #1518: Fix GS bucket regions
GS buckets have heterogenous regions and storage classes that the infra deploy system must accommodate. Buckets should be migrated to a consistent region.

dev bucket: Regional US-CENTRAL1
staging bucket: Multi-Regional  US
test bucket: Regional US-EAST1

xbrianh:
* All GS buckets are re-configured for "MULTI_REGION US", which requires either migration or replacement. We opt for replacement.
* For consistency and data integrity, replace all S3 buckets as well.
* `prod` data will be copied into the new buckets. For the other environments, we plan to start over with empty buckets (pending the OK from the other teams).
* Replacement will be performed on `dev` first, and later, during promotion, for the other environments.

Closes #1518


xbrianh:
Possibilities I considered:
* `org-hca-dss-even-dev` and `org-hca-dss-odd-dev`, since there will be binary swapping if we do this again
* `org-hca-dss-{ascending_int}-dev`
* `org-hca-dss-{creation_date}-dev`
* `org-hca-dss-{uuid}-dev`

PR #1591, opened 2018-09-19
Use the HCA DCP common Terraform backend bucket

xbrianh:
* Move TF state bucket to common HCA DCP bucket
* Put state files int s3://{bucket}/dss/{component}.state instead of s3://{bucket}/dss-{component}.state
* Make bucket account-id parameterization explicit


kislyuk:
Not sure if I understand the reasoning behind hardcoding the account id here. Why not leave it parameterized?

xbrianh:
Agree, leaving it parameterized.

PR #1596, opened 2018-09-24
Remove `local_root` from cloud_uploader

### LINKED ISSUE #1520: Get rid of local_root for the uploaders in tests/fixtures/cloud_uploader.py
They don't serve much of a purpose, and make the code much more confusing.

xbrianh:
Fixtures are only uploaded from `$DSS_HOME/tests/fixtures/datafiles`, so there's no point in parameterizing their location.

Closes #1520


xbrianh:
Turns out `local_root` is used more extensively than I realized.

PR #1602, opened 2018-09-25
[easy] Fix scripts/release.sh --help

kislyuk:

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -4,37 +4,39 @@ set -euo pipefail
 
 # This block discovers the command line flags `--force` and  `--no-deploy`,
 # and passes on positional arguments as $1, $2, etc.
-FORCE=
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -35,7 +35,6 @@ case $key in
     ;;
 esac
 done
-set -- "${POSITIONAL[@]}" # restore positional parameters
 
 if [[ $# != 2 ]]; then
```
xbrianh: With this change, `$#` evaluates to 0 here.

xbrianh:
* Running `scripts/release.sh` without args should show help.
* These should also display help: `scripts/release.sh --help`, `scripts/release.sh blargh`, scripts/release.sh master --force`
* Anything that does not involve two positional arguments should display help.

PR #1605, opened 2018-09-25
Fixing the errors returned by Auth failures

Bento007:
- updated the flask response
- update the swagger doc with unauthorized error
- updated test cases to verify error message returned
- added additional error messages when decoding JWT

### Test plan -->
checking the error response in test cases.


dss/error.py
```diff
@@ -37,12 +38,12 @@ def dss_exception_handler(e: DSSException) -> FlaskResponse:
         status=e.status,
         mimetype="application/problem+json",
         content_type="application/problem+json",
-        response={
+        response=json.dumps({
```
kislyuk: Why is this necessary? How did it work before?

PR #1608, opened 2018-09-26
[WIP] Adding debug log messages to subscriptions

Bento007:


dss/api/subscriptions.py
```diff
@@ -62,6 +63,8 @@ def find(replica: str):
         **{k: v for k, v in hit.to_dict().items() if not k.startswith('hmac_')}}
         for hit in search.scan()]
 
+    logger.debug(f"{len(responses)} Subscriptions found")
```
xbrianh: Maybe put the uuid here, too

Bento007: changing to info

PR #1609, opened 2018-09-26
restrict access to put bundle and files

Bento007:


Bento007:
Still finishing up the unit tests.

tests/test_subscriptions.py
```diff
@@ -14,6 +14,8 @@
 import connexion.apis.abstract
 import requests
 
+from tests.auth_tests import TestAuthMixin
```
xbrianh: Is `TestAuthMixin` used anywhere else?

Bento007: currently the tests are broken, I'm working on fixing them.

Bento007: it is now used many places

Issue #1623, opened 2018-10-03
Lambdas seem to have unnecessary, sweeping Get*/List* permissions.


DailyDreaming:
These were done for X-ray, however, this should only be necessary for EC2, not lambdas.

Documentation here: https://docs.aws.amazon.com/xray/latest/devguide/xray-permissions.html
AmazonS3ReadOnlyAccess **(Amazon EC2 only)** – Gives the instance permission to download the X-Ray daemon from Amazon S3.

Will be investigating this more and then updating.

Bento007:
#1631 resolves this

PR #1624, opened 2018-10-03
Restrict unnecessary lambda permissions.

DailyDreaming:
Connected to #1623 .

All lambdas were given robust s3 Get*/List* permissions for X-ray I believe, however I believe this is only needed for EC2 instances.  We're trying to tighten down privileges for our data store on the commons side and wanted to submit a PR.  Also to double-check and make sure it passes tests on the HCA side.

Documentation here: https://docs.aws.amazon.com/xray/latest/devguide/xray-permissions.html

> AmazonS3ReadOnlyAccess **(Amazon EC2 only)** – Gives the instance permission to download the X-Ray daemon from Amazon S3.

### DIFF
--- a/iam/policy-templates/dss-admin-lambda.json
+++ b/iam/policy-templates/dss-admin-lambda.json
@@ -35,8 +35,6 @@
       "Effect": "Allow",
       "Action": [
           "xray:PutTelemetryRecords",
-          "s3:Get*",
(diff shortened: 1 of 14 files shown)


DailyDreaming:
Tests failed, meaning I clearly didn't test this enough.  Will investigate.

PR #1637, opened 2018-10-09
Raise sync exception when s3/gs presigned urls fail to resolve

### LINKED ISSUE #1636: Oneshot copy sync s3->gs failure
32 mb fastq blob on `s3` synced to `gs` with content
```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Key>blobs/14c5b872e5b92ab7f941a09bb2a163401b6e6659bf62af17f06efbe79b6f47ed.dd95e19024c57bcfe7f68437596aa4fd0bc71541.363db3638fde9721b00cbf2132170201.1ee32d9b</Key><RequestId>80E081909C90BCF1</RequestId><HostId>VSvTMXyD0nofZ1woCDID2slYRM9lUAkdPrDQuQOmRQmqsiQgLnOK6mR2+bqtUVa28E5uCYnXBJE=</HostId></Error>
```

xbrianh:
This will cause the sync sfn to retry oneshot copies when presigned urls fail to resolve.

fixes #1636

### DIFF
--- a/dss/events/handlers/sync.py
+++ b/dss/events/handlers/sync.py
@@ -62,6 +62,10 @@ def sync_s3_to_gs_oneshot(source: BlobLocation, dest: BlobLocation):
         ExpiresIn=presigned_url_lifetime_seconds
     )
     with closing(http.request("GET", s3_blob_url, preload_content=False)) as fh:
+        if 200 != fh.status:
+            msg = f"failed to resolve s3 presigned url for {source.blob.key}"
+            logger.error(msg)
+            raise Exception(msg)
         gs_blob = dest.bucket.blob(source.blob.key, chunk_size=1024 * 1024)
         gs_blob.metadata = source.blob.metadata
         gs_blob.upload_from_file(fh)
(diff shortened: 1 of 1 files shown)


dss/events/handlers/sync.py
```diff
@@ -62,6 +62,10 @@ def sync_s3_to_gs_oneshot(source: BlobLocation, dest: BlobLocation):
         ExpiresIn=presigned_url_lifetime_seconds
     )
     with closing(http.request("GET", s3_blob_url, preload_content=False)) as fh:
+        if 200 != fh.status:
+            msg = f"failed to resolve s3 presigned url for {source.blob.key}"
+            logger.error(msg)
+            raise Exception(msg)
```
Bento007: We need a more specific exception. How about `DSSException`?

Bento007: look at https://github.com/HumanCellAtlas/data-store/blob/master/dss/error.py for ideas

kislyuk:
I have an idea why @xbrianh.

Listing the lambdas, it looks like we didn't destroy `dss-sync-integration` when deploying the SFN sync daemon to integration.

It must be destroyed, otherwise it will sync stuff on its own and interfere with the consistency guarantee.

This highlights the need for a consistent place to put notes to the operator. In this case I should have added notes that said, "after deployment, destroy X Y and Z for the old daemon"

PR #1646, opened 2018-10-18
Updating lambda IAM policies to denying delete on DSS_S3_BUCKET

Bento007:
This is a preventative measure to reduce the number of resource with permissions to perform destructive operations on the data store.

These same protections do not exists for GCP. We will need to create read and write only google service account for lambdas to use.

None


Bento007:
Closing this ticket, because using a bucket resource policy is a better option

PR #1650, opened 2018-10-19
Limits s3 deletion to DSS_AWS_DELETION_ROLE

Bento007:


kislyuk:
How does this integrate with the rest of the DSS deployment process? What do DSS daemons get out of being able to introspect the name of the privileged role?

In addition, this role should ideally be externally managed. In other words, it should be applied from outside the account (if we can find a good way to do that).

Instead of having this role inside DSS, consider saving this policy and role name in https://github.com/HumanCellAtlas/dcp-infra, and deploying it from there. To get the most out of this strategy, only `hca-id` account admins should be able to deploy/change this policy.

Bento007:
> What do DSS daemons get out of being able to introspect the name of the privileged role?

The `DSS_AWS_DELETION_ROLE` variable is not provided to the DSS daemons. It is only used when deploying the bucket.

kislyuk:
I see, thanks. That makes sense. Still, consider making this externally managed.

Bento007:
this PR was replaced by HumanCellAtlas/dcp-infra#68. closing

Issue #1648, opened 2018-10-19
`POST /search?output_format=raw` broken on staging


hannes-ucsc:
```
$ http "https://dss.staging.data.humancellatlas.org/v1/search?output_format=raw&replica=aws&per_page=100" es_query:={}
HTTP/1.1 502 Bad Gateway
Connection: keep-alive
Content-Length: 36
Content-Type: application/json
Date: Fri, 19 Oct 2018 03:18:45 GMT
x-amz-apigw-id: O_j66Ec3IAMFqTw=
x-amzn-RequestId: ae646ec4-d34d-11e8-9411-43b0356feac9

{
    "message": "Internal server error"
}
```

`per_page=10` works, interestingly.

xbrianh:
`dss-staging` is returning more data than API Gateway can handle:
"""(578440cb-d366-11e8-aaa0-b118a972298b) Lambda execution failed with status 200 due to customer function error: body size is too long. Lambda request id: 57852b5c-d366-11e8-a31b-a3a83baffc12""""

Bento007:
We should add a note, that when outputting raw page size may need to be adjusted, or we could add a better error message for the user? Maybe both?

Bento007:
create this ticket for future work #1675

PR #1652, opened 2018-10-22
Raw Search Page Limit Documented

Bento007:
Partial fix for #1648

### Release notes
- Documented limitation when using search with out format raw.


Bento007:
I think that is great idea but given our time limit and current requirements this would take too much time. How about if output is raw we limit the size of the return results to 10?

Bento007:
create this ticket for future work https://github.com/HumanCellAtlas/data-store/issues/1675

PR #1657, opened 2018-10-23
Update requirements.txt

### LINKED ISSUE #1639: Upgrade parmico to >= 2.4.2
Security vulnerability alert says that 2.4.1 is vulnerable

jnkrupp:
### Update requirements.txt
Upgrade dependencies excluding elasticsearch and elasticsearch-dsl. The goal is to run integration tests and shake out any bugs that might result from it.

Fixes #1639


Bento007:
what happened to requirements-dev.txt

kozbo:
@jnkrupp  please update this ticket

jnkrupp:
It turns out that `make refresh_all_requirements` was failing because the Makefile was creating a Python 2.7 virtual environment instead of 3.6. I'm working on fixing this, so that the packages are installed correctly and requirements-dev.txt doesn't end up empty.

Issue #1616, opened 2018-09-28
Async file upload needs to expose state better


kislyuk:
When an async PUT file fails, the reason for the error is completely opaque. If the copy op fails, the error needs to be forwarded through the API front end.

kislyuk:
cc @ttung

Bento007:
Proposed solution:

- Put file should return 201 when file is copied sync.
- Put file should return 202 when file is copied async.

- rename the step functions to use the file name and version

Get file responses:
- for a file copied return the file link.
- for a file being copied return retry later.
- for a file that failed to be copied return error response from step function.
- for a file this does not exist return file missing.

PR #1674, opened 2018-10-26
Send exceptions to fail state

xbrianh:
Catch gs copy client errors to an SFN fail task.

This will allow us to propagate stack traces to end users.

Connected to #1616


dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -65,6 +69,14 @@ def run(self) -> dict:
     return CopyWorkerTimedThread(lambda_context.get_remaining_time_in_millis() / 1000, event).start()
 
 
+def fail(event, lambda_context):
+    # Error and cause are available as:
+    # event['Error']
+    # event['Cause']
+    logger.error(f"Error: failed gscopyclient sfn execution on {event}")
```
Bento007: you don't need to include error in the message.

PR #1683, opened 2018-10-31
Refreshing requirements

Bento007:


brainstorm:
Might be interesting getting this merged in since I'm getting some warning(s) on a clean install:

`elasticsearch-dsl 6.2.1 has requirement elasticsearch<7.0.0,>=6.0.0, but you'll have elasticsearch 5.5.3 which is incompatible.`

Bento007:
that wont change since we are still using elasticsearch 5.5. Though we haven't encounter any compatibility issues yet.

PR #1696, opened 2018-11-09
Fixups for (#1691)

xbrianh:
A few missed adaptations to use SQS instead of SNS


daemons/dss-index/app.py
```diff
@@ -33,16 +33,18 @@ def dispatch_s3_indexer_event(event, context) -> None:
     if event.get("Event") == "s3:TestEvent":
         logger.info("DSS index daemon received S3 test event")
     else:
-        _handle_event(Replica.aws, event, context)
+        for event_record in event["Records"]:
```
Bento007: Why was this change needed?

Bento007: will indexer receive multiple records to process in one request?

xbrianh: SQS->Lambda sends messages in batches:
https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

Bento007: From https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html
> All messages in a failed batch return to the queue, so your function code must be able to process the same message multiple times without side effects.

Will this be a problem?

xbrianh: Not as long as the indexer is idempotent, which I believe it is.

Also, if needed, batch size may be configured.

Bento007: adding idempotent test for indexer #1697

Bento007: Fixes the failing tests and i'll review.

Issue #1702, opened 2018-11-12
Environment variable `DSS_VERSION` is not set


mweiden:
## Software version (git hash, date, branch)

* Dev 582c3d40
* Integration 56da699c
* Staging 56da699c

## Expected behavior

`GET /version` endpoint returns a git SHA1 prefix or version tag

## Actual behavior

```json
{
    "version_info": {
        "version": null
    }
}
```

## Steps to reproduce the behavior

Example: https://dss.dev.data.humancellatlas.org/version

mweiden:
Unfortunately, this seems to still be a problem.

xbrianh:
Fix from @mweiden worked, but `scripts/populate_lambda_ssm_parameters.py` needed to be executed. This should be done each time `environment` or `environment.$DSS_DEPOYMENT_STAGE` is updated.

PR #1703, opened 2018-11-12
Export `DSS_VERSION` environment variable to lambdas

mweiden:
Export `DSS_VERSION` to lambda.

Fixes https://github.com/HumanCellAtlas/data-store/issues/1702


mweiden:
@xbrianh @Bento007 I think this should do it 🤞

PR #1705, opened 2018-11-13
Store and recover chunks of state in dynamodb

xbrianh:
This will make it possible to catch copy sfn errors and expose them to the api.

It may also be useful as a lightweight caching mechanism.

related to #1616

### DIFF
--- a/dss/util/async_state.py
+++ b/dss/util/async_state.py
@@ -24,7 +24,7 @@ class AsyncStateItem:
 
     def __init__(self, key: str, body: dict) -> None:
         self.key = key
-        if not body.get('_type', None):
+        if not body.get('_type'):
             body['_type'] = type(self).__name__
         self.body = body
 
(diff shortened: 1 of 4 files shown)


dss/util/async_state.py
```diff
@@ -0,0 +1,112 @@
...
+        bar = AsyncStateItem.get("test_key")
+        type(foo) == MyAsyncSubclass  # True
+        type(bar) == MyAsyncSubclass  # True
+    """
+
+    table = f"dss-async-state-{os.environ['DSS_DEPLOYMENT_STAGE']}"
+
+    def __init__(self, key: str, body: dict) -> None:
+        self.key = key
+        if not body.get('_type', None):
```
Bento007: isn't the default None so you don't need to explicitly say `None`

xbrianh: yep, fixed

Issue #1689, opened 2018-11-07
Required GS test bucket permissions are not documented in README.md?


brainstorm:
Perhaps related with issue https://github.com/HumanCellAtlas/data-store/issues/1538?

Definitely related with issue #1356.

## Actual behavior
```
$ tests/fixtures/populate.py --s3-bucket $DSS_S3_BUCKET_TEST_FIXTURES --gs-bucket $DSS_GS_BUCKET_TEST_FIXTURES
Traceback (most recent call last):
  File "tests/fixtures/populate.py", line 20, in <module>
    populate(args.s3_bucket, args.gs_bucket)
  File "/home/ubuntu/data-store/tests/fixtures/populate_lib.py", line 190, in populate
    upload(uploader)
  File "/home/ubuntu/data-store/tests/fixtures/populate_lib.py", line 10, in upload
    uploader.reset()
  File "/home/ubuntu/data-store/tests/fixtures/cloud_uploader.py", line 128, in reset
    for blob in self.bucket.list_blobs():
  File "/home/ubuntu/hca-venv/lib/python3.7/site-packages/google/api_core/page_iterator.py", line 199, in _items_iter
    for page in self._page_iter(increment=False):
  File "/home/ubuntu/hca-venv/lib/python3.7/site-packages/google/api_core/page_iterator.py", line 230, in _page_iter
    page = self._next_page()
  File "/home/ubuntu/hca-venv/lib/python3.7/site-packages/google/api_core/page_iterator.py", line 349, in _next_page
    response = self._get_next_page_response()
  File "/home/ubuntu/hca-venv/lib/python3.7/site-packages/google/api_core/page_iterator.py", line 401, in _get_next_page_response
    query_params=params)
  File "/home/ubuntu/hca-venv/lib/python3.7/site-packages/google/cloud/_http.py", line 293, in api_request
    raise exceptions.from_http_response(response)
google.api_core.exceptions.Forbidden: 403 GET https://www.googleapis.com/storage/v1/b/org-umccr-hca-dss-test-fixtures/o?projection=noAcl: umccr-org-hca-store@hca-data-store-221707.iam.gserviceaccount.com does not have storage.objects.list access to org-umccr-hca-dss-test-fixtures.
```

## Steps to reproduce the behavior

[...]

PR #1711, opened 2018-11-14
README.md fixes after rearrangement of sections

brainstorm:
Address https://github.com/HumanCellAtlas/data-store/issues/1689#issuecomment-438520842 and `environment.local` vs `environment` possible confusion?

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,9 +173,6 @@ Run `./dss-api` in the top-level `data-store` directory to deploy the DSS API on
 
 ### Deploying the DSS
 
-Assuming the tests have passed above, the next step is to manually deploy. See the section below for information on
-CI/CD with Travis if continuous deployment is your goal.
-
 Several components in the DSS deployed separately as daemons, found in `$DSS_HOME/daemons`. Daemon deployment may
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -173,9 +173,6 @@ Run `./dss-api` in the top-level `data-store` directory to deploy the DSS API on
 
 ### Deploying the DSS
 
-Assuming the tests have passed above, the next step is to manually deploy. See the section below for information on
-CI/CD with Travis if continuous deployment is your goal.
-
```
xbrianh: Do you think the Travis section should be removed?

brainstorm: Oh, sorry about that, it was just the **ordering** of the sections that was off... 1) deploy, 2) passing tests, not the other way around.

xbrianh: The intention is that some tests should pass locally before deploying. Generally we have two types of tests:

- "standalone tests" which should not depend on deployed components
- "integration tests" which depend upon deployed components

It's reasonable to expect standalone tests to pass before attempting deployment.

xbrianh:
Closing this, but see:
https://github.com/HumanCellAtlas/data-store/issues/1966

PR #1725, opened 2018-11-15
updating patch collection to deduplicate contents

Bento007:
added code to test deduplication in collection patch and put.

### DIFF
--- a/tests/test_collections.py
+++ b/tests/test_collections.py
@@ -71,8 +71,22 @@ class TestCollections(unittest.TestCase, DSSAssertMixin, DSSUploadMixin):
 
     def test_put(self):
         "PUT new collection"
-        uuid, _ = self._put(self.contents)
-        self.addCleanup(self._delete_collection, uuid)
+        with self.subTest("with unique contents"):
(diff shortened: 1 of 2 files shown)


tests/test_collections.py
```diff
@@ -111,11 +125,19 @@ def test_patch(self):
...
-                              dict(description="foo", name="cn"),
-                              dict(description="bar", details={1: 2}),
-                              dict(add_contents=contents),
-                              dict(remove_contents=contents)]:
+        expected_contents = {'contents': [col_file_item,
+                                          col_ptr_item],
+                             'description': 'd',
+                             'details': {},
+                             'name': 'n',
+                             'owner': 'travis-test@human-cell-atlas-travis-test.iam.gserviceaccount.com'
```
kislyuk: This should not be hardcoded here. I think it's OK if the test would just not assume anything about the owner.

kislyuk:
Thanks @Bento007. To answer your earlier question, I intended to dedup input in both cases, so this looks like a bug in my code.

PR #1746, opened 2018-11-27
Configurable swagger security.

DailyDreaming:
Includes a class to change the auth in the swagger yaml based on a config specified by the user.

No change was made to the content (though the tags shifted position slightly) in the current yml, default is still the HCA security auth in place.

Users can run the script to modify the security currently in place.  No other parts of the swagger yaml are affected, so from an hca point of view, this whole feature can be safely ignored.

Added auth headers to insecure test calls so that tests can be run successfully on a secure system if the swagger auth is changed to auth everything.

Added a check in travis to ensure that the hca swagger auth is kept at its current default auth settings if pushed back.


mikebaumann:
Thanks for the progress on this, @DailyDreaming.
I think for both philosophical/principle and practical reasons, it may be better to start with the assumption that all APIs require auth, and auth can be waived for selected APIs by configuring them to be open. I think starting with all APIs by auth is consistent with the DSS being a secure (or at least highly securable) system.
From a practical perspective, almost all the HCA APIs require auth, and all the APIs for systems storing controlled access data need to be auth-ed, so having auth be the default and open being the configurable exception makes sense to me.

mikebaumann:
This comment may be beyond the scope of this specific PR, but it looks like an aspect related to configurable security that would need to be addressed is the `@security.authorized_group_required(['hca'])` decorators on specific APIs.

DailyDreaming:
@mikebaumann Yes, I'd just found that too.  Seems like it'll need to be reconfigured if anyone other than hca wants to use the dss.

tests/test_modify_swagger_auth.py
```diff
@@ -0,0 +1,88 @@
...
+            self.original_contents = f.readlines()
+
+    def tearDown(self):
+        # restore the original swagger contents
+        with open(os.path.join(pkg_root, 'dss-api.yml'), 'w') as f:
+            for line in self.original_contents:
+                f.write(line)
+
+    def test_auth_can_be_determined_from_swagger(self):
+        # assert that after modifying the swagger file to require auth on all endpoints,
```
Bento007: wrap this doc string triple quotes

auth.hca_default.json
```diff
@@ -0,0 +1,25 @@
+{
```
Bento007: remove hca from file name

scripts/swagger_auth.py
```diff
@@ -0,0 +1,152 @@
...
+            # If properly indented and we're in the correct section, this will be a call request.
+            elif line.startswith('    ') and not line.startswith('     ') and \
+                    path_section and line.strip().endswith(':'):
+                request_section = line.strip()[:-1]
+                if ignore_auth:
+                    security_endpoints[call_section].append(request_section)
+    return security_endpoints
+
+
+def ensure_hca_defaults_are_still_set():
```
Bento007: remove references to HCA

auth.default.json
```diff
@@ -0,0 +1,25 @@
+{
```
xbrianh: It seems unlikely that that these json files will be used outside the scope of `scripts/swagger_auth.py`,

Please consider in-lining this file and `auth.secure_all.json` into `scripts/swagger_auth.py`.

tests/test_modify_swagger_auth.py
```diff
@@ -8,27 +8,30 @@
...
             self.original_contents = f.readlines()
 
     def tearDown(self):
         # restore the original swagger contents
-        with open(os.path.join(pkg_root, 'dss-api.yml'), 'w') as f:
+        with open(self.swagger_path, 'w') as f:
             for line in self.original_contents:
                 f.write(line)
 
     def test_auth_can_be_determined_from_swagger(self):
```
Bento007: where do you test for an empty security config file, a security config file with non-existant paths, and a security config with duplicates?

Bento007:
can you please make a ticket for the additional work we discussed yesterday about adding it to the deploy and warning about some endpoints requiring security to function.

auth.default.json
Bento007: The idea is to allow customization on what endpoints to secure. Are you suggesting we don't allow users this level of customization?

DailyDreaming: We should talk about this on Monday!

environment
```diff
@@ -101,6 +101,8 @@ GOOGLE_APPLICATION_SECRETS_SECRETS_NAME="application_secrets.json"
 ES_ALLOWED_SOURCE_IP_SECRETS_NAME="es_source_ip"
 DSS_DEBUG=0
 
+AUTH_CONFIG_FILE="auth.default.json"
```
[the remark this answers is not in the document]
DailyDreaming: Thanks for catching that!  I hadn't seen that particular readme yet.

scripts/swagger_auth.py
```diff
@@ -0,0 +1,152 @@
...
+
+def determine_auth_configuration_from_swagger(ignore_auth=False):
+    """If ignore_auth is True, instead return all endpoints, regardless of auth."""
+    path_section = False  # bool flag to notify if we're in the section containing the API call definitions
+    call_section = None  # an api endpoint, e.g.: /subscription, /file/{uuid}, etc.
+    request_section = None  # a request call, e.g.: get, put, delete, etc.
+    security_endpoints = defaultdict(list)
+    with open(os.path.join(pkg_root, 'dss-api.yml'), 'r') as f:
+        for line in f:
+            # If not indented at all, we're in a new section, so reset.
```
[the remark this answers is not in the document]
DailyDreaming: Done.

tests/test_modify_swagger_auth.py
DailyDreaming: Done.

auth.hca_default.json
DailyDreaming: Done.

scripts/swagger_auth.py
DailyDreaming: Done.

DailyDreaming:
@Bento007 The PR include the warning about some endpoints requiring security to function.  As we discussed it should have more work done, though we should talk with Brian if it should be a part of the deploy rather than optional.

DailyDreaming:
@xbrianh @Bento007 This should address all concerns, please review!

Note: there's no longer an environment variable, if it's decided that we should add one for this and integrate the whole thing into deploy, it should be a separate PR I think.

Issue #1754, opened 2018-11-29
Add Encryption at rest


kozbo:
## User story
* As a system administrator
* I want to be sure that data is never available in a non-encrypted format
* so that some unauthorized person or program could access the data

## Definition of done
All data is encrypted at rest, for every copy of the data.

PR #1755, opened 2018-11-29
Encrypt S3 buckets at rest, should solve issue #1754 for AWS

brainstorm:
https://github.com/HumanCellAtlas/data-store/issues/1754

### DIFF
--- a/infra/buckets/s3.tf
+++ b/infra/buckets/s3.tf
@@ -1,6 +1,14 @@
 resource aws_s3_bucket dss_s3_bucket {
   count  = "${length(var.DSS_S3_BUCKET) > 0 ? 1 : 0}"
   bucket = "${var.DSS_S3_BUCKET}"
+
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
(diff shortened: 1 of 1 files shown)


Bento007:
For GCP it's on by default. It must be enabled for AWS S3  Thank you for the PR.

brainstorm:
You are welcome, feel free to merge ;)

Bento007:
If we could make this configurable via an environment variable that would be idea. @DailyDreaming would you to take this on?

DailyDreaming:
@Bento007 Sounds good to me, but I think we should talk to Kevin first.

brainstorm:
IMHO, you guys are overengineering the encryption at rest by making it parametrizable... just leave it on by default, why is that not reasonable to have it by design and that's it? It's offered for free, it's completely transparent and doesn't come with any (noticeable) performance impact.

All that is, unless you want to introduce custom/user-managed KMS, which is a different story.

DailyDreaming:
@brainstorm We're aware of the features involved.  We intend to discuss it.

PR #1766, opened 2018-12-04
Tagging terraform deployed resources with Terraform and DSS

Bento007:


infra/elasticsearch/main.tf
```diff
@@ -8,16 +8,28 @@ locals {
 resource "aws_cloudwatch_log_group" "es_index_log" {
   name = "/aws/aes/domains/${var.DSS_ES_DOMAIN}/es-index-${var.DSS_DEPLOYMENT_STAGE}-logs"
   retention_in_days = 90
+  tags {
+    CreatedBy = "Terraform"
+    Application = "DSS"
```
xbrianh: Downstream users may wish to change these values. Consider placing them in an env var.

Bento007: Created #1777 for future work.

PR #1767, opened 2018-12-05
sanitation checks for POST bundle checkout

amarjandu:

### DIFF
--- a/dss/api/bundles/checkout.py
+++ b/dss/api/bundles/checkout.py
@@ -14,7 +14,8 @@ def post(uuid: str, json_request_body: dict, replica: str, version: str = None):
     assert replica is not None
     _replica: Replica = Replica[replica]
     dst_bucket = json_request_body.get('destination', _replica.checkout_bucket)
-
+    if '/' in dst_bucket:
+        raise DSSException(400, "illegal_arguments", "Destination bucket invalid!")
     try:
         execution_id = start_bundle_checkout(
             _replica,
@@ -25,7 +26,6 @@ def post(uuid: str, json_request_body: dict, replica: str, version: str = None):
         )
     except BundleNotFoundError:
         raise DSSException(404, "not_found", "Cannot find bundle!")
-
     return jsonify(dict(checkout_job_id=execution_id)), requests.codes.ok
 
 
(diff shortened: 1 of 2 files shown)


dss/storage/checkout/__init__.py
```diff
@@ -34,11 +34,18 @@ def validate_file_dst(replica: Replica, dst_bucket: str, dst_key: str):
 
 def pre_exec_validate(
         replica: Replica, dss_bucket: str, dst_bucket: str, bundle_uuid: str, bundle_version: str) -> bool:
+    validate_dst_name(dst_bucket)
     validate_dst_bucket(replica, dst_bucket)
     validate_bundle_exists(replica, dss_bucket, bundle_uuid, bundle_version)
     return True
 
 
+def validate_dst_name(dst_bucket: str)-> bool:
```
xbrianh: The usage of `DestinationBucketNotFoundError` to propagate this error is misleading. The destination did not meet the expected format, so the bucket could not be discovered.

Consider performing this check in `dss/api/bundles/checkout.py` so that `DestinationBucketNotFoundError` is not needed.

dss/storage/checkout/__init__.py
```diff
@@ -34,11 +34,18 @@ def validate_file_dst(replica: Replica, dst_bucket: str, dst_key: str):
...
+    if len(dst_bucket.split(sep='/')) != 1:
```
xbrianh: This is equivalent to `if "/" in dst_bucket:`

tests/infra/server.py
```diff
@@ -74,7 +74,7 @@ def _inject_api_requests_methods(cls):
     def shutdown(self):
         if self._server is not None:
             self._server.server.shutdown()
-        self.join(timeout=30)
+        self.join(timeout=500)
```
xbrianh: What's the motivation for increasing this timeout?

dss/storage/checkout/__init__.py
amarjandu: changed to recommendation

tests/infra/server.py
amarjandu: Was left in by mistake for more debugging time, reverted.

dss/storage/checkout/__init__.py
amarjandu: I had placed the function within this file because it has other functions that are used within testing for dst_bucket validation. I have a work around, but I think the testing approach would have to be different from the other bucket tests that are in place.

amarjandu: check moved to dss/api/bundles/checkout.py

PR #1772, opened 2018-12-06
Adding user-agent to dispatch log messaged for auditing

Bento007:
### Release notes
Adding user-agent to dispatch log messaged for auditing

@mweiden will this break the monitor in graphana?


Bento007:
Not needed at this time

PR #1780, opened 2018-12-10
Fix Unwritable bucket

Bento007:


infra/buckets/s3.tf
```diff
@@ -92,7 +92,7 @@ resource aws_s3_bucket dss_s3_checkout_bucket_unwritable {
     },
     {
       "Action": [
-        "s3:Put*"
+        "s3:PutO*"
```
xbrianh: Please explain this in a comment

Bento007: Fixed

PR #1786, opened 2018-12-11
Remove url from GitLab environments

xbrianh:
Variable created in GitLab yml (outside of `variables`) are not interpolated, causing GitLabs environment mechanism to fail silently.

issue: https://gitlab.com/gitlab-com/support-forum/issues/2814
potential fix: https://gitlab.com/gitlab-org/gitlab-ce/issues/27921

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -24,11 +24,8 @@ before_script:
   - source ~/venv/bin/activate
   - pip install -r requirements-dev.txt
   - source environment
-  - if [[ master == $CI_COMMIT_REF_NAME ]]; then
-  -   DSS_ENV="dev"
-  - elif [[ -f "environment.$CI_COMMIT_REF_NAME" ]]; then
-  -   DSS_ENV=$CI_COMMIT_REF_NAME
-  -   source environment.$DSS_ENV
+  - if [[ -f "environment.$CI_COMMIT_REF_NAME" ]]; then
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -99,8 +96,14 @@ deploy:
...
-    name: $DSS_ENV
-    url: https://dss.$DSS_ENV.data.humancellatlas.org
+    name: $CI_COMMIT_REF_NAME
+    # TODO: incluude url when GitLab fixes it's wonky interpolation rules
+    # Variable created in this file (outside of `variables`) are not interpolated,
+    # causing GitLabs environment mechanism to fail silently.
+    # issue: https://gitlab.com/gitlab-com/support-forum/issues/2814
+    # fix: https://gitlab.com/gitlab-org/gitlab-ce/issues/27921
+    # - Brian Hannafiouus
+    # url: https://dss.$DSS_ENV.data.humancellatlas.org
```
DailyDreaming: The `$DSS_ENV` reference in the comment url was deleted and should be `$CI_COMMIT_REF_NAME` now.

xbrianh: Actually it should not be `CI_COMMIT_REF_NAME`, which evaluates to `master` instead of `dev`. `DSS_ENV` existed to overcome this limitation.

Anyway, comment fixed.

DailyDreaming: Okay, that makes sense.  Looks good now!

Issue #1792, opened 2018-12-12
Direct file download fails in prod with no error message


theathorn:
Go to the Files tab and click on the download icon for the first file in the list (e.g. AB-HE0202A-CZI-day0-DroNc_S4_R2_001.fastq.gz). The icon spins for a while and then turns to a red "x" with no error message.

┆Issue Number: HCAB-336

theathorn:
Still getting the CORS error (for any file) when attempting a single file download from the Data Browser in the Prod deploy only (dev, integration and staging are OK).

PR #1793, opened 2018-12-14
Set CORS policy on checkout bucket

mweiden:
Fixes https://github.com/HumanCellAtlas/data-store/issues/1792


DailyDreaming:
@mweiden I looked a little further into this and I think this policy looks appropriate as is.  I can't think of a reason we'd need to restrict the origins or headers and the paths look appropriate.

mweiden:
@DailyDreaming OK. Thanks for looking into this, Lon. Go ahead and merge and deploy when you're ready. Ask if you need help doing this.

DailyDreaming:
@mweiden I need to be added to the Google account to access those buckets.  Could you help set me up?

PR #1814, opened 2019-01-03
Parameterize SSM store name in environment

xbrianh:
This is needed for non-DCP deployments to use their own SSM parameter store.


amarjandu:
README for ENV values needs to be updated as well.

environment
```diff
@@ -95,6 +95,7 @@ DSS_CERTIFICATE_VALIDATION="EMAIL"
 DSS_ZONE_NAME="dev.data.humancellatlas.org."
 PYTHONWARNINGS=ignore:ResourceWarning,ignore::UserWarning:zipfile:
 DSS_SECRETS_STORE="dcp/dss"
+DSS_PARAMETER_STORE="dcp/dss"
```
amarjandu: README for ENV values needs to be updated as well.

xbrianh: fixed

PR #1817, opened 2019-01-03
Populating lambda SSM warns on undefined vars

xbrianh:
When population AWS Systems Manager parameters, warn instead of fail on undefined environment variables.

This allows variables such as DSS_VERSION, which is set during deployment, to be undefined when populating SSM values.

Make some readability improvements.

depends on #1814

### DIFF
--- a/scripts/populate_lambda_ssm_parameters.py
+++ b/scripts/populate_lambda_ssm_parameters.py
@@ -19,16 +19,17 @@ es_client = boto3.client("es")
 lambda_client = boto3.client("lambda")
 
 def get_local_lambda_environment():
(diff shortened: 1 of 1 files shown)


scripts/populate_lambda_ssm_parameters.py
```diff
@@ -62,6 +63,11 @@ def get_deployed_lambdas():
         except lambda_client.exceptions.ResourceNotFoundException:
             print(f"{name} not deployed, or does not deploy a Lambda function")
 
+def get_elasticsearch_endpoint():
```
Bento007: what is this change related to?

PR #1821, opened 2019-01-06
Add the current bucket policy into the terraform script.

DailyDreaming:
Not sure exactly why terraform would complain about this now, but this does fix the current deployment error by adding in the policy present on the buckets into the terrrafom script.

I may have missed something though, @xbrianh , could you take a look?


DailyDreaming:
Ah, so the tf file to set the bucket and to set the bucket's policy are not only separate files but separate repos.  Thanks for letting me know!

PR #1834, opened 2019-01-14
Use terraform to deploy custom domain natively

xbrianh:
Don't manage certificate via DSS terraform. If it is desirable to manage this certificate via terraform at all, it should be via dcp-infra.


infra/domain/main.tf
```diff
@@ -4,39 +4,25 @@ data aws_route53_zone selected {
...
-#   Terraform does not currently support regional API Gateway endpoints:
-#   https://github.com/terraform-providers/terraform-provider-aws/issues/2195
-resource null_resource dss_domain {
-  count  = "${length(var.DSS_CERTIFICATE_VALIDATION ) > 0 ? 1 : 0}"
-  provisioner "local-exec" {
-    command = "./create_regional_domain_name.py --domain-name ${var.API_DOMAIN_NAME} --certificate-arn ${aws_acm_certificate.cert.arn} --zone-id ${data.aws_route53_zone.selected.zone_id}"
-  }
-  depends_on = ["aws_acm_certificate_validation.cert_dns", "aws_acm_certificate_validation.cert_email"]
+resource "aws_route53_record" "dss" {
+  zone_id = "${data.aws_route53_zone.selected.zone_id}"
```
DailyDreaming: Should this be `regional_zone_id`?

xbrianh: nope, there is no such parameter for the `aws_route53_record` resource: https://www.terraform.io/docs/providers/aws/r/route53_record.html

PR #1838, opened 2019-01-16
Adding negative test cases to deleted from replicas

Bento007:
### Test plan
- Negative test cases have been added to verify that we are unable to delete from the main replica.

### DIFF
--- a/tests/test_gshcablobstore.py
+++ b/tests/test_gshcablobstore.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python
 # coding: utf-8
-
+import google
+import io
 import os
 import sys
 import unittest
 
+from cloud_blobstore import BlobNotFoundError
 from cloud_blobstore.gs import GSBlobStore
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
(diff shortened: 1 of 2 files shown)


tests/test_gshcablobstore.py
```diff
@@ -17,6 +19,7 @@
 
 class TestGSHCABlobStore(unittest.TestCase, HCABlobStoreTests):
     def setUp(self):
```
xbrianh: I'm a little concerned that these tests could become flaky. Let's upload `Deletion_test.txt` here to make sure it always exists.

infra/gcp_service_account/main.tf
```diff
@@ -20,12 +20,6 @@ resource "google_project_iam_member" "cloudruntimeconfiguratoradmin" {
   member  = "serviceAccount:${google_service_account.dss.email}"
 }
 
-resource "google_project_iam_member" "storageadmin" {
-  project = "${data.google_project.project.project_id}"
-  role    = "roles/storage.admin"
-  member  = "serviceAccount:${google_service_account.dss.email}"
-}
-
```
xbrianh: Consider separating `resource "google_project_iam_member" "storageadmin"` removal into it's own PR to make it prominent in the repo history.

Bento007:
the credential used to run the tests have deletion permissions. We need to create a service account that simulated the production service credentials permissions and use those to run the deletion test.

xbrianh:
What's the status of this PR?

Bento007:
unchanged

Issue #1667, opened 2018-10-25
Enable Audit logging for dss-api


Bento007:
## User story
* As an auditor
* I want to know the source IP, user-agent, route, time, response time, input and output bytes, status code, and any other essential request/response information
* so that I may follow requests through the system.

## Definition of done
enable audit logging for dss-api and verify that source IP, user-agent, route, time, response time, input and output bytes, status code, and any other essential request/response information are present in the audit logs for requests.

Bento007:
I've applied it manually using the console but It would be better to apply it using terraform. It will need to be applied after chalice deploys the lambda.

Bento007:
The data store for the humancellatlas already has audit logs setup for AWS. I will instead provide tools and documentions on how to setup audit logging for the datastore.

Bento007:
This is where the AWS audit log is specified in HCA. https://github.com/HumanCellAtlas/logs/blob/master/infrastructure/infra-aws.tf#L14

Bento007:
https://docs.google.com/document/d/1fnQINoch_baf-hseCOKXW4OWcM-TadMCwfpLWwRRhJ4/edit?ts=5c09b3d6

Bento007:
Work is on hold until GDPR and NIST requirements are finalized

Bento007:
Done

PR #1839, opened 2019-01-16
Adding Source IP and Decoded JWT to logs

Bento007:
### Release notes
- the user email, ip address, token issuer, token issue date, token expiration date, token issuer, and token claims will be logged for all requests.

Related to #1667

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -141,11 +141,11 @@ def get_chalice_app(flask_app) -> DSSChaliceApp:
         content_length = app.current_request.headers.get('content-length')
         user_agent = app.current_request.headers.get('user-agent')
         app.log.info(
-            """[request] "%s %s", \{"source_ip":"%s", "content_length":"%s", "user-agent":"%s"\}, %s""",
+            """[request] "%s %s" %s %s "%s" %s""",
             method,
             path,
             source_ip,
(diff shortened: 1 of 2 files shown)


chalice/app.py
```diff
@@ -137,6 +137,18 @@ def dispatch(*args, **kwargs):
         method = app.current_request.method
         query_params = app.current_request.query_params
         req_body = app.current_request.raw_body if app.current_request._body is not None else None
+        source_ip = app.current_request.context['identity']['sourceIp']
+        content_length = app.current_request.headers.get('content-length')
+        user_agent = app.current_request.headers.get('user-agent')
+        app.log.info(
+            """[request] "%s %s", \{"source_ip":"%s", "content_length":"%s", "user-agent":"%s"\}, %s""",
```
xbrianh: Can you explain this message format?

kislyuk: I'm curious too. Is this documented somewhere?

mweiden: This would be a concatenation of [common log format](https://en.wikipedia.org/wiki/Common_Log_Format) with a JSON object. This is not a standard format. I would recommend either 1) going full JSON or 2) following something standard like [combined log format](https://httpd.apache.org/docs/1.3/logs.html#combined).

PR #1852, opened 2019-01-22
[WIP] Fixing notify to send the tombstone doc when a bundle is deleted

Bento007:
Before this the notifier was sending the document that was tombstones, which result in an error when the users attempted to retrieve the doc.

### Release notes


Bento007:
duplicate of #1851

PR #1854, opened 2019-01-23
notify tombstone

Bento007:
This ensures that notifications are sent to a subscriber_id if the tombstoned bundle matches their query or if the subscriber is subscribed to tombstones.

### DIFF
--- a/tests/sample_search_queries.py
+++ b/tests/sample_search_queries.py
@@ -20,3 +20,17 @@ smartseq2_paired_ends_vx_query = \
             }
         }
     }
+
+tombstone_query = {
+    "query": {
+        'bool': {
+            'must': [
+                {
+                    'match': {
+                        "admin_deleted": True
+                    }
+                }
+            ]
+        }
+    }
+}
--- a/tests/sample_search_queries.py
+++ b/tests/sample_search_queries.py
@@ -26,7 +26,7 @@ tombstone_query = {
         'bool': {
             'must': [
                 {
-                    'match': {
+                    'term': {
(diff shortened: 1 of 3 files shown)


dss/index/es/backend.py
```diff
@@ -53,12 +53,17 @@ def remove_bundle(self, bundle: Bundle, tombstone: Tombstone):
...
         modified, index_name = doc.entomb(tombstone_doc, dryrun=self.dryrun)
         if self.notify or modified and self.notify is None:
-            self._notify(tombstone_doc, index_name)
+            self._notify_tombstone(doc, tombstone_doc, index_name)
 
     def _notify(self, bundle, index_name):
         subscription_ids = self._find_matching_subscriptions(bundle, index_name)
         self._notify_subscribers(bundle, subscription_ids)
 
+    def _notify_tombstone(self, bundle, bundle_tombstone, index_name):
```
hannes-ucsc:
```suggestion
    def _notify_tombstone(self, doc, tombstone_doc, index_name):
```

You might disagree with a naming convention but deviating from a convention only adds to the chaos. You'd have to refactor the codebase to your new naming convention.

tests/sample_search_queries.py
```diff
@@ -20,3 +20,17 @@
             }
         }
     }
+
+tombstone_query = {
+    "query": {
+        'bool': {
+            'must': [
+                {
+                    'match': {
```
hannes-ucsc:
```suggestion
                    'term': {
```

PR #1857, opened 2019-01-24
Remove "always" test mode

xbrianh:
We should not run some tests twice per CI/CD pipeline.


xbrianh:
Indexer tests have logic based on test mode. Closing for now.

PR #1862, opened 2019-01-25
Search when Version is None

Bento007:
in cases where manifest.version is not specified in the ES tombstone document we would still like to page through the results.

related to #1846 

### Test plan 
added test cases to page through tombstones with manifest.version and without and verified that multiple pages of results can be returned.

None

### Release notes
None

### DIFF
--- a/dss/api/search.py
+++ b/dss/api/search.py
@@ -99,24 +99,23 @@ def _es_search_page(es_query: dict,
         es_query['_source'] = False
 
     # https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-search-after.html
-    sort = [
-        "manifest.version:desc",
-        "uuid:desc"]
+    es_query['sort'] = [
+        {"uuid": {"order": "desc"}},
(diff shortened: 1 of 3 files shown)


dss/api/search.py
```diff
@@ -99,24 +99,23 @@ def _es_search_page(es_query: dict,
         es_query['_source'] = False
 
     # https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-search-after.html
-    sort = [
-        "manifest.version:desc",
-        "uuid:desc"]
+    es_query['sort'] = [
+        {"uuid": {"order": "desc"}},
+        {"manifest.version": {"missing": "last", "order": "desc"}}
```
xbrianh: Just so I can catch up on the context, can you explain the consequences of "missing"?

Bento007: if the field is missing the value will be "last" by default. The value of last will put it at the end of search results. This default value is not stored and only used for determining sort order, and for determining the point to search after.

PR #1868, opened 2019-01-30
L2 Health Checks

amarjandu:
Implements internal health checks for the DSS


daemons/dss-canary/app.py
```diff
@@ -0,0 +1,43 @@
...
+logger = logging.getLogger(__name__)
+configure_lambda_logging()
+
+app = domovoi.Domovoi(configure_logs=False)
+delete_timer = os.environ['DSS_CANARY_TIMER']
+create_timer = str(int(os.environ['DSS_CANARY_TIMER']) + 1)
+canaryKey = 'canary/health'
+
+
+@app.scheduled_function('rate({} minute)'.format(delete_timer))
```
xbrianh: Instead of creating a new daemon, consider using the GCP Cloud Schedule, which can inject messages into the dss-gs-event-relay pub/sub topic. This will not require the creation of a canary file: https://cloud.google.com/scheduler/docs/

amarjandu: @xbrianh i ended up removing the Canary and just keeping the check to specifically the Event Relay.

health
```diff
@@ -0,0 +1 @@
+{"Healthy": false, "canary": {"ElapsedTimeMinutes": 1405.9830637166667, "LastModified": "Thu, 31 Jan 2019 21:56:15 GMT"}}
```
Bento007: What is this used for?

amarjandu: removed,

dss/api/health.py
```diff
@@ -0,0 +1,78 @@
...
+    es_res = es_client.cluster.health()
+    if es_res['status'] == 'green':
+        es_status = True
+    return es_status, es_res
+
+
+def _get_dynamodb_status():
+    db_status = True
+    ddb_tables = list(filter(lambda x: x.startswith("dss") and x.endswith(os.getenv("DSS_DEPLOYMENT_STAGE")),
+                             dynamodb.list_tables()["TableNames"]))
```
xbrianh: The criteria used to find dynamo-db tables is not exclusive to the data-store. For now, perhaps we should use an explicit list.

dss/api/health.py
```diff
@@ -0,0 +1,78 @@
...
+    er_status = False
+    try:
+        er_name = "projects/{}/locations/{}/functions/{}".format(
+            os.environ["GCP_PROJECT_NAME"], os.environ["GCP_DEFAULT_REGION"], os.environ["DSS_EVENT_RELAY_NAME"])
+        service = discovery.build('cloudfunctions', 'v1', cache_discovery=False)
+        er_res = service.projects().locations().functions().get(name=er_name).execute()
+        if er_res['status'] == 'ACTIVE':
+            er_status = True
+            er_data = er_res
+    except Exception:
```
xbrianh: Catch-all try-except blocks may hide exceptions we wish to see.

Is there a narrower exception we can use?

dss/api/health.py
amarjandu: changed to explicit list

chalice/app.py
```diff
@@ -236,10 +237,11 @@ def version():
...
     @app.route("/internal/health")
     @time_limited(app)
-    def health():
+    def health_check(*args, **kwargs):
+        health_status = health.l2_health_checks(**kwargs)
         return chalice.Response(status_code=200,
-                                headers={"Content-Type": "text/html"},
-                                body="OK")
+                                headers={"Content-Type": "application/json"},
+                                body=health_status)
```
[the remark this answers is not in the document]
amarjandu: placed the logging into `dss/api/health.py/l2_health_check()`

amarjandu:
@xbrianh  i changed the response from a failure to what we spoke about earlier in the week, to just have a `{"Healthy" :False}`, and moving the actual failure status into the log. 
Also added changes to have more specific exception failures within the tests.

PR #1886, opened 2019-02-07
Add HTTP request handler preconfigured with timeout and retry policies

kislyuk:
This is needed for bundle scalability work and should be encouraged for use by any library that performs HTTP requests.

### DIFF
--- a/dss/util/networking.py
+++ b/dss/util/networking.py
@@ -1,6 +1,41 @@
-import contextlib
-import socket
+import contextlib, socket, threading, logging
 
+import requests
+from requests.adapters import HTTPAdapter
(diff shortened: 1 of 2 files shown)


kislyuk:
Yes - for now this can serve as a sort of reference implementation. In the future I think DSS might need to replace the Google HTTP client bindings with something very similar to this or building upon this. Also, if DSS needs to contact other HTTP APIs besides S3/GS, this implementation could be used.

kislyuk:
Actually, I guess this should live in https://github.com/HumanCellAtlas/dcplib. I'll back this out and resubmit there.

PR #1908, opened 2019-02-21
removed TEST_USER buckets, added mocking to tests

amarjandu:
PR for removing TEST checkout buckets from cache_flow, added mocking for testing.

### DIFF
--- a/dss/storage/checkout/cache_flow.py
+++ b/dss/storage/checkout/cache_flow.py
@@ -19,8 +19,8 @@ Metadata Caching RFC: https://docs.google.com/document/d/1PQBO5qYUVJFAXFNaMdgxq8
 
 
 def is_dss_bucket(dst_bucket: str):
-    return dst_bucket in (os.environ['DSS_GS_CHECKOUT_BUCKET_TEST'], os.environ['DSS_S3_CHECKOUT_BUCKET_TEST'],
-                          os.environ['DSS_S3_CHECKOUT_BUCKET'], os.environ['DSS_GS_CHECKOUT_BUCKET'])
+    """Lambdas only have access to the standard checkout buckets, should not modify non-DSS buckets"""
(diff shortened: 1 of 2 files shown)


dss/storage/checkout/cache_flow.py
```diff
@@ -19,8 +19,8 @@
 
 
 def is_dss_bucket(dst_bucket: str):
-    return dst_bucket in (os.environ['DSS_GS_CHECKOUT_BUCKET_TEST'], os.environ['DSS_S3_CHECKOUT_BUCKET_TEST'],
-                          os.environ['DSS_S3_CHECKOUT_BUCKET'], os.environ['DSS_GS_CHECKOUT_BUCKET'])
+    """Lambdas only have access to the standard checkout buckets, should not modify non-DSS buckets"""
```
DailyDreaming: This docstring seems out of context (what the function does and why), is this function necessarily directly associated with the lambdas?

tests/test_checkout_caching.py
```diff
@@ -39,24 +40,32 @@ def tearDownClass(cls):
 
     def setUp(self):
         Config.set_config(BucketConfig.TEST)
+        self.aws_checkout_bucket = os.environ['DSS_S3_CHECKOUT_BUCKET']
```
DailyDreaming: Should these be test buckets?

tests/test_checkout_caching.py
```diff
@@ -66,15 +75,15 @@ def test_aws_user_checkout_doesnt_create_tag(self):
         pass
```
DailyDreaming: `pass`?  Must have missed this before.  Probably should be taken out.

tests/test_checkout_caching.py
amarjandu: removed, not needed

tests/test_checkout_caching.py
amarjandu: removed all references in file

DailyDreaming:
LGTM.

PR #1916, opened 2019-02-22
Document subscriptions in swagger

### LINKED ISSUE #1909: Subscriptions are not properly documented anywhere in the DSS API docs
The Swagger docs contain some details about the optional parameters, but there is no overview of the basic subscription/webhook functionality and no example notification delivery payload.

xbrianh:
This documents JMESPath subscriptions, which differ slightly to Elasticsearch subscriptions.

It is our intention to migrate all subscriptions to JMESPath.

fixes #1909

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -51,8 +51,37 @@ info:
 
     ```
 
-    # Special String Formats
+    # Subscrptions
 
+    DSS supports webhook subscriptions for data events like bundle creation and deletion. Webhooks are callbacks to a public
+    HTTPS endpoint provided by your application. When an event matching your subscription occurs, DSS will send a push
+    notification (via an HTTPS POST or PUT request), giving your application an up-to-date stream of system activity.
+
+    Subscriptions are delivered with the payload format
+
+    ```
+
+    {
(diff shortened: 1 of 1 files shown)


dss-api.yml
```diff
@@ -51,8 +51,37 @@ info:
...
+    notification (via an HTTPS POST or PUT request), giving your application an up-to-date stream of system activity.
+
+    Subscriptions are delivered with the payload format
+
+    ```
 
+    {
+      'transaction_id': {uuid},
+      'subscription_id': {uuid},
+      'event_type': "CREATE"|"TOMBSTONE"|"DELETE",
```
rexwangcc: This is a good point, as far as I can tell, the current analysis listener doesn't distinguish these types 🤔 Maybe this is why we had some duplicated workflows recently? @samanehsan

dss-api.yml
```diff
@@ -51,8 +51,37 @@ info:
 
     ```
 
-    # Special String Formats
+    # Subscrptions
```
DailyDreaming: Just a spelling nit: `Subscrptions`

PR #1917, opened 2019-02-25
[WIP] Increase parallelization of PATCH /collections

xbrianh:
The primary bottle neck in PATCH /collections is collection verification. This can be partially overcome by increasing the parallelization of verification logic.

Via testing from local API server on EC2 instance, we find:
- Lambda memory needs to be scaled to 704MB to take advantage of concurrency. Note that Lambda CPU scales automatically with memory.
- collections can grow to millions of items via the aws replica
- gcp not as scalable due to cross-cloud network calls

### DIFF
--- a/chalice/.chalice/config.json
+++ b/chalice/.chalice/config.json
@@ -13,5 +13,5 @@
     }
   },
   "lambda_timeout": 300,
-  "lambda_memory_size": 256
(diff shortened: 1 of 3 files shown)


xbrianh:
resolved by #1995

PR #1926, opened 2019-02-28
Change regex for PUT file/{uuid}.

DailyDreaming:
Going back to the original regex I had which should fix things.  Also added one of the breaking paths to the testing.


DailyDreaming:
Thanks!

PR #1961, opened 2019-03-08
Update swagger with explicit retry-after headers.

DailyDreaming:
This PR explicitly adds Retry-After headers to [500, 502, 503, 504] responses for all methods except `POST /bundles/{uuid}/checkout`.

This allows the client to know which codes to retry on, even if it is a POST request (currently there is a desire to retry on POST search, see: https://github.com/HumanCellAtlas/dcp-cli/issues/222 ).

*Current CLI behavior:*
Do not retry on POST methods.  Retry on 500, 502, 503, & 504 for all non-POST methods.

*Updating the swagger would:*
Allow changes in the `dcp-cli` to retry on POST search but not `POST /bundles/{uuid}/checkout`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ meeting notes, and [this Zenhub board](https://app.zenhub.com/workspace/o/humanc
 
 ## Overview
 
-The DSS is a replicated data storage system designed for hosting large sets of scientific experimental data on 
+The DSS is a replicated data storage system designed for hosting large sets of scientific experimental data on
 [Amazon S3](https://aws.amazon.com/s3/) and [Google Storage](https://cloud.google.com/storage/). The DSS exposes an API
 for interacting with the data and is built using [Chalice](https://github.com/aws/chalice),
 [API Gateway](https://aws.amazon.com/api-gateway/) and [AWS Lambda](https://aws.amazon.com/lambda/). The API also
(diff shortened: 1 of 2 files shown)


DailyDreaming:
@xbrianh @kislyuk Is this a better solution than https://github.com/HumanCellAtlas/dcp-cli/pull/267 ?

I'd want to follow this up with deleting this line (setting `status_forcelist=set()`): https://github.com/HumanCellAtlas/dcp-cli/blob/master/hca/util/__init__.py#L207

And changing `retry_after_status_codes={301}` to `retry_after_status_codes={301, 500, 502, 503, 504}` here: https://github.com/HumanCellAtlas/dcp-cli/blob/master/hca/util/__init__.py#L118

DailyDreaming:
@xbrianh @kislyuk Is this a better solution than https://github.com/HumanCellAtlas/dcp-cli/pull/267 ?

I'd want to follow this up with deleting this line (setting `status_forcelist=set()`): https://github.com/HumanCellAtlas/dcp-cli/blob/master/hca/util/__init__.py#L207

And changing `retry_after_status_codes={301}` to `retry_after_status_codes={301, 500, 502, 503, 504}` here: https://github.com/HumanCellAtlas/dcp-cli/blob/master/hca/util/__init__.py#L118

kislyuk:
This is better, but is there a way to de-duplicate these definitions so they can be defined once and referenced using `$ref` or something like that?

As a last resort, you could use a YAML anchor+reference, although I prefer not to use those.

DailyDreaming:
@kislyuk I'll look into it, thank you.

DailyDreaming:
@kislyuk Should be addressed.

DailyDreaming:
@kislyuk How's this?

DailyDreaming:
@kislyuk Changed.  Thanks for taking a look!

PR #1967, opened 2019-03-11
Add script to ensure consistent credentials.

DailyDreaming:
Addresses #1800.

I don't have (and prefer not having XD) access to `prod` so this doesn't test that, though I could add it.

Second, I wasn't sure if I should add a unit test for this.  I set it to run whenever anyone does a `make deploy`.  Feedback would be appreciated!

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,7 +30,6 @@ before_script:
   - scripts/fetch_secret.sh application_secrets.json > application_secrets.json
   - scripts/fetch_secret.sh gcp-credentials.json > gcp-credentials.json
   - export GOOGLE_APPLICATION_CREDENTIALS=$(pwd -P)/gcp-credentials.json
-  - scripts/check_deployment_secrets.py
 
 unit_tests:
   stage: test
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,6 +30,7 @@ before_script:
   - scripts/fetch_secret.sh application_secrets.json > application_secrets.json
   - scripts/fetch_secret.sh gcp-credentials.json > gcp-credentials.json
   - export GOOGLE_APPLICATION_CREDENTIALS=$(pwd -P)/gcp-credentials.json
+  - scripts/check_deployment_secrets.py
 
 unit_tests:
   stage: test
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,8 @@ smoketest:
 scaletest:
 	./tests/scalability/scale_test_runner.py -r 10 -d 30
 
+force-deploy: deploy-chalice deploy-daemons
+
 deploy: check-secrets deploy-chalice deploy-daemons
 
 deploy-chalice:
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ smoketest:
 scaletest:
 	./tests/scalability/scale_test_runner.py -r 10 -d 30
 
-deploy: check_secrets deploy-chalice deploy-daemons
+deploy: check-secrets deploy-chalice deploy-daemons
 
 deploy-chalice:
 	$(MAKE) -C chalice deploy
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ smoketest:
 scaletest:
 	./tests/scalability/scale_test_runner.py -r 10 -d 30
 
-deploy: deploy-chalice deploy-daemons
+deploy: check_secrets deploy-chalice deploy-daemons
 
 deploy-chalice:
 	$(MAKE) -C chalice deploy
@@ -78,6 +78,9 @@ plan-infra:
 
 deploy-infra:
 	$(MAKE) -C infra apply-all
+
+check-secrets:
+	scripts/check_deployment_secrets.py
 
 create-github-deployment:
 	$(eval REMOTE=$(shell git remote get-url origin | perl -ne '/([^\/\:]+\/.+?)(\.git)?$$/; print $$1'))
--- a/infra/gcp_service_account/outputs.tf
+++ b/infra/gcp_service_account/outputs.tf
@@ -1,3 +1,0 @@
-output "email" {
-  value = ${google_service_account.dss.email}
-}
--- a/infra/gcp_service_account/outputs.tf
+++ b/infra/gcp_service_account/outputs.tf
@@ -1,0 +1,3 @@
+output "email" {
+  value = ${google_service_account.dss.email}
+}
--- a/scripts/check_deployment_secrets.py
+++ b/scripts/check_deployment_secrets.py
@@ -39,8 +39,7 @@ class SecretsChecker(object):
         self.stage = stage
         self.stages = {'dev': 'environment',
                        'integration': 'enviroment.integration',
-                       'staging': 'environment.staging',
-                       'prod': 'environment.prod'}
+                       'staging': 'environment.staging'}
 
         self.missing_secrets = []
         self.malformed_secrets = []
--- a/scripts/check_deployment_secrets.py
+++ b/scripts/check_deployment_secrets.py
@@ -10,14 +10,10 @@ import os
 import json
 
 
-class SecretFetchingError(Exception):
-    pass
-
-
 class SecretsChecker(object):
     def __init__(self, stage):
         self.stage = stage
-        self.stages = ('dev', 'integration', 'staging')
+        self.stages = ('dev', 'integration', 'staging', 'prod')
         self.service_account = self.fetch_terraform_output("service_account", "gcp_service_account").strip()
 
         self.email = [f'{self.service_account}@human-cell-atlas-travis-test.iam.gserviceaccount.com']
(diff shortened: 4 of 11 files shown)


scripts/check_deployment_secrets.py
```diff
@@ -0,0 +1,61 @@
...
+                         f'The following secret                : {secret}\n'
+                         f'Had the unexpected setting          : {current}\n'
+                         f'When one of these items was expected: {expected}\n')
+
+
+def main(stage=None):
+    if not stage:
+        stage = os.environ['DSS_DEPLOYMENT_STAGE']
+
+    app_secret_name = 'application_secrets.json'
```
amarjandu: `GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME` and `GOOGLE_APPLICATION_SECRETS_SECRETS_NAME` are defined within environment

scripts/check_deployment_secrets.py
```diff
@@ -0,0 +1,61 @@
...
+import subprocess
+import os
+import json
+
+
+auth_uri = ['https://auth.data.humancellatlas.org/authorize', 'https://auth.dev.data.humancellatlas.org/authorize']
+token_uri = ['https://auth.data.humancellatlas.org/oauth/token','https://auth.dev.data.humancellatlas.org/oauth/token']
+dev_email = ['travis-test@human-cell-atlas-travis-test.iam.gserviceaccount.com']
+integration_email = ['org-humancellatlas-integration@human-cell-atlas-travis-test.iam.gserviceaccount.com']
+staging_email = ['org-humancellatlas-staging@human-cell-atlas-travis-test.iam.gserviceaccount.com']
```
xbrianh: These values should be grabbed from our Terraform state file, which can be accomplished via output values.

See:
https://www.terraform.io/docs/configuration/outputs.html

DailyDreaming:
@xbrianh I agree that it should be decided when to run this.  I went ahead and implemented it as a run upon deploy, but I'm not opposed to including it as a part of the canary test if you think that would be valuable.  Otherwise, running it as a unit test could also be an option.

scripts/check_deployment_secrets.py
DailyDreaming: Addressed!  Thanks @amarjandu .

scripts/check_deployment_secrets.py
DailyDreaming: Should be addressed!  Thanks @xbrianh .

infra/gcp_service_account/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "email" {
```
xbrianh: This can be pulled into `main.tf`

Makefile
```diff
@@ -60,7 +60,7 @@ smoketest:
 scaletest:
 	./tests/scalability/scale_test_runner.py -r 10 -d 30
 
-deploy: deploy-chalice deploy-daemons
+deploy: check-secrets deploy-chalice deploy-daemons
```
xbrianh: It may desirable, occasionally, to deploy manually without checking secrets, and it should be possible to do this without hacking build files.

Lets move this change to the deploy stages in `.gitlab-ci.yml` and `.gitlab-prod-ci.yml`. This way it's enshrined in our CI/CD process while preserving our ability to perform manual operations.

tests/test_secrets.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class TestSecretCheck(unittest.TestCase):
+    @needs_terraform
+    def test_dev_secrets(self):
+        """Checks if the secrets in dev are gucci."""
+        s = SecretsChecker(stage='dev')
```
xbrianh: Can `stage` be pulled in from `DSS_DEPLOYMENT_STAGE`? Then we would only need a single `test_secrets` method.

Makefile
DailyDreaming: Sounds good!

tests/test_secrets.py
DailyDreaming: Can do.

Makefile
DailyDreaming: I'd argue that it's not a hack and a necessary dependency if you're deploying.  That said, I think a check in gitlab would probably accomplish the same goal as far as flagging someone to change them back.

tests/test_secrets.py
DailyDreaming: Also a win of deleting my terrible comments.  XD

scripts/check_deployment_secrets.py
```diff
@@ -0,0 +1,156 @@
...
+        self.gcp_cred_secret = self.fetch_secret(self.gcp_cred_secret_name)
+
+    def run_cmd(self, cmd, cwd=os.getcwd(), shell=True):
+        p = subprocess.Popen(cmd,
+                             shell=shell,
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE,
+                             cwd=cwd,
+                             env=self.stage_env)
+        stdout, stderr = p.communicate()
```
Bento007:
```suggestion
        stdout, _ = p.communicate()
```

Bento007: use underscore to indicate the value returned is not needed.

tests/test_secrets.py
```diff
@@ -0,0 +1,48 @@
...
+from tests import needs_terraform
+
+
+logger = logging.getLogger(__name__)
+
+
+class TestSecretCheck(unittest.TestCase):
+    @needs_terraform
+    def test_secrets(self):
+        """Checks if the current stage's secrets are gucci."""
```
Bento007: Also use the "result when condition" format for your test strings.

tests/test_secrets.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        This should not test other stages because we have no way of knowing what
+        those secrets should be, so this should always pass for other stage names.
+        """
+        s = SecretsChecker(stage='somenonsensenamelikeprod')
+        s.run()
+
+    @needs_terraform
+    def test_invalid_secrets(self):
+        """This should raise a ValueError and warn the user since it will check dev against an unqualified email."""
```
Bento007: A ValueError is raised and the user is warned when an unqualified email is stored in a secret.

scripts/check_deployment_secrets.py
```diff
@@ -0,0 +1,156 @@
...
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.PIPE,
+                             cwd=cwd,
+                             env=self.stage_env)
+        stdout, _ = p.communicate()
+        return stdout.decode('utf-8')
+
+    def get_stage_env(self, env_file):
+        dump = '/usr/bin/python -c "import os, json; print(json.dumps(dict(os.environ)))"'
+        cmd = ['/bin/bash', '-c', f'source {env_file} && {dump}']
```
amarjandu: may want to have `#!/usr/bin/env bash` to avoid breaks if `bash` is not in the `/bin/`. Not sure how applicable this is to our user-base.

DailyDreaming: The equivalent to the shebang in a straight cli command would be just using `bash` instead of `/bin/bash` since they both pull the first `bash` command off of the top of the available paths.  It's not a bad point though.  I'll change them from absolute paths.

scripts/check_deployment_secrets.py
```diff
@@ -0,0 +1,155 @@
...
+                self.error_message += f'The following secret seems malformed: {s}\n'
+            raise ValueError(self.error_message)
+
+
+def main(stage=None):
+    if len(sys.argv) > 1:
+        stage = sys.argv[1]
+    elif not stage:
+        stage = os.environ['DSS_DEPLOYMENT_STAGE']
+    s = SecretsChecker(stage)
```
amarjandu: perhaps should ensure that `if stage in STAGES` where `STAGES` are  `['dev','integration','staging','prod']` just to make sure the `DSS_DEPLOYMENT_STAGE` is not set to something arbitrary.

.gitlab-ci.yml
```diff
@@ -30,6 +30,7 @@ before_script:
   - scripts/fetch_secret.sh application_secrets.json > application_secrets.json
   - scripts/fetch_secret.sh gcp-credentials.json > gcp-credentials.json
   - export GOOGLE_APPLICATION_CREDENTIALS=$(pwd -P)/gcp-credentials.json
+  - scripts/check_deployment_secrets.py
```
xbrianh: or here

tests/__init__.py
```diff
@@ -112,3 +113,11 @@ def get_auth_header(real_header=True, authorized=True, group='hca', email=True,
         info = UNAUTHORIZED_GCP_CREDENTIALS
     token = get_service_jwt(info, group, email=email, email_claim=email_claim) if real_header else str(uuid.uuid4())
     return {"Authorization": f"Bearer {token}"}
+
+
+def needs_terraform(test_item):
```
xbrianh: Terraform is a requirement for working with the DSS. Shouldn't these tests fail if it is not installed?

scripts/check_deployment_secrets.py
DailyDreaming: It does this within the class itself.

tests/__init__.py
DailyDreaming: Travis doesn't have terraform.  I'll change it to only skip on travis since that's the only bogus use-case.

Issue #1901, opened 2019-02-19
Scale PUT /bundle to support ~10^6 files (PATCH /bundle)


xbrianh:
- [x] Implement paging for GET /bundle
- [x] Support bundle agglomeration via PATCH
- [ ] Mark intermediate bundles as non-indexable

connects to #1891

PR #1978, opened 2019-03-13
Add support for PATCH /bundle

### LINKED ISSUE #1940: Add bundle /PATCH
This allows large bundle support via agglomeration.

connects to #1901

xbrianh:
This allows bundles to grow to arbitrary size via /PATCH

connects to #1901 
closes #1940

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1594,10 +1594,10 @@ paths:
             properties:
               uuid:
                 type: string
-                description: A RFC4122-compliant ID of this collection.
+                description: A RFC4122-compliant ID of this bundle.
               version:
                 type: string
-                description: New version of the updated collection.
+                description: New version of the updated bundle.
                 format: DSS_VERSION
             required:
               - uuid
@@ -1621,7 +1621,7 @@ paths:
                   - code
         403:
           description: >
-            Returned when credentials presented do not grant access to this collection.
+            Returned when credentials presented do not grant access to this bundle.
           schema:
             allOf:
               - $ref: '#/definitions/Error'
@@ -1632,7 +1632,7 @@ paths:
                     description: >
                       Machine-readable error code.  The types of return values should not be changed lightly.
 
-                      The code `Forbidden` indicates that the user does not have access to this collection.
+                      The code `Forbidden` indicates that the user does not have access to this bundle.
 
                       The code `OAuthResponseProblem` indicates that the OAuth credentials presented were not
                       understood.
@@ -1641,7 +1641,7 @@ paths:
                   - code
         404:
           description: >
-            Returned when the server could not find the collection to patch.
+            Returned when the server could not find the bundle to patch.
           schema:
             allOf:
               - $ref: '#/definitions/Error'
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1452,35 +1452,7 @@ paths:
                     'indexed': False}]
                 type: array
                 items:
-                  type: object
-                  properties:
-                    uuid:
-                      description: A RFC4122-compliant ID for the file.
-                      type: string
-                      pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
-                    version:
-                      description: Timestamp of file creation in DSS_VERSION format.
-                      type: string
-                      format: DSS_VERSION
-                    name:
-                      description: >
-                        Name of the file.
-
-                        - Absolute paths (beginning with a slash) are not allowed.
-                        - Relative paths must not have unix shortcuts (i.e. ".", "..", "~/")
-                        - No "\" (back)slashes allowed.  e.g. "C:\\path" or "path\file.json"
-                        - Only relative paths with "/" (forward)slashes allowed.  e.g. "path/file.json" or "file.json"
-                        - Please upload using "/" (forward)slashes, even if on Windows.
-                      type: string
-                      pattern: "(?=^((?!\\.\\/).)*$)(?=^((?!\\\\).)*$)(?=^((?!\\/\\.).)*$)(?=^(?!\\/).*$)(?=^(?!~\\/).*$)(?=^(?!\\.\\.$).*)(?=^(?!\\.$).*)"
-                    indexed:
-                      description: True iff this file should be indexed.
-                      type: boolean
-                  required:
-                    - uuid
-                    - version
-                    - name
-                    - indexed
+                  $ref: '#/definitions/BundleFile'
               creator_uid:
                 description: User ID who is creating this bundle.
                 type: integer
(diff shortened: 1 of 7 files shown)


dss/api/bundles/__init__.py
```diff
@@ -288,6 +262,76 @@ def delete(uuid: str, replica: str, json_request_body: dict, version: str = None
...
+            ).to_key()
+            if 'file_metadata' not in file:
+                try:
+                    file_metadata = handle.get(replica.bucket, metadata_key)
+                except BlobNotFoundError:
+                    continue
+                file['file_metadata'] = json.loads(file_metadata)
+
+        # check to see if any file metadata is still not yet loaded.
+        for file in files:
```
Bento007: Move the files from the `files` list into a different list say `complete_files` when they have `file_metadata` added. Then you only need to check if `files` is true. It will also let you return a list of missing files at the end.

dss/api/bundles/__init__.py
```diff
@@ -244,7 +187,38 @@ def put(uuid: str, replica: str, json_request_body: dict, version: str):
         )
     status_code = requests.codes.created if created else requests.codes.ok
 
-    return jsonify(dict(version=version, manifest=bundle_metadata)), status_code
+    return status_code
+
+
+def bundle_file_id_metadata(bundle_file_metadata):
+    return hashabledict({
+        # 'indexed': bundle_file_metadata['indexed'],
```
Bento007: saving for later?

dss-api.yml
```diff
@@ -1566,6 +1538,152 @@ paths:
...
+        Add or remove files from a bundle. A specific version of the bundle to update must be provided, and a
+        new version will be written.
+      parameters:
+        - name: uuid
+          in: path
+          description: A RFC4122-compliant ID of the bundle to update.
+          required: true
+          type: string
+          pattern: "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
+        - name: replica
```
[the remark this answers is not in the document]
ttung: My guess is that you want to hit a specific replica consistently, as to avoid propagation delays.

dss-api.yml
```diff
@@ -1566,6 +1538,152 @@ paths:
...
+          schema:
+            type: object
+            properties:
+              add_files:
+                type: array
+                description: >
+                  List of new files to add to the bundle. File names must be unique.
+                items:
+                  $ref: '#/definitions/BundleFile'
+                maxItems: 1000
```
ttung: please make sure that this payload size is compatible with the restrictions of API gateway.

dss/api/bundles/__init__.py
```diff
@@ -288,6 +262,76 @@ def delete(uuid: str, replica: str, json_request_body: dict, version: str = None
 
     return jsonify(response_body), status_code
 
+def build_bundle_file_metadata(replica: Replica, user_supplied_files: dict):
+    handle = Config.get_blobstore_handle(replica)
+
+    time_left = nestedcontext.inject("time_left")
+
+    # decode the list of files.
+    files = [{'user_supplied_metadata': file} for file in user_supplied_files]
```
ttung: pls avoid using python keywords as variable names.

dss-api.yml
xbrianh: planning on tuning this number during profiling. I expect we can handle more than 1000/patch.

dss/api/bundles/__init__.py
Bento007: I saw that github highlighted it but it is really a python keyword?

dss-api.yml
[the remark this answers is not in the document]
xbrianh: @Bento007 seems reasonable to use `replica` similarly to the other endpoints, where it is required.

dss-api.yml
```diff
@@ -1566,6 +1538,152 @@ paths:
...
+                    type: string
+                    description: >
+                      The code `illegal_version` is returned when version is not a DSS_VERSION format-compliant timestamp.
+                      The code `duplicate_filename` is returned when the bundle contains two files of the same name.
+                    enum: [duplicate_filename, illegal_version]
+                required:
+                  - code
+        403:
+          description: >
+            Returned when credentials presented do not grant access to this collection.
```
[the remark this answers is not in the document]
xbrianh: fixed

dss-api.yml
```diff
@@ -1566,6 +1538,152 @@ paths:
...
+                      The code `Forbidden` indicates that the user does not have access to this collection.
+
+                      The code `OAuthResponseProblem` indicates that the OAuth credentials presented were not
+                      understood.
+                    enum: [Forbidden, OAuthResponseProblem]
+                required:
+                  - code
+        404:
+          description: >
+            Returned when the server could not find the collection to patch.
```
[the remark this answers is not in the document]
xbrianh: fixed

dss/api/bundles/__init__.py
xbrianh: This is a carryover of existing verification logic. I'd prefer to refactor in a separate PR.

dss/api/bundles/__init__.py
xbrianh: removed

dss/api/bundles/__init__.py
xbrianh: vim highlights it, but:
```
In [1]: import keyword                                                                                                                           
In [2]: keyword.iskeyword("file")                                                                                                                
Out[2]: False
```

tests/test_bundle.py
```diff
@@ -866,6 +867,103 @@ def delete_bundle(
...
+        res = self.app.patch("/v1/bundles/{}".format(bundle_uuid),
+                             headers=get_auth_header(authorized=True),
+                             params=dict(replica="aws"),
+                             json=dict())
+        self.assertEqual(res.status_code, requests.codes.bad_request)
+
+    def test_patch(self):
+        bundle_uuid, bundle_version = self._put_bundle()
+        files = self._patch_files()
+        tests = [
```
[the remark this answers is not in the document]
xbrianh: tests added

tests/test_bundle.py
```diff
@@ -866,6 +867,103 @@ def delete_bundle(
...
+        res = self.app.patch(
+            f"/v1/bundles/{bundle_uuid}",
+            headers=get_auth_header(authorized=True),
+            params=dict(version=bundle_version, replica="aws"),
+            json=dict(add_files=files),
+        )
+        self.assertEqual(res.status_code, requests.codes.bad_request)
+
+    def test_patch_name_collision(self):
+        "BAD REQUEST is returned for file name collisions."
```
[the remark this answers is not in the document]
xbrianh: It is not an error to remove files not in the bundle.

tests/test_bundle.py
```diff
@@ -934,13 +934,22 @@ def test_patch(self):
...
         bundle_uuid, bundle_version = self._put_bundle()
         files = self._patch_files(number_of_files=2000)
-        res = self.app.patch(
-            f"/v1/bundles/{bundle_uuid}",
-            headers=get_auth_header(authorized=True),
-            params=dict(version=bundle_version, replica="aws"),
-            json=dict(add_files=files),
-        )
-        self.assertEqual(res.status_code, requests.codes.bad_request)
+        with self.subTest("Bad request is returned when adding > 1000 files in a single request."):
```
Bento007: why does line 936 say 2000, but the tests say 1000?

tests/test_bundle.py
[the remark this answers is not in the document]
Bento007: Are those tests else where?

dss/api/bundles/__init__.py
Bento007: Nit: how about you change it to `_file` to make the machine happy?

dss-api.yml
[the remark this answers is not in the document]
Bento007: I'm concerned that simultaneous modification of a FQID on different replicas could result in inconsistent replicas when syncing.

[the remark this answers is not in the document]
xbrianh: Versions are minted on the backend, making a collision causing replica inconsistency extraordinarily unlikely even in the case of hitting PATCH /bundle on multiple replicas simultaneously.

dss/api/bundles/__init__.py
xbrianh: changed

tests/test_bundle.py
xbrianh: The number of files (2000) needs to be greater than the limit (1000), to trigger the error.

Changed to 1001.

tests/test_bundle.py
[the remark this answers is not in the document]
xbrianh: tests added.

Issue #1975, opened 2019-03-13
Warn when Lambda contains an incomplete set of env vars


xbrianh:
This can be a post-deployment warning for now.

PR #1979, opened 2019-03-13
Verify deployment  environment variables

amarjandu:
Address the need to verify env variables across deployments or from
local context.
Resolves: #1975


scripts/populate_lambda_ssm_parameters.py
```diff
@@ -82,6 +85,43 @@ def get_admin_user_emails():
...
+    missing, diff = compare_env_dicts(ssm_env_1, ssm_env_2)
+    if len(missing) > 0:
+        print(f"missing values when comparing {stage} to {compare_stage}\n {missing}")
+    if len(diff) > 0:
+        print(f"Found different values when comparing comparing {stage} to {compare_stage}\n{diff}")
+
+
+def compare_env_dicts(d1: dict, d2: dict):
+    missing_env_values = set.symmetric_difference(set(d1.keys()), set(d2.keys()))
+    different_env_values = {k: d2[k] for k in d2 if k in d1 and d2[k] != d1[k]}
```
DailyDreaming: `different_env_values` should probably return both values as a diff for printing a comparison.

scripts/populate_lambda_ssm_parameters.py
```diff
@@ -102,12 +142,21 @@ def get_admin_user_emails():
         default=None,
         help="Remove a single environment variable in SSM parameters and all deployed lambdas"
     )
+    parser.add_argument("--verify",
+        nargs='*',
```
DailyDreaming: I'd recommend not using nargs and just using a comma-delimited string.  nargs tend to be a bit more ambiguous.

scripts/populate_lambda_ssm_parameters.py
```diff
@@ -102,12 +142,21 @@ def get_admin_user_emails():
...
+        help='Verify the deployed Lambda Environment Variables for a current stage'
+    )
     args = parser.parse_args()
 
     if args.print:
         ssm_env = get_ssm_lambda_environment()
         for name, val in ssm_env.items():
             print(f"{name}={val}")
+    elif args.verify:
+        print(f"Deployment Stage: {os.getenv('DSS_DEPLOYMENT_STAGE')}")
```
DailyDreaming: Should include something more descriptive like `Verifying env vars in deployment stage: `.

scripts/populate_lambda_ssm_parameters.py
```diff
@@ -102,12 +142,21 @@ def get_admin_user_emails():
...
+        compare_local()
+        compare_stages(os.getenv('DSS_DEPLOYMENT_STAGE')) # TODO adjust argument parsing to allow specification of stage
```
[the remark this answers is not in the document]
amarjandu: i took the idea for the dev to "earn their milkshake" and forced them to specify what they want to compare, makes life easier for testing.

scripts/populate_lambda_ssm_parameters.py
```diff
@@ -82,6 +85,43 @@ def get_admin_user_emails():
     admin_user_emails.append(gcp_service_account_email)
     return ",".join(admin_user_emails)
 
+
+def compare_local():
```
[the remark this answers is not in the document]
amarjandu: Id rather have the dev manually source their env, then perform the action, it forces us to be deliberate, and removes the guesswork // potential side effects.

scripts/populate_lambda_ssm_parameters.py
amarjandu: Noted, will do.

scripts/populate_lambda_ssm_parameters.py
```diff
@@ -82,6 +85,43 @@ def get_admin_user_emails():
...
+    ssm_env = get_ssm_lambda_environment()
+    local_env = get_local_lambda_environment()
+    missing_local, diff_local = compare_env_dicts(local_env, ssm_env)
+    if len(missing_local) > 0:
+        print(f"missing values when comparing local env to deployment\n {missing_local}")
+    if len(diff_local) > 0:
+        print(f"Found different values when comparing local env to deployment \n{diff_local}")
+
+
+def compare_stages(stage: str, previous_stage: str = None):
```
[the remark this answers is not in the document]
amarjandu: naming changed, see above for env discussion.

scripts/populate_lambda_ssm_parameters.py
amarjandu: kept nargs added support for len(args._action)

scripts/verify_env.py
```diff
@@ -0,0 +1,44 @@
...
+Does not check
+
+"""
+import os
+import boto3
+import json
+
+
+ssm_client = boto3.client("ssm")
+stages = ['dev', 'integration', 'staging', 'prod']
```
xbrianh: What does this variable do?

scripts/verify_env.py
```diff
@@ -0,0 +1,44 @@
...
+    local_env = get_local_lambda_environment_keys()
+    return compare_env_lists(local_env, ssm_env.keys())
+
+
+def compare_env_lists(l1: list, l2: list):
+    return list(set.symmetric_difference(set(l1), set(l2)))
+
+
+if __name__ == '__main__':
+    print(compare_local_stages())
```
xbrianh: Please add text explaining this output. For instance, something like "WARNING: inconsistent local and deployed lambda environment variables:", and list which variables are where.

scripts/verify_env.py
amarjandu: it was hitchhiking, just got on a train of of town..... *removed*

scripts/verify_env.py
amarjandu: Added text for output + added source for the missing env values in output.
@xbrianh this makes the output not-pretty  although there is a json for the source of the missing env, the warning string makes this non-parsable by `jq`; an alternative would be to add a {"Status": "WARNING" } to the output.

scripts/verify_env.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+"""
+Used to check missing deployed lambda env variables with local environment variables.
+Does not check the values of the variables, limited to a check if they key is present.
```
DailyDreaming: `if they key`

scripts/verify_env.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def get_ssm_lambda_environment(stage):
+    parms = lambda_client.get_function_configuration(
+        FunctionName=f"dss-{stage}"
+    )['Environment']['Variables']
+    return parms
+
+
+def compare_local_stage(stage=None):
```
DailyDreaming: If stage is anything other than `None`, this could run into a conflict with the currently sourced environment.  Either delete it as an input arg and have this script only ever be used for the current stage, or source the env for the stage it's checking.

scripts/verify_env.py
amarjandu: changed

scripts/verify_env.py
```diff
@@ -0,0 +1,47 @@
...
+    """Compares local env values to deployed lambda end value"""
+    stage = os.environ['DSS_DEPLOYMENT_STAGE']
+    ssm_env = set(get_lambda_environment(stage))
+    local_env = set(get_local_lambda_environment_keys())
+    in_ssm = list(ssm_env - local_env)
+    in_local = list(local_env - ssm_env)
+    return {stage: in_ssm, "local": in_local}
+
+
+def compare_env_lists(l1: list, l2: list):
```
Bento007: Remove this code if it Is not used anywhere

scripts/verify_env.py
```diff
@@ -0,0 +1,47 @@
...
+    )['Environment']['Variables']
+    return parms
+
+
+def compare_local_stage():
+    """Compares local env values to deployed lambda end value"""
+    stage = os.environ['DSS_DEPLOYMENT_STAGE']
+    ssm_env = set(get_lambda_environment(stage))
+    local_env = set(get_local_lambda_environment_keys())
+    in_ssm = list(ssm_env - local_env)
```
Bento007: why do you convert it back to a list?

scripts/verify_env.py
```diff
@@ -0,0 +1,47 @@
...
+    in_local = list(local_env - ssm_env)
```
Bento007: same as above

scripts/verify_env.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def compare_env_lists(l1: list, l2: list):
+    return list(set.symmetric_difference(set(l1), set(l2)))
+
+
+if __name__ == '__main__':
+    missing_keys = compare_local_stage()
+    number_of_missing_keys = sum(map(lambda v: len(v), missing_keys.values()))
+    if number_of_missing_keys > 0:
```
Bento007: this can be simplified to `if any(missing_keys.values()):`

scripts/verify_env.py
amarjandu: complete

scripts/verify_env.py
amarjandu: changed! thanks :)

scripts/verify_env.py
amarjandu: @Bento007 ah was working with sets and set.symmetrical_difference(), changed to list comprehension.

Bento007:
```suggestion
    in_ssm = ssm_env - local_env
```

scripts/verify_env.py
```diff
@@ -0,0 +1,48 @@
...
+        try:
+            env[name] = os.environ[name]
+        except KeyError:
+            print(f"Warning: {name} not defined")
+    return env.keys()
+
+
+def get_lambda_environment(stage):
+    parms = lambda_client.get_function_configuration(
+        FunctionName=f"dss-{stage}"
```
DailyDreaming: This only checks `f"dss-{stage}"`, should it check the other lambdas as well?

amarjandu: removed lambda checks, went to SSM instead.

PR #1985, opened 2019-03-19
Implement trufflehog test check.

DailyDreaming:
Need some feedback on this.

Might be good to centralize the regex list on `dcp` (right now I've put the `regex.json` in a test repo) for each hca repo to use, and it might also be good to have a unit test on `dcp` that verifies that this catches known secrets.


xbrianh:
This looks like a reasonable implementation. The `dcp` repo is geared towards dcp-wide integration tests, so I wonder if `dcplib` might be a better home for the shared trufflehog configuration file.

DailyDreaming:
Thanks for taking a look Brian!  I agree, `dcplib` sounds like a good place for it.  Will make the PR.

DailyDreaming:
The `regexes.json` been added to `dcplib` and the url updated here.

PR #1987, opened 2019-03-21
Set Elasicsearch log retention to 5 years

xbrianh:


DailyDreaming:
I guess this accounts for leap years.

PR #1995, opened 2019-03-26
Increase collection verification parallelization

xbrianh:
Use a few more connections to marginally increase verification performance.

depends on #1993 #1994 
connected to #1901


kislyuk:
Consider scaling this with the Lambda size (memory allocation).

xbrianh:
@kislyuk Scaling parallelization with lambda size will be handled in a future PR, see: https://github.com/HumanCellAtlas/data-store/issues/1996

Issue #1997, opened 2019-03-28
Resource Tagging


amarjandu:
- [x] Establish tags/labels on resources
- [x] turn on tags to build report in cost explorer
- [x] write Draft RFC for Tech arch meeting (3/5/2019) 
- [ ] modify on-boarding docs for how to tag properly ( complete after RFC?)

Resolves #1980

amarjandu:
## Notes 4/5/19 : 3:28 PM

### Complete
added changes to `infra` deployment in [2010](https://github.com/HumanCellAtlas/data-store/pull/2010) 
update resources with tagging

### In Progress
NA

### TODO
see changes in Cost Explorer, build example to show

### Awaiting 
Awaiting information to show up in Cost Explorer (may take a few days)

xbrianh:
It appears resource tagging is causing conflicts in CI/CD, since the tags will always need to be updated.

Is it possible to:
 - Only update resource tagging when there are other changes. Otherwise tags will need to be updated every time a new users executes `make deploy-infra`.
 - Ignore resource tagging changes in CI/CD scripting. Otherwise we won't be able to detect infra changes from our build scripting.

amarjandu:
## Notes 4/11/19 : 11:39 AM
Owner was changed to `dss-team@data.humancellatlas.org`
updates to enviornment variables + build scripts
updated resources within {dev,ntegration,staging}
### Complete


### In Progress

### TODO

### Awaiting

PR #2010, opened 2019-04-02
[WIP] Resource Tagging with Terraform

amarjandu:
Uses local tagging variables for all resources
Apply tagging to resources

Resolves: #1997


infra/build_deploy_config.py
```diff
@@ -89,8 +110,16 @@
     "GCP_DEFAULT_REGION",
 ]
 
+if not validate_aws_config():
+    exit(1)
+
+caller_info = boto3.client("sts").get_caller_identity()
+owner = caller_info['UserId'].split(':')[1]
```
amarjandu: @kislyuk  this would allow the `role_session_name` to be populated into owner; might need to update travis_test credentials with a `role_session_name` value within the CD.

infra/build_deploy_config.py
```diff
@@ -11,10 +11,31 @@
 infra_root = os.path.abspath(os.path.dirname(__file__))
 
 
+def validate_aws_config():
+    current_caller = boto3.client('sts').get_caller_identity()
+    if '@' not in current_caller['UserId'].split(':')[1]:
+        print('~/.aw/config needs to have an email under the role_session_name')
```
DailyDreaming: Should this just raise?  For example (also including the suggestion from below to declare the variables in one place):
```
caller_info = boto3.client("sts").get_caller_identity()
owner = caller_info['UserId'].split(':')[1]

def validate_aws_config():
    if '@' not in owner:
        raise ValueError('~/.aw/config needs to have an email under the role_session_name')
```
The call would also be less code overall and just `validate_aws_config()`.

infra/build_deploy_config.py
```diff
@@ -89,8 +110,16 @@
     "GCP_DEFAULT_REGION",
 ]
 
+if not validate_aws_config():
+    exit(1)
+
+caller_info = boto3.client("sts").get_caller_identity()
```
DailyDreaming: `boto3.client("sts").get_caller_identity()` and `caller_info['UserId'].split(':')[1]` are already declared above and could be reused in `validate_aws_config()` if they're going to be naked vars.

infra/build_deploy_config.py
```diff
@@ -111,6 +140,12 @@
         val = os.environ[key]
         fp.write(terraform_variable_template.format(name=key, val=val))
 
+    fp.write(terraform_variable_template.format(name="project", val=project))
```
DailyDreaming: I'm not sure you want to be writing the same variable into the file multiple times.  Also, `.format()` can take in all variables at once as a dictionary.

infra/elasticsearch/main.tf
```diff
@@ -123,11 +114,12 @@ resource aws_elasticsearch_domain elasticsearch {
...
   }
 
-  tags {
-    Domain = "${var.DSS_ES_DOMAIN}"
-    CreatedBy = "Terraform"
-    Application = "DSS"
-  }
+    tags = "${merge(
+        local.common_tags,
+        map(
```
DailyDreaming: Is writing `Domain = "${var.DSS_ES_DOMAIN}"` clearer than a 3 line map?

infra/build_deploy_config.py
amarjandu: We would have to rewrite the same string to file multiple times to populate the TF variables, just like on ln.141 ...

unfortunately env_vars_to_infra is a list and not a dict, otherwise the values could have been added, and the same loop could do all the work....

infra/build_deploy_config.py
amarjandu: thanks, ill refactor

infra/build_deploy_config.py
DailyDreaming: A loop would be nicer but in the end just a nit.  I'll resolve this.

infra/build_deploy_config.py
xbrianh: In our system, infrastructure changes are always performed by an operator. Travis or Allspark should not be updating these tags.

amarjandu: Noted: ensures operator has role_session_name defined :)

infra/build_deploy_config.py
amarjandu: i scrapped the function, and just threw it inline, with the raise, and var reuse.

infra/elasticsearch/main.tf
amarjandu: from what i understand through what TF has available, there is not.  its possible to break this into 1 line, but there its gonna be harder to manage the resource-specfic values

infra/build_deploy_config.py
```diff
@@ -11,10 +11,25 @@
 infra_root = os.path.abspath(os.path.dirname(__file__))
 
 
+
+
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument("component")
 args = parser.parse_args()
 
+terraform_aws_tagging_template = """
```
xbrianh: It would be preferable to not generate these files.

infra/elasticsearch/main.tf
DailyDreaming: Alright, sounds best to leave it as is then.

infra/build_deploy_config.py
amarjandu: Changed tagging.tf to module located in `/infra`

infra/build_deploy_config.py
```diff
@@ -89,8 +87,11 @@
     "GCP_DEFAULT_REGION",
 ]
 
+caller_info = boto3.client("sts").get_caller_identity()
+if '@' not in caller_info['UserId']:
+    raise ValueError('~/.aw/config needs to have an email under the role_session_name')
```
DailyDreaming: Has `~/.aw/config` been verified to have an email on `gitlab` during automated deploy?  I know running this it errors on `dcp-developer` and I have to hack in my email.

amarjandu: I removed this section of the code; i was having issues with 'assumed-role' w/ api 2fa. There will be other ways to enforce this, which is going to be handled with removed resources with invalid tags.

environment
```diff
@@ -107,6 +107,9 @@ GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME="gcp-credentials.json"
 GOOGLE_APPLICATION_SECRETS_SECRETS_NAME="application_secrets.json"
 ES_ALLOWED_SOURCE_IP_SECRETS_NAME="es_source_ip"
 DSS_DEBUG=0
+TF_VAR_PROJECT=dcp
```
DailyDreaming: Update `https://github.com/HumanCellAtlas/data-store/blob/master/docs/environment/README.md` for each.

environment
```diff
@@ -107,6 +107,9 @@ GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME="gcp-credentials.json"
...
+TF_VAR_SERVICE=dss
```
DailyDreaming: Are the `TF_VAR_SERVICE`, etc. names used?  I only see a ref to `var.SERVICE`, etc.

amarjandu: they are, terraform requires the env variables to be defined as `TF_VAR_{x}` and they get overridden when TF is parsing the values for the variables that are marked with the same `x` value. This is why the variables are defined, but dont have defaults attached to them.

environment
amarjandu: shall do

infra/async_state_db/main.tf
```diff
@@ -1,3 +1,10 @@
+module "tagging" {
```
xbrianh: Can we use a `locals` definition here instead of a module? It can include the `owner` and `common_tags` keys, and would not require a modification of the infra build system.

environment
xbrianh: If we move away from using a TF module, we can give these variables canonical DSS names, e.g. `DSS_INFRA_TAG_PROJECT`.

infra/async_state_db/main.tf
amarjandu: We should be able to do that, let me remove the tagging.tf module.

environment
```diff
@@ -73,6 +73,9 @@ DSS_GS_CHECKOUT_BUCKET_TEST_USER=org-hca-dss-checkout-test-user  # for tests w/n
 DSS_GS_CHECKOUT_BUCKET_INTEGRATION=org-humancellatlas-dss-checkout-integration
 DSS_GS_CHECKOUT_BUCKET_STAGING=org-humancellatlas-dss-checkout-staging
 DSS_GS_CHECKOUT_BUCKET_PROD=org-humancellatlas-dss-checkout-prod
+DSS_INFRA_TAG_PROJECT=dcp
+DSS_INFRA_TAG_SERVICE=dss
+DSS_INFRA_TAG_ENV=${DSS_DEPLOYMENT_STAGE}
```
xbrianh: This is already available to `/infra` scripting as `var.DSS_DEPLOYMENT_STAGE`

environment.integration
```diff
@@ -15,6 +15,9 @@ DSS_GCP_SERVICE_ACCOUNT_NAME="org-humancellatlas-integration"
 DSS_ZONE_NAME="integration.data.humancellatlas.org."
 ACM_CERTIFICATE_IDENTIFIER="390e82ea-a684-49f4-a4b3-857f22cee874"
 DSS_CHECKOUT_BUCKET_OBJECT_VIEWERS="serviceAccount:1037839730885-compute@developer.gserviceaccount.com,serviceAccount:caas-account@broad-dsde-mint-test.iam.gserviceaccount.com,serviceAccount:caas-prod-account-for-test@broad-dsde-mint-test.iam.gserviceaccount.com"
+TF_VAR_PROJECT=dcp
+TF_VAR_SERVICE=dss
+TF_VAR_ENV=${DSS_DEPLOYMENT_STAGE}
```
xbrianh: variable names need to be updated.

Environment variables that are the same across all deployments do not need to be included in `environment.*`

infra/build_deploy_config.py
```diff
@@ -89,8 +90,9 @@
     "GCP_DEFAULT_REGION",
 ]
 
+caller_info = boto3.client("sts").get_caller_identity()
```
xbrianh: This can be removed?

environment
amarjandu: changed in `local.common.tag` + removed

infra/build_deploy_config.py
amarjandu: reverted file, added `DSS_INFRA_TAG_*`

environment.integration
amarjandu: removed references within these files

Issue #1920, opened 2019-02-27
GET /collection should support paging


xbrianh:
This is needed to overcome 6MB payload response limits on AWS Lambda.

PR #2013, opened 2019-04-02
Add paging to collections.

DailyDreaming:
Addresses #1920.

This adds paging to the new `GET /collections` root endpoint, with some caveats.

Right now this is really expensive as all collections in a bucket need to be searched and opened to determine ownership before returning the information to the user.

To prevent timing out, I've put in a max paging size of 100 (min 50).  This will check each chunk of 100 `collection` files in the bucket and return any belonging to the user per page (takes about 15s per page when the bucket contains about 2000 collections).

This means instead of returning 100 collections per page, it will return between 0 and 100 collections per page since it needs to check them all, but only some will belong to the user.

Collections need to be indexed to make this scalable.

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -133,7 +133,7 @@ class TestBundleApi(unittest.TestCase, TestAuthMixin, DSSAssertMixin, DSSUploadM
                 self._test_bundle_get_paging(replica, list(), 501, codes=requests.codes.bad_request)
 
     def _test_bundle_get_paging(self,
-                                replica: str,
+                                replica,
                                 expected_files: list,
                                 per_page: int,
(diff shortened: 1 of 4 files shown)


tests/test_bundle.py
```diff
@@ -105,7 +105,7 @@ def test_bundle_paging(self):
             ))
             expected_files = manifest['files']
 
-        for replica in Replica:
+        for replica in (Replica.aws, Replica.gcp):
```
amarjandu: was this changed to allow the gcp replica to be tested as well?

DailyDreaming: Yep!

DailyDreaming:
Merging then with the understanding that I'll need to implement indexing soon.  Ticket here: #2014

Issue #1199, opened 2018-04-30
Production Smoke Test


mweiden:
# User story (Why?)
As a DCP operator,
I want to know if the production deployment can process core API requests
to know if the service is truly available. This test should not generate any test data.

# DoD (What?)
A production smoke test is in place which performs the following operations
 - [x] Upload a bundle
 - [x] Delete a bundle
 - [x] Download a bundle
 - [x] Subscribe
- [x] Unsubscribe
- [ ] Receive notifications
 - [x] Find a bundle
 - [x] Checkout a bundle

amarjandu:
## Notes 4/9/19 : 1:35 PM

### Complete
implemented most of the testing, with the exception of the notification system.
unable to test notifications without generating data; passive notification might be available, but might hold processes in a waiting state. 

### In Progress
NA, in review
### TODO

### Awaiting
Review from team

PR #2029, opened 2019-04-09
Production Smoke Tests

amarjandu:
Asserts production system functionality on both replicas.

Resolves: #1199


tests/test_prod_smoketest.py
```diff
@@ -0,0 +1,203 @@
...
+    except subprocess.CalledProcessError as e:
+        parser.exit(RED(f'{parser.prog}: Exit status {e.returncode} while running "{command}". Stopping.'))
+
+
+def run_for_json(command, **kwargs):
+    return json.loads(run(command, stdout=subprocess.PIPE, **kwargs).stdout.decode(sys.stdout.encoding))
+
+
+@testmode.integration
+class ProdSmoketest(unittest.TestCase):
```
DailyDreaming: This should inherit `Smoketest()` from `test/test_smoketest.py` and only change what's needed from there (and the functions above should be imported from `test/test_smoketest.py`).

Either that or modify `test/test_smoketest.py` to run on all stages as appropriate.

amarjandu: created base class for regular + prod smoketest to inherit from.

tests/test_smoketest.py
```diff
@@ -124,94 +67,34 @@ def smoketest(self, starting_replica, checkout_bucket, test_bucket):
...
-                url = s3.generate_presigned_url(ClientMethod='put_object',
-                                                Params=dict(Bucket=self.notification_bucket,
-                                                            Key=notification_key,
-                                                            ContentType='application/json'))
-                put_response = run_for_json([f'{self.venv_bin}hca', 'dss', 'put-subscription',
-                                             '--callback-url', url,
-                                             '--method', 'PUT',
-                                             '--es-query', json.dumps(query),
-                                             '--replica', replica.name])
+                url = BaseSmokeTest.generate_presigned_url(self, self.notification_bucket, notification_key)
```
DailyDreaming: Is there a reason to call the base inherited class instead of self?

Meaning:
`url = BaseSmokeTest.generate_presigned_url(self, self.notification_bucket, notification_key)`
instead of:
`url = self.generate_presigned_url(self, self.notification_bucket, notification_key)`

tests/test_smoketest.py
```diff
@@ -124,94 +67,34 @@ def smoketest(self, starting_replica, checkout_bucket, test_bucket):
...
+                put_response = BaseSmokeTest.subscription_create_es(self, replica, query, url)
```
DailyDreaming: Same as above, etc. etc.  Not commenting for all of them.

tests/test_smoketest.py
amarjandu: resolved.

tests/smoktest/base_smoketest.py
```diff
@@ -0,0 +1,215 @@
...
+                object_key = get_dst_bundle_prefix(bundle_uuid, bundle_version)
+                print(f"Checking bucket {checkout_bucket} "
+                      f"object key: {object_key}")
+                files = list(blob_handle.list(checkout_bucket, object_key))
+                self.assertEqual(len(files), file_count)
+                break
+        else:
+            self.fail("Timed out waiting for checkout job to succeed")
+
+    def post_search_es(self, replica, es_query):
```
DailyDreaming: I know this is a deviation from the original smoketest, but I'm realizing that we test some `es_query` functionality but there's no reciprocal `jmespath_query` test.  Since we're promoting `jmespath_query` as the default and intended user experience, should we add those tests symmetrically?

tests/test_prod_smoketest.py
```diff
@@ -0,0 +1,79 @@
...
+    params = BaseSmokeTest.params
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.url = "https://www.example.com"
+        cls.query = {"query": {"bool": {"must": [{"term": {"admin_deleted": "true"}}]}}}
+
+    @classmethod
+    def tearDownClass(cls):
```
xbrianh: Can this method be moved into the base class?

tests/test_smoketest.py
```diff
@@ -1,107 +1,50 @@
...
         # Prepare the bundle using stock metadata and random data
         #
         cls.bundle_dir = os.path.join(cls.workdir.name, "bundle")
-        shutil.copytree("tests/fixtures/datafiles/example_bundle", cls.bundle_dir)
+        shutil.copytree(os.environ["DSS_HOME"] + "/tests/fixtures/datafiles/example_bundle", cls.bundle_dir)
         with open(os.path.join(cls.bundle_dir, "async_copied_file"), "wb") as fh:
             fh.write(os.urandom(ASYNC_COPY_THRESHOLD + 1))
 
+    @classmethod
+    def tearDownClass(cls):
```
xbrianh: Can this method be moved into the base class?

tests/test_prod_smoketest.py
amarjandu: changed

tests/test_smoketest.py
amarjandu: changed.

tests/smoktest/base_smoketest.py
xbrianh: We should, but in a different PR

.gitlab-ci-prod.yml
```diff
@@ -39,3 +39,14 @@ deploy:
   only:
     - prod
   when: manual
+
+testing:
+  stage: deploy
```
xbrianh: Please change `stage` to `test`, and add a `test` stage to `stages` on line 12

see https://github.com/HumanCellAtlas/data-store/blob/7af5da0cd11f158bcf4645c89099df448a24cbe6/.gitlab-ci.yml#L12

.gitlab-ci-prod.yml
```diff
@@ -39,3 +39,14 @@ deploy:
...
+  script:
+    -make smoketest-prod
+  environment:
+    name: prod
+    url: https://dss.data.humancellatlas.org
+  only:
+    - prod
+  when: on_sucess
```
xbrianh: This line can be removed.

jobs are performed in the order of `stages`

amarjandu: resolved

.gitlab-ci-prod.yml
amarjandu: resolved

.gitlab-ci-prod.yml
```diff
@@ -39,3 +40,13 @@ deploy:
   only:
     - prod
   when: manual
+
+test:
+  stage: test
+  script:
+    -make smoketest-prod
+  environment:
```
xbrianh: The `environment` block should be omitted for non-deployment stages

amarjandu: addressed

PR #2034, opened 2019-04-10
Parallelize JMESPath notification delivery

xbrianh:
This parallelizes JMESPath filtering and notification delivery, and scales up the notification Lambda.

Future notification scaling tracked via #2034

### DIFF
--- a/daemons/dss-notify-v2/app.py
+++ b/daemons/dss-notify-v2/app.py
@@ -109,6 +109,7 @@ def _notify_subscribers(replica: Replica, key: str):
         if should_notify(replica, subscription, metadata_document, key):
             notify_or_queue(replica, subscription, metadata_document, key)
 
+    # TODO: Consider scaling parallelization with Lambda size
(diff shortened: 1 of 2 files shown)


daemons/dss-notify-v2/app.py
```diff
@@ -105,5 +102,15 @@ def launch_from_notification_queue(event, context):
...
             logger.warning(f"Recieved queue message with no matching subscription:{message}")
 
+def _notify_subscribers(replica: Replica, key: str):
+    metadata_document = build_bundle_metadata_document(replica, key)
+
+    def _func(subscription):
+        if should_notify(replica, subscription, metadata_document, key):
+            notify_or_queue(replica, subscription, metadata_document, key)
+
+    with ThreadPoolExecutor(max_workers=20) as e:
```
DailyDreaming: Factor of the number of available cores?

xbrianh: added todo comment

PR #2039, opened 2019-04-12
Update IAM to allow `dss-notify-v2-{stage}` access

natanlao:

### DIFF
--- a/iam/policy-templates/ci-cd.json
+++ b/iam/policy-templates/ci-cd.json
@@ -162,7 +162,11 @@
         "arn:aws:sqs:*:$account_id:dss-notify-test-*",
         "arn:aws:sqs:*:$account_id:dss-notify-dev-*",
         "arn:aws:sqs:*:$account_id:dss-notify-integration-*",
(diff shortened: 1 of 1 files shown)


iam/policy-templates/ci-cd.json
```diff
@@ -162,7 +162,11 @@
         "arn:aws:sqs:*:$account_id:dss-notify-test-*",
         "arn:aws:sqs:*:$account_id:dss-notify-dev-*",
         "arn:aws:sqs:*:$account_id:dss-notify-integration-*",
-        "arn:aws:sqs:*:$account_id:dss-notify-staging-*"
+        "arn:aws:sqs:*:$account_id:dss-notify-staging-*",
+        "arn:aws:sqs:*:$account_id:dss-notify-v2-test-*",
```
amarjandu: we can remove the `-*` after the stage, we also should add the event relay sqs in here; it looks something like `dss-notify-v2-event-relay-dev`

natanlao:
Dupe of #2040

PR #2043, opened 2019-04-15
Bundle manifest memcaching

### LINKED ISSUE #2011: Cache /PATCH versions in memory on save
This will improve performance of bundle and collection agglomeration.

xbrianh:
Previously, bundle manifests were cached in memory during GET. This extends that functionality to PUT and PATCH /bundle, improving PATCH performance.

closes #2011

### DIFF
--- a/dss/storage/bundles.py
+++ b/dss/storage/bundles.py
@@ -117,7 +117,8 @@ def idempotent_save(blobstore: BlobStore, bucket: str, key: str, data: dict) -> 
         else:
             blobstore.upload_file_handle(bucket, key, io.BytesIO(_d))
 
-        _bundle_manifest_cache[key] = data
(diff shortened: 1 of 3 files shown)


dss/storage/bundles.py
```diff
@@ -1,21 +1,44 @@
...
 
 
-@lru_cache(maxsize=32)
 def get_bundle_manifest(
         uuid: str,
         replica: Replica,
         version: typing.Optional[str],
         *,
         bucket: typing.Optional[str] = None) -> typing.Optional[dict]:
+    key = BundleFQID(uuid, version).to_key()
```
DailyDreaming: Typing is still ` -> typing.Optional[dict]` but it looks like this function doesn't return a dictionary anymore.

PR #2046, opened 2019-04-15
Modifed Description for /PUT/Collection

amarjandu:
modify description for /PUT/Collection, remove reference for generated timestamp

Resolves: [dcp/cli#299](https://www.youtube.com/watch?v=Nw4Sng23YH0)

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1116,7 +1116,7 @@ paths:
         - name: version
           in: query
           description: >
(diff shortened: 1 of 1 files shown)


amarjandu:
> LGTM, although... having used this, I don't see why a user needs to manually mint their own. I feel like the DSS should be able to do this for you. Wait to merge this til we can talk about it at stand up tomorrow?

yeah I'm unsure of the reason why this timestamp is user specified, perhaps @xbrianh can give some insight. 🕵️

PR #2050, opened 2019-04-16
Revert "Clarify and fix test expectations"

natanlao:
Reverts HumanCellAtlas/data-store#2038


natanlao:
`master` was force pushed

PR #2052, opened 2019-04-17
Dynamic DynamoDB Table Fetching

amarjandu:
Add dynamic fetching of of dynamodb tables to health checks

Resolves: #2021

### DIFF
--- a/dss/api/health.py
+++ b/dss/api/health.py
@@ -7,8 +7,15 @@ from googleapiclient import discovery, errors
 from elasticsearch.exceptions import ConnectionError
 from dss.index.es import ElasticsearchClient
 from dss.util.aws.clients import dynamodb  # type: ignore
+from dss.util.aws.clients import resourcegroupstaggingapi  # type: ignore
 
 logger = logging.getLogger(__name__)
+
+
+def get_resource_by_tag(resource_string: str, tag_filter: dict):
+    dss_resources = resourcegroupstaggingapi.get_resources(ResourceTypeFilters=[resource_string],
+                                                           TagFilters=[tag_filter])
+    return dss_resources
 
 
 def _get_es_status(host: str = "localhost", port: int = None):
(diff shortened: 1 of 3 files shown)


tests/test_health.py
```diff
@@ -56,6 +56,16 @@ def test_full_health_check(self, mock_er, mock_ddb, mock_es):
...
 
+    @testmode.standalone
+    def test_resource_fetch(self):
+
+        service_tags = {"Key": "service", "Values": ["dss"]}
+        resource_list = health.get_resource_by_tag(resource_string='dynamodb:table', tag_filter=service_tags)
+        ddb_tables = [x['ResourceARN'].split('/')[1] for x in resource_list['ResourceTagMappingList'] if
+                      os.environ.get('DSS_DEPLOYMENT_STAGE') in x['ResourceARN']]
+        print(ddb_tables)
+        self.assertGreater(len(ddb_tables), 0)
```
xbrianh: This test may be more future proof using `s3` instead of `dynamodb`.

dss/api/health.py
```diff
@@ -33,10 +40,10 @@ def _get_dynamodb_status():
     """Checks dynamoDB table status, tables are explicitly specified within the function"""
     db_status = True
     stage = os.getenv("DSS_DEPLOYMENT_STAGE")
-    ddb_tables = ["dss-async-state-{}".format(stage),
-                  "dss-subscriptions-v2-aws-{}".format(stage),
-                  "dss-subscriptions-v2-gcp-{}".format(stage)
-                  ]
+    service_tags = {"Key": "service", "Values": ["dss"]}
+    resource_list = get_resource_by_tag(resource_string='dynamodb:table', tag_filter=service_tags)
```
xbrianh: I think the same effect can be achieved replacing `dynamodb:table` with `dynamodb`.

tests/test_health.py
amarjandu: Updated test to work with s3

dss/api/health.py
amarjandu: I'm going to keep it as tables to ensure the resource filter is kept to tables in this function.

PR #2060, opened 2019-04-19
Modify file names in metadata document

xbrianh:
This keeps JMESPath querying consistent with current indexing schemes.

### DIFF
--- a/dss/events/handlers/notify_v2.py
+++ b/dss/events/handlers/notify_v2.py
@@ -196,8 +196,10 @@ def build_bundle_metadata_document(replica: Replica, key: str) -> dict:
             except json.decoder.JSONDecodeError:
                 logging.info(f"{file_metadata['name']} not json decodable")
             else:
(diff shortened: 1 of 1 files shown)


dss/events/handlers/notify_v2.py
```diff
@@ -220,6 +222,19 @@ def build_deleted_bundle_metadata_document(key: str) -> dict:
         "version": version,
     }
 
+def _dot_to_underscore_and_strip_numeric_suffix(name: str) -> str:
```
amarjandu: LGTM: is the modification for the name due to jmespath allowing [SubExpressions](http://jmespath.org/specification.html#subexpressions)?

xbrianh: Yes, this change makes it easier to use JMESPath sub-expressions. It was possible before but the syntax is unfortunate.

The second reason is this makes metadata file names the same in ES and JMESPath index documents.

PR #2069, opened 2019-04-23
Abstract out dynamoDB functionality from subscriptions.

DailyDreaming:
Separated from the collections PR into it's own, but ultimately for use as a dynamoDB base class for each of these cases.


dss/util/dynamodb.py
```diff
@@ -0,0 +1,65 @@
+from dss.util.aws.clients import dynamodb  # type: ignore
+
+
+class DynamoLookup(object):
```
xbrianh: A class with nothing but static methods may not deserve to be a class.

Note that modules can provides descriptive name spacing.
```
imporot my_cool_module

my_cool_module.my_cool_method()
```

dss/subscriptions_v2/__init__.py
```diff
@@ -22,64 +18,35 @@ class SubscriptionData:
...
-        Key={
-            'hash_key': {
-                'S': owner
-            },
-            'sort_key': {
-                'S': uuid
-            }
-        }
-    )
+class SubscriptionLookup(DynamoLookup):
```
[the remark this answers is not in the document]
DailyDreaming: The main benefit is inheriting the base class methods.  I tend to prefer classes for collections of related functions and especially when using a base class set of functions.

I could program this functionally and avoid using a class if that's preferable, but there are a few things that Subscriptions does that I don't want in Collections.  For example, the `body` value is a dictionary in Subscriptions and a string in `Collections`.  The type should be handled by the outer class and not strictly enforced by the base class.

So you can do something like:
`class BaseClass:
    def get():
        get some db query...
        return value.get('item', None)`

`class Subscriptions(BaseClass):
    def get_subscription():
        return json.loads(self.get())`

`class Collections(BaseClass):
    def get_collection():
        return self.get()`

dss/util/dynamodb.py
DailyDreaming: It's not just for descriptive namespacing.  I prefer an ABC to importing bare functions because it clearly defines which methods are intended to be inherited and some methods will only need to be wrapped and returned.

I'd be willing to make this purely functional if you prefer, I think it's more of a style preference.  I will argue that it makes more sense alongside the collections class also inheriting the same methods, at least to me it looks nicer.

DailyDreaming:
Updated to the functional style.

dss/util/dynamodb.py
```diff
@@ -0,0 +1,63 @@
+from dss.util.aws.clients import dynamodb  # type: ignore
```
xbrianh: Let's move this file to `dss/dynamodb/__init__.py`

dss/subscriptions_v2/__init__.py
```diff
@@ -22,64 +18,38 @@ class SubscriptionData:
...
-        TableName=_ddb_table_template.format(replica.name),
-        KeyConditionExpression="#hash_key=:owner",
-        ExpressionAttributeNames={'#hash_key': "hash_key"},
-        ExpressionAttributeValues={':owner': {'S': owner}}
-    )
-    subscriptions = [json.loads(item['body']['S']) for item in db_resp['Items']]
-    return subscriptions
+    items = dynamodb.get_primary_key_items(table=subscription_db_table.format(replica.name),
+                                           key=owner)
+    return [json.loads(item['body']['S']) for item in items]
```
xbrianh: Can we move `['body']['S']` into `dynamodb.get_primary_key_items`?

dss/util/dynamodb.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    return db_resp.get('Items', [])
+
+
+def get_all_table_items(table: str) -> list:
+    db_resp = dynamodb.scan(TableName=table)
+    return db_resp.get('Items', [])
+
+
+def delete_item(table: str, key1: str, key2: str = None):
```
xbrianh: Please think of more descriptive names for `key1` and `key2`.

dss/util/dynamodb.py
DailyDreaming: Alright.

dss/subscriptions_v2/__init__.py
DailyDreaming: Alright.

DailyDreaming:
@xbrianh Changes made.

daemons/dss-notify-v2/app.py
```diff
@@ -18,10 +16,12 @@
 import dss
 from dss import Config, Replica
 from dss.logging import configure_lambda_logging
-from dss.subscriptions_v2 import get_subscriptions_for_replica, get_subscription
 from dss.events.handlers.notify_v2 import (should_notify, notify_or_queue, notify, build_bundle_metadata_document,
                                            build_deleted_bundle_metadata_document)
 from dss.events.handlers.sync import exists
+from dss.subscriptions_v2 import (get_subscription,
+                                  get_subscriptions_for_replica)
```
xbrianh: Probably don't need the parentheses

dss/subscriptions_v2/__init__.py
```diff
@@ -22,64 +18,35 @@ class SubscriptionData:
...
-                'S': owner
-            },
-            'sort_key': {
-                'S': uuid
-            }
-        }
-    )
+    dynamodb.delete_item(table=subscription_db_table.format(replica.name),
+                         key1=owner,
+                         key2=uuid)
```
xbrianh: keyword arguments need updating

dss/subscriptions_v2/__init__.py
```diff
@@ -22,64 +18,35 @@ class SubscriptionData:
...
+    dynamodb.put_item(table=subscription_db_table.format(doc[SubscriptionData.REPLICA].name),
+                      key1=doc[SubscriptionData.OWNER],
+                      key2=doc[SubscriptionData.UUID],
+                      value=json.dumps(doc))
+
+
+def get_subscription(replica: Replica, owner: str, uuid: str):
+    item = dynamodb.get_item(table=subscription_db_table.format(replica.name),
+                             key1=owner,
+                             key2=uuid)
```
xbrianh: keyword arguments need updating

dss/subscriptions_v2/__init__.py
```diff
@@ -22,64 +18,35 @@ class SubscriptionData:
...
-                'S': doc[SubscriptionData.UUID]
-            },
-            'body': {
-                'S': json.dumps(doc)
-            },
-        }
-    )
+    dynamodb.put_item(table=subscription_db_table.format(doc[SubscriptionData.REPLICA].name),
+                      key1=doc[SubscriptionData.OWNER],
+                      key2=doc[SubscriptionData.UUID],
```
xbrianh: keyword arguments need updating

dss/dynamodb/__init__.py
```diff
@@ -0,0 +1,67 @@
+from dss.util.aws.clients import dynamodb as db  # type: ignore
```
xbrianh: The methods in this class could use docstrings.

For instance, when must `sort_key` be passed in?

PR #2072, opened 2019-04-24
Use dynamoDB methods in async_state.

DailyDreaming:
Refactor stage 2.

### DIFF
--- a/dss/dynamodb/__init__.py
+++ b/dss/dynamodb/__init__.py
@@ -3,52 +3,83 @@ from typing import Generator
 from dss.util.aws.clients import dynamodb as db  # type: ignore
 
 
-def put_item(table: str, value: str, hash_key: str, sort_key: str):
+def put_item(table: str, value: str, hash_key: str, sort_key: str=None):
     """
     Put an item into a dynamoDB table.
+
+    Will determine the type of db this is being called on by the number of keys provided (omit
+    sort_key to PUT into a db with only 1 primary key)
 
     :param table: Name of the table in AWS.
     :param str value: Value stored by the two primary keys.
     :param str hash_key: 1st primary key that can be used to fetch associated sort_keys and values.
     :param str sort_key: 2nd primary key, used with hash_key to fetch a specific value.
+                         Note: If not specified, this will PUT only 1 key (hash_key) and 1 value.
     :return: None
     """
-    db.put_item(
-        TableName=table,
-        Item={
-            'hash_key': {
-                'S': hash_key
-            },
-            'sort_key': {
-                'S': sort_key
-            },
-            'body': {
-                'S': value
-            },
-        }
-    )
+    if sort_key:
+        db.put_item(
+            TableName=table,
+            Item={
+                'hash_key': {
+                    'S': hash_key
+                },
+                'sort_key': {
+                    'S': sort_key
+                },
+                'body': {
+                    'S': value
+                },
+            }
+        )
+    else:
+        db.put_item(
+            TableName=table,
+            Item={
+                'key': {
+                    'S': hash_key
+                },
+                'body': {
+                    'S': value
+                },
+            }
+        )
 
 
-def get_item(table: str, hash_key: str, sort_key: str):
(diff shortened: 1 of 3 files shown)


dss/dynamodb/__init__.py
```diff
@@ -3,52 +3,83 @@
...
+            TableName=table,
+            Item={
+                'key': {
+                    'S': hash_key
+                },
+                'body': {
+                    'S': value
+                },
+            }
+        )
```
xbrianh: if the async ddb schema is changed to use `hash_key` instead of `key`, this will work:
```
item = {
    'hash_key': {
        'S': hash_key
    },
    'body': {
        'S': value
    }
if sort_key is not None:
    item['sort_key'] = dict(S=sort_key)
db.put_item(TableName=table, Item=item)
```

dss/dynamodb/__init__.py
```diff
@@ -88,23 +119,37 @@ def get_all_table_items(table: str, return_key: str='body') -> Generator[str, No
...
+    else:
+        db.delete_item(
+            TableName=table,
+            Key={
+                'key': {
+                    'S': hash_key
+                },
+            }
+        )
```
xbrianh: ditto

dss/dynamodb/__init__.py
DailyDreaming: Alright, can do.  What additional steps should I take when tearing down the table and building the new one?

dss/dynamodb/__init__.py
```diff
@@ -3,53 +3,54 @@
 from dss.util.aws.clients import dynamodb as db  # type: ignore
 
 
-def put_item(table: str, value: str, hash_key: str, sort_key: str):
+def format_item(value, hash_key, sort_key):
```
xbrianh: Consider making this method private, and making `value` and `sort_key` keywords.

dss/dynamodb/__init__.py
```diff
@@ -3,53 +3,54 @@
...
+    item = {'hash_key': {'S': hash_key}}
+    if value:
+        item['body'] = {'S': value}
+    if sort_key:
+        item['sort_key'] = {'S': sort_key}
+    return item
+
+
+def put_item(table: str, value: str, hash_key: str, sort_key: str=None, overwrite: str=None):
```
xbrianh: Is it safe to pass in `None` for `value`? Ideally `value` would come after `hash_key` and `sort_key`.

dss/dynamodb/__init__.py
```diff
@@ -3,53 +3,54 @@
...
 
+    Will determine the type of db this is being called on by the number of keys provided (omit
+    sort_key to PUT into a db with only 1 primary key).
+
     :param table: Name of the table in AWS.
     :param str value: Value stored by the two primary keys.
     :param str hash_key: 1st primary key that can be used to fetch associated sort_keys and values.
     :param str sort_key: 2nd primary key, used with hash_key to fetch a specific value.
+                         Note: If not specified, this will PUT only 1 key (hash_key) and 1 value.
+    :param str overwrite: Don't overwrite if this parameter exists.  For example, setting this
```
xbrianh: What happens if `overwrite` exists? e.g., is an exception thrown?

DailyDreaming: That's a good question.  I may have added this too early since this was from the collections portion.  Will take this out and test it when I add it with the collections portion.

dss/dynamodb/__init__.py
DailyDreaming: Will test.

dss/dynamodb/__init__.py
DailyDreaming: Alright!

DailyDreaming: Made private and separated the function into 2 instead for the value keyword.

dss/dynamodb/__init__.py
DailyDreaming: It can't pass value as `None` anymore unless explicitly done.  It couldn't before either, but this makes it clearer.

PR #2079, opened 2019-04-26
Increase GS client connection pool to 20

xbrianh:
This matches S3 connection pool

Also addresses `WARNING:urllib3.connectionpool:Connection pool is full, discarding connection: www.googleapis.com`:
https://travis-ci.com/HumanCellAtlas/data-store/jobs/195750957

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -13,6 +13,8 @@ from cloud_blobstore.gs import GSBlobStore
 from google.cloud.storage import Client
 from google.auth.transport.requests import AuthorizedSession
 from google.oauth2 import service_account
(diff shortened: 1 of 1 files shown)


dss/config.py
```diff
@@ -13,6 +13,8 @@
 from google.cloud.storage import Client
 from google.auth.transport.requests import AuthorizedSession
 from google.oauth2 import service_account
+from requests.adapters import HTTPAdapter, DEFAULT_POOLSIZE
```
Bento007: keeping these imports local reduces the spin up of our lambdas. If lambda start up becomes an issue we should start localizing imports again.

PR #2080, opened 2019-04-26
Add basic indexing for collections.

DailyDreaming:
- [x] Add collections/dynamoDB API for faster user lookups.
- [x] Increase the paging testing to be more robust now that it's faster.
- [x] Test individual collections db calls.
- [x] Deleted files should be updated in the db, but handle files that are unexpectedly not there.
- [x] Have dynamoDB return paged uuid responses.
- [x] Script to populate the collections db from scratch (read everything in the collections key of that bucket).


dss/api/collections.py
```diff
@@ -46,43 +47,47 @@ def get_impl(uuid: str, replica: str, version: str = None):
...
     # single response returning all collections (or those remaining)
     else:
-        collections = fetch_collections(handle, bucket, collection_keys[start_at:])
-        return jsonify({'collections': collections}), requests.codes.ok
+        collection_page = collections[start_at:]
+        return jsonify({'collections': collection_page}), requests.codes.ok
+
 
 @dss_handler
 @security.authorized_group_required(['hca'])
```
amarjandu: This `GET` will get updated for sharability at a later time?

dss/collections/__init__.py
```diff
@@ -0,0 +1,35 @@
+import os
+
+from dss import dynamodb  # type: ignore
+
+
+collection_db_table = f"dss-collections-db-{os.environ['DSS_DEPLOYMENT_STAGE']}"
+
+
+def put_collection(owner: str, uuid: str, permission_level: str = 'owner'):
+    assert permission_level in ('owner', 'read-only')
```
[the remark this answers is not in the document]
DailyDreaming: `read-only` isn't implemented yet.  It's there for sharing functionality and planning to be compatible since that is coming soon.  I wasn't sure whether to include it, but felt it wouldn't hurt.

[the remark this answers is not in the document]
DailyDreaming: Also, `read-only` users won't have write functionality obviously, but it's still a put operation into the database with the item containing the user and their permission level for that collection uuid.

dss/api/collections.py
DailyDreaming: It will.  This PR got broken up into a number of smaller earlier PRs refactoring dynamodb in general, and the sharing functionality will also become a separate PR.

dss/collections/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    dynamodb.put_item(table=collection_db_table,
+                      hash_key=owner,
+                      sort_key=uuid,
+                      value=permission_level,
+                      dont_overwrite='sort_key')
+
+
+def get_collection(owner: str, uuid: str) -> str:
```
DailyDreaming: Just a note, this is unused, but will be used for sharing collections.  I can take it out if necessary and add it later when necessary.  @xbrianh Let me know.

dss/collections/__init__.py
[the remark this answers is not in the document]
DailyDreaming: Removed for now.

dss/collections/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def get_collection_uuids_for_owner(owner: str) -> list:
+    items = dynamodb.get_primary_key_items(table=collection_db_table,
+                                           key=owner,
+                                           return_key='sort_key')
+    return [item for item in items]
+
+
+def get_all_collection_uuids():
```
xbrianh: Redundant with `dss.collections. get_all_collection_uuids`?

DailyDreaming: `get_all_collection_uuids` gets all uuids (all owners) in the table for repopulating it from the `update_collections_db.py` script.  `get_collection_uuids_for_owner` fetches only the current owner related uuids for the API.

Does that address your comment?

scripts/update_collection_db.py
```diff
@@ -0,0 +1,108 @@
...
+                self.key_set[uuid] = version
+
+        self.all_bucket_uuids = set(self.key_set.keys())
+        self.all_database_uuids = set(get_all_collection_uuids())
+        self.uuids_not_in_db = self.all_bucket_uuids.symmetric_difference(self.all_database_uuids)
+
+        bucket_name = f'{Replica[replica].storage_schema}://{self.bucket}'
+        print(f'Found {str(len(self.all_bucket_uuids))} collections in {bucket_name}.')
+        print(f'Found {str(len(tombstone_uuids))} tombstoned collections in {bucket_name}.')
+        print(f'Found {str(len(self.all_database_uuids))} collections in database.')
```
xbrianh: f-strings convert variables to `str` for you.

scripts/update_collection_db.py
```diff
@@ -0,0 +1,108 @@
+"""
```
xbrianh: It may be preferable to refactor this file so that consistency verification and repair happens in one direction at a time:
  - check that s3 is consistent with ddb
  - check that ddb is consistent with s3

Each operation, separately, is parallelizable.

dss/dynamodb/__init__.py
```diff
@@ -12,7 +12,7 @@ def _format_item(hash_key, sort_key, value):
     return item
 
 
-def put_item(*, table: str, hash_key: str, sort_key: str=None, value: str, overwrite: str=None):
+def put_item(*, table: str, hash_key: str, sort_key: str=None, value: str, dont_overwrite: str=None):
```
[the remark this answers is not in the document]
DailyDreaming: I believe the default behavior is to silently overwrite it.

dss/collections/__init__.py
```diff
@@ -0,0 +1,32 @@
+import os
+
+from dss import dynamodb  # type: ignore
+
+
+collection_db_table = f"dss-collections-db-{os.environ['DSS_DEPLOYMENT_STAGE']}"
+
+
+def put_collection(owner: str, uuid: str, permission_level: str = 'owner'):
```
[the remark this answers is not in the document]
DailyDreaming: Alright, I'd based them on the subscriptions naming scheme used previously.  Should I rename those as well?

scripts/update_collection_db.py
```diff
@@ -0,0 +1,108 @@
...
+
+class CollectionDatabaseTools(object):
+    def __init__(self, replica='aws'):
+        self.replica = replica
+        self.bucket = Replica[replica].bucket
+        self.handle = Config.get_blobstore_handle(Replica[replica])
+
+        raw_keys = self.handle.list(self.bucket, prefix='collections')
+        tombstone_uuids = [i[len('collections/'):-len('.dead')] for i in
+                           self.handle.list(self.bucket, prefix='collections') if i.endswith('.dead')]
```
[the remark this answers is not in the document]
DailyDreaming: Alright.

scripts/update_collection_db.py
DailyDreaming: So, current behavior consists of these separate steps (though these 3 are all in `def __init__`):
1. Look at all collections in s3 and make a set of uuids.
2. Look at all table items in dynamodb and make a set of uuids.
3. Make a set of uuids in s3 that aren't in the dynamodb table.

Then you can either hard reset:
4a. Delete all items in the DB and rebuild all collections with everything from the bucket.

Or update:
4b. Delete nothing, add all collection entries from s3 that aren't already in the database.
Note: GET /collections operations check the file before returning and delete the entry from the database if it doesn't exist.  If an entry is somehow missed, it won't break anything and should eventually be removed through normal use if it's not in the bucket.

If you want to guarantee the entries are 1-to-1, the hard reset seems fairly efficient to me (it's always going to be expensive; every file will need to be opened once to check ownership no matter what you do).  Otherwise, updating the table with only the diff can be much faster and should never break anything.

We should talk on Monday!

dss/subscriptions_v2/__init__.py
```diff
@@ -29,10 +29,13 @@ def put_subscription(doc: dict):
...
-    item = dynamodb.get_item(table=subscription_db_table.format(replica.name),
-                             hash_key=owner,
-                             sort_key=uuid)
-    return json.loads(item) if item else None
+    try:
+        item = dynamodb.get_item(table=subscription_db_table.format(replica.name),
+                                 hash_key=owner,
+                                 sort_key=uuid)
+        return json.loads(item)
+    except dynamodb.ItemNotFoundInDatabase:
```
amarjandu: perhaps logger should be filled with what was missing in this context.

dss/api/collections.py
```diff
@@ -46,43 +47,41 @@ def get_impl(uuid: str, replica: str, version: str = None):
...
+        {'collections': [{'uuid': uuid, 'version': version}, {'uuid': uuid, 'version': version}, ... , ...]}
+    """
+    # TODO: Replica is unused, so this does not use replica.  Appropriate?
+    owner = security.get_token_email(request.token_info)
 
-    # expensively list every collection file in the bucket, even those not belonging to the user (possibly 1000's... )
-    collection_keys = [i for i in handle.list(bucket, prefix='collections') if not i.endswith('dead')]
+    collections = []
+    for collection in owner_lookup.get_collection_versioned_uuids_for_owner(owner):
+        uuid, version = collection.split('.', 1)
```
xbrianh: Might as well use `CollectionFQID` from `dss/storage/identifiers.py`

dss-api.yml
```diff
@@ -2456,19 +2448,16 @@ definitions:
...
         type: string
         description: A UUID identifying the collection.
         pattern: "[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}"
-      collection_version:
-        type: string
+      version:
         description: The version of the UUID identifying the collection.
-      collection:
-        $ref: '#/definitions/Collection'
+        type: string
```
xbrianh: What does it mean to drop `dss/storage/identifiers.py`?

dss/collections/owner_lookup.py
```diff
@@ -0,0 +1,49 @@
+import os
+
+from dss import dynamodb  # type: ignore
+
+
+collection_db_table = f"dss-collections-db-{os.environ['DSS_DEPLOYMENT_STAGE']}"
+
+
+def put_collection(owner: str, versioned_uuid: str, permission_level: str = 'owner'):
```
xbrianh: Please change `versioned_uuid` to `collection_fqid`

This should be applied throughout this PR.

dss/dynamodb/__init__.py
```diff
@@ -1,8 +1,13 @@
 from typing import Generator
+from botocore.exceptions import ClientError
 
 from dss.util.aws.clients import dynamodb as db  # type: ignore
 
 
+class ItemNotFoundInDatabase(Exception):
```
xbrianh: Maybe `DynamoDBItemNowFound`?

dss/dynamodb/__init__.py
```diff
@@ -24,18 +29,23 @@ def put_item(*, table: str, hash_key: str, sort_key: str=None, value: str, overw
...
-        query['ConditionExpression'] = f'attribute_not_exists({overwrite})'
-    db.put_item(**query)
+    if dont_overwrite:
+        query['ConditionExpression'] = f'attribute_not_exists({dont_overwrite})'
 
+    try:
+        db.put_item(**query)
+    except ClientError as e:
+        if e.response['Error']['Code'] != 'ConditionalCheckFailedException':
+            raise
```
xbrianh: I feel like encountering an existing item while `dont_overwrite != None` should raise.

If I were writing client methods, I'd want to be able to react when a DDB operation is not idempotent. What do you think?

scripts/update_collection_db.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        Returns any (owner, versioned_uuid) present in the bucket but not in the collections table.
+
+        Returns an iterable tuple of strings: (owner, versioned_uuid) representing the item's key pair.
+
+        The returned keys can then be added to the collections dynamodb table.
+        """
+        for collection_key in self.handle.list(self.bucket, prefix=COLLECTION_PREFIX):
+            self.total_bucket_collection_items += 1
+            if collection_key != f'{COLLECTION_PREFIX}/':
```
xbrianh: This check can be avoided using `prefix=f"{COLLECTION_PREFIX}/"` on line 89

scripts/update_collection_db.py
```diff
@@ -0,0 +1,157 @@
...
+            added += 1
+        print(f'{added} collection items added from: {self.bucket_name}')
+        return added
+
+
+def main():
+    start = time.time()
+    c = CollectionDatabaseTools(replica='aws')
+
+    if len(sys.argv) > 1:
```
xbrianh: please use `argparse`

see https://github.com/HumanCellAtlas/data-store/blob/4e0a61c388013bb0499777a573c43ada7dd386ee/scripts/populate_lambda_ssm_parameters.py#L86

dss/dynamodb/__init__.py
DailyDreaming: I think you're right.  Will put into the collection method.

dss-api.yml
DailyDreaming: That object is the whole collection list, like a bundle.  It has unrestricted length so it was decided to not include them.

scripts/update_collection_db.py
DailyDreaming: I usually only use argparse if there's more than one option, like for this one that I did: https://github.com/HumanCellAtlas/data-store/blob/master/scripts/swagger_auth.py#L160

This is more based on browsing the scripts to determine the conventions in current script usage.  I'll start including argparse in all scripts then.

dss/subscriptions_v2/__init__.py
DailyDreaming: This is server side so the user would never see it.

dss/collections/owner_lookup.py
```diff
@@ -0,0 +1,55 @@
...
+from botocore.exceptions import ClientError
+
+from dss import dynamodb  # type: ignore
+
+
+collection_db_table = f"dss-collections-db-{os.environ['DSS_DEPLOYMENT_STAGE']}"
+
+
+def put_collection(owner: str, collection_fqid: str, permission_level: str = 'owner'):
+
```
xbrianh: spurious newline

DailyDreaming:
@xbrianh Yes, we'll need to create a table and run the indexing script for each stage it promotes to.  I'll include that in the release notes.

dss/api/collections.py
```diff
@@ -46,43 +47,41 @@ def get_impl(uuid: str, replica: str, version: str = None):
...
-                                    'collection_version': version,
-                                    'collection': collection})
-    return all_collections
 
 @dss_handler
 @security.authorized_group_required(['hca'])
-def listcollections(replica: str, per_page: int, start_at: int = 0):
-    bucket = Replica[replica].bucket
-    handle = Config.get_blobstore_handle(Replica[replica])
+def listcollections(per_page: int, start_at: int = 0):
```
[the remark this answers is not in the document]
DailyDreaming: Addressed.  Thanks for taking a look!

Issue #1968, opened 2019-03-11
Move TF configuration to stage-specific buckets


xbrianh:
Currently the DSS TF state files are stored in a dcp-shared bucket parameterized with the AWS account id:  `org-humancellatlas-{account_id}-terraform`.

State files should be moved to DSS stage specific buckets parameterized with both account_id and stage, e.g. `org-humancellatlas-dss-{account_id}-{DSS_DEPLOYMENT_STAGE}-terraform`

amarjandu:
Would need to update the lambda/cloud function build scripts + scripts for `infra`

PR #2086, opened 2019-04-29
Isolate Terraform State File Buckets

amarjandu:
Isolate terraform state file buckets by stage for DSS

Resolves: #1968


amarjandu:
use this spell to move statefiles quickly.... if needed. replace {account_id} as well as the stage within the grep and destination bucket stage. 

`aws s3 ls s3://org-humancellatlas-{account_id}-terraform/dss/ | grep dev | awk '{print $4}' | while read line; do aws s3 cp s3://org-humancellatlas-{account_id}-terraform/dss/$line s3://org-humancellatlas-dss-861229788715-dev-terraform/$line; done`

PR #2087, opened 2019-04-29
Remove trufflehog dependency.

DailyDreaming:
This is a travis only dependency.

Updating via `make requirements-dev.txt` doesn't remove dependencies it seems.

Blanking the file and regenerating the whole thing seems to update a lot of libraries and breaks in a few ways.  So just removing the trufflehog-only portions for now.


DailyDreaming:
@kislyuk Will do, thanks for taking a look.

PR #2090, opened 2019-04-30
Prod Smoke Test Query

amarjandu:
Remove possibility for tombstoned bundles to be in the prod-smoketest

### DIFF
--- a/tests/test_prod_smoketest.py
+++ b/tests/test_prod_smoketest.py
@@ -2,7 +2,7 @@
 """
 A prod test for the DSS, checks core API Requests to know whats available. Works with data present.
 """
-import os, sys, argparse, unittest
+import os, sys, argparse, unittest, json
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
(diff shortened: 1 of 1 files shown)


tests/test_prod_smoketest.py
```diff
@@ -24,7 +24,7 @@ class ProdSmoketest(BaseSmokeTest):
     def setUpClass(cls):
         super().setUpClass()
         cls.url = "https://www.example.com"
-        cls.query = {"query": {"bool": {"must": [{"term": {"admin_deleted": "true"}}]}}}
+        cls.query = {"query": {"bool": {"must_not": [{"term": {"admin_deleted": "true"}}]}}}
```
xbrianh: Won't this change interfere with the subscription notification testing?

We may need to define two queries.

amarjandu: Due to the test not creating/destroying data, it only checks if the subscription can successfully be created/retrieved. From my tests locally it works successfully with the new query. I can add in the older one if requested.  

Screen Shot 2019-04-30 at 12 46 36 PM

xbrianh: OK, I understand.

Please make it clear (via comment or naming scheme) that the subscription query is not expected to be used.

amarjandu: added comment

PR #2091, opened 2019-05-01
Simplify GitLab job definitions

xbrianh:
- provide flexible scale and performance test job

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,6 +12,7 @@ stages:
   - test
   - deploy
   - integration_test
+  - scale_and_performance
   - release
 
 before_script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -35,63 +36,40 @@ before_script:
...
-  stage: test
-  variables:
-    DSS_TEST_MODE: "integration"
-  script:
-    - python tests/test_sync.py TestSyncDaemon.test_sync_large_blob
-  only:
-    refs:
-      - schedules
-    variables:
-      - $DSS_TEST_LARGE_FILE_SYNC
```
amarjandu: test_sync_large_blob was removed, intentional? Otherwise looks much cleaner. 👍

xbrianh: The large blob sync test will be performed under the new `scale_and_performance` stage.

PR #2092, opened 2019-05-01
Fix multipart sync completion criteria

xbrianh:
When object size is an integer multiple of part size, `object_size // part_size) + 1` evaluates to the incorrect number of parts.


xbrianh:
unit test added

PR #2104, opened 2019-05-06
Use one keyword for iterables.

DailyDreaming:
PR is for discussion.

@hannes-ucsc @kislyuk There would obviously have to be corresponding code changes if the swagger doc changes in this PR were to go through, but would it make sense to just have one keyword (`contents`) and return each item in that array rather than the current behavior of using several hard-coded keywords (only `contents` would be hard-coded into the CLI with this change, though we could comment in the swagger that this is its intended usage)?

For reference, the current CLI iterator returns: https://github.com/HumanCellAtlas/dcp-cli/commit/a84fae932fe96770a88501aeae9eb31d40ff2785

Which would change to simply:
```
for item in page.json()["contents"]:
    yield item
```

Hannes's comment: https://github.com/HumanCellAtlas/dcp-cli/issues/331

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1036,7 +1036,7 @@ paths:
           schema:
             type: object
             properties:
-              collections:
+              contents:
                 description: A user's collection UUIDs and versions.
                 type: array
                 items:
@@ -1046,7 +1046,7 @@ paths:
           schema:
             type: object
             properties:
-              collections:
+              contents:
                 description: A user's collection UUIDs and versions.
                 type: array
                 items:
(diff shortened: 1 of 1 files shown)


kislyuk:
I agree that it's not appropriate to introduce a breaking change for this.

I would like to see a new metadata item introduced into the route swagger/openapi description, for those routes that support pagination, indicating the name of the key that points to the array of contents.

I emphasize that this is not urgent work and can be postponed. The small issue of hardcoding/trying the various names for the array key in the client is tolerable for the foreseeable future.

DailyDreaming:
@kislyuk @hannes-ucsc You're right, for all of the reasons Hannes listed, though I thought coding it would be a faster, clearer example.  Closing this in favor of potentially adding a new field hinting at the suggested key to iterate through in the future.

PR #2112, opened 2019-05-08
Add retries to PUT collection method when testing.

DailyDreaming:


tests/test_collections.py
```diff
@@ -7,6 +7,7 @@
 import io
 import json
 import boto3
+import time
```
xbrianh: Is `time` used anywhere?

tests/test_collections.py
```diff
@@ -82,6 +83,7 @@ def upload_file(app, contents):
         resp_obj.raise_for_status()
         return file_uuid, resp_obj.json()["version"]
 
+    @eventually(timeout=60, interval=1, errors={requests.exceptions.HTTPError})
```
xbrianh: Can this error be restricted further to `requests.codes.unprocessable`?

DailyDreaming: I don't think it can with the eventually function (I can't think of a non-hacky way to add it).  It goes back to a custom retry since you need the response to get the code.  I'll make the change.

tests/test_collections.py
DailyDreaming: Changed.

PR #2113, opened 2019-05-08
Verify checksums on `PUT /file/{uuid}`

natanlao:

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -34,6 +34,14 @@ retry request after the specified interval."""
 RETRY_AFTER_INTERVAL = 10
 
 logger = logging.getLogger(__name__)
+checksum_format = {
+    'hca-dss-crc32c': r'^[a-z0-9]{8}$',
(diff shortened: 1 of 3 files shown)


amarjandu:
you have failing checks for synk, please investigate.

dss/api/files.py
```diff
@@ -34,6 +34,14 @@
 RETRY_AFTER_INTERVAL = 10
 
 logger = logging.getLogger(__name__)
+checksum_format = {
+    'hca-dss-crc32c': r'^[a-z0-9]{8}$',
+    'hca-dss-s3_etag': r'^[a-z0-9]{32}(-([2-9]|[1-8][0-9]|9[0-9]|[1-8][0-9]{2'
```
amarjandu: possible to give context on this?

natanlao: I've updated the code in question to describe its origin and purpose. For reference, here is the part of the Swagger spec that this was copied from: https://github.com/HumanCellAtlas/data-store/blob/95908ef36dd1c10aaa34704f8ec49ae2647824d3/dss-api.yml#L293

PR #2111, opened 2019-05-08
Operations CLI framework

xbrianh:
This PR
  1. Provides a central entrypoint for DSS operations scripting, and
  2. Provides a framework for command forwarding to Lambda

connected to #2096, #2111

PR #2114, opened 2019-05-08
Operatios cloud executor

xbrianh:
This daemon can execute operational commands via SQS-Lambda integration. For instance, referential integrity of bundles may be checked by sending queuing a command into SQS for each bundle in the bucket.

depends on #2111 
connected to #2096, #2111


iam/policy-templates/dss-operations-lambda.json
```diff
@@ -0,0 +1,41 @@
...
+      "Action": [
+        "logs:CreateLogGroup",
+        "logs:CreateLogStream",
+        "logs:PutLogEvents"
+      ],
+      "Resource": "arn:aws:logs:*:*:*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": "s3:*",
```
amarjandu: We should consider not giving the lambda deletion abilities.

xbrianh: Deletions on the main S3 buckets are denied externally to the data-store repo:
https://github.com/HumanCellAtlas/dcp-infra/blob/master/terraform/accounts/humancellatlas/policy_aws_replica.tf

PR #2123, opened 2019-05-09
Operation to repair file blob metadata

xbrianh:
This operation
  1. Verifies that file metadata is equivalent on all replicas (where file metadata is referenced by key=`files/{uuid}`
  2. sets the blob content-type equal to the file metadata content-type

### DIFF
--- a/dss/operations/storage.py
+++ b/dss/operations/storage.py
@@ -133,16 +133,18 @@ class repair_blob_content_type(StorageOperationHandler):
     def process_key(self, key):
         try:
             file_metadata = json.loads(self.handle.get(self.replica.bucket, key))
-        except BlobNotFoundError:
-            self.log_warning(BlobNotFoundError.__name__, dict(key=key))
-        blob_key = compose_blob_key(file_metadata)
-        blob_content_type = self.handle.get_content_type(self.replica.bucket, blob_key)
-        client = Config.get_native_handle(self.replica)
-        if blob_content_type != file_metadata['content-type']:
-            if Replica.aws == self.replica:
-                update_aws_content_type(client, self.replica.bucket, blob_key, file_metadata['content-type'])
-            elif Replica.gcp == self.replica:
-                update_gcp_content_type(client, self.replica.bucket, blob_key, file_metadata['content-type'])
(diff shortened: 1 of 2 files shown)


dss/operations/storage.py
```diff
@@ -108,6 +108,26 @@ def process_key(self, key):
                                       file_metadata_content_type=file_metadata['content-type'],
                                       blob_content_type=blob_content_type))
 
+@storage.action("repair-file-blob-metadata",
+                arguments={"--entity-type": dict(default="file", choices=["file"])})
```
DailyDreaming: What other choices other than file will we need to account for in the future?

dss/operations/storage.py
```diff
@@ -108,6 +108,26 @@ def process_key(self, key):
...
+class repair_blob_content_type(StorageOperationHandler):
+    """
+    Make blob metadata consistent with file metadata.
+    """
+    def process_key(self, key):
```
DailyDreaming: Should there be a function to crawl through the entire bucket and make repairs?

dss/operations/storage.py
xbrianh: None, this line is there to restrict the options of `--entity-type` to just `file`.

dss/operations/storage.py
xbrianh: The operator should decide what to do based on the output of `scripts/dss-ops.py storage verify-file-blob-metadata`.

Once a list of repair candidates is generated, scripting may be used to either call the repair commands locally or SQS forward them for Lambda execution:
```
for key in $keys; do
  scripts/dss-ops.py storage repair-file-blob-metadata --replica $replica --keys $key --forward-to-lambda
done
```

DailyDreaming: If you use the command above on only one replica as it seems to be doing, will the changes sync?

dss/operations/storage.py
```diff
@@ -124,6 +124,26 @@ def process_key(self, key):
...
+        try:
```
Bento007: Is this code run anywhere in the tests or an integration test?

PR #2135, opened 2019-05-14
Resume unfinished GS copies

xbrianh:
This:
  1. Avoids lost state by adding choice step to SFN flow
  2. Provides 10 seconds of padding for thread to time out before Lambda times out

### DIFF
--- a/dss/stepfunctions/gscopyclient/implementation.py
+++ b/dss/stepfunctions/gscopyclient/implementation.py
@@ -88,7 +88,7 @@ def copy_worker(event, lambda_context):
                     state[_Key.TOKEN] = response[0]
                     self.save_state(state)
 
-    return CopyWorkerTimedThread(lambda_context.get_remaining_time_in_millis() / 1000, event).start()
+    return CopyWorkerTimedThread(lambda_context.get_remaining_time_in_millis() / 1000 - 10, event).start()
 
 
 def fail(event, lambda_context):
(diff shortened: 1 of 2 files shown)


ttung:
I don't think I'm qualified to review this any more.

dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -140,6 +140,19 @@ def _sfn():
                 "Resource": copy_worker,
                 "Retry": _retry_default(),
                 "Catch": _catch_default(),
+                "Next": "TestFinished",
+            },
+            f"TestFinished": {
```
amarjandu: is this supposed to be an f-string?

dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -139,6 +139,19 @@ def _sfn():
...
+                "Type": "Choice",
+                "Choices": [{
+                    "Variable": "$.finished",
```
amarjandu: please excuse this dumb question: if the variable value is not marked as "true" what mechanism saves the context of the state so it's not lost?

dss/stepfunctions/gscopyclient/implementation.py
xbrianh: will fix

dss/stepfunctions/gscopyclient/implementation.py
xbrianh: The state is saved by the AWS step function (SFN) execution engine.

The SFN calls Lambda functions on some state and saves the state returned. The [SFN language](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-amazon-states-language.html) also provides some flow control steps, such as the `Choice` step used here.

PR #2136, opened 2019-05-14
Async state convenience methods and bug fixes

xbrianh:
This:
  1. Adds convenience method to async state to capture state absent private keys
  2. Prevents accidental modification of state data before putting into database

### DIFF
--- a/dss/util/async_state.py
+++ b/dss/util/async_state.py
@@ -21,6 +21,7 @@ class AsyncStateItem:
 
     def __init__(self, key: str, body: dict) -> None:
         self.key = key
+        body = body.copy()
         if not body.get('_type'):
             body['_type'] = type(self).__name__
         self.body = body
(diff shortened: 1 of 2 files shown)


dss/util/async_state.py
```diff
@@ -21,16 +21,23 @@ class MyAsyncSubclass(AsyncStateItem):
...
             body['_type'] = type(self).__name__
         self.body = body
 
     def _put(self) -> typing.Any:
         return put_item(table=self.table, value=json.dumps(self.body), hash_key=self.key)
 
+    @property
+    def data(self):
+        data = self.body.copy()
+        del data['_type']
```
amarjandu: dumb q: why the deletion operation here?

xbrianh: To make this assertion pass:
```
in_data = dict(my_key=foo)
AsyncStateItem.put(key, in_data)
assert in_data == AsyncStateItem.get(key).data
```

PR #2137, opened 2019-05-14
[Spike] Prevent multiple concurrent checkout jobs

xbrianh:
This is an experimental solution to the following problem: If multiple clients initiate GET requests against the same file, a checkout job will be initiated for each client.

This mechanism:
  1. Places a marker in a DynamoDB table when a file checkout job is started
  2. On subsequent requests:
    - If object exists in checkout bucket, return as usual
    - If not, check for marker. If marker exists, do not start new checkout job
  3. Marker is deleted upon checkout success.
  4. If checkout fails, marker contains the exception.

### DIFF
--- a/dss/api/files.py
+++ b/dss/api/files.py
@@ -37,16 +37,16 @@ logger = logging.getLogger(__name__)
 
 
 @dss_handler
-def head(uuid: str, replica: str, version: str = None, token: str = None):
-    return get_helper(uuid, Replica[replica], version, token)
+def head(uuid: str, replica: str, version: str = None):
+    return get_helper(uuid, Replica[replica], version)
 
 
 @dss_handler
-def get(uuid: str, replica: str, version: str = None, token: str = None):
-    return get_helper(uuid, Replica[replica], version, token)
+def get(uuid: str, replica: str, version: str = None):
+    return get_helper(uuid, Replica[replica], version)
 
 
-def get_helper(uuid: str, replica: Replica, version: str = None, token: str = None):
+def get_helper(uuid: str, replica: Replica, version: str = None):
     with tracing.Subsegment('parameterization'):
         handle = Config.get_blobstore_handle(replica)
         bucket = replica.bucket
@@ -90,7 +90,7 @@ def get_helper(uuid: str, replica: Replica, version: str = None, token: str = No
         blob_path = compose_blob_key(file_metadata)
 
     if request.method == "GET":
-        token, ready = _verify_checkout(replica, token, file_metadata, blob_path)
+        ready = _verify_checkout(replica, file_metadata, blob_path)
         if ready:
             response = redirect(handle.generate_presigned_GET_url(
                 replica.checkout_bucket,
@@ -98,7 +98,6 @@ def get_helper(uuid: str, replica: Replica, version: str = None, token: str = No
         else:
             with tracing.Subsegment('make_retry'):
                 builder = UrlBuilder(request.url)
-                builder.replace_query("token", token)
                 response = redirect(str(builder), code=301)
                 headers = response.headers
                 headers['Retry-After'] = RETRY_AFTER_INTERVAL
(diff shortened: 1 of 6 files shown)


dss/storage/checkout/bundle.py
```diff
@@ -72,50 +72,47 @@ def verify_checkout(
...
+        raise token
+
     if token is None:
         execution_id = start_bundle_checkout(replica, bundle_uuid, bundle_version, dst_bucket=replica.checkout_bucket)
 
-        decoded_token = {
+        token = {
             CheckoutTokenKeys.EXECUTION_ID: execution_id,
             CheckoutTokenKeys.START_TIME: time.time(),
             CheckoutTokenKeys.ATTEMPTS: 0,
```
mikebaumann: It appears the ATTEMPTS value is no longer being updated or used, and could/should be removed.

dss/storage/checkout/bundle.py
```diff
@@ -72,50 +72,47 @@ def verify_checkout(
...
+    :return: is_checkout_ready
     """
     if _is_checkout_valid(replica, bundle_uuid, bundle_version):
         if _is_checkout_stale(replica, bundle_uuid, bundle_version):
             start_bundle_checkout(replica, bundle_uuid, bundle_version, dst_bucket=replica.checkout_bucket)
         return "", True
 
     decoded_token: dict
+
+    async_key = f"{replica.name}/{bundle_uuid}.{bundle_version}"
```
mikebaumann: It seems the composition of the `async_key` would need to include the bucket name to avoid state getting conflated in the following cases of concurrent checkouts of the same bundle/version:
* Concurrent checkout to DSS checkout bucket and to a user-provided bucket
* Concurrent checkouts to two different user-provided buckets

dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -84,6 +84,8 @@ def run(self) -> dict:
                 response = dst_blob.rewrite(src_blob, token=state.get(_Key.TOKEN, None))
                 if response[0] is None:
                     state[Key.FINISHED] = True
+                    async_key = f"{Replica.gcp.name}/{self.destination_key}"
```
mikebaumann: I noticed an inconsistency between the s3copyclient and gscopyclient in that the s3copyclient has a function to construct the `_async_key` that provides a default value, but the gscopyclient does not. This may (or may not) be correct and appropriate, I did not dig in deeply to determine that. Here I am just noting the apparent difference/inconsistency.

dss/api/files.py
```diff
@@ -121,12 +120,22 @@ def get_helper(uuid: str, replica: Replica, version: str = None, token: str = No
...
-def _verify_checkout(
-        replica: Replica, token: typing.Optional[str], file_metadata: dict, blob_path: str,
-) -> typing.Tuple[str, bool]:
+def _verify_checkout(replica: Replica, file_metadata: dict, blob_path: str,) -> typing.Tuple[str, bool]:
     cloud_handle = Config.get_blobstore_handle(replica)
     hca_handle = Config.get_hcablobstore_handle(replica)
 
+    # Lookup existing checkout job. token is an exception if checkout failed
+    async_key = f"{replica.name}/{blob_path}"
+    token = AsyncStateItem.get(async_key)
```
mikebaumann: There are multiple places within this PR that multiple small changes are scattered through and across various functions/methods. Doing so is expedient for testing implementation ideas, yet it can make potentially make the code brittle as it is evolved by other authors. IMO, it would make the code more maintainable to have a helper class, perhaps `CheckoutTracker`, that provides a well-defined interface and encapsulates the details of key formation, `AsynchStateItem` use, etc.

dss/api/files.py
```diff
@@ -121,12 +120,22 @@ def get_helper(uuid: str, replica: Replica, version: str = None, token: str = No
...
+    if isinstance(token, S3CopyEtagError):
+        raise DSSException(
+            requests.codes.unprocessable,
+            "missing_checksum",
+            "Incorrect s3-etag"
+        )
+    elif isinstance(token, AsyncStateError):
+        raise token
```
mikebaumann: Once a checkout has failed (e.g. due to a checksum error), it is never possible for a user to get the file again without administrative intervention to clear the error from the DynamoDB table.
We had some good discussion regarding desired behavior options and some additional/alternate ways of handling this.

mikebaumann (changes requested):
Thanks, @xbrianh!

I think this is a beneficial change that should go forward, pending some relatively small modifications. Some of the benefits this change provides are:
* More efficient operation in general.
* Reduced costs of moving a file out of cold storage. Access costs are higher for cold storage, so preventing concurrent copies is more beneficial.
* DRS 1.0 delay handling does not use redirects, and does not provide a clear/convenient way to pass the token around.
* Preventing removal and replacement of file in checkout bucket that may be in use as multiple concurrent checkouts complete asynchronously. I don't know if this creates data access problems or not, yet still seems worth avoiding.

In addition to the comments added inline, some additional tests for concurrent checkout behavior are likely needed as well. It is good that the existing tests are continuing to pass without modification, yet I suspect that some additional tests are needed.

Issue #2045, opened 2019-04-15
Optionally provide a cloud native URL as output from GET /files/{uuid}


kozbo:
Some consumers of HCA file data require a cloud native (e.g. `gs://` or `s3://`) URL instead of a presigned URLs. One such consumer is the Broad's Terra (formerly FireCloud) system, which does not support presigned URLs but instead requires cloud native URLs for cloud data access.
This request is motivated by the current HCA Browser handoff to Terra development effort.

A proposed design is as follows:
- Enhance the DSS `GET /files/{uuid}` with an optional `directurl` query parameter, with the values `true` and `false` (default). This is consistent with the `GET /bundles/{uuid}` `directurls` query parameter.
- When `directurl=true`, in the final response from the `GET /files/{uuid}` operation (the response with HTTP code 302), returns the cloud native URL to the file in the checkout bucket in the `Location` header, instead of the presigned URL, for that same file/object.

@hannes-ucsc and I collaborated on this proposal, and consider it consistent with the definition of the HTTP 302 response in [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.4.3).

We are open to alternate designs, yet this seems simple and consistent.

It is understood that the with only the changes above, the cloud native URL path would be the "blob key" path (composed of a series of checksums) rather than the more user-meaningful cloud native paths produced by `GET /bundles/{uuid}`. This is "okay" for the purposes of the HCA Data Browser handoff to Terra, yet it _may_ be worth considering support for user meaningful names for file checkout as well (perhaps based on file UUID + version?), mostly for other use cases. This should be addressed in a separate issue(/PR), as it is a bigger "can of worms" and much lower priority.

Future Note:
This behavior is generally/broadly useful when HCA has only open access data and the checkout buckets have public access and #2055 is available. When HCA supports controlled access data, and the checkout buckets do not have public access, a means of providing credentials to use with the cloud native URL will also be needed.

Done When:
The DSS API provides cloud native URLs that are requested with per-file (not per bundle) granularity.

mikebaumann:
@kislyuk, are you fine with this change as currently described above?
The need for this was discussed at the Blue Box Architecture meeting on 4/8/19 (https://docs.google.com/document/d/1qqr8z-5OaFZX3uhGc10hnsrKwNuOb0vvzg88dbRK8Bc/edit#heading=h.5iphltimkd7).
[...]

Issue #2100, opened 2019-05-03
Create presigned URLs suitable for workflow execution


mikebaumann:
As an external workflow service provider (or researcher using an external workflow service),
I want to access HCA data using suitable presigned URLs,
so that I can run my workflows/analysis on the data.

Presigned URLs for data access have many benefits, yet there are some important limitations of presigned URLs wrt workflow execution, some in general and some specific to the URLs currently produced by the DSS. The issues presented here were primarily stem from collaboration with the Broad on the "HCA handoff to Terra", yet are generally applicable as well.

Terra/Cromwell does not currently support presigned URLs, yet according Alex Baumann (Broad) the Broad is open to considering such support (at some point in the future) if the key limitations of presigned URLs can be addressed/resolved. Notes from a meeting with Alex summarizing these limitations are available here:
https://docs.google.com/document/d/18WDYr3uS82Eu1GhHmeH6PX7tFY6Zk8piKd3ixOoE8Bg/edit#heading=h.i4ag7sau4uy

The relative priority of this issue depends on the needs of end-users, the "HCA handoff to Terra" will use cloud native URLs, for which other issues have been created already.

In short, these limitations are:

## Expiration Duration
Currently the presigned URLs produced by the DSS use an (implicit) expiration time of 1 hour for AWS and 1 day for GCP, yet analysis/workflows may require more time (Terra caps workflow execution at 7 days). This could be addressed by enhancing the DSS `GET /files` and `GET /bundles` (with `presignedurls=true` with an optional query parameter which specifies the presigned URL duration (up to 7 days, the maximum duration supported by AWS). The longer the expiration of the presigned URL the greater the security risk, so it is appropriate to allow the user to specify only the expiration duration they need.
Also, I have heard and read that the actual presigned URL expiration of the DSS presigned URLs for AWS may actually be (substantially) less than one hour due to being signed by the assumed role of the Lambda.

## Signed by the Requester
For security/audibility, it is important that the presigned URLs be signed by the user's account, and not by a DSS user/role, to provide an audit trail of who is accessing the data. This will be more important when the DSS provides controlled access data, yet will likely be important even for security compliance. 
For controlled data access, it is also important to be able to revoke user access promptly (e.g. within 24 hours), and the mechanisms for doing that for presigned URLs are generally lacking.

## Requester Pays
[...]

PR #2138, opened 2019-05-14
Direct URLs for Files

amarjandu:
Return direct url option for `GET /file/{uuid}` endpoint

Addresses: #2045
This PR returns a cloud native URL with the same path as is currently used for the presigned URL, the blob key path (with checksums), and refers to the same object in the checkout bucket.
There is a current question regarding whether that is the best/right path to return going forward, yet this current change is sufficient to unblock the HCA handoff to Terra effort.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -368,7 +368,7 @@ paths:
         - name: directurl
           in: query
           required: false
-          description: When set to true, the response will contain API-specific URLs that are tied to the specified replica
+          description: When set to true, the response will contain API-specific URLs that are tied to the specified replica, for example `gs://bucket/object` or `s3://bucket/object`
           type: boolean
       responses:
         301:
@@ -1452,7 +1452,7 @@ paths:
         - name: directurls
           in: query
           description: >
-            Include API-specific direct-access URLs in the response.  This is mutually exclusive with the presignedurls parameter.
+            When set to true, the response will contain API-specific URLs that are tied to the specified replica, for example `gs://bucket/object` or `s3://bucket/object`. This parameter is mutually exclusive with the presigned urls parameter.
           required: false
           type: boolean
         - name: presignedurls
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -368,11 +368,13 @@ paths:
         - name: directurl
           in: query
           required: false
-          description: When set to true, the response will contain API-specific URLs that are tied to the specified
-            replica, for example `gs://bucket/object` or `s3://bucket/object`. The use of presigned URLs is recommended
-            for data access. Cloud native URLs are currently provided for a limited set of use cases and may not be
-            provided in the future. If cloud native URLs are required, please contact the data store team regarding the
-            credentials necessary to use them.
+          description: >
+            When set to true, the response will contain API-specific URLs that are tied to the specified replica,
+            for example `gs://bucket/object` or `s3://bucket/object`.
+
+            The use of presigned URLs is recommended for data access. Cloud native URLs are currently provided for a
+            limited set of use cases and may not be provided in the future. If cloud native URLs are required, please
+            contact the data store team regarding the credentials necessary to use them.
           type: boolean
       responses:
         301:
@@ -1457,10 +1459,12 @@ paths:
           in: query
           description: >
             When set to true, the response will contain API-specific URLs that are tied to the specified replica,
-            for example `gs://bucket/object` or `s3://bucket/object`. This parameter is mutually exclusive with the
-            presigned urls parameter. The use of presigned URLs is recommended for data access. Cloud native URLs are
-            currently provided for a limited set of use cases and may not be provided in the future. If cloud native
-            URLs are required, please contact the data store team regarding the credentials necessary to use them.
+            for example `gs://bucket/object` or `s3://bucket/object`.
+
+            This parameter is mutually exclusive with the presigned urls parameter. The use of presigned URLs is
+            recommended for data access. Cloud native URLs are currently provided for a limited set of use cases and
+            may not be provided in the future. If cloud native URLs are required, please contact the data store team
+            regarding the credentials necessary to use them.
           required: false
           type: boolean
         - name: presignedurls
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -368,7 +368,10 @@ paths:
         - name: directurl
           in: query
           required: false
-          description: When set to true, the response will contain API-specific URLs that are tied to the specified replica, for example `gs://bucket/object` or `s3://bucket/object`
+          description: When set to true, the response will contain API-specific URLs that are tied to the specified replica,
+            for example `gs://bucket/object` or `s3://bucket/object`
+            The use of presigned URLs is recommended for data access. Cloud native URLs are currently provided for a limited set of use cases and may not be provided in the future.
+            If cloud native URLs are required, please contact the data store team regarding the credentials necessary to use them.
           type: boolean
       responses:
         301:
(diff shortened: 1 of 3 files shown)


amarjandu:
@mikebaumann the native url that returns takes the form of `gs://org-hca-dss-checkout-test/blobs/9cdc9050cecf59381fed55a2433140b69596fc861bee55abeafd1f9150f3e2da.15684690e8132044f378b4d4af8a7331c8da17b1.7f54939b30ae7b6d45d473a4c82a41b0.114dee2c`

let me know if you think more in depth testing is required.

tests/test_file.py
```diff
@@ -382,6 +382,47 @@ def _test_file_get_latest(self, replica: Replica):
...
+
+        native_url = str(UrlBuilder()
+                         .set(path="/v1/files/" + file_uuid)
+                         .add_query("replica", replica.name)
+                         .add_query("directurl", "True"))
+        presigned_url = str(UrlBuilder()
+                            .set(path="/v1/files/" + file_uuid)
+                            .add_query("replica", replica.name))
+        with override_bucket_config(BucketConfig.TEST_FIXTURE):
+            native_resp_obj = self.assertGetResponse(
```
mikebaumann: Additional/different testing is needed. Let's talk about this.

dss-api.yml
```diff
@@ -365,6 +365,11 @@ paths:
           description: Token to manage retries.  End users constructing queries should not set this parameter.
           required: false
           type: string
+        - name: directurl
+          in: query
+          required: false
+          description: When set to true, the response will contain API-specific URLs that are tied to the specified replica
```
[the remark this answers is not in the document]
amarjandu: resolved

dss-api.yml
```diff
@@ -1447,7 +1452,7 @@ paths:
         - name: directurls
           in: query
           description: >
-            Include direct-access URLs in the response.  This is mutually exclusive with the presignedurls parameter.
+            Include API-specific direct-access URLs in the response.  This is mutually exclusive with the presignedurls parameter.
```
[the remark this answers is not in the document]
amarjandu: resolved

tests/test_file.py
amarjandu: Added tests with blob_size from native url

amarjandu: added tests for schema + checkout bucket in url

tests/test_file.py
```diff
@@ -382,6 +382,57 @@ def _test_file_get_latest(self, replica: Replica):
...
+            )
+
+            verify_headers = ['X-DSS-VERSION', 'X-DSS-CREATOR-UID', 'X-DSS-S3-ETAG', 'X-DSS-SHA256',
+                              'X-DSS-SHA1', 'X-DSS-CRC32C']
+            native_headers_verify = {k: v for k, v in native_resp_obj.response.headers.items() if k in verify_headers}
+            presigned_headers_verify = {k: v for k, v in resp_obj.response.headers.items() if k in verify_headers}
+            self.assertDictEqual(native_headers_verify, presigned_headers_verify)
+
+            self.assertTrue(
+                native_resp_obj.response.headers['Location'].split('//')[0].startswith(replica.storage_schema))
```
mikebaumann: nit: If you split on "://" then you could just check that `replica.storage_schema` equals the 0th chunk, instead of using startswith.

xbrianh:
Some comments:
  - even though this will allow the DSS to provide cloud native urls to objects in the checkout bucket, it does not provide the authorization to use them. 
  - on GCP, authorization is granted via a [whitelisted set of service accounts](https://github.com/HumanCellAtlas/data-store/blob/ab1de5ace2ab0d62d0eb3c54be572832b7c5d720/infra/buckets/gs.tf#L54-L59)
  - on AWS, it appears that authorization is not granted at all. Is this correct?

It may be confusing to provide URLs to users who are not authorize to use those URLs. Perhaps we should consider providing this information in our API spec, or not providing S3 native urls at all.

What do you think, @mikebaumann

mikebaumann:
Yes, I think which access methods (preseigned URLs, cloud native URLs) are available/supported for HCA data is a large and important HCA program level issue. As has been discussed recently within the DSS team, only supporting presigned URLs would avoid the complexities of providing cloud specific credentials for each replica, which would become even more complex with the introduction of controlled access data.

However, a key partner's system currently only supports cloud native URLs, and for the HCA program to make meaningful progress in the near-term, cloud native URLs must be provided at this point. This is why `GET /bundles/{uuid}` provides `directurls` and why a symmetric change for `GET /files/{uuid}` is being made here.

If the DSS enhances its presigned URLs to be suitable for workflow/analysis use (#2100), and our key partner(s) add support for them, the question then becomes whether presigned URLs are sufficient for all end-users of HCA data, or are there other contingents that still *require* cloud native URLs.

Given this uncertainty, I am for making this change (consistent with `GET /bundles/{uuid}`) and perhaps adding a note to the Swagger that the `directurl(s)` option may not be supported in the future (for `GET /files/{uuid}` or `GET /bundles/{uuid}`).

Constraining the current `directurl(s)` support to only the Google replica would further reduce exposure, yet given the current uncertainty, I am not sure it is worth the additional code complexity and development effort (albeit small) to do so at this point. If directurls are provided for the S3 replica and an end-user tries to use them but cannot, that may help to surface important use cases to consider. And, if it is a valid use case that should be supported, we could provide credentials, could we not? Perhaps a simple and symmetric approach that could be considered for AWS for the near-term would be to create a specific, limited access user account (analogous to a Google Service Account) that is whitelisted on the bucket and for which keys could be provided for read access, would that work?

mikebaumann:
@xbrianh/team, what do you think about including the following text in the Swagger descriptions for `directurl(s)`?

> The use of presigned URLs is recommended for HCA data access. Cloud native URLs are currently provided for a limited set of use cases and may not be provided in the future. If cloud native URLs are required, please contact the HCA data-store team regarding the credentials necessary to use them.

xbrianh:
@mikebaumann That description of directurl usage in the DSS looks great, I agree 100%. Perhaps we can add it to this PR and then merge.

mikebaumann:
I'm glad you like the text above, @xbrianh. Perhaps it should be generalized for the case of the HCA data-store being used by other projects. Other projects have have their own preferred access policies and end-up editing this description anyway, yet we could at least provide something generic to start with. How about:

> The use of presigned URLs is recommended for data access. Cloud native URLs are currently provided for a limited set of use cases and may not be provided in the future. If cloud native URLs are required, please contact the data store team regarding the credentials necessary to use them.

xbrianh:
That's a good point about generality, @mikebaumann. It may be helpful to projects forking the HCA data-store to avoid referring to the HCA specifically in our documentation, if we can avoid it. And it seems we mostly avoid it currently. Your updated description LGTM.

dss-api.yml
```diff
@@ -1452,7 +1455,11 @@ paths:
         - name: directurls
           in: query
           description: >
-            When set to true, the response will contain API-specific URLs that are tied to the specified replica, for example `gs://bucket/object` or `s3://bucket/object`. This parameter is mutually exclusive with the presigned urls parameter.
+            When set to true, the response will contain API-specific URLs that are tied to the specified replica,
+             for example `gs://bucket/object` or `s3://bucket/object`. This parameter is mutually exclusive with the presigned urls parameter.
+            The use of presigned URLs is recommended for data access.
```
mikebaumann: Ditto comment above.

amarjandu: reformatted yml, verified in presentation in Swagger UI page using http://editor.swagger.io/

PR #2144, opened 2019-05-15
Retry-After header values added for 500 responses.

DailyDreaming:
Edit:  Just returning the Retry-After header value as a string.

~I'm having trouble getting the Retry-After header value to return as an `int` (it returns as as string despite the code in this PR).  Some explanation:~

~This is the response given (at least 90%+ of the time) for 500, 502, 503, 504 errors:~ https://github.com/HumanCellAtlas/data-store/blob/master/chalice/app.py#L203

~Chalice sends headers using self.send_header to push it as a string into a buffer:~ https://github.com/aws/chalice/blob/d0a278086b6c02197057fb71af8c402c436091d3/chalice/local.py#L586~

~Deep down it uses the BaseHTTPRequestHandler:~ https://docs.python.org/3/library/http.server.html#http.server.BaseHTTPRequestHandler

~Specifically: BaseHTTPRequestHandler.send_header:~ https://github.com/python/cpython/blob/3.7/Lib/http/server.py#L510

~Where each header key-pair is appended as: ```self._headers_buffer.append(("%s: %s\r\n" % (keyword, value)).encode('latin-1', 'strict'))```~

~At which point the int type appears to be lost and it is encoded as a string.~


DailyDreaming:
Yes.

chalice/app.py
```diff
@@ -200,6 +189,15 @@ def maybe_fake_504() -> typing.Optional[chalice.Response]:
         res_headers.pop("Content-Length", None)
         res_headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains; preload"
         res_headers["X-AWS-REQUEST-ID"] = app.lambda_context.aws_request_id
+
+        if status_code in (requests.codes.server_error,         # 500 status code
+                           requests.codes.bad_gateway,          # 502 status code
+                           requests.codes.service_unavailable,  # 503 status code
+                           requests.codes.gateway_timeout       # 504 status code
+                           ) \
+                and not (app.current_request.method == 'POST' and path.startswith('/v1/bundles')):
```
amarjandu: can we define the checks at the top of the function?
```
def maybe_fakie_504:
  eligible_status_codes = (requests.codes.server_error,         # 500 status code
                           requests.codes.bad_gateway,          # 502 status code
                           requests.codes.service_unavailable,  # 503 status code
                           requests.codes.gateway_timeout )          # 504 status code
                          
  ignore_uri_types =  ('/v1/bundles')
  ignore_request_methods = ('POST')

# some time later
    if status code in eligible_status_codes: ... blah
```
not sure if this might be easier to read this way....
nit.

dss/error.py
```diff
@@ -81,7 +117,16 @@ def wrapper(*args, **kwargs):
             code = "read_only"
             title = "The DSS is currently read-only"
             stacktrace = ""
-            headers = {'Retry-After': 600}
+            headers = {'Retry-After': '600'}
+
+        # These errors may be returned by the chalice app before reaching here.
+        if status in (requests.codes.server_error,         # 500 status code
```
amarjandu: nit. might be useful to have a method that returns this object, used twice so far....

tests/test_api.py
```diff
@@ -12,15 +12,107 @@
...
 from dss.util import UrlBuilder
 from dss.util.version import datetime_to_version_format
-from tests.infra import DSSAssertMixin, DSSUploadMixin, DSSStorageMixin, TestBundle, testmode
+from dss.config import Replica, DeploymentStage
+from tests.infra import DSSAssertMixin, DSSUploadMixin, DSSStorageMixin, TestBundle, testmode, ExpectedErrorFields
 from tests.infra.server import ThreadedLocalServer
 from tests import get_auth_header
 
 
+class TestApiErrors(unittest.TestCase, DSSAssertMixin):
```
amarjandu: testmode?

tests/test_api.py
```diff
@@ -12,15 +12,107 @@
...
+        cls.app.start()
+        cls.app._chalice_app._override_exptime_seconds = 15.0
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.app.shutdown()
+
+    # all endpoints should return Retry-After headers for 50x responses, though we only test GET /bundle/{uuid} here
+
+    def test_retry_after_response_500_GET_bundle(self):
```
amarjandu: this could get refactored into 2 base functions, then iterate through a set of possibilities, easier read.

dss/error.py
```diff
@@ -81,7 +117,16 @@ def wrapper(*args, **kwargs):
             code = "read_only"
             title = "The DSS is currently read-only"
             stacktrace = ""
-            headers = {'Retry-After': 600}
+            headers = {'Retry-After': '600'}
+
+        # These errors may be returned by the chalice app before reaching here.
```
xbrianh: Can this occur for methods decorated with `dss_handler`?

dss/error.py
```diff
@@ -81,7 +117,16 @@ def wrapper(*args, **kwargs):
...
+        if status in (requests.codes.server_error,         # 500 status code
+                      requests.codes.bad_gateway,          # 502 status code
+                      requests.codes.service_unavailable,  # 503 status code
+                      requests.codes.gateway_timeout       # 504 status code
+                      ) \
+                and not (func.__qualname__.upper() == 'POST' and 'uuid' in kwargs):  # not POST /bundle/{uuid}/checkout
```
xbrianh: Can routing information from connexion be used here?

dss/error.py
```diff
@@ -37,6 +37,22 @@ def __init__(self, title: str = "User is not authorized to access this resource"
                          title,
                          *args, **kwargs)
 
+
+def maybe_fake_error(headers, code) -> bool:
```
xbrianh: Why were probabilistic errors moved here? I think the previous mechanism with a single error (504) seems sufficient.

DailyDreaming: First, to conform to the previous standard.  Since the mechanism is in place, it made sense to me to extend it.

Second, using `mock` doesn't raise the error here (where we want to test) but in `chalice/app.py`.

Third, using `mock` has difficulties.  Using one call throws the intended error and passes, but using a second call does not.

So I chose this because it's pretty straight forward, it uses an existing mechanism (which I need to use anyway to properly raise a `504`), and I think most importantly hits `dss/errors.py`, which is what should be propagating these errors.

dss/error.py
DailyDreaming: I can look into doing that.

dss/error.py
DailyDreaming: I believe so?  Since this was originally caused by `mock`, I'd have to find a different way to check.

I'm guessing it might return a 500 directly, if say, other components were down (e.g. API gateway) or were erroring/backed up for some other reason.

dss/error.py
xbrianh: Let's please not extend the random errors, and not raise random errors in `dss_handler`.

It's OK to omit the probabilistic chalice exception from these tests for now. We can address it in the future.

DailyDreaming: Alright then.

chalice/app.py
DailyDreaming: Should be addressed.  Let me know what you think.

dss/error.py
DailyDreaming: I've tested this a bit now and haven't encountered it hitting this portion of the code other than when using `mock`.

dss/error.py
DailyDreaming: Done.  Not sure why I didn't do that before, thanks for the catch.

chalice/app.py
```diff
@@ -200,6 +203,10 @@ def maybe_fake_504() -> typing.Optional[chalice.Response]:
         res_headers.pop("Content-Length", None)
         res_headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains; preload"
         res_headers["X-AWS-REQUEST-ID"] = app.lambda_context.aws_request_id
+
+        if include_retry_after_header(return_code=status_code, method=app.current_request.method, uri=path):
```
xbrianh: This appears to attach a `Retry-After` header to all responses except those listed in `include_retry_after_header`, including 200 responses.

dss/error.py
```diff
@@ -37,6 +36,26 @@ def __init__(self, title: str = "User is not authorized to access this resource"
                          title,
                          *args, **kwargs)
 
+
+def include_retry_after_header(return_code, method, uri):
+    uuid_pattern = '[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}'
+    bundle_checkout_pattern = f'/v1/bundles/{uuid_pattern}/checkout'
+    bundle_checkout = re.compile(bundle_checkout_pattern)
```
xbrianh: If we're going to use regex, it should probably be compiled once per Lambda cold start.

The uuid pattern can also be pulled in from `dss/storage/identifiers.py`

DailyDreaming: I just discovered that the uuid pattern in `dss/storage/identifiers.py` doesn't work, but the one I copy-pasted from the swagger.yml does.  >.<

chalice/app.py
DailyDreaming: sigh

DailyDreaming: Will fix.

dss/error.py
```diff
@@ -81,7 +101,10 @@ def wrapper(*args, **kwargs):
             code = "read_only"
             title = "The DSS is currently read-only"
             stacktrace = ""
-            headers = {'Retry-After': 600}
+            headers = {'Retry-After': '600'}
+
+        if include_retry_after_header(return_code=status, method=request.method, uri=request.path):
```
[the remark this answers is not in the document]
DailyDreaming: I implemented old testing that hit this, but you didn't want it to trigger errors inside of `dss/error.py` and I may have misunderstood, but thought you said to leave additional testing to a future ticket.

It will hit either `dss/error.py`, or `chalice/app.py` in different circumstances I believe.  Therefore, I'd strongly argue that we need both to be included as long as we have two error handling points (technically 3, since a 504 is it's own thing).

I'll try and make some fake endpoints in the test and see if that's adequate for testing.

chalice/app.py
```diff
@@ -121,7 +137,13 @@ def wrapper(*args, **kwargs):
                     chalice_response = future.result(timeout=time_remaining_s)
                     return chalice_response
                 except TimeoutError:
-                    return timeout_response()
+                    if determine_internal_method(method):
+                        # used for "/internal/{endpoint}" methods
+                        request_method, request_path = determine_internal_method(method)
```
xbrianh: `determine_internal_method` should not be called twice.

DailyDreaming: Fixed.

dss/error.py
```diff
@@ -49,12 +97,18 @@ def dss_exception_handler(e: DSSException) -> FlaskResponse:
...
         }))
 
+
+def determine_method_path(func):
+    if 'mock' in func.__name__:
+        return MOCK_METHOD, MOCK_PATTERN
+    return request.method, request.path
+
+
 def dss_handler(func):
```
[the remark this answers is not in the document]
DailyDreaming: I've avoided regex now, though in a different (simpler) way.

dss-api.yml
```diff
@@ -2055,7 +2055,7 @@ paths:
                   code:
                     type: string
                     description: Machine-readable error code.  The types of return values should not be changed lightly.
-                    enum: [unhandled_exception, illegal_arguments, not_found]
+                    enum: [unhandled_exception, illegal_arguments, not_found, service_unavailable]
```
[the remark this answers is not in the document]
DailyDreaming: Removed.

dss/error.py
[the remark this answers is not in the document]
DailyDreaming: I've avoided regex now, though in a different (simpler) way.

dss/error.py
```diff
@@ -81,7 +98,10 @@ def wrapper(*args, **kwargs):
             code = "read_only"
             title = "The DSS is currently read-only"
             stacktrace = ""
-            headers = {'Retry-After': 600}
+            headers = {'Retry-After': '600'}
```
xbrianh: It is not an error to use an integer.

chalice/app.py
```diff
@@ -121,7 +138,8 @@ def wrapper(*args, **kwargs):
                     chalice_response = future.result(timeout=time_remaining_s)
                     return chalice_response
                 except TimeoutError:
-                    return timeout_response()
+                    request_method, request_path = determine_path(method)
+                    return timeout_response(method=request_method, uri=request_path)
```
xbrianh: Consider moving the logic of `determine_path` into `timeout_response`, since that is the only place it's used.

dss/error.py
DailyDreaming: I wanted to keep it consistent.

chalice/app.py
DailyDreaming: Will do.

dss/error.py
DailyDreaming: It also errors if the header is added in `app.py` because swagger validates it:
```
ERROR:dss:Internal Error for <function get_chalice_app.<locals>.dispatch at 0x7fcb972786a8>
  File "/home/quokka/git/data-store/v3nv/lib/python3.6/site-packages/chalice/app.py", line 845, in _get_view_function_response
    self._validate_response(response)
  File "/home/quokka/git/data-store/v3nv/lib/python3.6/site-packages/chalice/app.py", line 873, in _validate_response
    if '\n' in value:
TypeError: argument of type 'int' is not iterable
```

I'd like to keep them all `str` rather than some `str` and some not.  What do you think?

xbrianh: str is fine

Issue #2145, opened 2019-05-16
For GS checkout caching, use `STANDARD` instead of `DURABLE_REDUCED_AVAILABILITY`


xbrianh:
Background information:
  - A test was performed using a Google bucket lifecycle policy set to delete all objects with storage class `STANDARD` in a multi-regional bucket. It was not obvious this would work since Google’s (often misleading) documentation indicated that `STANDARD` is an alias for `MULTI_REGIONAL`
  - However, it _did_ work: `STANDARD` objects were deleted while `MULTI_REGIONAL` objects remain
  - This is relevant for our meta-data checkout caching strategy, which uses GS bucket lifecycle rules for cache expiration.
  - This sort of storage class change, `MULTI_REGIONAL` -> `STANDARD`, makes GS bucket-to-bucket copies fast.

Related to #2124

mikebaumann:
Any hope of achieving similar checkout performance gains on AWS?
It seems large file checkouts on AWS currently take ~30 seconds.

xbrianh:
@mikebaumann I am not aware of S3 capabilities analogous to the apparent copy by reference behavior observed on GS. Unfortunately, unless new information comes to light, current S3 checkout performance is unlikely to be improved in the foreseeable future.

mikebaumann:
I am confident in the data from @xbrianh that this will resolve the problem.
All the same, I am willing/available to rerun my simple checkout performance test script that produced the data reported in #2124 and provide updated results. If interested, please ping me when this change reaches the DSS `integration` deployment.

xbrianh:
@mikebaumann We would be remiss to not verify our GS copy results. We'll ping you when this lands.

xbrianh:
@mikebaumann This change has landed in the DSS integration deployment.

Issue #2082, opened 2019-04-29
Formalize Brian's Test Scripts


amarjandu:
Review scripts brian created,
integrate them within gitlab jobs

Issue #2081, opened 2019-04-29
Formalize Trent's Scale Test to Gitlab


amarjandu:
go over scale testing located in https://github.com/HumanCellAtlas/data-store-scale-test
migrate it over as a job within allspark

PR #2152, opened 2019-05-17
Use STANDARD storage class to mark non-cached checkout objects

xbrianh:
addresses #2145

### DIFF
--- a/dss/stepfunctions/gscopyclient/implementation.py
+++ b/dss/stepfunctions/gscopyclient/implementation.py
@@ -69,9 +69,9 @@ def copy_worker(event, lambda_context):
             if not will_cache:
                 logger.info("Not caching %s with content-type %s size %s", self.source_key, content_type, self.size)
 
-            # TODO: DURABLE_REDUCED_AVAILABILITY is being phased out by Google; use a different method in the future
+            # TODO: STANDARD is being phased out by Google; use a different method in the future
             if not will_cache and is_dss_bucket(self.destination_bucket):
-                # the DURABLE_REDUCED_AVAILABILITY storage class marks (short-lived) non-cached files
+                # the STANDARD storage class marks (short-lived) non-cached files
                 dst_blob._patch_property('storageClass', 'STANDARD')
                 # setting the storage class explicitly seems like it blanks the content-type, so we add it back
                 dst_blob._patch_property('contentType', content_type)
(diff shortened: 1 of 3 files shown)


infra/buckets/gs.tf
```diff
@@ -41,7 +41,7 @@ resource google_storage_bucket dss_gs_checkout_bucket {
     }
     condition {
       age = "${var.DSS_BLOB_TTL_DAYS}"
-      matches_storage_class = ["DURABLE_REDUCED_AVAILABILITY"]
+      matches_storage_class = ["STANDARD"]
```
amarjandu: can we change this for the test buckets too?

xbrianh: The test bucket lifecycle rule is not based on storage class, so it doesn't need to be updated.

xbrianh:
@DailyDreaming Thanks for pointing those out. Fixed.

dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -69,10 +69,10 @@ def run(self) -> dict:
             if not will_cache:
                 logger.info("Not caching %s with content-type %s size %s", self.source_key, content_type, self.size)
 
-            # TODO: DURABLE_REDUCED_AVAILABILITY is being phased out by Google; use a different method in the future
+            # TODO: STANDARD is being phased out by Google; use a different method in the future
```
DailyDreaming: This TODO can be deleted now.

xbrianh: will fix

tests/test_checkout_caching.py
```diff
@@ -137,13 +137,13 @@ def test_google_cached_checkout_creates_multiregional_storage_type(self, mock_ch
     def test_google_uncached_checkout_creates_durable_storage_type(self, mock_check):
```
mikebaumann: nit: You could also update the test name, perhaps:
```
test_google_uncached_checkout_creates_uncached_storage_type
```

xbrianh: fixed

xbrianh:
@mikebaumann Performance testing is WIP, and should be handled as part of #2081 and #2082, although it would be useful, I think, to create a checklist of DSS operations that should undergo regular performance testing.

PR #2154, opened 2019-05-18
Change how content-type is propagated when caching.

DailyDreaming:
Hopefully addresses intermittent failures like this one: https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/31047 .

### DIFF
--- a/dss/stepfunctions/gscopyclient/implementation.py
+++ b/dss/stepfunctions/gscopyclient/implementation.py
@@ -60,6 +60,7 @@ def copy_worker(event, lambda_context):
             state = self.get_state_copy()
             src_blob = self.gcp_client.bucket(self.source_bucket).get_blob(self.source_key)
             dst_blob = self.gcp_client.bucket(self.destination_bucket).blob(self.destination_key)
+            src_blob.reload()
             content_type = src_blob.content_type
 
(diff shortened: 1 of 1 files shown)


dss/stepfunctions/gscopyclient/implementation.py
```diff
@@ -60,7 +60,7 @@ def run(self) -> dict:
             state = self.get_state_copy()
             src_blob = self.gcp_client.bucket(self.source_bucket).get_blob(self.source_key)
             dst_blob = self.gcp_client.bucket(self.destination_bucket).blob(self.destination_key)
-            content_type = src_blob._get_content_type(None)
+            content_type = src_blob.content_type
```
xbrianh: Google client libraries do not automatically populate blob properties.

Please add `src_blob.reload()` before this line.

xbrianh: Nevermind, @DailyDreaming  pointed out that `get_blob` [does reload for you](https://github.com/googleapis/google-cloud-python/blob/4166ddbf1f8b07c2a5e2db8f7955457b5cedd82d/storage/google/cloud/storage/bucket.py#L705)

PR #2155, opened 2019-05-18
Fetch content type from blobstore.

DailyDreaming:
Blobs can't be relied upon to already have `content-type` apparently.

### DIFF
--- a/dss/stepfunctions/gscopyclient/implementation.py
+++ b/dss/stepfunctions/gscopyclient/implementation.py
@@ -1,11 +1,11 @@
 import logging
+from cloud_blobstore.gs import GSBlobStore
 
 from dss import Config, Replica
 from dss.stepfunctions.lambdaexecutor import TimedThread
 from dss.storage.files import write_file_metadata
 from dss.util.async_state import AsyncStateError
 from dss.storage.checkout.cache_flow import should_cache_file, is_dss_bucket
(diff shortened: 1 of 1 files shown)


xbrianh:
The test uploads a blob with `content-type=None`. When this is the case, private method `src_blob._get_content_type(None)` returns default content-type of "application/octet-stream".

The public property `src_blob.content_type` reports content-type honestly. However, since content-type was None, an exception was raised in `should_cache_file`, which expects a string. Unfortunately this exception was swallowed by the threading logic from `TimedThread`, and the step function reported a succesful completion.

I have pushed a fix directly onto master replacing line 63 with `content_type = src_blob.content_type or "application/octet-stream"`. This PR is no longer needed.

DailyDreaming:
@xbrianh I believe it defaults to guessing the mimetype from the actual contents of the file: https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/blob.py#L723

Would a better solution here be to set the content-type further upstream so that both Google and AWS could be guaranteed the same content-type?  Is it possible that defaulting to "application/octet-stream" in one replica without knowing what the other replica chose might cause problems?

xbrianh:
The content-type is specified in the file metadata. Here I'm referring to the DSS entities that live under the `files` prefix in the main bucket. This is consistent across replicas.

PR #2162, opened 2019-05-20
Separate GCP and AWS collections paging testing.

DailyDreaming:


tests/test_collections.py
```diff
@@ -129,6 +129,7 @@ def fetch_collection_paging_response(self, codes, replica: str, per_page: int):
 
     def test_collection_paging(self):
         min_page = 10
+        time.sleep(4)  # wait for collection replicas to sync
         self.create_temp_user_collections(num=min_page + 1)  # guarantee at least one paging response
```
xbrianh: This solution may make this test less intermittent, but highlights a more fundamental issue.

Can test dependency on sync daemon success be removed by making `create_temp_user_collections` replica specific?

DailyDreaming: That sounds good to me.  Should make it shorter too and I'm always for shorter tests.

DailyDreaming: Done.

tests/test_collections.py
```diff
@@ -47,7 +47,7 @@ def setUpClass(cls):
         cls.contents = [cls.col_file_item] * 8 + [cls.col_ptr_item] * 8
         cls.uuid, cls.version = cls._put(cls, cls.contents)
         cls.invalid_ptr = dict(type="foo", uuid=cls.file_uuid, version=cls.file_version, fragment="/xyz")
-        cls.replicas = ('aws', 'gcp')
+        cls.replicas = ('aws', )  # only necessary to test in AWS
```
xbrianh: Is this variable only used for paging tests? If so, can we please change the variable name to `cls.paging_test_replicas`.

DailyDreaming: Done.

PR #2180, opened 2019-05-29
Add operation to trigger sync

### LINKED ISSUE #2171: Create sync command operation
Write a `dss-ops` command to initiate sync on DSS entities. This tool should use the sync daemon.

xbrianh:
This allows DSS operators to initiate sync for replica repair, and eventually replica standup.

The new functionality is:
  1. Adding an operational SQS queue entry point to the sync daemon, and
  2. Pushing sqs messages to the operational queue

This has been tested manually.

closes #2171

### DIFF
--- a/daemons/dss-sync-sfn/app.py
+++ b/daemons/dss-sync-sfn/app.py
@@ -87,7 +87,7 @@ def launch_from_forwarded_event(event, context):
     return executions
 
 # This entry point is for operator initiated replication
-@app.sqs_queue_subscriber("dss-sync-operation-" + os.environ['DSS_DEPLOYMENT_STAGE'], batch_size=1)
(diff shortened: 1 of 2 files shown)


daemons/dss-sync-sfn/app.py
```diff
@@ -85,6 +86,37 @@ def launch_from_forwarded_event(event, context):
             raise NotImplementedError()
     return executions
 
+# This entry point is for operator initiated replication
+@app.sqs_queue_subscriber("dss-sync-operation-" + os.environ['DSS_DEPLOYMENT_STAGE'], batch_size=1)
```
kislyuk: Consider increasing `batch_size` to amortize the startup cost in this SFN dispatcher. When triggered by an operator, the latency of aggregating a batch in the queue doesn't seem to be as much of an issue.

xbrianh: `batch_size` changed to default.

PR #2182, opened 2019-05-30
Remove indexer retries for non-existent objects

### LINKED ISSUE #2181: Indexer is handling large numbers of AWS bucket events for non-existent objects.
Bucket events for non-existent objects cause unhandled exceptions in the indexer, leading to slowdowns as these events are processed multiple times during the Lambda-SQS retry behavior.

xbrianh:
closes #2181

### DIFF
--- a/dss/index/indexer.py
+++ b/dss/index/indexer.py
@@ -11,6 +11,7 @@ from dss.storage.identifiers import BundleFQID, FileFQID, ObjectIdentifier, Obje
 from dss.util.time import RemainingTime
 from .backend import IndexBackend
 from .bundle import Bundle, Tombstone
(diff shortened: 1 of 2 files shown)


daemons/dss-notify-v2/.chalice/config.json
```diff
@@ -7,8 +7,44 @@
   "stages": {
     "dev": {
       "api_gateway_stage": "dev",
-      "tags": {},
-      "environment_variables": {}
+      "tags": {
```
DailyDreaming: This file was changed by accident.

xbrianh: sigh

PR #2185, opened 2019-05-30
[easy] Use the same regex for all uuids.

DailyDreaming:
Noticed this when testing.  Some of the uuids in the swagger differ from each other as well.  This ensures we only use version 4 RFC4122-compliant IDs.

We seem to have variations between `A-Za-z0-9` and `a-z0-9` in our swagger doc, and the regex in `storage/identifiers.py` only allows `A-Fa-f0-9`.

The actual standard (for version 4) is `a-f0-9`, which is none of these.  I'm wondering which regex makes sense, because I think they should all be the same regex.

I opted for `A-Za-z0-9` based on @kislyuk 's response.

### DIFF
--- a/dss/storage/identifiers.py
+++ b/dss/storage/identifiers.py
@@ -28,10 +28,6 @@ DSS_BUNDLE_TOMBSTONE_REGEX = re.compile(
 DSS_OBJECT_NAME_REGEX = re.compile(
     f"^({BUNDLE_PREFIX}|{FILE_PREFIX}|{COLLECTION_PREFIX})/({UUID_PATTERN})(?:\.({VERSION_PATTERN}))?(\.{TOMBSTONE_SUFFIX})?$")  # noqa
 
-# API endpoint patterns
-BUNDLE_CHECKOUT_PATTERN = f'/v1/bundles/{UUID_PATTERN}/checkout'
-BUNDLE_CHECKOUT_URI_REGEX = re.compile(BUNDLE_CHECKOUT_PATTERN)
-
 
 class ObjectIdentifierError(ValueError):
     pass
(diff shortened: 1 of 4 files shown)


DailyDreaming:
@amarjandu It wasn't much, I just posted a question in `#data-store-eng`.

tests/test_identifiers.py
```diff
@@ -0,0 +1,30 @@
...
+sys.path.insert(0, pkg_root)  # noqa
+
+from dss.storage.identifiers import UUID_REGEX
+from tests.infra import testmode
+
+
+@testmode.standalone
+class TestRegexIdentifiers(unittest.TestCase):
+    def test_REGEX_MATCHING(self):
+        chars = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
```
kislyuk: Consider `string.ascii_letters + string.digits`

dss/storage/identifiers.py
```diff
@@ -10,7 +10,7 @@
 # versioned tombstones are indexed after all bundles during a reindex operation.
 TOMBSTONE_SUFFIX = "dead"
 
-UUID_PATTERN = "[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}"
+UUID_PATTERN = "[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}"
```
xbrianh: Since this pattern is used for storage identifiers, should upper case be disallowed?

DailyDreaming: Sounds reasonable.

PR #2188, opened 2019-05-31
Miscellaneous fixes

kislyuk:


kislyuk:
Shell scripts should generally be agnostic to which directory they are run from. It's generally not a good idea to use `cd` in shell scripts at all, especially in functions, since you're changing global state.

If you do use `cd`, you should make sure you come back to the directory you came from, which is what `pushd` and `popd` are for. `cd ..` does not do that.

Issue #2176, opened 2019-05-29
Simple Usage Notifications


amarjandu:
- [ ] Number of invocations per Lambda over the past 24 hours
  - [ ] Cumulative runtime per Lamba over the past 24 hours

PR #2198, opened 2019-06-04
DSS Lambda Monitoring

amarjandu:
Creates Fargate Task that performs:
Aggregation of Lambda metrics from AWS CloudWatch logs.
Output of  Metrics to slack via Webhook.

Resolves: #2176


environment
```diff
@@ -47,6 +47,7 @@ AWS_DEFAULT_OUTPUT=json
 AWS_DEFAULT_REGION=us-east-1
 GCP_DEFAULT_REGION=us-central1
 GCP_PROJECT_NAME=human-cell-atlas-travis-test
+DSS_AVAILABILITY_ZONES="${AWS_DEFAULT_REGION}a ${AWS_DEFAULT_REGION}c ${AWS_DEFAULT_REGION}d"
```
DailyDreaming: This will fail for most regions that are not `us-east-1`.  Can this be written generically so that changing the region will allow it to work?

These also appear to be the zones used by allspark; any particular reason we don't use all zones?

infra/fargate/main.tf
```diff
@@ -0,0 +1,210 @@
...
+  cpu = "256"
+  memory = "512"
+  # TODO add logging below.
+  container_definitions = <<DEFINITION
+[
+  {
+    "family": "dss-monitor",
+    "name": "dss-monitor-lambda",
+    "image": "humancellatlas/dss-monitor-image",
+    "cpu": 256,
```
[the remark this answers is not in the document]
amarjandu: addressed

infra/fargate/main.tf
```diff
@@ -0,0 +1,210 @@
...
+# Route the public subnet trafic through the IGW
+resource "aws_route" "internet_access" {
+  route_table_id         = "${aws_vpc.dss_fargate.main_route_table_id}"
+  destination_cidr_block = "0.0.0.0/0"
+  gateway_id             = "${aws_internet_gateway.gw.id}"
+}
+
+
+
+resource "aws_ecs_cluster" "dss_monitor" {
```
xbrianh: Let's use the default cluster for now. We can add this back in the future if needed.
```
data "aws_ecs_cluster" "default" {
  cluster_name = "default"
}
```

infra/fargate/main.tf
```diff
@@ -0,0 +1,210 @@
...
+  availability_zone = "${local.availability_zones[count.index]}"
+  cidr_block = "${cidrsubnet("10.0.0.0/16", "4", count.index)}"
+}
+
+resource "aws_internet_gateway" "gw" {
+  vpc_id = "${aws_vpc.dss_fargate.id}"
+}
+
+# Route the public subnet trafic through the IGW
+resource "aws_route" "internet_access" {
```
xbrianh: The default vpc has an available route. This can be removed.

infra/fargate/task-definitions/lambda-stats.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+    "name": "dcp-monitor",
+    "image": "humancellatlas/dss-monitor-image",
+    "cpu": 10,
+    "memory": 512,
+    "essential": true,
+    "portMappings": [
```
xbrianh: These port mappings can be removed.

infra/fargate/task-definitions/lambda-stats.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+    "name": "dcp-monitor",
+    "image": "humancellatlas/dss-monitor-image",
+    "cpu": 10,
```
xbrianh: `cpu` and `memory` don't need to be defined at the container level.

infra/fargate/main.tf
```diff
@@ -0,0 +1,210 @@
...
+resource "aws_vpc" "dss_fargate" {
+  cidr_block = "10.0.0.0/16"
+  enable_dns_support = true
+  enable_dns_hostnames = true
+  tags = "${local.common_tags}"
+}
+
+# These subnets need to be associated with a route table providing
+# internet access.
+resource "aws_subnet" "dss_monitor" {
```
xbrianh: subnets can be looked up on the default vpc:
```
data "aws_availability_zones" "available" {}

data "aws_subnet" "default" {                                                                            
  count             = 3                                                                                  
  vpc_id            = "${data.aws_vpc.default.id}"                                                       
  availability_zone = "${data.aws_availability_zones.available.names[count.index]}"
  default_for_az    = true                                                                               
}
```

environment
xbrianh: Availability zones can be introspected in TF scripting. This variable can be removed.

infra/fargate/entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# default entry point for monitoring scripts
+
+git clone https://github.com/HumanCellAtlas/data-store.git
+cd data-store
+git fetch
+git checkout amar-monitorLiz
```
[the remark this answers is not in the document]
amarjandu: yes, shall be changed on before final merge, just here to test.

infra/fargate/task-definitions/lambda-stats.json
```diff
@@ -0,0 +1,15 @@
...
+    "memory": 512,
```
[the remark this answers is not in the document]
amarjandu: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/AWS_Fargate.html

the lowest from what i can see.

PR #2223, opened 2019-06-28
Large bundle manifest sync test

xbrianh:
This adds a sync test for a manifest containing 1000000 files.

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -12,21 +12,30 @@ import hashlib
 import unittest
 from unittest import mock
 import uuid
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
+import flask
 import boto3
 import crcmod
 from botocore.vendored import requests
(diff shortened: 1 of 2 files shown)


tests/test_sync.py
```diff
@@ -307,6 +316,81 @@ def check_s3_dest(dest_blob):
...
+        cls.app = ThreadedLocalServer()
+        cls.app.start()
+        cls.flask_app = flask.Flask(__name__)
+
+    @classmethod
+    def teardownClass(cls):
+        cls.app.shutdown()
+
+    def setUp(self):
+        dss.Config.set_config(dss.BucketConfig.NORMAL)
```
DailyDreaming: Should this be `dss.BucketConfig.TEST`?

xbrianh: This is an integration test of the deployed sync daemon, so it will be performed in the main bucket.

Issue #2073, opened 2019-04-24
Metadata files in GCP are stored (or served) without Content-Type header


hannes-ucsc:
This can be most easily reproduced in the CLI

```
host:temp hannes$ python3 -m venv .venv
host:temp hannes$ . .venv/bin/activate
(.venv) host:temp hannes$ pip install hca
Collecting hca
  Downloading https://files.pythonhosted.org/packages/4a/3d/468a35203dd9fb5e42c2952301d2c97317759e47cbc5a89f53932ece409b/hca-5.0.2-py2.py3-none-any.whl (86kB)
    100% |████████████████████████████████| 92kB 4.9MB/s
... 
OMITTED 
...
Successfully installed Jinja2-2.10.1 MarkupSafe-1.1.1 PyJWT-1.7.1 PyYAML-3.13 argcomplete-1.9.5 argparse-1.4.0 asn1crypto-0.24.0 atomicwrites-1.3.0 awscli-1.16.145 boto3-1.9.135 botocore-1.12.135 cachetools-3.1.0 certifi-2019.3.9 cffi-1.12.3 chardet-3.0.4 colorama-0.3.9 commonmark-0.7.5 crc32c-1.7 crcmod-1.7 cryptography-2.3.1 dcplib-1.6.5 docutils-0.14 future-0.17.1 google-auth-1.6.3 google-auth-oauthlib-0.3.0 hca-5.0.2 idna-2.8 jmespath-0.9.4 jsonpointer-1.14 jsonschema-2.6.0 oauthlib-3.0.1 puremagic-1.4 pyasn1-0.4.5 pyasn1-modules-0.2.5 pycparser-2.19 python-dateutil-2.8.0 requests-2.21.0 requests-oauthlib-1.2.0 rsa-3.4.2 s3transfer-0.2.0 six-1.12.0 tenacity-5.0.4 tweak-1.0.2 urllib3-1.24.2
You are using pip version 9.0.3, however version 19.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
(.venv) host:temp hannes$ python
Python 3.6.5 (default, Jun 17 2018, 12:13:06)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from hca.dss import DSSClient
>>> dss = DSSClient()
[...]

PR #2232, opened 2019-07-02
Don't change part layout when fixing object metadata

xbrianh:
This fixes a bug in storage operations that causes part layout to change when fixing metadata on multipart objects in s3.

connected to #2073


environment
```diff
@@ -54,26 +54,26 @@ DSS_S3_BUCKET_TEST=org-hca-dss-test
 DSS_S3_BUCKET_TEST_FIXTURES=org-hca-dss-test-fixtures
 DSS_S3_BUCKET_INTEGRATION=org-humancellatlas-dss-integration
 DSS_S3_BUCKET_STAGING=org-humancellatlas-dss-staging
-DSS_S3_BUCKET_PROD=org-humancellatlas-dss-prod
+DSS_S3_BUCKET_PROD=org-hca-dss-prod
```
DailyDreaming: Looks like this file might have been changed accidentally?

dss/operations/storage.py
```diff
@@ -146,12 +158,16 @@ def process_key(self, key):
                     update_aws_content_type(client, self.replica.bucket, blob_key, file_metadata['content-type'])
                 elif Replica.gcp == self.replica:
                     update_gcp_content_type(client, self.replica.bucket, blob_key, file_metadata['content-type'])
+            elif Replica.aws == self.replica:
+                blob_etag = self.handle.get_cloud_checksum(self.replica.bucket, blob_key)
+                if blob_etag != file_metadata['s3-etag']:
+                    update_aws_content_type(client, self.replica.bucket, blob_key, file_metadata['content-type'])
```
DailyDreaming: Should this `assert` that the new etag matches the metadata etag on the next line?

environment
xbrianh: will fix

dss/operations/storage.py
xbrianh: will add

Issue #2197, opened 2019-06-04
Return HMAC key, if present, during GET /subscription


xbrianh:
This is requested by some clients to improve subscription management.
See: https://github.com/HumanCellAtlas/data-store/issues/2190#issuecomment-497919188

PR #2237, opened 2019-07-04
/GET Subscriptions HMAC Key

amarjandu:
add `hmac_secret_key` to be returned on /GET Subscription for ES Queries. 

Addresses #2197


dss-api.yml
```diff
@@ -2295,6 +2295,10 @@ definitions:
           number of attempts is reached.
         type: string
         pattern: 'https?://'
+      hmac_secret_key:
+        description: >
+          Key used to verify authenticity of subscription
+          type: string
```
DailyDreaming: Accidental indent.

amarjandu: resolved

PR #2238, opened 2019-07-05
/FIND Subscriptions HMAC_SECRET_KEY

### LINKED ISSUE #2197: Return HMAC key, if present, during GET /subscription

amarjandu:
return `hmac_secret_key` in  get-subscriptions operation (find all for owner)
resolves #2197


dss/api/subscriptions_v1.py
```diff
@@ -59,7 +59,7 @@ def find(replica: str):
         'uuid': hit.meta.id,
         'replica': replica,
         'owner': owner,
-        **{k: v for k, v in hit.to_dict().items() if not k.startswith('hmac_')}}
+        **{k: v for k, v in hit.to_dict().items() if not k.startswith('hmac_key_id')}}
```
amarjandu: @xbrianh im not sure why there was a filter initially placed on the `hmac` data keys; if only subscriptions that are located/returned must belong to the owner. sub.v2 does not have this restriction.

xbrianh: I don't know why that restriction was put into place. Either @hannes-ucsc or @mikebaumann may be able to shed some light on this.

PR #2243, opened 2019-07-10
Adjust smoketest to wait a bit longer when downloading after a sync.

DailyDreaming:
This has been flaky lately.

### DIFF
--- a/tests/infra/base_smoketest.py
+++ b/tests/infra/base_smoketest.py
@@ -205,7 +205,7 @@ class BaseSmokeTest(unittest.TestCase):
                     except:
                         print(f'Waiting for {replica.name} bundle uuid: {bundle_uuid}')
                         time.sleep(1)
-                    else:
+                    if i == 2:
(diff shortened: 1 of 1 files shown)


tests/infra/base_smoketest.py
```diff
@@ -198,7 +198,12 @@ def _test_replica_sync(self, current_replica, bundle_uuid, ):
                 else:
                     parser.exit(RED(f"Failed to replicate bundle from {current_replica.name} to {replica.name}"))
             with self.subTest(f"{current_replica.name}: Download bundle from {replica}"):
-                run(f"{self.venv_bin}hca dss download --replica {replica.name} --bundle-uuid {bundle_uuid}")
+                for i in range(30):
+                    try:
+                        run(f"{self.venv_bin}hca dss download --replica {replica.name} --bundle-uuid {bundle_uuid}")
+                        break
+                    except:
+                        time.sleep(1)
```
amarjandu: can we throw a print statement here indicating there was an error?

xbrianh: The CLI has built in retries. If there is a problem, this may take much longer than 30 seconds.

PR #2246, opened 2019-07-15
Checkout cached files when PUT file is called.

DailyDreaming:


dss/storage/bundles.py
```diff
@@ -90,6 +85,8 @@ def save_bundle_manifest(replica: Replica, uuid: str, version: str, bundle: dict
     data = json.dumps(bundle).encode("utf-8")
     fqid = BundleFQID(uuid, version).to_key()
     created, idempotent = idempotent_save(handle, replica.bucket, fqid, data)
+    if should_cache_file(content_type='application/json', size=len(data)):
```
xbrianh: Files should have been pre-cached already in PUT /file, so we don't need to do it again here.

DailyDreaming: Removed.

tests/test_file.py
```diff
@@ -113,6 +122,43 @@ def _test_put_auth_errors(self, scheme, test_bucket):
...
+            handle = Config.get_blobstore_handle(replica)
+            src_key = generate_test_key()
+            src_data = b'{"status":"valid"}'
+            source_url = f"{scheme}://{test_bucket}/{src_key}"
+            file_uuid = str(uuid.uuid4())
+            bundle_uuid = str(uuid.uuid4())
+            version = datetime_to_version_format(datetime.datetime.utcnow())
+
+            # write dummy file and upload to upload area
+            with tempfile.NamedTemporaryFile(delete=True) as fh:
```
xbrianh: You can avoid using temporary files with
```
with io.BytesIO(src_data) as fh:
    uploader.checksum_and_upload_file("test_file_name", src_key, "application/json")
```

dss/api/files.py
```diff
@@ -329,5 +328,7 @@ class CopyMode(Enum):
                 f"file with UUID {uuid} and version {version} already exists")
         status_code = requests.codes.ok
 
-    return jsonify(
-        dict(version=version)), status_code
+    if should_cache_file(content_type=content_type, size=size):
+        start_file_checkout(replica=replica, blob_key=dst_key)
```
kislyuk: Does `start_file_checkout` react appropriately when the copy is asynchronous (CopyMode.COPY_ASYNC above)?

kislyuk:
This LGTM overall, but I'm worried about the situation when PUT file results in an async put (the request returns `202 accepted` and a SFN is kicked off to copy the file asynchronously). This is done when the file is larger than `ASYNC_COPY_THRESHOLD`. If a checkout is triggered before the async copy completes, what happens?

You can easily sidestep this problem for small files by only triggering checkout when `should_cache_file()` returns True **and** the file size is less than `ASYNC_COPY_THRESHOLD`. Fixing this behavior for large files could be postponed for expediency. One possible fix would be to let the copy step function kick off the checkout when it's done, when COPY_MODE is async.

dss/api/files.py
DailyDreaming: I've made the (temporary) check to account for this.

DailyDreaming:
@kislyuk I believe I've made the appropriate change.  Thanks for taking a look.  I'll follow up with handling the larger files in a separate PR.

PR #2258, opened 2019-07-16
Verify content type.

DailyDreaming:
Multipart file uploads weren't specifying the content-type and so this was creating mismatches.  Tested and working when the step function is manually deployed.

### DIFF
--- a/dss/stepfunctions/s3copyclient/implementation.py
+++ b/dss/stepfunctions/s3copyclient/implementation.py
@@ -66,6 +66,7 @@ def setup_copy_task(event, lambda_context):
         mpu = s3_blobstore.s3_client.create_multipart_upload(
             Bucket=destination_bucket,
             Key=destination_key,
+            Metadata=s3_blobstore.get_all_metadata(source_bucket, source_key),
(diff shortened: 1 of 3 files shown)


dss/stepfunctions/s3copyclient/implementation.py
```diff
@@ -63,7 +63,11 @@ def setup_copy_task(event, lambda_context):
     if part_count * part_size < source_size:
         part_count += 1
     if part_count > 1:
-        mpu = s3_blobstore.s3_client.create_multipart_upload(Bucket=destination_bucket, Key=destination_key)
+        mpu = s3_blobstore.s3_client.create_multipart_upload(
+            Bucket=destination_bucket,
+            Key=destination_key,
+            Metadata=s3_blobstore.get_all_metadata(source_bucket, source_key),
```
xbrianh: I'm including this out-of-date comment for posterity:
The cloud-blobstore method `get_user_metadata` merges object tagging with metadata, which is not what we want for this operation. We should use `get_all_metadata` instead.

btw, if the names are confusing we can make changes to cloud-blobstore to alleviate confusion.

PR #2260, opened 2019-07-17
Use a tempdir in smoketest when downloading.

DailyDreaming:


tests/infra/base_smoketest.py
```diff
@@ -183,7 +183,14 @@ def _test_get_subscriptions(self, replica, requested_subscription):
         list_of_subscription_uuids = [x['uuid'] for x in list_of_subscriptions if x['uuid']]
         self.assertIn(requested_subscription, list_of_subscription_uuids)
 
-    def _test_replica_sync(self, current_replica, bundle_uuid, ):
+    def _download_bundle(self, replica_name: str, bundle_uuid: str):
+        tempdir = tempfile.mkdtemp()
```
amarjandu: can the tempdir take the `cls.workdir` into context, and keep everything contained inside?

DailyDreaming: I suppose so.  I'll add it in.

xbrianh: Agreeing with Amar:
```
with tempfile.TemporaryDirectory() as tempdir:
    pass
```

From the [documentation](https://docs.python.org/3.6/library/tempfile.html#tempfile.TemporaryDirectory):
> On completion of the context or destruction of the temporary directory object the newly created temporary directory and all its contents are removed from the filesystem.

tests/infra/base_smoketest.py
```diff
@@ -183,7 +183,14 @@ def _test_get_subscriptions(self, replica, requested_subscription):
         list_of_subscription_uuids = [x['uuid'] for x in list_of_subscriptions if x['uuid']]
         self.assertIn(requested_subscription, list_of_subscription_uuids)
 
-    def _test_replica_sync(self, current_replica, bundle_uuid, ):
+    def _download_bundle(self, replica_name: str, bundle_uuid: str):
+        tempdir = tempfile.mkdtemp(prefix=f'{cls.workdir.name}/')
```
xbrianh: Agreeing with Amar:
```
with tempfile.TemporaryDirectory() as tempdir:
    pass
```

From the [documentation](https://docs.python.org/3.6/library/tempfile.html#tempfile.TemporaryDirectory):
> On completion of the context or destruction of the temporary directory object the newly created temporary directory and all its contents are removed from the filesystem.

DailyDreaming: Addressed.

tests/infra/base_smoketest.py
```diff
@@ -183,7 +183,12 @@ def _test_get_subscriptions(self, replica, requested_subscription):
         list_of_subscription_uuids = [x['uuid'] for x in list_of_subscriptions if x['uuid']]
         self.assertIn(requested_subscription, list_of_subscription_uuids)
 
-    def _test_replica_sync(self, current_replica, bundle_uuid, ):
+    def _download_bundle(self, replica_name: str, bundle_uuid: str):
+        with tempfile.TemporaryDirectory(prefix=f'{self.workdir.name}/') as tempdir:  # type: ignore
+            run(f"{self.venv_bin}hca dss download --replica {replica_name} --bundle-uuid {bundle_uuid}"  # type: ignore
```
xbrianh: Can we get rid of the `type: ignore`s?

DailyDreaming: MyPy says that `self.workdir` and `self.venv_bin` don't exist.

PR #2282, opened 2019-07-25
[Easy] Correctly detect CI in chalice build

xbrianh:
Use a GitLab CI variable to detect CI environment.


Bento007:
Should I fix this in Fusillade as well?

xbrianh:
@Bento007 Yes, I think so.

xbrianh:
@Bento007 Sorry, this was correct as is. Closing.

Issue #2270, opened 2019-07-23
Travis subscription clean up.


DailyDreaming:
Smoketest may not be cleaning up after itself.  There were 495 travis subscriptions sitting in `dev`.

PR #2304, opened 2019-07-30
Delete Test Subscription

amarjandu:
Deletes the subscriptions that the test_subscriptions make. 
Resolves: #2270

### DIFF
--- a/tests/test_subscriptions.py
+++ b/tests/test_subscriptions.py
@@ -162,6 +162,7 @@ class TestSubscriptionsBase(ElasticsearchTestCase, TestAuthMixin, DSSAssertMixin
         self.assertEqual(self.endpoint, Endpoint.from_subscription(json_response))
         self.assertEquals(self.hmac_key_id, json_response['hmac_key_id'])
         self.assertNotIn('hmac_secret_key', json_response)
+        self._delete_subscription(uuid=json_response['uuid'])
 
         # File not found request
         url = str(UrlBuilder()
@@ -175,8 +176,9 @@ class TestSubscriptionsBase(ElasticsearchTestCase, TestAuthMixin, DSSAssertMixin
 
     def test_find(self):
         num_additions = 25
+        uuids = list()
         for _ in range(num_additions):
-            self._put_subscription()
+            uuids.append(self._put_subscription())
         url = str(UrlBuilder()
                   .set(path="/v1/subscriptions")
                   .add_query("replica", self.replica.name)
@@ -191,6 +193,9 @@ class TestSubscriptionsBase(ElasticsearchTestCase, TestAuthMixin, DSSAssertMixin
         self.assertEqual(self.endpoint, Endpoint.from_subscription(json_response['subscriptions'][0]))
         self.assertNotIn('hmac_secret_key', json_response['subscriptions'][0])
         self.assertEqual(num_additions, len(json_response['subscriptions']))
+
+        for uuid in uuids:
+            self._delete_subscription(uuid=uuid)
 
     def test_delete(self):
         find_uuid = self._put_subscription()
@@ -226,6 +231,15 @@ class TestSubscriptionsBase(ElasticsearchTestCase, TestAuthMixin, DSSAssertMixin
             headers=get_auth_header()
         )
         return resp_obj.json.get('uuid')
+
+    def _delete_subscription(self, uuid, subscription_type=None):
+        if not subscription_type:
+            subscription_type = 'elasticsearch'
+        url = (UrlBuilder()
+               .set(path=f"v1/subscriptions/{uuid}")
+               .add_query("replica", self.replica.name)
+               .add_query('subscription_type', subscription_type))
+        self.assertDeleteResponse(url, requests.codes.ok, headers=get_auth_header())
 
 
 class TestGCPSubscription(TestSubscriptionsBase):
--- a/tests/test_subscriptions.py
+++ b/tests/test_subscriptions.py
@@ -167,7 +167,7 @@ class TestSubscriptionsBase(ElasticsearchTestCase, TestAuthMixin, DSSAssertMixin
             self.assertEquals(self.hmac_key_id, json_response['hmac_key_id'])
             self.assertNotIn('hmac_secret_key', json_response)
         finally:
-            self._cleanup_subscription(json_response['uuid'])
+            self._cleanup_subscription(find_uuid)
 
         # File not found request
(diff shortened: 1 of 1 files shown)


tests/test_subscriptions.py
```diff
@@ -162,6 +163,7 @@ def test_get(self):
         self.assertEqual(self.endpoint, Endpoint.from_subscription(json_response))
         self.assertEquals(self.hmac_key_id, json_response['hmac_key_id'])
         self.assertNotIn('hmac_secret_key', json_response)
+        self._delete_subscription(uuid=json_response['uuid'])
```
xbrianh: Please use a `try: finally:` block. This will allow the subscription to be deleted even if the assertions fail.

This should be applied throughout this PR.

Bento007: [addcleanup](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup) would be the preferred functions, and it should be used close to the beginning of the test.

tests/test_subscriptions.py
```diff
@@ -227,6 +233,15 @@ def _put_subscription(self, endpoint=None, expect_code=None, attachments=None):
         )
         return resp_obj.json.get('uuid')
 
+    def _delete_subscription(self, uuid, subscription_type=None):
```
Bento007: You put [addCleanup](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup) here if this function is only used for cleanup. If you do put it here, change the name of the function to `_cleanup_subscription`

tests/test_subscriptions.py
amarjandu: added

Bento007: They functionally accomplish the same thing. It's your preference.

tests/test_subscriptions.py
```diff
@@ -191,6 +195,8 @@ def test_find(self):
         self.assertEqual(self.endpoint, Endpoint.from_subscription(json_response['subscriptions'][0]))
         self.assertNotIn('hmac_secret_key', json_response['subscriptions'][0])
         self.assertEqual(num_additions, len(json_response['subscriptions']))
+        for _uuid in uuids:
+            self.addCleanup(self._cleanup_subscription, _uuid)
```
Bento007: This needs to put at the beginning of the function. I suggest around line 183. Alternatively you could put in `_put_subscription` so that every time a subscriptions is created a cleanup subscription is added to the test teardown.

tests/test_subscriptions.py
```diff
@@ -227,6 +233,15 @@ def _put_subscription(self, endpoint=None, expect_code=None, attachments=None):
...
         return resp_obj.json.get('uuid')
 
+    def _cleanup_subscription(self, uuid, subscription_type=None):
+        if not subscription_type:
+            subscription_type = 'elasticsearch'
+        url = (UrlBuilder()
+               .set(path=f"/v1/subscriptions/{uuid}")
+               .add_query("replica", self.replica.name)
+               .add_query('subscription_type', subscription_type))
+        self.assertDeleteResponse(url, requests.codes.okay, headers=get_auth_header())
```
Bento007: If you use addCleanup don't thrown an assertion in the cleanup, otherwise the rest of the teardown will fail.

tests/test_subscriptions.py
```diff
@@ -158,6 +159,8 @@ def test_get(self):
             requests.codes.okay,
             headers=get_auth_header())
         json_response = resp_obj.json
+        self.addCleanup(self._cleanup_subscription, json_response['uuid'])
```
Bento007: this should go before your first `assert` and preferable as soon as you create the subscription.

tests/test_subscriptions.py
amarjandu: ah, i dont like that, i wanna test that they all get deleted, if they are not deleting thats an issue, ill move back to the commit with `try: finally:`

tests/test_subscriptions.py
amarjandu: reverted to try:finally: need verification of deletion.

tests/test_subscriptions.py
```diff
@@ -146,22 +149,25 @@ def test_subscription_registration_succeeds_when_query_does_not_match_mappings(s
...
+                url,
+                requests.codes.okay,
+                headers=get_auth_header())
+            json_response = resp_obj.json
+            self.assertEqual(self.sample_percolate_query, json_response['es_query'])
+            self.assertEqual(self.endpoint, Endpoint.from_subscription(json_response))
+            self.assertEquals(self.hmac_key_id, json_response['hmac_key_id'])
+            self.assertNotIn('hmac_secret_key', json_response)
+        finally:
+            self._cleanup_subscription(json_response['uuid'])
```
DailyDreaming: If it's `self._cleanup_subscription(find_uuid)`, then it will be cleaned up if the `PUT` succeeds and the `GET` fails.

amarjandu: resolved, good catch, i guess it is possible for the assertEquals() check to fail here.....

tests/test_subscriptions.py
```diff
@@ -174,23 +180,28 @@ def test_get(self):
...
-            requests.codes.okay,
-            headers=get_auth_header())
-        json_response = resp_obj.json
-        self.assertEqual(self.sample_percolate_query, json_response['subscriptions'][0]['es_query'])
-        self.assertEqual(self.hmac_key_id, json_response['subscriptions'][0]['hmac_key_id'])
-        self.assertEqual(self.endpoint, Endpoint.from_subscription(json_response['subscriptions'][0]))
-        self.assertNotIn('hmac_secret_key', json_response['subscriptions'][0])
-        self.assertEqual(num_additions, len(json_response['subscriptions']))
+        try:
+            num_additions = 25
```
xbrianh: Is there rationale for making 25 subscriptions? Would 3 work?

amarjandu: I dont see why this would be an issue, i can change that.

DailyDreaming: @amarjandu Yesterday we wanted to limit them to 5 per user.  That's since increased, but they can be weighty.

amarjandu: @DailyDreaming what i ment to say in that was, Im not opposed to decreasing the number of subscriptions inserted, i dont see why it would be an issue from a testing perspective.

amarjandu: Updated `num_additions=3`

PR #2307, opened 2019-07-30
Use chalice request context

xbrianh:
Flask request context is not generally available during timeout exceptions. Request contextual information is passed into `timeout_response` from chalice instead.

### DIFF
--- a/chalice/app.py
+++ b/chalice/app.py
@@ -66,7 +66,7 @@ class DSSChaliceApp(chalice.Chalice):
         self._override_exptime_seconds = None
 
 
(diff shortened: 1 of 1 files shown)


chalice/app.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, *args, **kwargs):
         self._override_exptime_seconds = None
 
 
-def timeout_response(method, uri: str) -> chalice.Response:
+def timeout_response(method: str=None, path: str=None) -> chalice.Response:
```
DailyDreaming: Nit: I would probably require using `(method: str, path: str)` here.

PR #2309, opened 2019-07-30
Limit user v1 subscriptions to 100 per replica.

DailyDreaming:


dss-api.yml
```diff
@@ -673,9 +673,12 @@ paths:
     put:
       security:
         - dcpAuth: []
-      summary: Create a event subscription.
+      summary: Create an event subscription.
       description: >
         Register an HTTP endpoint that is to be notified when a given event occurs.
+
+        Each user is allowed 100 subscriptions, a limit that may be increased in the future. Concerns about
+        notification service limitations should be routed to the DSS development team.
```
xbrianh: @kislyuk Do you have any comments on our documentation of subscription limitations?

DailyDreaming:
Rerunning tests now that indexing is reset and back up before merging.

dss/api/subscriptions_v1.py
```diff
@@ -16,8 +15,8 @@
 from dss.notify import attachment
 from dss.util import security
 
-
 logger = logging.getLogger(__name__)
+SUBSCRIPTION_LIMIT = 100
```
Bento007: This value should live in DSS.Config.

PR #2310, opened 2019-07-30
Limit a user's v2 subscriptions to 100 per replica.

DailyDreaming:


dss/api/subscriptions_v2.py
```diff
@@ -39,6 +40,11 @@ def find(replica: str):
 
 @security.authorized_group_required(['hca', 'public'])
 def put(json_request_body: dict, replica: str):
+    owner = security.get_token_email(request.token_info)
+    if len([s for s in get_subscriptions_for_owner(Replica[replica], owner) if owner == s['owner']]):
```
xbrianh: Instead of retrieving the entire list of subscriptions, please look into `Select COUNT`: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html#DDB-Query-request-Select

This may require extending the DSS DynamoDB interface, as well as the v2 subscriptions interface.

tests/test_subscriptions.py
```diff
@@ -75,6 +77,7 @@ def setUp(self):
         self.sample_percolate_query = smartseq2_paired_ends_vx_query
         self.hmac_key_id = 'dss_test'
         self.hmac_secret_key = '23/33'
+        self.replicas = ['aws', 'gcp']
```
chmreid: Is there a programmatic way to get all replica names from a Replica object?

dss/api/subscriptions_v2.py
```diff
@@ -29,16 +31,21 @@ def get(uuid: str, replica: str):
...
+        if 'hmac_secret_key' in subscription:
+            subscription.pop('hmac_secret_key')
+    return {'subscriptions': subscriptions}, requests.codes.ok
 
 
 @security.authorized_group_required(['hca', 'public'])
 def put(json_request_body: dict, replica: str):
+    owner = security.get_token_email(request.token_info)
+    if count_subscriptions_for_owner(Replica[replica], owner) > SUBSCRIPTION_LIMIT:
+        raise DSSException(requests.codes.not_acceptable, "not_acceptable",
```
xbrianh: Does this error code need to be included in `dss-api.yml`?

DailyDreaming: Good point.  Will update.

tests/test_subscriptions.py
xbrianh: Yes:
```
self.replicas = [replica.name for replica in Replica]
```

amarjandu: `self.replicas = [replica.name for replica in Replica]` 🤔

DailyDreaming: There is.  Will add it.  Thanks for catching that.

Issue #2292, opened 2019-07-29
Bundle Enumeration Endpoint


xbrianh:
Bundle enumeration endpoints should be implemented according to https://github.com/HumanCellAtlas/dcp-community/blob/bhannafi-dss-bundle-enumeration.md/rfcs/text/0000-dss-bundle-enumeration.md

PR #2318, opened 2019-08-01
Bundle: enumeration

amarjandu:
Adds ability to enumerate through primary bucket for bundles, filters out tombstoned bundles.
Adds ability to specify prefix for enumeration, allowing a more specific list of bundles to be returned.

Closes: #2292


dss/api/bundles/__init__.py
```diff
@@ -132,6 +136,50 @@ def get(
         response.headers['Link'] = link
         return response
 
+@dss_handler
+def enumerate(replica: str, prefix: str, token: str = None,
+              per_page: int = PerPageBounds.per_page_max,
+              search_after: typing.Optional[str] = None):
+
+    storage_handler = Config.get_blobstore_handle(Replica[replica])
+    key_prefix = prefix if prefix else 'bundles/'  # might need to add filtering over here (no files/blobs allowed)
```
xbrianh: This should only list bundles. The purpose of the prefix is to allow parallelized bundle listings.
```
key_prefix = f"bundles/{prefix}"
```

dss/api/bundles/__init__.py
```diff
@@ -132,6 +136,50 @@ def get(
...
+    """Does this even need to have prefix? the only option is bundles, only in the checkout bucket does the
+        Bundles/ prefix actually have files, and thats after a --presigned url, otherwise contents after bundles/ are not
+        exposed........"""
+    kwargs = dict(bucket=Replica[replica].bucket, prefix=key_prefix, k_page_max=per_page)
+    if search_after:
+        kwargs['start_after_key'] = search_after
+    if token:
+        kwargs['token'] = token
+    prefix_iterator = storage_handler.list_v2(**kwargs)
```
xbrianh: `list_v2` accepts an argument to limit the number of keys returned: `k_page_max`. This can be used to avoid the `StopIteration` logic below.

amarjandu:
clou-blobstore needs to be updated to always have the token value returned.....

dss-api.yml
```diff
@@ -1433,6 +1433,62 @@ paths:
...
+      parameters:
+        - name: replica
+          in: query
+          description: Replica to fetch from.
+          required: true
+          type: string
+          enum: [aws, gcp]
+        - name: prefix
+          in: query
+          description: Prefix to for searching through specific keys
```
chmreid: typo "Prefix to for"

dss-api.yml
```diff
@@ -1433,6 +1433,62 @@ paths:
                     enum: [unhandled_exception, Forbidden, Unauthorized, OAuthResponseProblem, not_found, read_only]
                 required:
                   - code
+  /bundles:
+    get:
+      operationId: dss.api.bundles.enumerate
+      summary: Retreive all bundles or bundle contents within a primary bucket.
```
natanlao:
```suggestion
      summary: Retrieve all bundles or bundle contents within a primary bucket.
```

dss/storage/bundles.py
```diff
@@ -119,3 +120,41 @@ def _latest_version_from_object_names(object_names: typing.Iterator[str]) -> str
...
+    prefix_iterator = storage_handler.list_v2(**kwargs)
+    keys = dict()  # type: dict
+    uuid_list = list()
+    total_keys = 0
+
+    for key, meta in prefix_iterator:
+        uuid, version = key.split('.', 1)
+        if not version.endswith(TOMBSTONE_SUFFIX):
+            search_after = key
+            if keys.get(uuid) is None:
```
natanlao: keys.setdefault(uuid, []).append(version)

dss/api/bundles/__init__.py
```diff
@@ -132,6 +133,40 @@ def get(
         response.headers['Link'] = link
         return response
 
+@dss_handler
+def enumerate(replica: str, prefix: str = 'bundles/', token: str = None,
```
chmreid: `token: str` should be `token: typing.Optional[str]` since it can take on a None value

dss/storage/bundles.py
```diff
@@ -119,3 +120,41 @@ def _latest_version_from_object_names(object_names: typing.Iterator[str]) -> str
             version = current_version
 
     return version
+
+
+def list_available_uuids(replica: str = None, prefix: str = None, per_page: int = PerPageBounds.per_page_max,
```
chmreid: types for `prefix`, `search_after`, and `token` should be `typing.Optional[str]`

dss-api.yml
amarjandu: thx!

dss/storage/bundles.py
amarjandu: yikes, i think my merge got messed up......... this file was not supposed to be committed to change :(

dss/api/bundles/__init__.py
```diff
@@ -132,6 +133,41 @@ def get(
...
+              per_page: int = PerPageBounds.per_page_max,
+              search_after: typing.Optional[str] = None):
+
+    if prefix and not prefix.startswith('bundles/'):
```
DailyDreaming: Should prefix be optional?  We could ensure the user doesn't input a bad prefix.  For example, something like:

```
from dss.storage.identifiers import BUNDLE_PREFIX


def enumerate(replica, prefix=None, args, ...):
    search_prefix = f'{BUNDLE_PREFIX}/{prefix}' if prefix else f'{BUNDLE_PREFIX}/'
    ... other stuff ...

File: dss-api.yml [hunk 9]
```diff
@@ -1433,6 +1433,62 @@ paths:
...
+            $ref: '#/definitions/SearchResult'
+        206:
+          description: A single page of results was retrieved.
+          schema:
+            $ref: '#/definitions/SearchResult'
+          headers:
+            Link:
+              type: string
+              description: URL to retrieve the next page of results.
+        default:
```
DailyDreaming: This should include 400, 500, 502, 503, and 504 errors.

dss/api/bundles/__init__.py
```diff
@@ -132,6 +133,41 @@ def get(
         response.headers['Link'] = link
         return response
 
+@dss_handler
+def enumerate(replica: str, prefix: str = 'bundles/', token: str = None,
```
DailyDreaming: Consider using `from dss.storage.identifiers import BUNDLE_PREFIX`.

dss/api/bundles/__init__.py
```diff
@@ -132,6 +133,41 @@ def get(
...
+              per_page: int = PerPageBounds.per_page_max,
```
DailyDreaming: Consider adding a docstring to describe the use of `prefix`, `token`, `search_after`, etc.

dss/api/bundles/__init__.py
```diff
@@ -132,6 +133,41 @@ def get(
...
+        response = make_response(jsonify(payload), requests.codes.ok)
+    else:
+        next_url = UrlBuilder(request.url)
+        next_url.replace_query("search_after", payload['search_after'])
+        next_url.replace_query("token", payload['token'])
+        link = f"<{next_url}>; rel='next'"
+        payload.update(dict(has_more=True, token=payload['token'], link=f'{next_url}'))
+        response = make_response(jsonify(payload), requests.codes.partial)
+        response.headers['Link'] = link
+    return response
```
DailyDreaming: Should return a status code too.  Something like `return jsonify(response_body), status_code`.

dss/api/bundles/__init__.py
DailyDreaming: This.  ^

dss/api/bundles/__init__.py
DailyDreaming: I just realized that I made a similar comment.  @amarjandu Please address comments like these before asking for additional reviews.

DailyDreaming:
@amarjandu It should also return `"uuid": "uuid"` rather than `"uuid": "bundles/uuid"`

dss/storage/bundles.py
amarjandu: complete

dss/storage/bundles.py
```diff
@@ -119,3 +120,41 @@ def _latest_version_from_object_names(object_names: typing.Iterator[str]) -> str
...
+    prefix_iterator = storage_handler.list_v2(**kwargs)  # note dont wrap this in enumerate() it looses the token
+    keys = dict()  # type: dict
+    uuid_list = list()
+    total_keys = 0
+
+    for key, meta in prefix_iterator:
+        uuid, version = key.split('.', 1)
+        if not version.endswith(TOMBSTONE_SUFFIX):
+            search_after = key
+            keys.setdefault(uuid.strip(f'{BUNDLE_PREFIX}/'), []).append(version)
```
natanlao: `uuid.strip(..)` should be replaced with a splice if the bundle prefix is always expected to be present, since `str.split` doesn't work the way one might expect. The [pydoc](https://docs.python.org/3/library/stdtypes.html#str.strip) says it better than I can.

For example:

```
>>> from dss.storage.identifiers import BUNDLE_PREFIX
>>> test_uuid = "bundles/bbbebbbb-2101-4e52-94d9-29eebebebebe"
>>> test_uuid.strip(f'{BUNDLE_PREFIX}/')
'-2101-4e52-94d9-29'
```

Were `uuid` to contain a character in `BUNDLE_UUID` (i.e. `b` or `e`) in the wrong place, unexpected behavior would occur.

dss/storage/bundles.py
```diff
@@ -119,3 +120,41 @@ def _latest_version_from_object_names(object_names: typing.Iterator[str]) -> str
...
+            total_keys += 1
+        elif TOMBSTONE_SUFFIX == version:
+            if keys.get(uuid) is not None:
```
natanlao: This is a nitpick, but is it possible for `keys[uuid]` to be set to `None`? I am not familiar enough with the code to say. If not, then `uuid in keys` would be a clearer membership test.

dss-api.yml
```diff
@@ -1449,6 +1449,70 @@ paths:
                     enum: [unhandled_exception, Forbidden, Unauthorized, OAuthResponseProblem, not_found, read_only]
                 required:
                   - code
+  /bundles:
+    get:
+      operationId: dss.api.bundles.enumerate
+      summary: List through all bundles available to the data-store
+      description: >
+        Given a prefix
```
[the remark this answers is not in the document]
amarjandu: done

dss/storage/bundles.py
amarjandu: nice catch! addressed.

dss-api.yml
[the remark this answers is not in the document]
xbrianh: Please provide a more complete description.

dss-api.yml
```diff
@@ -1449,6 +1449,70 @@ paths:
...
+      parameters:
+        - name: replica
+          in: query
+          description: Replica to fetch from.
+          required: true
+          type: string
+          enum: [aws, gcp]
+        - name: prefix
+          in: query
+          description: used to search for keys under a specific "bundle/" prefix
```
xbrianh: Please provide a description that does not include the jargon words "keys" or "bundle/".

dss-api.yml
```diff
@@ -1449,6 +1449,70 @@ paths:
...
+        - name: token
+          in: query
+          description: Token to manage retries.  End users constructing queries should not set this parameter.
+          required: false
+          type: string
+        - name: search_after
+          in: query
+          description: >
+            **Search-After-Context**.
+            An internal state pointer parameter for use with pagination. This parameter is used as a reference to what information the previous page has, and what needs to be returned next. This value should not be set by the user.
```
xbrianh: Please use the same description provided in the POST /search endpoint.

dss/storage/bundles.py
xbrianh: Has this been addressed? `split` should be used instead of `strip`

dss/storage/bundles.py
```diff
@@ -119,3 +120,41 @@ def _latest_version_from_object_names(object_names: typing.Iterator[str]) -> str
             version = current_version
 
     return version
+
+
+def list_available_uuids(replica: str = None,
+                         prefix: typing.Optional[str] = None,
+                         per_page: int = PerPageBounds.per_page_max,
+                         search_after: typing.Optional[str] = None,
+                         token: typing.Optional[str] = None):
```
xbrianh: This method does not return a list of `uuid`s. Please think of a more descriptive name.

amarjandu: complete

dss-api.yml
amarjandu:
```
- name: prefix
          in: query
          description: Is used to specify the beginning of a particular bundle UUID. Specifying a character(s) would provide all avaliable bundles starting with that character(s). 
          required: false
          type: string
```

how is this?

dss-api.yml
amarjandu: complete

amarjandu:
> The API endpoint for bundle enumeration deserves more consideration. For instance, GET /collections returns collections owned by the current user. To be consistent, GET /bundles should return bundles owned by the current user.
> 
> Consider moving the bundle enumeration endpoint to GET /bundles/list_all, or similar. @DailyDreaming @kislyuk Do you have opinions?

 I would want to have the endpoint for enumeration kept at `v1/bundles`  to perform a list_all operation, while using `v1/bundles/:user_id` to perform an enumeration on bundles owned by a `user_id`. 

By having the endpoint be changed to `v1/bundles/list_all`; seems like we are introducing a move away from using a hierarchical identifier in the URI path.. ex `v1/bundles/:uuid`

This however would require a change to be made to the `collections` endpoint, allowing enumeration on `v1/collections` and user-collection enumeration on `v1/collections/:user_id`.

Changing the API this way would however put `v1/subscriptions` in a weird spot, but because subscriptions can be seen as more private, the change might not need to be applied here.  

with that being said, I think there should be a change made before this is merged in, and perhaps some feedback to the proposal above.

dss/api/bundles/__init__.py
amarjandu: addressed

dss/storage/bundles.py
amarjandu: yeah, the commit history was not up-to-date when your review came in, but its been changed.

amarjandu:
I've opted to use `/v1/bundles/list-all` as the end point, it will allow us to have an easier experience later on with enumerating bundles owned by a user in the CLI, and within the API.

tests/test_smoketest.py
```diff
@@ -122,6 +128,24 @@ def smoketest(self, starting_replica, checkout_bucket, test_bucket):
                 else:
                     self.fail("Timed out waiting for notification to arrive")
 
+        for replica in self.replicas:
+            with self.subTest(f'Testing Bundle Enumeration on {replica.name}'):
+                enumerate_bundles = list()
+                page_size = 10
+                first_page = self.get_bundle_enumerations(replica.name, page_size)
+                self.assertEquals(first_page['per_page'], 10)
```
natanlao: `assertEquals` is a deprecated alias
```suggestion
                self.assertEqual(first_page['per_page'], 10)
```

natanlao: Does this test pass on Python 3.x? It should fail as of 3.2, I thought...

dss/storage/bundles.py
```diff
@@ -119,3 +120,45 @@ def _latest_version_from_object_names(object_names: typing.Iterator[str]) -> str
             version = current_version
 
     return version
+
+
+def enumerate_avaliable_bundles(replica: str = None,
```
xbrianh: I'd prefer the contents of this method to be inlined into `dss.bundles.__init__.py enumerate`. You can consider a nested method.

This can be handled in a future PR.

tests/test_bundle.py
```diff
@@ -984,6 +984,12 @@ def test_patch_name_collision(self):
         )
         self.assertEqual(res.status_code, requests.codes.bad_request)
 
+    def test_enumeration_bundles(self):
+        bundle_uuid, bundle_version = self._put_bundle()
```
xbrianh: This test may take a long time to run if a larger number of bundles are present in the test bucket. Consider using a random UUID of the format `00000000-xxx-xxxx-xxxx-xxxxxxxxxxxx`, where the x's are random hex digits. This should help the bundle appear near the beginning of the results.

Listing should halt when:
1. The bundle is found
2. a non-zeros prefix is encountered in the UUID (this will occur upon test failure.

xbrianh: This change should probably be made in the smoketest as well.

dss/storage/bundles.py
```diff
@@ -119,3 +120,45 @@ def _latest_version_from_object_names(object_names: typing.Iterator[str]) -> str
...
+                                prefix: typing.Optional[str] = None,
+                                per_page: int = PerPageBounds.per_page_max,
+                                search_after: typing.Optional[str] = None,
+                                token: typing.Optional[str] = None):
+    """
+    :return: returns dictionary with bundles that are available, provides context of cloud providers internal pagination
```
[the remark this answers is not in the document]
amarjandu: done

tests/test_smoketest.py
```diff
@@ -122,6 +128,24 @@ def smoketest(self, starting_replica, checkout_bucket, test_bucket):
...
+                self.assertIs(first_page['has_more'], True)
```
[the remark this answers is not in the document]
amarjandu: done

tests/test_smoketest.py
amarjandu: it passes, i changed it tho

dss/storage/bundles.py
amarjandu: addressed

dss-api.yml
```diff
@@ -1449,6 +1449,73 @@ paths:
...
+            directly; it should instead directly fetch the URL given in the `Link` header.
+          required: false
+          type: string
+      responses:
+        200:
+          description: All results retrieved.
+          schema:
+            $ref: '#/definitions/BundleEnumerationResult'
+        206:
+          description: A single page of results was retrieved.
```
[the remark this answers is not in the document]
amarjandu: done

dss/storage/bundles.py
amarjandu: had to revert this, it was breaking unit tests.

PR #2322, opened 2019-08-05
Tagging Updates: Name

amarjandu:
change resource tagging `Name` to reference sub-component of the DSS
adds labels to gcp buckets

### DIFF
--- a/daemons/build_deploy_config.sh
+++ b/daemons/build_deploy_config.sh
@@ -38,7 +38,7 @@ cat "$config_json" | jq ".stages.$stage.layers=[env.layer_version_arn]" | sponge
 
 export DEPLOY_ORIGIN="$(whoami)-$(hostname)-$(git describe --tags --always)-$(date -u +'%Y-%m-%d-%H-%M-%S').deploy"
 cat "$config_json" | jq ".stages.$stage.tags.DSS_DEPLOY_ORIGIN=\"$DEPLOY_ORIGIN\" | \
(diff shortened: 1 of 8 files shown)


daemons/build_deploy_config.sh
```diff
@@ -13,6 +13,7 @@ if [[ -z $DSS_DEPLOYMENT_STAGE ]]; then
 fi
 
 export daemon_name=$1
+export daemon_tag=$(echo $daemon_name | cut -d '-' -f2-)
```
DailyDreaming: Why do you edit the daemon name instead of just using `daemon_name`?

amarjandu: its to strip out dss from the name to avoid `dss-dss-operations-dev` as the Name.

infra/buckets/s3.tf
```diff
@@ -3,7 +3,7 @@ data "aws_caller_identity" "current" {}
 locals {
   common_tags = "${map(
     "managedBy" , "terraform",
-    "Name"      , "${var.DSS_INFRA_TAG_PROJECT}-${var.DSS_DEPLOYMENT_STAGE}-${var.DSS_INFRA_TAG_SERVICE}",
+    "Name"      , "${var.DSS_INFRA_TAG_SERVICE}-cloudstorage",
```
xbrianh: Can we specify which cloud?, e.g. something like `-s3-storage` and `-gs-storage`

amarjandu: added tagging to google + differentiation of cloud

PR #2325, opened 2019-08-05
Add secrets management to dss-ops script

### LINKED ISSUE #2296: Move secrets handling into dss/operations
The secrets handling scripts `scripts/set_secret.py` and `fetch_secret.sh` should be moved into a new operational framework `dss/operations/secrets.py`.

This work should result in three commands:
  - [x] `scripts/dss-ops.py secrets list-secrets`
  - [x] `scripts/dss-ops.py secrets get-secret --secret-name`
  - [x] `scripts/dss-ops.py secrets set-secret --secret-name`. This should accept data from stdin

chmreid:
This pull request adds functionality to dss-ops that allows it to manage secrets. There are three verbs added:

* `dss-ops.py secrets list` to list secrets (names only)
* `dss-ops.py secrets get --secret-name <X>` to get a secret variable's value
* `dss-ops.py secrets set --secret-name <X>` to set a secret variable's value
* `dss-ops.py secrets delete --secret-name <X>` to delete a secret variable from the secrets manager

The `set` action can take input piped in from stdin.

This PR closes #2296

### DIFF
--- a/dss/operations/secrets.py
+++ b/dss/operations/secrets.py
@@ -1,0 +1,103 @@
+"""
+Get/set secret variable values from the AWS Secrets Manager
+"""
+import os
+import sys
+import click
+import boto3
+import select
+import typing
+import argparse
+import json
+import logging
+import subprocess
+
+from dss.operations import dispatch
+
+
+logger = logging.getLogger(__name__)
+
+
+events = dispatch.target("secrets",
+                         help=__doc__)
+
+
+@events.action("list-secrets")
+def list_secrets(argv: typing.List[str], args: argparse.Namespace):
+    """Print a list of all secrets"""
+    pass
+
+@events.action("get-secret",
+               arguments={
+                   "--secret-name": dict(
+                       required=True,
+                       nargs="*",
+                       help="name of secret to retrieve")})
+def get_secret(argv: typing.List[str], args: argparse.Namespace):
+    """Get the value of the secret variable specified by the --secret-name flag"""
+    SM = boto3.client('secretsmanager')
+    stage = os.environ['DSS_DEPLOYMENT_STAGE']
+    secrets_store = os.environ['DSS_SECRETS_STORE']
+    secret_id = f'{secrets_store}/{stage}/{args.secret_name}'
+
+    try:
+        # Start by trying to get the secret variable
+        secret_val = SM.get_secret_value(SecretId=secret_id)
+
+    except SM.exceptions.ResourceNotFoundException:
+        # The secret variable does not exist
+        print("Resource Not Found: {}".format(secret_id))
+
+    else:
+        # Get operation was successful, secret variable exists
+        print('Resource Found: {} = {}'.format(secret_id, secret_val))
+
+@events.action("set-secret",
+               arguments={
+                   "--secret-name": dict(required=True,
+                                         nargs="*",
+                                         help="name of secret to retrieve"),
+                   "--dry-run": dict(help="do a dry run of the actual operation")})
+def set_secret(argv: typing.List[str], args: argparse.Namespace):
+    """Set the value of the secret variable specified by the --secret-name flag"""
+    SM = boto3.client('secretsmanager')
+    stage = os.environ['DSS_DEPLOYMENT_STAGE']
+    secrets_store = os.environ['DSS_SECRETS_STORE']
+    secret_id = f'{secrets_store}/{stage}/{args.secret_name}'
+
+    # Use the `select` module to obtain the value that is passed
+    # to this script via stdin (i.e., piped to this script).
+    # Note: user provides secret variable *value* via stdin,
+    # user provides secret variable *name* via --secret-name flag.
+    if not select.select([sys.stdin, ], [], [], 0.0)[0]:
+        print(f"No data in stdin, exiting without setting {secret_id}")
+        sys.exit()
+    val = sys.stdin.read()
+
+    print("Setting", secret_id)
+
+    try:
+        # Start by trying to get the secret variable
+        _ = SM.get_secret_value(SecretId=secret_id)
+
+    except SM.exceptions.ResourceNotFoundException:
+        # The secret variable does not exist, so create it
+        if args.dry_run:
+            # Create it for fakes
+            print("Resource Not Found: Creating {}".format(secret_id))
+        else:
+            # Create it for real
+            _ = SM.create_secret(
+                Name=secret_id,
+                SecretString=val
+            )
+
+    else:
+        # Get operation was successful, secret variable exists
+        if args.dry_run:
+            print('Resource Found: Updating {}'.format(secret_id))
+        else:
+            _ = SM.update_secret(
+                SecretId=secret_id,
+                SecretString=val
+            )
--- a/dss/operations/secrets.py
+++ b/dss/operations/secrets.py
@@ -10,11 +10,10 @@ import typing
 import argparse
 import json
 import logging
-from pprint import pprint
 
 from dss.operations import dispatch
 from dss.util.aws import ARN as arn
-from dss.util.aws.clients import secretsmanager # note: ssm = param store, not secrets manager
+from dss.util.aws.clients import secretsmanager
 
 
 logger = logging.getLogger(__name__)
@@ -22,18 +21,20 @@ logger = logging.getLogger(__name__)
 
 def get_long_name(secret_name, arn_prefix, store_prefix):
     """
-    Given a (user-provided) name of a secret variable, 
-    determine whether the ARN prefix and store/stage 
+    Given a (user-provided) name of a secret variable,
+    determine whether the ARN prefix and store/stage
     prefix must be added to that variable to get its
     long name.
 
-    Users can specify variable names in any of the 
+    Users can specify variable names in any of the
     following forms, and this function will return
     the full ARN resource name:
 
     - es_source_ip-q2baD1 (no store/stage prefix present)
     - dcp/dss/dev/es_source_ip-q2baD1 (no ARN prefix present)
-    - arn:aws:secretsmanager:us-east-1:861229788715:secret:dcp/dss/dev/es_source_ip-q2baD1 (target output; store/stage prefix and ARN prefix present)
+    - arn:aws:secretsmanager:us-east-1:861229788715:secret:(continued)
+      dcp/dss/dev/es_source_ip-q2baD1 (target output, store/stage prefix
+      and ARN prefix present)
     """
     # Figure out what kind of prefix the user provided
     # with the provided list of list
@@ -53,12 +54,12 @@ def get_long_name(secret_name, arn_prefix, store_prefix):
 
 def get_short_name(secret_name, arn_prefix, store_prefix):
     """
-    Given a (user-provided) name of a secret variable, 
-    determine whether the ARN prefix and store/stage 
+    Given a (user-provided) name of a secret variable,
+    determine whether the ARN prefix and store/stage
     prefix must be removed from that variable to get
     its short name.
 
-    Users can specify variable names in any of the 
+    Users can specify variable names in any of the
     following forms, and this function will return
     the store/stage prefixed resource name only:
 
@@ -118,7 +119,7 @@ def list_secrets(argv: typing.List[str], args: argparse.Namespace):
             long_name = secret['ARN']
             short_name = long_name.split(":")[-1]
             if short_name.startswith(prefix):
-                # If the prefix matches the store and stage specified, 
+                # If the prefix matches the store and stage specified,
                 # store this secret's name for later
                 if args.long:
                     secret_names.append(long_name)
@@ -158,7 +159,7 @@ def get_secret(argv: typing.List[str], args: argparse.Namespace):
     if args.stage is None:
         stage_name = os.environ['DSS_DEPLOYMENT_STAGE']
 
-    if args.secret_name is None or len(args.secret_name)==0:
+    if args.secret_name is None or len(args.secret_name) == 0:
         logger.error("Unable to get secret: no secret was specified! Use the --secret-name flag.")
         sys.exit()
 
@@ -205,7 +206,7 @@ def set_secret(argv: typing.List[str], args: argparse.Namespace):
         stage_name = os.environ['DSS_DEPLOYMENT_STAGE']
 
     # Make sure a secret name was specified
-    if args.secret_name is None or len(args.secret_name)==0:
+    if args.secret_name is None or len(args.secret_name) == 0:
         logger.error("Unable to set secret: no secret name was specified! Use the --secret-name flag.")
         sys.exit()
     secret_name = args.secret_name
@@ -216,7 +217,7 @@ def set_secret(argv: typing.List[str], args: argparse.Namespace):
     # user provides secret variable *name* via --secret-name flag.
     if not select.select([sys.stdin, ], [], [], 0.0)[0]:
         err_msg = f"No data in stdin, cannot set secret {secret_name} without a value from stdin!"
-        logger.error()
+        logger.error(err_msg)
         sys.exit()
     secret_val = sys.stdin.read()
 
@@ -242,8 +243,7 @@ def set_secret(argv: typing.List[str], args: argparse.Namespace):
             # Create it for real
             print("Secret variable {} not found in secrets manager, creating it".format(short_secret_name))
             _ = secretsmanager.create_secret(
-                    Name=short_secret_name,
-                    SecretString=secret_val
+                Name=short_secret_name, SecretString=secret_val
             )
     else:
         # Get operation was successful, secret variable exists
@@ -254,8 +254,7 @@ def set_secret(argv: typing.List[str], args: argparse.Namespace):
             # Update it for real
             print("Secret variable {} found in secrets manager, updating it".format(short_secret_name))
             _ = secretsmanager.update_secret(
-                    SecretId=short_secret_name,
-                    SecretString=secret_val
+                SecretId=short_secret_name, SecretString=secret_val
             )
 
 
@@ -273,7 +272,7 @@ def set_secret(argv: typing.List[str], args: argparse.Namespace):
                        help="do a dry run of the actual operation")})
 def del_secret(argv: typing.List[str], args: argparse.Namespace):
     """
-    Delete the value of the secret variable specified by the 
+    Delete the value of the secret variable specified by the
     --secret-name flag from the secrets manager
     """
     store_name = os.environ['DSS_SECRETS_STORE']
@@ -281,7 +280,7 @@ def del_secret(argv: typing.List[str], args: argparse.Namespace):
         stage_name = os.environ['DSS_DEPLOYMENT_STAGE']
 
     # Make sure a secret name was specified
-    if args.secret_name is None or len(args.secret_name)==0:
+    if args.secret_name is None or len(args.secret_name) == 0:
         logger.error("Unable to set secret: no secret name was specified! Use the --secret-name flag.")
         sys.exit()
     secret_name = args.secret_name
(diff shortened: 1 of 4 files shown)


dss/operations/secrets.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+events = dispatch.target("secrets",
+                         help=__doc__)
+
+
+@events.action("list-secrets")
+def list_secrets(argv: typing.List[str], args: argparse.Namespace):
+    """Print a list of all secrets"""
+    pass
```
amarjandu: What does this function do? does it attempt to all the secrets under the `{secret_store}/{stage}` ?

dss/operations/secrets.py
```diff
@@ -0,0 +1,103 @@
...
+
+@events.action("set-secret",
+               arguments={
+                   "--secret-name": dict(required=True,
+                                         nargs="*",
+                                         help="name of secret to retrieve"),
+                   "--dry-run": dict(help="do a dry run of the actual operation")})
+def set_secret(argv: typing.List[str], args: argparse.Namespace):
+    """Set the value of the secret variable specified by the --secret-name flag"""
+    SM = boto3.client('secretsmanager')
```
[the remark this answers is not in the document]
chmreid: Done 👍

dss/operations/secrets.py
chmreid: Forgot to fill in this placeholder!

chmreid:
The way I implemented the list-secrets method was to split it into two functions:

* "name secrets" prints the names of all of the secret variables in the secrets manager
* "list secrets" prints the names and values of all of the secret variables in the secrets manager

A few open questions:

* Was the list variables functionality requested in #2296 intended to print variable names only, or variable values? 
* Currently this prints all variables in a given secrets manager; should we filter on secret store and stage, as @amarjandu asked/mentioned in his review comment above?
* I am still working on testing this out on my local machine, will probably need to bug someone for suggestions on doing that
* I also need to add unit tests for the new functionality. Are unit tests for `dss-ops.py` script located in `tests/test_operations.py`? Again, will probably need to bug someone to talk about a good testing strategy

xbrianh:
@chmreid 

  - This CLI command to list secrets may give you some implementation hints: `aws secretsmanager list-secrets | jq -r '.SecretList[].Name' | grep $DSS_SECRETS_STORE | grep $DSS_DEPLOYMENT_STAGE`
  - The `list-secrets` command should _not_ print values
  - I think it's sufficient to perform manual testing for this functionality

chmreid:
A note on secret names (probably similar for names of parameters as well):

The secret names have a long form and a short form. The long form is what the API returns when we ask for a list of secrets, but the short version is probably what people would type or write down.

Long: `arn:aws:secretsmanager:us-east-1:861229788715:secret:dcp/dss/dev/es_source_ip-q2baD1`
 Short: `dcp/dss/dev/es_source_ip-q2baD1`

I'm planning to lop off the `arn:aws:secretsmanager:us-east-1:861229788715:secret` portion and only keep the short label for the secret when we print it out (for the `get` or `list` actions). Same thing when we ask the user to provide secrets - we (and the AWS API) will expect something that's in the short form.

dss/operations/secrets.py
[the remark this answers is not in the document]
chmreid: Now using `dss.util.aws.clients` instead of getting boto clients directly ourselves.

chmreid:
This pull request is revamped and is now ready for a final review. I have tested all four functions (set, get, list, and delete) and they are working with the secrets store.

Note:
* the delete action also has an interactive dialog to prompt the user and make sure they _reeeeeally_ want to delete the given variable
* the list action only lists the names of variables, no values

Note that I changed the names of the actions to exclude the `-secrets` suffix, since that would get repetitive. Now usage just looks like:

    dss-ops.py secrets list
    dss-ops.py secrets get --secret-name <X>
    dss-ops.py secrets set --secret-name <X>
    dss-ops.py secrets delete --secret-name <X>

dss/operations/secrets.py
```diff
@@ -0,0 +1,339 @@
...
+import boto3
+import select
+import typing
+import argparse
+import json
+import logging
+
+from dss.operations import dispatch
+from dss.util.aws import ARN as arn
+from dss.util.aws.clients import secretsmanager
```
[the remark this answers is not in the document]
chmreid: done

dss/operations/secrets.py
```diff
@@ -0,0 +1,339 @@
...
+    # in case the user pipes the JSON output of this script to something else
+    store_name = os.environ['DSS_SECRETS_STORE']
+    if args.stage is None:
+        stage_name = os.environ['DSS_DEPLOYMENT_STAGE']
+
+    if args.secret_name is None or len(args.secret_name) == 0:
+        logger.error("Unable to get secret: no secret was specified! Use the --secret-name flag.")
+        sys.exit()
+
+    # Necessary because AWS requires full resource identifiers to fetch secrets
```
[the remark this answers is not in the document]
chmreid: See https://github.com/HumanCellAtlas/data-store/pull/2331

dss/operations/secrets.py
```diff
@@ -0,0 +1,339 @@
...
+    secret_val = sys.stdin.read()
+
+    # Necessary because AWS requires full resource identifiers to fetch secrets
+    region_name = arn.get_region()
+    account_id = arn.get_account_id()
+    arn_prefix = f"arn:aws:secretsmanager:{region_name}:{account_id}:secret:"
+    store_prefix = f"{store_name}/{stage_name}/"
+
+    full_secret_name = get_long_name(secret_name, arn_prefix, store_prefix)
+    short_secret_name = get_short_name(full_secret_name, arn_prefix, store_prefix)
```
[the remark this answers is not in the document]
chmreid: Addressed this with commit 3ccbc8c

dss/operations/secrets.py
```diff
@@ -0,0 +1,339 @@
...
+    secret_name = args.secret_name
+
+    # Necessary because AWS requires full resource identifiers to delete secrets
+    region_name = arn.get_region()
+    account_id = arn.get_account_id()
+    arn_prefix = f"arn:aws:secretsmanager:{region_name}:{account_id}:secret:"
+    store_prefix = f"{store_name}/{stage_name}/"
+
+    full_secret_name = get_long_name(secret_name, arn_prefix, store_prefix)
+    short_secret_name = get_short_name(full_secret_name, arn_prefix, store_prefix)
```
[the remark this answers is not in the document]
chmreid: Addressed with commit 3ccbc8c

chmreid:
Travis tests for branch passed 🎉
Travis tests for pull request failed 😕 
I believe this is unrelated to my commits, though:

```
======================================================================

FAIL: test_file_get_checkout (__main__.TestFileApi) [Replica.aws: Initiates checkout and returns 302 immediately for GET on stale checkout file.]

  File "tests/test_file.py", line 721, in _test_file_get_checkout

    test_creation_date_updated(file_key, old_creation_date)

  File "/home/travis/build/HumanCellAtlas/data-store/tests/__init__.py", line 73, in call

    return func(*args, **kwargs)

  File "tests/test_file.py", line 709, in test_creation_date_updated

    self.assertTrue(prev_creation_date < handle.get_creation_date(replica.checkout_bucket, key))

AssertionError: False is not true

Ran 19 tests in 169.695s

FAILED (failures=1)

Makefile:45: recipe for target 'tests/test_file.py' failed

make: *** [tests/test_file.py] Error 1

The command "make $MAKE_ARGS" exited with 2.
```

dss/operations/secrets.py
```diff
@@ -0,0 +1,313 @@
...
+
+    elif user_provided_store:
+        short_secret_name = secret_name
+        long_secret_name = arn_prefix + secret_name
+
+    else:
+        short_secret_name = store_prefix + secret_name
+        long_secret_name = arn_prefix + store_prefix + secret_name
+
+    return short_secret_name, long_secret_name
```
amarjandu: nice 🥇

dss/operations/secrets.py
```diff
@@ -0,0 +1,312 @@
...
+    """
+    # Note: this function should not print anything except the final JSON,
+    # in case the user pipes the JSON output of this script to something else
+    store_name, stage_name = get_secretsmanager_names(args)
+    arn_prefix, store_prefix = get_secretsmanager_prefixes(args)
+
+    # Make sure a secret name was specified
+    if args.secret_name is None or len(args.secret_name) == 0:
+        logger.error("Unable to set secret: no secret name was specified! Use the --secret-name flag.")
+        sys.exit()
```
xbrianh: Consider using `RuntimeError` instead of `sys.exit()`:
```
raise RuntimeError("Unable to set secret: no secret name was specified! Use the --secret-name flag.")
```

dss/operations/secrets.py
```diff
@@ -0,0 +1,312 @@
...
+    secret_names = args.secret_name
+
+    for secret_name in secret_names:
+
+        # Get resource IDs
+        short_secret_name, long_secret_name = long_short_resource_names(secret_name, arn_prefix, store_prefix)
```
xbrianh: I'm not sure I understand why the ARN is being constructed. It is not required to fetch the secret:
```
In [6]: import boto3                                                                                                                
In [7]: s = boto3.client("secretsmanager").get_secret_value(SecretId="dcp/dss/dev/es_source_ip")
```

The boto3 documentation also indicates this:
> You can specify either the Amazon Resource Name (ARN) or the friendly name of the secret.

chmreid: @xbrianh to provide full context, I know now that you can provide  a short name for a variable. However, AWS adds an alphanumeric suffix to each variable short name, and when you include the alphanumeric suffix with the variable short name, it can't find the variable.

For example, if you print all secrets, the `es_source_ip` variable is actually called

```
dcp/dss/dev/es_source_ip-q2baD1
```

and if you try and look up the variable using that exact short name, including the suffix, it will not find the variable.

See https://github.com/HumanCellAtlas/data-store/pull/2331 for more details - I hard-coded some calls to `get_secret_value()` with different forms of secret names to demonstrate what worked and what didn't. I discovered the suffix thing after I created that pull request.

dss/operations/secrets.py
chmreid: Okay, done!

dss/operations/secrets.py
xbrianh: The suffix is included in the ARN, not the secret name. Check out this response to `list_secrets`:
```
{'ARN': 'arn:aws:secretsmanager:us-east-1:861229788715:secret:dcp/dss/dev/es_source_ip-q2baD1',
 'Name': 'dcp/dss/dev/es_source_ip',
 'LastChangedDate': datetime.datetime(2019, 8, 8, 19, 12, 6, 153000, tzinfo=tzlocal()),
 'LastAccessedDate': datetime.datetime(2019, 8, 8, 0, 0, tzinfo=tzlocal()),
 'SecretVersionsToStages': {'658c3b41-0806-48b9-b05d-ea7dc2dbf237': ['AWSCURRENT'],
  'f37ccfe2-16e0-4305-a250-ef89d2c47ece': ['AWSPREVIOUS']}}
```

chmreid: I have removed the functions and infrastructure for handling either long or short secret names separately, and instead converted everything in the script to use short names _only_ for retrieving secrets.

chmreid:
Travis tests finally passing 🎉
Re-triggering tests from Travis website worked.

dss/operations/secrets.py
```diff
@@ -0,0 +1,280 @@
...
+            help="name of secret or secrets to retrieve (list values can be separated by a space)",
+        ),
+        "--json": dict(
+            default=False,
+            action="store_true",
+            help="format the output as JSON if this flag is present",
+        ),
+    },
+)
+def get_secret(argv: typing.List[str], args: argparse.Namespace):
```
Bento007: Maybe I'm being overly paranoid, but we should never return the encrypted secret with operation tasks. Especially if these operation will be running mostly our local machines. on At most we should be able to list and set. If we need to view key which should be rare, it should be done through the AWS console.

xbrianh: This is intended to replace `scripts/fetch-secret.py`, which we use programmatically in our CI/CD scripting.

chmreid: @Bento007 I agree we need to think hard about whether the get functionality is necessary, and that the AWS console is the optimal place to be viewing secret values. It is a bit too much power to be able to `print all the secrets`. 

@xbrianh this does beg the question - where is the need for this functionality to get the value of a secret in the CI/CD scripting coming from, and can we address that by removing the need somehow? In a best case scenario, it would be as simple as replacing shell scripts with Python scripts to eliminate the need to pass secrets around and use the secrets manager directly.

amarjandu: I don't think there is an easy way to remove the need to pull secrets for our deployments. This prevents secrets from being stored on an operators machine, which would be a bigger threat target..

xbrianh: Operators can already dump secrets to STDOUT with `aws secretsmanager get-secret-value`. The current (and future) scripting facilitates the identification and usage of DSS secrets.

The CI/CD jobs require secrets for authorization to access infrastructure during testing, and to install authorization keys into deployed DSS services.

Bento007: This can be a future feature, but we should fetch the secret in the lambda rather than baking in during deployment. This will allow us to rotate the secret with out requiring a redeployment.

In general this is an improvement to operations and security compared to what we currently have.

xbrianh: @Bento007 There are tradeoffs to calling secretsmanager from deployed Lambdas. This is something we should consider, but each network call reduces the reliability and performance of our Lambdas in ways that are not under our control.

Bento007: The secret can be retrieve once per lambda when the lambda is spun up. This would only hurt the reliability of starting a new lambda which only happens when we scale horizontally. I think it's worth exploring in the future.

chmreid:
@DailyDreaming The tests are finished and ready for your approval 🎉

dss/operations/secrets.py
```diff
@@ -0,0 +1,326 @@
...
+        # stdin provides secret value, flag --secret-name provides secret name.
+        if not select.select([sys.stdin], [], [])[0]:
+            err_msg = f"No data in stdin, cannot set secret {secret_name} without "
+            err_msg += "a value from stdin or without --secret-value flag!"
+            raise RuntimeError(err_msg)
+        secret_val = sys.stdin.read()
+
+    # Determine if we are doing a dry run
+    dry_run = False
+    if hasattr(args, "dry_run"):
```
DailyDreaming: Same.

dss/operations/secrets.py
```diff
@@ -0,0 +1,326 @@
...
+        # Use stdin (input piped to this script) as secret value.
+        # stdin provides secret value, flag --secret-name provides secret name.
+        if not select.select([sys.stdin], [], [])[0]:
+            err_msg = f"No data in stdin, cannot set secret {secret_name} without "
+            err_msg += "a value from stdin or without --secret-value flag!"
+            raise RuntimeError(err_msg)
+        secret_val = sys.stdin.read()
+
+    # Determine if we are doing a dry run
+    dry_run = False
```
DailyDreaming:
```
dry_run = False
        if args.dry_run:
            dry_run = True
```

Consider shortening to just:
```
dry_run = args.dry_run
```

dss/operations/secrets.py
```diff
@@ -0,0 +1,310 @@
...
+            secret_name = secret["Name"]
+
+            # Only save secrets for this store and stage
+            if secret_name.startswith(store_prefix):
+                secret_names.append(secret_name)
+
+    secret_names.sort()
+
+    if hasattr(args, "json") and args.json is True:
+        print(json.dumps(secret_names))
```
DailyDreaming: Consider `json.dumps(secret_names, indent=4)`.

dss/operations/secrets.py
```diff
@@ -0,0 +1,310 @@
...
+        try:
+            response = secretsmanager.get_secret_value(SecretId=secret_name)
+            secret_val = response["SecretString"]
+        except ClientError:
+            # A secret variable with that name does not exist
+            logger.warning(f"Resource not found: {secret_name}")
+        else:
+            # Get operation was successful, secret variable exists
+            if use_json:
+                print(json.dumps({secret_name: secret_val}))
```
DailyDreaming: Consider using `, indent=4` again, and also, sometimes the `secret_val` is a dictionary, but this will print it as a string, so consider using something like:
```
try:
    secret_val = json.loads(secret_val)
except:
    pass

print(json.dumps({secret_name: secret_val}, indent=4))
```

For reference, the output of the command: `scripts/dss-ops.py secrets get --secret-name dcp/dss/dev/admin_user_emails dcp/dss/dev/application_secrets.json --json`

dss/operations/secrets.py
```diff
@@ -0,0 +1,326 @@
...
+            help="do a dry run of the actual operation",
+        ),
+    },
+)
+def set_secret(argv: typing.List[str], args: argparse.Namespace):
+    """Set the value of the secret variable specified by the --secret-name flag"""
+    store_prefix = get_secretsmanager_prefix(args)
+
+    # Make sure a secret name was specified
+    if args.secret_name is None or len(args.secret_name) == 0:
```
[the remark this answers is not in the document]
chmreid: This and several other comments were caused by writing tests incorrectly. I've fixed them and will mark related comments as resolved.

dss/operations/secrets.py
chmreid: done!

dss/operations/secrets.py
chmreid: Done, thanks!

chmreid:
@DailyDreaming a helpful description could be added, but no other subcommands have such descriptions. I suggest waiting for another PR. That PR would either add descriptions to all subcommands, or modify the help functionality to print the docstring for a given subcommand when you type that subcommand plus `--help`.

dss/operations/secrets.py
```diff
@@ -0,0 +1,326 @@
...
+            print(secret_name)
+
+
+@events.action(
+    "get",
+    arguments={
+        "--stage": dict(
+            required=False, help="the stage for which secrets should be listed"
+        ),
+        "--secret-name": dict(
```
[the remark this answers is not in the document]
chmreid: Done

DailyDreaming:
@chmreid Sounds good to me.  It's low priority, so we might have the new dev ops team member do this when they join.

tests/__init__.py
```diff
@@ -120,3 +122,41 @@ def skip_on_travis(test_item):
         return unittest.skip("Test doesn't run on travis.")(test_item)
     else:
         return test_item
+
+class CaptureStdout(list):
```
xbrianh: This is a cool construction. However, please consider using existing [contextlib.redirect_stdout](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout) functionality from the contextlib library.

chmreid: Opened #2355 so we can address this when Python 2 support is dropped

dss/operations/secrets.py
```diff
@@ -0,0 +1,292 @@
...
+        if not select.select([sys.stdin], [], [])[0]:
+            err_msg = f"No data in stdin, cannot set secret {secret_name} without "
+            err_msg += "a value from stdin or without --secret-value flag!"
+            raise RuntimeError(err_msg)
+        secret_val = sys.stdin.read()
+
+    # Create or update
+    try:
+        # Start by trying to get the secret variable
+        _ = secretsmanager.get_secret_value(SecretId=secret_name)
```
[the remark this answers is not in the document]
chmreid: Done!

tests/__init__.py
xbrianh: In general, it is preferable to keep our code as simple as possible to limit potential bugs and technical debt. In this case, the data-store can use `contextlib.redirect_stdout` without waiting for Python 2 support to be dropped from the  CLI.

PR #2331, opened 2019-08-07
[DO NOT MERGE] Demonstrate that retrieval of secrets using short names is broken

chmreid:
In PR #2325 we had to jump through some hoops to assemble the full ARN number of the secrets, instead of using the (much simpler) secret store/stage prefix.

This pull request implements a simple `foobar` action in the dss operations script that proves the full ARN is required for each secret we retrieve from the secrets manager; simple/short names do not cut it.

Example of how to run this new action using the new branch, and example output:

```
$ git clone -b chmreid-short-arn-broken git@github.com:HumanCellAtlas/data-store

$ cd data-store

$ ./scripts/dss-ops.py foobar plink
------------------
Full secret name:
arn:aws:secretsmanager:us-east-1:861229788715:secret:dcp/dss/dev/es_source_ip-q2baD1
{'ARN': 'arn:aws:secretsmanager:us-east-1:861229788715:secret:dcp/dss/dev/es_source_ip-q2baD1',
 'CreatedDate': datetime.datetime(2019, 8, 1, 13, 41, 53, 237000, tzinfo=tzlocal()),
 'Name': 'dcp/dss/dev/es_source_ip',
 'ResponseMetadata': {'HTTPHeaders': {'connection': 'keep-alive',
                                      'content-length': '753',
                                      'content-type': 'application/x-amz-json-1.1',
                                      'date': 'Wed, 07 Aug 2019 21:47:19 GMT',
                                      'x-amzn-requestid': '3b78adf1-9b04-4407-81e7-6ce9111e9385'},
                      'HTTPStatusCode': 200,
                      'RequestId': '3b78adf1-9b04-4407-81e7-6ce9111e9385',
                      'RetryAttempts': 0},
 'SecretString': '<REDACTED>',
 'VersionId': '658c3b41-0806-48b9-b05d-ec7dc2dbf237',
 'VersionStages': ['AWSCURRENT']}
------------------
Short secret name:
dcp/dss/dev/es_source_ip-q2baD1
No resource with the name dcp/dss/dev/es_source_ip-q2baD1 exists
```

### DIFF
--- a/dss/operations/foobar.py
+++ b/dss/operations/foobar.py
@@ -31,12 +31,13 @@ def plink(argv: typing.List[str], args: argparse.Namespace):
     stage_name = os.environ['DSS_DEPLOYMENT_STAGE']
 
     tiny_secret_name = "es_source_ip-q2baD1"
-    short_secret_name = "dcp/dss/dev/es_source_ip-q2baD1"
+    short_secret_name_broken = "dcp/dss/dev/es_source_ip-q2baD1"
+    short_secret_name = "dcp/dss/dev/es_source_ip"
     full_secret_name = "arn:aws:secretsmanager:us-east-1:861229788715:secret:dcp/dss/dev/es_source_ip-q2baD1"
 
     try:
         print("------------------")
-        print("Full secret name:")
+        print(f"Full secret name {full_secret_name}:")
         print(full_secret_name)
         response1 = secretsmanager.get_secret_value(SecretId=full_secret_name)
         pprint(response1)
@@ -45,7 +46,7 @@ def plink(argv: typing.List[str], args: argparse.Namespace):
 
     try:
         print("------------------")
-        print("Short secret name:")
+        print(f"Short secret name {short_secret_name}:")
         print(short_secret_name)
         response2 = secretsmanager.get_secret_value(SecretId=short_secret_name)
         pprint(response2)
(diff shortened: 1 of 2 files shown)


chmreid:
Tagging @amarjandu

chmreid:
I think I've reconciled where the problem is coming from.

If we list a secret, and we ask for the full name, we will see something like:

```
arn:aws:secretsmanager:us-east-1:861229788715:secret:dcp/dss/dev/es_source_ip-q2baD1
```

Note that the suffix `-q2baD1` is _automatically generated_ by AWS when we create this variable.

If we try and look up a variable from the secrets manager, we can either use the full resource identifier (shown above), or we can use the short variable name. 

Amar stated the short variable name could also be used by boto3/the AWS API. However, you have to use the following form of the short variable name (_without_ the suffix):

```
dcp/dss/dev/es_source_ip
```

If you try and use `dcp/dss/dev/es_source_ip-q2baD1` it will not find the variable in the secrets manager.

@amarjandu I think this resolves why we were getting different behavior asking for secrets using the short name - I was using the short name with the suffix and the variable was not being found, you were using the short name without the suffix and the variable was being found.

amarjandu:
Ah i see! 👀 
We can keep this in the script, perhaps if we need it for resource tagging, or for within an IAM policy. 
Feel free to close out this branch as needed.

PR #2332, opened 2019-08-07
Add pagination response headers (closes #2287)

### LINKED ISSUE #2287: Add explicit pagination headers
Currently pagination must be inferred from a Link header. I propose that in DSS and other DCP OpenAPI based services, methods that support pagination explicitly return a header that indicates it, like this (for `GET bundle`):
```
X-OpenAPI-Pagination: true
X-OpenAPI-Paginated-Content-Key: files
```
The second header refers to the response JSON body key containing an array of results that are paginated (vary between pages).

natanlao:


kislyuk:
@DailyDreaming @natanlao this commit was incomplete. It does not properly set `X-OpenAPI-Paginated-Content-Key` for all routes. Please review PRs thoroughly.

PR #2334, opened 2019-08-09
Add tests for dss-ops secrets functionality

chmreid:
This commit has two parts:
- additions to `tests/__init__.py` that add context manager objects to capture stdout and to feed a string to stdin
- additions to `test_operations.py` that test specific functionality added to dss operations (secrets verb)

Note: this PR is to merge into the `dss-secrets` branch (#2325), not the `master` branch

### DIFF
--- a/dss/operations/secrets.py
+++ b/dss/operations/secrets.py
@@ -22,7 +22,10 @@ def get_secretsmanager_prefix(args):
     the SecretsManager.
     """
     store_name = os.environ["DSS_SECRETS_STORE"]
-    if args.stage is None:
+
+    if hasattr(args,'stage') and args.stage is not None:
+        stage_name = args.stage
+    else:
         stage_name = os.environ["DSS_DEPLOYMENT_STAGE"]
 
     store_prefix = f"{store_name}/{stage_name}/"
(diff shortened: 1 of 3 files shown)


chmreid:
@amarjandu this PR adds two context managers, one to replace stdin and one to replace stdout

chmreid:
:warms hands over fire: just waitin for these tests to pass

chmreid:
Tests are passing when run locally/manually, but Travis CI/CD does not have permission to test out the secrets functionality (cannot create/update/remove secret variables). Waiting to see where discussion about that will lead...

dss/operations/secrets.py
```diff
@@ -36,7 +39,7 @@ def get_secretsmanager_prefix(args):
     "list",
     arguments={
         "--stage": dict(
-            required=False, help="the stage for which secrets should be listed"
+            required=True, help="the stage for which secrets should be listed"
```
xbrianh: The code in `dss/*` and `scripts/*` uses the stage defined in the env var `DSS_DEPLOYMENT_STAGE`. Passing in a different stage as an argument to dss-ops scripting may result it conflicts or other undefined behavior.

This may not be the case for secrets handling, but even so it would be preferable to maintain consistency with the rest of the dss operational scripting.

chmreid: Good catch - will make this a non-required flag. The cod already looks for the stage name in the `DSS_DEPLOYMENT_STAGE` env var if the flag is undefined.

chmreid:
This PR is _finally_ ready to merge into the main dss-ops secrets PR, and get some testing of that new functionality going...

chmreid:
squash-n-merge

PR #2337, opened 2019-08-09
Terraform v0.12.6 compatibility updates

### LINKED ISSUE #2336: Update terraform.
The latest version of terraform breaks on our files.  If we ever want to use a newer version of terraform, we'll have to update accordingly (mostly seems to be minor syntax; equal signs and commas).

xbrianh:
This updates the Terraform scripting that defines DSS deployed infrastructure for compatibility with Terraform v0.12.6

closes #2336


DailyDreaming:
@amarjandu The terraform commands tell you.  It looks a bit like when you run `flake8`.

PR #2339, opened 2019-08-12
Check the correct DCP branch status during release

xbrianh:
This checks the correct target branch for dcp-wide integration test status, allowing the normal release button (not force release) to work in allspark.


amarjandu:
Does this change SOP to using the `release` button, and not using the `force release`?

xbrianh:
@amarjandu  yes

PR #2344, opened 2019-08-13
Fix 500 error when deleting nonexistent bundle (closes #1918)

### LINKED ISSUE #1918: DELETE /bundle with non-existent uuid raises 500
```
[ERROR]	2019-02-27T17:44:16.669Z	5a255441-c0d1-4ed9-9085-2905dbe5c46e	
https://dss.dev.data.humancellatlas.org/v1/bundles/5f900625-55ba-4a72-af06-6c1b4d5e42f7?replica=gcp&version=2019-02-27T174218.411307Z
validation error: 'code' is a required property  Failed validating 'required' in schema['allOf'][1]:     {'properties': {'code': {'description': 'Machine-readable error code.  '                                             'The types of return values '                                             'should not be changed '                                             'lightly.',                              'enum': ['unhandled_exception',                                       'illegal_arguments',                                       'read_only'],                              'type': 'string'}},      'required': ['code'],      'type': 'object'}  On instance:     {'title': 'bundle not found'}
```

natanlao:


dss/api/bundles/__init__.py
```diff
@@ -238,7 +238,7 @@ def delete(uuid: str, replica: str, json_request_body: dict, version: str = None
         response_body = dict()  # type: dict
     else:
         status_code = requests.codes.not_found
-        response_body = dict(title="bundle not found")
+        response_body = dict(title="bundle not found", code=status_code)
```
xbrianh: `DSSException` should be raised here. See, [for example](https://github.com/HumanCellAtlas/data-store/blob/62c60d96a096ef3b2c80810ddc619c9ad23a2397/dss/api/bundles/__init__.py#L54)

`not_found` should also needs to be included in the API spec for DELETE /bundle [here](https://github.com/HumanCellAtlas/data-store/blob/62c60d96a096ef3b2c80810ddc619c9ad23a2397/dss-api.yml#L1996)

natanlao:
[Passing on Allspark](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/pipelines/14793)

PR #2346, opened 2019-08-13
Update default runner to python 3.7.

DailyDreaming:
Please post if there are any good reasons not to do so?

Edit: Python libraries were updated to reflect the current AWS lambda runtime versions.

Edit: Changes were made to `tests/test_indexer.py` since the tests were incompatible with the updated `HTTPSignatureAuth.verify()` function, which has several new asserts that break our mocked stub tests.


kislyuk:
LGTM but please make sure gitlab, travis and the deployed lambdas are all running 3.7. We did this upgrade in the Query Service previously, in case you want to use our configuration as a template. Ubuntu 19.04 is the first Ubuntu to use 3.7 by default.

kislyuk:
I believe you also need "sudo: required" in Travis in order to use Python 3.7.

xbrianh:
`README.md` also needs to be updated.

DailyDreaming:
@kislyuk This was actually started as a direct result of realizing `python3.6-dev` is no longer a part of the default packages upon upgrading my personal laptop to Ubuntu 19.04.  Will look at the `query-service` and update accordingly, thanks for the feedback!

@xbrianh Will do, thank you!

kislyuk:
Are the tests in `tests/test_indexer.py` here intentionally? If so, please add a note in the PR description/final commit message.

DailyDreaming:
@kislyuk They're intentional.  Changes were made to `tests/test_indexer.py` since the tests were incompatible with the updated `HTTPSignatureAuth.verify()` function, which has several new asserts that break our mocked stub tests.

Added to the description.

amarjandu:
Was this planning on being merged in? just curious...

DailyDreaming:
@amarjandu Of course you'd message just before I took a long trip.  XD

This will be merged in after ES is removed, since it only fails on the indexing tests and it isn't worth fixing tests that will be removed.  It should be a simple update once that's gone through.

amarjandu:
There is going to be some work involved with the lambda layers. if we continue with this.
https://github.com/HumanCellAtlas/data-store/blob/4b05c1b83fb31354e2b64a6ed89600c95e9a2dd9/scripts/generate_upload_requirements_layer.sh#L18
needs to be change to have the version number (3.7) in the path, we should also probably move to have the python version number included within the name of the lambda layer to avoid annoying debugging. 

Optionally we can move the layers into different bucket prefix to allow for a lookup based on the operators version....

`common.mk` is also going to have to change.

DailyDreaming:
@amarjandu `common.mk` was updated a while back, not sure if you just missed that?

The py3.6 -> py3.7 mention has been addressed, thanks!

amarjandu:
> @amarjandu `common.mk` was updated a while back, not sure if you missed that?
> 
> The py3.6 -> py3.7 mention has been addressed, thanks!

Yes! i should have used `cmd+f` the first time ha.

PR #2354, opened 2019-08-16
Relax sync size limit

xbrianh:
This relaxes the sync metadata size limit to 50 MB. Bundle manifests containing tens of thousands of files are expected to be in this range.


xbrianh:
@kislyuk Done! my physicist brain is in conflict with my software brain on this issue  😂

PR #2363, opened 2019-08-19
Add ttl to async db items.

DailyDreaming:


DailyDreaming:
@chmreid There is, I'll add a test!

DailyDreaming:
Test added but it won't work until the new table is deployed.

dss/dynamodb/__init__.py
```diff
@@ -7,16 +7,18 @@ class DynamoDBItemNotFound(Exception):
...
         item['body'] = {'S': value}
     if sort_key:
         item['sort_key'] = {'S': sort_key}
+    if ttl:
+        item['ttl'] = {'N': str(ttl)}
     return item
 
 
-def put_item(*, table: str, hash_key: str, sort_key: str=None, value: str, dont_overwrite: str=None):
+def put_item(*, table: str, hash_key: str, sort_key: str=None, value: str, dont_overwrite: str=None, ttl: int=None):
```
chmreid: The keys `sort_key`, `dont_overwrite`, and `ttl` should be changed from `str` to `typing.Optional[str]` (respectively, `int` to `typing.Optional[int]`), since these values can be set to None (see note [here](https://docs.python.org/3/library/typing.html#typing.Optional) on `typing.Optional` - it is appropriate to use when an argument can take on a value of None). Hat tip to @amarjandu (I noticed [this line](https://github.com/HumanCellAtlas/data-store/pull/2318/files#diff-d24709f50152a3e6b9d338cdc6df6c3bR139) in the bundle enumeration pull request)

dss/dynamodb/__init__.py
```diff
@@ -7,16 +7,18 @@ class DynamoDBItemNotFound(Exception):
     pass
 
 
-def _format_item(hash_key, sort_key, value):
+def _format_item(hash_key: str, sort_key: str, value: str, ttl: int=None):
```
chmreid: [This comment](https://github.com/HumanCellAtlas/data-store/pull/2363/files#r315452815) about using `typing.Optional[<type>]` applies here too

Issue #1837, opened 2019-01-16
Add content disposition header to checked-out objects


hannes-ucsc:
The Data Browser offers a direct download feature which uses the DSS' `GET /files` to check out a file and then download it to the users disk. By default, most user agents treat URLs whose response has known content type like `image/jpeg`, `application/pdf`, `text/tab-separated-values` by opening them in a tab rather than just downloading and saving them to disk. That default user agent behavior can be overridden by including the `Content-Disposition: attachment` header in the response. A server can optionally suggest a file name for the downloaded file by specifying `Content-Disposition: attachment; filename="filename.jpg"`. Without the `filename` part, browsers default to the last component of the URL path as the suggested local file name. In case of signed URLs to files checked out by the DSS, this last component is the blob checksum. It is very long and meaningless to users of the Data Browser.

S3 can be coaxed into including this header in responses to requests for a signed URL to an object in two ways: the header can be persisted with the object as metadata during `PUT object` or the header can be incorporated into the signed URL.

Ideally, we would like to specify the file name, as in `Content-Disposition: attachment; filename="filename.jpg"`, but this would require adding an optional parameter to the DSS' `PUT /files` since blobs can have multiple files and each file can be assigned different names in bundle manifests. To allow for sharing a single checked-out blob between multiple files, the header should be incorporated in the signed URL instead of being persisted as object metadata.

hannes-ucsc:
@kozbo, where is this on your list of priorities? We have a horrible hack in Azul to work around this.

xbrianh:
It seems content disposition can be specified when creating signed GET urls for S3 and GS.

Perhaps the DSS can allow clients to indicate content-disposition with an optional parameter on GET /files.

kislyuk:
I agree content-disposition should be accommodated in DSS to specify a fixed disposition (attachment?) and a filename. It's important to make sure this behavior is consistent between replicas (GS and S3, and whatever we end up using in the future). Unfortunately, I can't think of a clean solution to the filename ambiguity using our current facilities. I think it's time to allow specifying a "default" filename that's intrinsic to the file, as Hannes suggests.

One thing we need to thoroughly de-risk before we nail this down is how this impacts webapps that want to build on top of DSS. In other words I want to make sure the Content-Disposition header does not interfere with javascript-based apps that want to render DSS file contents in the browser.

xbrianh:
[...]

PR #2371, opened 2019-08-24
Add response-content-disposition to GET /file.

DailyDreaming:
@hannes-ucsc Is this what you were thinking for #1837 ?

Will add tests if this is the right idea.  Not sure if an associated `parent_bundle_fqid` is the right direction for knowing where to get the file metadata for the `filename` (can `filename` be passed directly in your use-case?).


dss-api.yml
```diff
@@ -381,6 +381,10 @@ paths:
 
             The use of presigned URLs is recommended for data access. Cloud native URLs are currently provided for a limited set of use cases and may not be provided in the future. If cloud native URLs are required, please contact the data store team regarding the credentials necessary to use them.
           type: boolean
+        - name: parent_bundle_fqid
```
hannes-ucsc: I'd drop `parent_` because DSS doesn't really use that term. And I would make it so that a user can just pass in a bundle UUID to exploit the DSS convention of defaulting to latest version. It would require some special validation logic to disallow passing only a version and no uuid.

dss/api/files.py
```diff
@@ -110,9 +114,18 @@ def get_helper(uuid: str, replica: Replica, version: str = None, token: str = No
                     path=get_dst_key(blob_path)
                 )))
             else:
+                # TODO: Add this into the cloud blobstore
+                filename = determine_filename(parent_bundle_fqid, file_fqid) or blob_path[len('blobs/'):]
```
hannes-ucsc: Handle case where filename is None. You don't want to set the CD header if user doesn't pass in a bundle fqid. Right now you'd be setting the header to a file name of `"None"` which is not right.

dss/api/files.py
```diff
@@ -139,6 +152,16 @@ def get_helper(uuid: str, replica: Replica, version: str = None, token: str = No
     return response
 
 
+def determine_filename(bundle_fqid, file_fqid):
+    """
+    Returns the filename for a file from its associated bundle metadata or None if not found in the bundle metadata.
+    """
+    if bundle_fqid:
```
hannes-ucsc: I'd pull this conditional up. See above comment

dss/api/files.py
DailyDreaming: It should be handling it, it's defaulting to `blob_path[len('blobs/'):]` (your current default in the browser according to #1837) when `parent_bundle_fqid == None`.

dss-api.yml
DailyDreaming: What I'm worried about is that the file name might change from version to version in the bundle (or worse, no longer exist in the bundle).

dss/api/files.py
DailyDreaming: Can do.

dss-api.yml
DailyDreaming: Oh and I can drop `parent_`, I just thought it added some clarity.

dss/api/files.py
hannes-ucsc: Huh? If `parent_bundle_fqid` is `None`, `determine_filename` will return `None` and `filename` will be set to `None`. Three lines down, `filename` is interpolated into the CD header, causing `kwargs` to be set to `{'ResponseContentDisposition': f'attachment; filename=None'}`.

dss-api.yml
hannes-ucsc: > What I'm worried about is that the file name might change from version to version

How is that worrisome?

> (or worse, no longer exist in the bundle).

Lon, the file could be missing in ANY version of the bundle. The file may even be absent in EVERY version of the bundle. Your are not preventing that sort of failure by requiring the version. The only thing you are preventing is a valid use case: returning the file name in the LATEST version of the bundle.

dss/api/files.py
DailyDreaming: The line is:
`filename = determine_filename(parent_bundle_fqid, file_fqid) or blob_path[len('blobs/'):]`

You need to scroll right.

dss-api.yml
DailyDreaming: If the user is using a bundle as the basis for iterating `GET /file`, it will contain the filename that the user is expecting.

So say I am user John, and I go to download an old bundle based on a paper that I want to reproduce.  This bundle is uuid=1 and version=2.

Bundle: `uuid=1` and `version=2` contains:
- File 0: a.txt (file version 1)
- File 1: b.txt (file version 1)
- File 2: c.txt (file version 1)

Bundle: `uuid=1` and `version=3` contains:
- File 0: x.txt (file version 2)
- File 1: y.txt (file version 2)
- File 2: c.txt (file version 1)

I run the following psuedocode:
```
for file in get_bundle.paginate():
    response = get_file()
```

The file names returned for the first bundle would be:
- File 0: a.txt
- File 1: b.txt
- File 2: c.txt

But for the second (latest) would be:
- File 0: the.gigantic.blobstore.hash.indentifier
- File 1: the.gigantic.blobstore.hash.indentifier
- File 2: c.txt

So in this case, fetching the newest version leads to a confusing response, to me.  I hope that makes my point clearer.  Fetching the correct bundle version a file was in is the only guarantee that you will be able to fetch the metadata (and thus a valid filename) for that file.

So I see it as (assuming no user error in submitting the correct `bundle_fqid`):
 - Input only a uuid: Only sometimes fetches the correct metadata
 - Input uuid + version: Always fetch the correct metadata

That doesn't exclude user error in entering a bad `bundle_fqid`, but in that case it just defaults to returning the blob hash.

dss/api/files.py
hannes-ucsc: I see. So I was wrong. But does it give you pause that a I didn't pick up on this? Why not just 

```suggestion
                if parent_bundle_fqid:
                    filename = determine_filename(parent_bundle_fqid, file_fqid)
                else:
                    filename = blob_path[len('blobs/'):]
```

dss-api.yml
hannes-ucsc: Thank you for the structured response. I think your pseudo code omits some important detail which makes it hard for me to follow the conclusions you lay out. Let me add that detail:

```
for file in get_bundle(uuid="1", version="2").paginate():
    response = get_file(uuid=file.uuid, version=file.version, parent_bundle_fqid="1.2")
```

This is what any educated user interested in a specific bundle version would write. With my proposed dropping of `parent_` and splitting up the bundle FQID into separate parameters it would read:

```
for file in get_bundle(uuid="1", version="2").paginate():
    response = get_file(uuid=file.uuid, version=file.version, bundle_uuid="1", bundle_version="2")
```

A user interested in the latest version of a bundle would write

```
for file in get_bundle(uuid="1").paginate():
    response = get_file(uuid=file.uuid, version=file.version, bundle_uuid="1")
```

That's a valid use case which your proposal prohibits. Are you worried about a race where a new bundle is published while the iteration is going on?

> That doesn't exclude user error in entering a bad bundle_fqid, but in that case it just defaults to returning the blob hash.

It should definitely not do that. If the user passes in the FQID of a bundle that doesn't contain the requested file, a 404 error should be raised.

DailyDreaming:
@hannes-ucsc Changes made.

xbrianh:
@DailyDreaming @hannes-ucsc Downstream clients will have different preferences for the value of Content-Disposition. For the case of presigned urls, clients should pass in their preferred value via a `content-disposition` parameter, as per [this comment](https://github.com/HumanCellAtlas/data-store/issues/1837#issuecomment-487641015).

Providing a default value for Content-Disposition by asking clients to pass in the associated bundle seems reasonable. However, this overlaps with proposed access control design issues that are still work in progress. At this stage, we should consider restricting changes to GET /file to only the `content-disposition` parameter. Does this serve the needs of the browser?

hannes-ucsc:
> Does this serve the needs of the browser?

Yes. 

It would have been helpful to give this guidance before a team member embarks on a task. I assumed the current implementation in this PR was driven by guidance from the lead and spent considerable time reviewing it.

xbrianh:
@hannes-ucsc @DailyDreaming The code implementing default content-disposition based upon bundle affiliation should be retained somewhere. Even if it's not used in this PR, it is likely to be useful in the future when the DSS adds controlled access features. It has value.

DailyDreaming:
@xbrianh I'll retain it for the future in a gist then.  It should be easy to reimplement.

DailyDreaming:
@xbrianh Changes have been made but will not pass until a new `cloud-blobstore` has been released and we update the `dss` to include the new version.

tests/test_file.py
```diff
@@ -477,6 +477,37 @@ def _test_file_get_latest(self, replica: Replica):
...
+        with override_bucket_config(BucketConfig.TEST_FIXTURE):
+            resp_obj = self.assertGetResponse(
+                url,
+                requests.codes.found,
+                headers=get_auth_header(),
+                redirect_follow_retries=FILE_GET_RETRY_COUNT,
+                min_retry_interval_header=RETRY_AFTER_INTERVAL,
+                override_retry_interval=1
+            )
+            if resp_obj.response.status_code == requests.codes.found:
```
xbrianh: Not sure I understand how this test is retried when the response code is not found.

DailyDreaming: It isn't, except internally.  Three other `data-store` tests in `test_file.py` follow this convention:
https://github.com/HumanCellAtlas/data-store/blob/master/tests/test_file.py#L439
https://github.com/HumanCellAtlas/data-store/blob/master/tests/test_file.py#L478
https://github.com/HumanCellAtlas/data-store/blob/master/tests/test_file.py#L848

I believe:

```
with override_bucket_config(BucketConfig.TEST_FIXTURE):
    resp_obj = self.assertGetResponse(
        url,
        requests.codes.found,
        headers=get_auth_header(),
        redirect_follow_retries=FILE_GET_RETRY_COUNT,
        min_retry_interval_header=RETRY_AFTER_INTERVAL,
        override_retry_interval=1
    )
    if resp_obj.response.status_code == requests.codes.found:
        url = resp_obj.response.headers['Location']
        response = requests.get(url)
        self.assertEqual(response.headers['Content-Disposition'],
                         'attachment; filename=test-data.json')
        return
self.fail(f"Failed after {FILE_GET_RETRY_COUNT} retries.")
```

Is equivalent to:

```
with override_bucket_config(BucketConfig.TEST_FIXTURE):
    resp_obj = self.assertGetResponse(
        url,
        requests.codes.found,
        headers=get_auth_header(),
        redirect_follow_retries=FILE_GET_RETRY_COUNT,
        min_retry_interval_header=RETRY_AFTER_INTERVAL,
        override_retry_interval=1
    )
    assert resp_obj.response.status_code == requests.codes.found, f"Failed after {FILE_GET_RETRY_COUNT} retries."

    url = resp_obj.response.headers['Location']
    response = requests.get(url)
    self.assertEqual(response.headers['Content-Disposition'],
                     'attachment; filename=test-data.json')
```

I can refactor it if necessary.

xbrianh: As far as I can tell, `resp_obj.response.status_code == requests.codes.found` will always evaluate to `True` if this point in the code is reached.

DailyDreaming:
@xbrianh I agree.  I've made appropriate changes.

Issue #2333, opened 2019-08-09
Significant notification message loss in production


hannes-ucsc:
While researching https://github.com/DataBiosphere/azul/issues/1212 I found some disconcerting evidence that the DSS notifier is dropping messages.

Every notification message goes into the first notification queue (queue 0). Every notification message that goes into queue 0 should come out of that queue 0 again. Invariably, there should be as many message receptions for queue 0 as there are message additions to that queue. The notifier code emits a log message right before it adds a message and after it receives a message. The number of log messages should be equal for both message types. It isn't.

<img width="1494" alt="Screen Shot 2019-08-08 at 9 52 38 PM" src="https://user-images.githubusercontent.com/5143256/62755260-1f69db00-ba28-11e9-89f5-8f309d0b27c6.png">

<img width="1497" alt="Screen Shot 2019-08-08 at 9 52 50 PM" src="https://user-images.githubusercontent.com/5143256/62755269-25f85280-ba28-11e9-8830-0fc2da03220c.png">

Note the total number of hits: 228,475 additions against 44,629 receptions. Also notice how stretched out over time the reception graph is. The rate of the message consumption is also surprisingly low: during the first 4 days or so it is just 500 messages per hour, that's less than 10 per minute. IIRC, when I wrote the notifier it was able to handle many more messages, but the consumption rate was configurable via the number of worker threads in the pool.

My theory is that SQS is simply expiring message (the default retention period is 4 days and I am not sure I wrote code to extend the retention period.

This CloudWatch metric underlines that:

<img width="1422" alt="Screen Shot 2019-08-08 at 10 17 31 PM" src="https://user-images.githubusercontent.com/5143256/62756015-0d3d6c00-ba2b-11e9-88ae-4c9e43b73edb.png">

Note how the message age is capped at ~ 346k seconds which is 4 days.

TL;DR It looks as if queue consumption rate and message retention period are ill configured in production, causing SQS to expire notification messages before they even get their first shot of being delivered to the subscriber.

DailyDreaming:
We can increase the number of elasticsearch instances to accommodate this.  However, it might be costly.

I don't think the lambdas are the bottleneck.

[...]

PR #2373, opened 2019-08-26
Increase the number of DSS notify workers.

DailyDreaming:
Addresses #2333 .

@hannes-ucsc Simply running the mock test in `tests/test_notifier.py::TestNotifierConfig::test_notifier_from_config`  seems to set the default number of workers to four.  I'm sorry if there was any miscommunication earlier, and you may have mentioned the default and I missed it, but can you confirm that that sounds accurate?  If so, this PR should octuple the number of workers assigned.

Note: Manually change the queue retention time through the browser once this is merged.


DailyDreaming:
@hannes-ucsc I'll update them on `dev` and update our release tags, which are the instructions followed whenever we do a release.  I'll remember.  Thanks!

Issue #2375, opened 2019-08-26
Create roles (IAM policy format)


kozbo:
Check Fusillade docs
dcp-fusillade

chmreid:
Was able to successfully create the new roles by using the `roles.json` that @Bento007 provided in #2395 using the `setup_fusillade.py` script in the dcp-fusillade repo.

```
$ ./setup_fusillade.py dev --file roles.json
The following roles do not exist:
The following roles will be created:
	- dss_wrangler
	- dss_collector
	- dss_subscriber
	- dss_admin
	- dss_reader
The following roles will be updated:
The following groups will be created:
The following users will be created:
The following changes will be made:
[...snip...]
```

Now I would like to figure out how to enumerate roles and groups. Should be straightforward to do with a client using the following API endpoints:

* `/v1/groups` - list all groups
* `/v1/roles` - list all roles

chmreid:
In case it's useful, here's a simple script to enumerate all users, groups, or roles in a given stage:

https://gist.github.com/chmreid/632276c38b8f8119e965b4b1f24483b7

chmreid:
See [this comment](https://github.com/HumanCellAtlas/data-store/issues/2377#issuecomment-527569730) and [this gist](https://gist.github.com/chmreid/632276c38b8f8119e965b4b1f24483b7) for a script to enumerate fusillade users, groups, and roles.

PR #2395, opened 2019-08-27
Adding a document to explain the action and resource relationship for…

Bento007:
… defining authorization rules.

This is a recommended a guide line for defining your authorization policies based on the endpoints defines in the openapi spec.

relates to #2375


Bento007:
The fusillade config should live in the DSS since it is the DSS that manages these roles.

PR #2400, opened 2019-08-27
Untombstone Script (in progress)

jeffrey856:
Created a script to untombstone.

(Currently pseudo code)

### DIFF
--- a/scripts/untombstone.py
+++ b/scripts/untombstone.py
@@ -1,150 +1,101 @@
 import os, re
 import logging as logger
 
-from dss import Config, Replica, dynamodb
+from dss import Replica, dynamodb
 from dss.dynamodb import get_item, put_item, delete_item, DynamoDBItemNotFound
 from dss.index.es import ElasticsearchClient
-from cloud_blobstore import BlobNotFoundError
-from dss.index.es.backend.ElasticsearchIndexBackend import index_bundle
-from dss import DSSException, dss_handler, DSSForbiddenException
 from dss.collections import owner_lookup
-
-from hca.dss import DSSClient
-from hca.util.exceptions import SwaggerAPIException
+from dss.events.handlers.sync import dependencies_exist
+from dcplib.aws.sqs import SQSMessenger, get_queue_url
 from dss.storage.identifiers import DSS_BUNDLE_TOMBSTONE_REGEX as dead_template
 
+
+
+sns = boto3.client('sns')
+sns.meta.config.max_pool_connections = 100
+sns_topic_run = "domovoi-s3-events-org-humancellatlas-dss-dev"
+gs_topic_run = "dss-gs-events-org-humancellatlas-dss-dev"
 
 # --------------------------------------------------------------
 # untombstone bundles
 # --------------------------------------------------------------
 
-
-def untombstone_bundle(uuid, replica, version=None):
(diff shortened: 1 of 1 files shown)


dss/api/bundles/__init__.py
```diff
@@ -31,7 +37,61 @@
...
+            body={"query": {"terms": {"_id": ["{}.{}.dead".format(uuid, version)]}}},
+        )
+
+        logger.debug(f"removed dead bundle {uuid}.{version}.dead from es")
+        es_client.update_by_query(
+            index="all",
+            body={"query": {"terms": {"_id": ["{}.{}".format(uuid, version)]}}},
+        )
+        logger.debug("Restored bundle {uuid}.{version}")
+        es_back.index_bundle(restore_key)
```
xbrianh: Indexing a formerly tombstoned bundle should invoke the normal DSS indexing machinery by sending an SQS message to the index queue. You can use [this snippet](https://github.com/HumanCellAtlas/data-store/blob/fbb073543290082aab4f2950c5a71858bde4ba4f/dss/operations/elasticsearch.py#L37-L39) from the manual indexing operation as an example.

scripts/untombstone.py
```diff
@@ -0,0 +1,150 @@
...
+
+    if re.match(dead_template, key):
+        # matches dead fqid and deletes it from es
+        es_client.delete_by_query(
+            index="_all", body={"query": {"terms": {"_id": [key]}}}
+        )
+    logger.debug(f"removed dead bundle {key} from es")
+
+
+def update_og_bundle(fqid):
```
xbrianh: What is the purpose of this method?

scripts/untombstone.py
```diff
@@ -0,0 +1,150 @@
...
+        key = f"{uuid}.{version}.dead"
+    else:
+        key = f"{uuid}.dead"
+    if tombstoned_or_not_bundle(key, replica) is False:
+        pass
+    else:
+        deindex_dead_bundle(key)
+        update_og_bundle(key)
+        # indexs and notifys subs
+        index_bundle(key)
```
xbrianh: This should not be called directly. Instead, the normal DSS indexing machinery should be invoked by sending an SQS message to the index queue. You can use [this DSS operations snippet](https://github.com/HumanCellAtlas/data-store/blob/master/dss/operations/elasticsearch.py#L37-L39) as an example.

xbrianh:
Several DSS operations are invoked during bundle creation, such as the ES indexer, sync daemon, and v2 notifier. The most straightforward way to restore a bundle to it's normal state is by populating object creation events into the SNS object creation topic for each replica. For s3, the SNS topic is `domovoi-s3-events-org-humancellatlas-dss-dev`,  and on GS it is `dss-gs-events-org-humancellatlas-dss-dev`. This will cause all of the usual bundle daemons to be invoked.

You should verify the referential integrity of the bundle to be restored. This can be performed with the [existing dependency check used during sync operations](https://github.com/HumanCellAtlas/data-store/blob/fbb073543290082aab4f2950c5a71858bde4ba4f/dss/events/handlers/sync.py#L231). It should be invoked on the bundle key, and then each file key in the bundle.

Also note that a bundle/collection should be restored on either all replicas or none.

PR #2410, opened 2019-08-30
Requirements Update

amarjandu:
updates Domovoi to `2.0.2` to include asset.id for downstream tagging.


DailyDreaming:
Just a heads up, `httpie==1.0.3` is probably going to cause problems that need to fixed: https://github.com/HumanCellAtlas/data-store/issues/2404

PR #2412, opened 2019-09-04
Periodically check for replica consistency

natanlao:
Current limitations:
- Only checks if GCP is missing anything on AWS (not if AWS is missing anything on GCP)

How it works:
- The script `scripts/verify_consistency.sh` will list all the objects in the given S3 bucket and compare files, objects, and bundles present in GCP.
- The datetime of the oldest blob/file/bundle verified is stored in SSM.
- If any inconsistencies are found, the pipeline will fail.
- `.gitlab-ci.yml` is configured with two jobs - one job that verifies all blobs/files/bundles, and a job that verifies only new blobs/files/bundles.
- Gitlab will be configured to notify someone if either of the verification jobs fail. (I would appreciate feedback on who should be notified.)
- Gitlab will be configured to run both of these jobs periodically. (I would appreciate feedback on how often each check should be run.)

Does not cl*se #1680; Gitlab will need to be configured after this PR is merged.


.gitlab-ci.yml
```diff
@@ -171,3 +172,28 @@ force_release_prod:
...
   only:
     - staging
+
+verify_sync_partial:on-schedule:
+  stage: verify_sync
+  only:
+    refs:
+      - schedules
+  script:
+    - FILES_LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names dcp/dss/${DSS_DEPLOYMENT_STAGE}/files_last_verified | jq -r '.[]')"
```
amarjandu: can we change the secret name to `file_last_verified_time` or something similar? also these secrets need to be generated off the bad for a new instance of the DSS that should be reflected in the README, or there should be some way to create the secrets required if they dont exists.

natanlao:
> Does the pipeline run every time we update infra? or is it going to be scheduled to run on a weekly basis?

@amarjandu The latter, though I am not set on the interval yet.

.gitlab-ci.yml
natanlao: > also these secrets need to be generated off the bad for a new instance of the DSS that should be reflected in the README, or there should be some way to create the secrets required if they dont exists.

Would the appropriate way to approach this to be to set these parameters with Terraform?

xbrianh:
@natanlao The frequency can be stage dependent. Probably quarterly for the production deployment.

natanlao:
How about this for a schedule? 

|         | Full           | Quick         |
|---------|----------------|---------------|
| prod    | every 3 months | every month   |
| staging | every 6 weeks  | every 3 weeks |
| dev     | none           | every 6 hours |

.gitlab-ci.yml
```diff
@@ -171,3 +173,28 @@ force_release_prod:
...
+    - bash scripts/verify_consistency.sh bundles "${BUNDLE_LAST_VERIFIED}" | tee bundle_verify_log
+    - test "$(cat *_verify_log | wc -l)" -eq 0  # If any inconsistencies, fail pipeline
+
+verify_sync_full:on-schedule:
+  stage: verify_sync
+  only:
+    refs:
+      - schedules
+  script:
+    - bash scripts/verify_consistency.sh files 0 | tee file_verify_log
```
Bento007: where do these log files go?

Bento007: Is there a way of knowing which verify failed based on this test without running the script again?

natanlao: > where do these log files go?

They don't go anywhere - they're written so that we can see if any discrepancies were logged (see [L200](https://github.com/HumanCellAtlas/data-store/pull/2412/files/f50ec3d2fb5220398323245f020a5642e2e151f0#diff-96edf7a6f008de9e928d04e1ae5e12a5R200)) and are also printed to stdout. If any discrepancies are found, it will be listed in one of those files, and caught by L200 causing a failed build, which in turn will notify someone who can check the build log.

> Is there a way of knowing which verify failed based on this test without running the script again?

Yes, this should be clear from the build log.

.gitlab-ci.yml
amarjandu: Terraform might be a good location to house this information, but i think we would need to let Terraform know that we just need to instantiate these variables, and not care about the changes that are made to them, otherwise the change the value may cause breaks in the deployment scripting. 

We might be able to do this by using `ignore_changes` by Terraform https://www.terraform.io/docs/configuration/resources.html#ignore_changes

.gitlab-ci.yml
```diff
@@ -171,3 +173,31 @@ force_release_prod:
...
+  script:
+    - FILES_SECRETS_NAME="${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/{DSS_FILES_VERIFY_SECRETS_NAME}"
+    - BLOBS_SECRETS_NAME="${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/{DSS_BLOBS_VERIFY_SECRETS_NAME}"
+    - BUNDLES_SECRETS_NAME="${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/{DSS_BUNDLES_VERIFY_SECRETS_NAME}"
+    - FILES_LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names ${FILES_SECRETS_NAME} jq -r '.[]')"
+    - BLOBS_LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names ${BLOBS_SECRETS_NAME} | jq -r '.[]')"
+    - BUNDLE_LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names ${BUNDLES_SECRETS_NAME} | jq -r '.[]')"
+    - bash scripts/verify_consistency.sh files "${FILES_LAST_VERIFIED}" | tee file_verify_log
+    - bash scripts/verify_consistency.sh blobs "${BLOBS_LAST_VERIFIED}" | tee blob_verify_log
+    - bash scripts/verify_consistency.sh bundles "${BUNDLE_LAST_VERIFIED}" | tee bundle_verify_log
```
amarjandu: possible to print out the logs here? i assume the get lost when the image is destroyed?

scripts/verify_consistency.sh
```diff
@@ -0,0 +1,36 @@
...
+fi
+
+OBJECT_TYPE="$1"
+CUTOFF_TIME="'$2'"
+SECRET_NAME="dcp/dss/${DSS_DEPLOYMENT_STAGE}/${OBJECT_TYPE}_last_verified"
+
+# We want to get a list of all objects of prefix $TYPE that were created after
+# $CUTOFF. To do that, we have to list all objects in $DSS_S3_BUCKET and sort
+# them ourselves, locally.
+aws s3api list-objects-v2 \
```
amarjandu: this has a 1000 object limit for how many things get returned, is that problematic?

infra/secretsmanager/main.tf
```diff
@@ -0,0 +1,50 @@
+locals {
+  file_secret_name   = "${var.DSS_SECRETS_STORE}/${var.DSS_DEPLOYMENT_STAGE}/${var.DSS_FILE_VERIFY_SECRETS_NAME}"
+  blob_secret_name   = "${var.DSS_SECRETS_STORE}/${var.DSS_DEPLOYMENT_STAGE}/${var.DSS_BLOB_VERIFY_SECRETS_NAME}"
+  bundle_secret_name = "${var.DSS_SECRETS_STORE}/${var.DSS_DEPLOYMENT_STAGE}/${var.DSS_BUNDLE_VERIFY_SECRETS_NAME}"
+}
+
+resource "aws_secretsmanager_secret" "files_last_verified" {
```
amarjandu: please apply resource tagging if available, see `infra/buckets/s3.tf` for help.

.gitlab-ci.yml
natanlao: Yes, `tee` will write the logs to both file and stdout

scripts/verify_consistency.sh
natanlao: No, `aws s3api list-objects-v2` is transparently paginated

infra/secretsmanager/main.tf
natanlao: Done!

infra/secretsmanager/main.tf
```diff
@@ -2,21 +2,32 @@ locals {
   file_secret_name   = "${var.DSS_SECRETS_STORE}/${var.DSS_DEPLOYMENT_STAGE}/${var.DSS_FILE_VERIFY_SECRETS_NAME}"
   blob_secret_name   = "${var.DSS_SECRETS_STORE}/${var.DSS_DEPLOYMENT_STAGE}/${var.DSS_BLOB_VERIFY_SECRETS_NAME}"
   bundle_secret_name = "${var.DSS_SECRETS_STORE}/${var.DSS_DEPLOYMENT_STAGE}/${var.DSS_BUNDLE_VERIFY_SECRETS_NAME}"
+  tags = "${map(
+  "Name"      , "${var.DSS_INFRA_TAG_SERVICE}-s3-storage",
```
amarjandu: the name needs to be changed! `"${var.DSS_INFRA_TAG_SERVICE}-secrets"`

scripts/verify_consistency.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo "CUTOFF_TIME determines which objects are selected for review and"
+    echo "    is formatted as an iso8601 datetime. To verify all present"
+    echo "    objects, provide a CUTOFF_TIME of zero (or something else"
+    echo "    absurdly small)"
+    exit 1
+fi
+
+OBJECT_TYPE="$1"
+CUTOFF_TIME="'$2'"
+SECRET_NAME="dcp/dss/${DSS_DEPLOYMENT_STAGE}/${OBJECT_TYPE}_last_verified"
```
amarjandu: the secret store variable should be used here.

scripts/verify_consistency.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# We want to get a list of all objects of prefix $TYPE that were created after
+# $CUTOFF. To do that, we have to list all objects in $DSS_S3_BUCKET and sort
+# them ourselves, locally.
+aws s3api list-objects-v2 \
+    --bucket "$DSS_S3_BUCKET" \
+    --prefix "${OBJECT_TYPE}/" \
+    --query "Contents[?LastModified > ${CUTOFF_TIME} && !ends_with(Key, '.dead')]" \
```
DailyDreaming: If someone, say, adds new tags to all objects in a PR, will that change the `LastModified` time?  Also, I don't expect versioned bundles to change, but does the root `bundle` ever change in a way that might throw this off?

scripts/verify_consistency.sh
```diff
@@ -0,0 +1,37 @@
...
+
+OBJECT_TYPE="$1"
+CUTOFF_TIME="'$2'"
+SECRET_NAME="dcp/dss/${DSS_DEPLOYMENT_STAGE}/${OBJECT_TYPE}_last_verified"
+
+# We want to get a list of all objects of prefix $TYPE that were created after
+# $CUTOFF. To do that, we have to list all objects in $DSS_S3_BUCKET and sort
+# them ourselves, locally.
+aws s3api list-objects-v2 \
+    --bucket "$DSS_S3_BUCKET" \
```
DailyDreaming: Since this only checks `s3`, if there's a google file that exists in the `gcp` bucket, but not in the `s3` bucket, will this catch it?

natanlao:
Closed after discussion with @xbrianh

PR #2428, opened 2019-09-16
DSS SmokeTest: Test Order

amarjandu:
Reorder smoke test to allow more time between replication and prefix enumeration

### DIFF
--- a/tests/test_smoketest.py
+++ b/tests/test_smoketest.py
@@ -129,7 +129,7 @@ class Smoketest(BaseSmokeTest):
                     self.fail("Timed out waiting for notification to arrive")
 
         for replica in self.replicas:
-            # Enumerations against the replicas should be done after the test_replica_sync to ensure consistency. 
+            # Enumerations against the replicas should be done after the test_replica_sync to ensure consistency.
             with self.subTest(f'Testing Bundle Enumeration on {replica.name}'):
                 enumerate_bundles = list()
                 page_size = 10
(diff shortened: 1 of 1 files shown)


DailyDreaming:
Actually, if this is addressing:
```
======================================================================
FAIL: test_smoketest (__main__.Smoketest) [gcp: Enumerate on that bundle uuid]
  File "tests/test_smoketest.py", line 88, in smoketest
    self.assertIn(bundle_match, resp['bundles'])
AssertionError: {'uuid': '580361e4-3251-48e3-ab4a-af26eac7fa48', 'version': '2019-09-16T213142.666530Z'} not found in []
```
from: https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/46374

It might be better to do an intelligent wait until it appears (within, for example, 5 seconds or so).

amarjandu:
> Actually, if this is addressing:
> 
> ```
> ======================================================================
> FAIL: test_smoketest (__main__.Smoketest) [gcp: Enumerate on that bundle uuid]
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "tests/test_smoketest.py", line 88, in smoketest
>     self.assertIn(bundle_match, resp['bundles'])
> AssertionError: {'uuid': '580361e4-3251-48e3-ab4a-af26eac7fa48', 'version': '2019-09-16T213142.666530Z'} not found in []
> ```
> 
> from: https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/46374
> 
> It might be better to do an intelligent wait until it appears (within, for example, 5 seconds or so).

Performing the enumeration as the last operation before deletion would ensure that its cross synced, see `tests/infra/base_smoketest.py` _test_replica_sync. i can add a delay, but placing the operation here should be fine, otherwise the sync tests would fail.

DailyDreaming:
@amarjandu Just add a comment then.

PR #2324, opened 2019-08-05
Add Lambda SSM environment variable management to dss-ops script

chmreid:
This pull request adds a new class of actions to the `dss-ops.py` operational tooling script, called `params`. Namely:

- [x] `./dss-ops.py params` to access the new menu of actions for manipulating lambda/SSM parameters
- [x] `./dss-ops.py params ssm-list` to list all environment variables in the SSM (store)
- [x] `./dss-ops.py params ssm-set` to add a new environment variable into the SSM (store)
- [x] `./dss-ops.py params ssm-unset` to unset an environment variable in the SSM (store)
- [x] `./dss-ops.py params lambda-list` to list all environment variables for all deployed lambda functions
- [x] `./dss-ops.py params lambda-set` to set the value of an environment variable for all deployed lambda functions
- [x] `./dss-ops.py params lambda-unset` to unset the value of an environment variable for all deployed lambda functions
- [x] `./dss-ops.py params lambda-update` to update the lambda environment stored in the parameter store (and optionally, update the environment of all deployed lambda functions)

Supplemental changes to documentation:

- [x] README updates
- [ ] wiki documents in allspark (dcp-ops wiki)

This PR closes #2299 

(Please merge this commit by squashing)

PR #2432, opened 2019-09-17
Add SSM parameter store functionality to dss-ops

### LINKED ISSUE #2299: Create operational tooling for Lambda SSM environment variables
The functionality of `scripts/populate_lambda_ssm_parameters.py` should be moved into the DSS operational framework, providing capability to:
  - [x] Print the current parameters in the SSM store
  - [x] Push new parameters into the SSM store
  - [ ] Print the current environment variables in the deployed Lambda
  - [ ] Set a single variable into each deployed Lambda
  - [ ] Remove a single variable from each deployed Lambda

chmreid:
This PR adds functionality to the `dss-ops.py` script that allows operators to list, set, and unset parameters stored in the SSM store under the "environment" variables. It also adds tests for the new SSM parameter functionality.

This PR is part of #2299 and is a reworked version of (part of) a PR that was abandoned (#2324). This PR is much simpler and should be easier/faster to review.

### DIFF
--- a/dss/operations/ssm_params.py
+++ b/dss/operations/ssm_params.py
@@ -37,7 +37,7 @@ def get_ssm_variable_prefix() -> str:
 def fix_ssm_variable_prefix(param_name: str) -> str:
     """Add the variable store and stage prefix to the front of an SSM param name"""
     prefix = get_ssm_variable_prefix()
-    if not param_name.startswith(prefix):
+    if not (param_name.startswith(prefix) or param_name.startswith("/"+prefix)):
         param_name = f"{prefix}/{param_name}"
     return param_name
 
@@ -152,7 +152,7 @@ def ssm_set(argv: typing.List[str], args: argparse.Namespace):
         print(f'Name: {name}')
         print(f'Value: {val}')
     else:
-        set_ssm_var(name, val)
+        set_ssm_parameter(name, val)
 
 
 @ssm_params.action(
@@ -175,4 +175,4 @@ def ssm_unset(argv: typing.List[str], args: argparse.Namespace):
     if args.dry_run:
         print(f'Dry-run deleting variable "{name}" from SSM store')
     else:
-        unset_ssm_var(name)
+        unset_ssm_parameter(name)
--- a/dss/operations/ssm_params.py
+++ b/dss/operations/ssm_params.py
@@ -1,12 +1,17 @@
 """
 Get and set parameters in the "environment" variable in the SSM parameter store.
-These parameters are used in a way that is similar to environment variables.
+
+Parameters in the SSM store are utilized similar to environment variables. Any parameter set with
+this script will be stored in the SSM parameter store, in the "environment" variable, which is
+stored as a serialized JSON object with key-value pairs.
 """
 import os
 import sys
 import select
 import json
 import argparse
+import logging
+import typing
 
 from botocore.exceptions import ClientError
 
--- a/dss/operations/ssm_params.py
+++ b/dss/operations/ssm_params.py
@@ -94,3 +94,80 @@ def unset_ssm_parameter(environment: dict, set_env_fn, env_var: str) -> None:
         print(f'Previous value: {prev_value}')
     except KeyError:
         print(f'Nothing to unset for variable "{env_var}" in SSM param store environment')
+
+
+ssm_params = dispatch.target("params", arguments={}, help=__doc__)
+
+
+@ssm_params.action(
+    "list",
+    arguments={
+        "--json": dict(
+            default=False,
+            action="store_true",
+            help="format the output as JSON"
+        )
+    }
+)
+def ssm_list(argv: typing.List[str], args: argparse.Namespace):
+    """Print out all variables stored in the SSM store"""
+    ssm_env = get_ssm_environment()
+    if args.json:
+        print(json.dumps(ssm_env, indent=4))
+    else:
+        for name, val in ssm_env.items():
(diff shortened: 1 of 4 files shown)


dss/operations/ssm_params.py
```diff
@@ -0,0 +1,164 @@
...
+import typing
+
+from botocore.exceptions import ClientError
+
+from dss.operations import dispatch
+from dss.util.aws.clients import secretsmanager  # type: ignore
+import dss.operations.util as util
+
+from dss.util.aws.clients import ssm as ssm_client  # type: ignore
+from dss.util.aws.clients import secretsmanager as sm_client  # type: ignore
```
chmreid: FIXME: just noticed secretsmanager is imported twice

dss/operations/ssm_params.py
```diff
@@ -0,0 +1,161 @@
...
+
+def set_ssm_parameter(env_var: str, value) -> None:
+    """
+    Set a parameter in the SSM param store variable "environment".
+
+    :param env_var: the name of the environment variable being set
+    :param value: the value of the environment variable being set
+    """
+    environment = get_ssm_environment()
+    try:
```
natanlao: Is there a reason this is not `prev_value = environment.get(env_var)`?

tests/__init__.py
```diff
@@ -160,3 +160,31 @@ def __exit__(self, *args, **kwargs):
         # Clean up by setting sys.stdout back to what it was before we opened
         # up this context
         sys.stdout = self._stdout
+
+class SwapStdin(object):
+    """Utility object using a context manager to swap out stdin with user-provided data."""
+    def __init__(self, *, input=None):
```
natanlao: `input` is a builtin, you may want to consider renaming the argument

tests/__init__.py
```diff
@@ -160,3 +160,31 @@ def __exit__(self, *args, **kwargs):
...
+        To swap out stdin properly (so it still works with the select module) requires a "real" file
+        with an os-level file descriptor. Fortunately os.pipe() will create a file descriptor for
+        the pipe, so we create a pipe and fill it with mock data, then swap it out with stdin.
+        """
+        fdr, fdw = os.pipe()
+        os.write(fdw, self.input)
+        os.close(fdw)
+        f = os.fdopen(fdr)  # use the file descriptor directly
+
+        self._stdin = sys.stdin
```
natanlao: Nit: the original `stdin` (at start of program execution) is also stored in `sys.__stdin__`

tests/test_operations.py
```diff
@@ -423,6 +424,78 @@ def paginate(self):
...
+        testvar_value2 = "Goodbye world!"
+
+        # Assemble an old and new environment to return
+        old_env = {"dummy_key": "dummy_value"}
+        new_env = dict(**old_env)
+        new_env[testvar_name] = testvar_value
+        ssm_old_env = _wrap_ssm_env(old_env)
+        ssm_new_env = _wrap_ssm_env(new_env)
+
+        with self.subTest("Create a new SSM parameter"):
```
natanlao: Nit: to reduce nesting, these context managers can all be invoked in the same `with` statement a la:

    with x() as ctx, \
         y() as ctx2, \
         z() as ctx3(():

etc.

dss/operations/ssm_params.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+@ssm_params.action(
+    "list",
+    arguments={
+        "--json": dict(default=False, action="store_true", help="format the output as JSON")
+    },
+)
+def ssm_list(argv: typing.List[str], args: argparse.Namespace):
+    """Print out all variables stored in the SSM store"""
```
xbrianh: Here and throughout this PR, docstrings could be expanded to indicate which SSM store we're dealing with:
```
"""Print out all variables stored in the SSM store $DSS_DEPLOYMENT_STAGE/environment"""
```

dss/operations/ssm_params.py
```diff
@@ -0,0 +1,161 @@
...
+    environment[env_var] = value
+    set_ssm_environment(environment)
+    print(f"Success! Set variable in SSM parameter store environment:")
+    print(f"Name: {env_var}")
+    print(f"Value: {value}")
+    if prev_value:
+        print(f"Previous value: {prev_value}")
+
+
+def unset_ssm_parameter(env_var: str) -> None:
```
xbrianh: nit: This and `set_ssm_parameter` are only used in one place. Consider inlining them into the respective operations methods.

dss/operations/ssm_params.py
chmreid: No, just some inefficient code. Thank you for the suggestion!

tests/__init__.py
chmreid: Forgot to do that, fixed!

tests/test_operations.py
chmreid: Very nice!

tests/__init__.py
chmreid: Great! Got rid of `self._stdin` and using `sys.__stdin__` instead.

dss/operations/ssm_params.py
chmreid: Updated - thanks for the suggestion! I wasn't too happy with "SSM store" or "SSM param store environment" so this works better.

dss/operations/ssm_params.py
chmreid: Done! thanks for pointing this out

Issue #2293, opened 2019-07-29
Event journaling and replay


xbrianh:
Event journaling and replay should be implemented according to https://github.com/HumanCellAtlas/dcp-community/blob/bhannafi-dss-event-journaling.md/rfcs/text/0000-dss-event-journaling.md

In order to implement event recording and playback:
  - [x] Integration of FlashFlood into the DSS to record new events
  - [x] Creat daemon to perform event journaling 
  - [x] Implement GET /events and GET /event API endpoints

PR #2433, opened 2019-09-17
Define the bucket hosting DSS event data

xbrianh:
This also defines the canonical interfaces to the event bucket and replica prefixes that will contain event data.

connected to #2293


environment
```diff
@@ -66,6 +67,10 @@ DSS_S3_CHECKOUT_BUCKET_UNWRITABLE=org-hca-dss-checkout-unwritable
 DSS_S3_CHECKOUT_BUCKET_INTEGRATION=org-humancellatlas-dss-checkout-integration
 DSS_S3_CHECKOUT_BUCKET_STAGING=org-humancellatlas-dss-checkout-staging
 DSS_S3_CHECKOUT_BUCKET_PROD=org-humancellatlas-dss-checkout-prod
+DSS_FLASHFLOOD_BUCKET=org-humancellatlas-dss-events-dev
+DSS_FLASHFLOOD_BUCKET_INTEGRATION=org-humancellatlas-dss-events-integration
```
amarjandu: Should these be set in the corresponding environment.{stage} file?

infra/buckets/s3.tf
```diff
@@ -150,3 +150,12 @@ resource aws_s3_bucket dss_s3_checkout_bucket_unwritable {
 }
 POLICY
 }
+
+resource aws_s3_bucket dss_s3_events_bucket {
+  count = "${length(var.DSS_S3_BUCKET) > 0 ? 1 : 0}"
```
amarjandu: does the var evaluated in the count need to be the event bucket?

environment
xbrianh: yep, fixed

infra/buckets/s3.tf
xbrianh: yes, fixed

infra/buckets/s3.tf
```diff
@@ -150,3 +150,12 @@ resource aws_s3_bucket dss_s3_checkout_bucket_unwritable {
...
 POLICY
 }
+
+resource aws_s3_bucket dss_s3_events_bucket {
+  count = "${length(var.DSS_FLASHFLOOD_BUCKET) > 0 ? 1 : 0}"
+  bucket = "${var.DSS_FLASHFLOOD_BUCKET}"
+  server_side_encryption_configuration {
+    rule {apply_server_side_encryption_by_default {sse_algorithm = "AES256"}}
+  }
+  tags = "${local.common_tags}"
```
natanlao: Is there a reason this isn't `tags = "${merge(local.common_tags, local.aws_tags)}"`?

tests/test_config.py
```diff
@@ -43,6 +43,14 @@ def test_s3_checkout_bucket(self):
         Config.set_config(BucketConfig.TEST_FIXTURE)
         self.assertEquals(Config.get_s3_checkout_bucket(), os.environ["DSS_S3_CHECKOUT_BUCKET_TEST"])
 
+    def test_s3_events_bucket(self):
+        Config.set_config(BucketConfig.NORMAL)
+        self.assertEquals(Config.get_flashflood_bucket(), os.environ["DSS_FLASHFLOOD_BUCKET"])
```
natanlao: Nit: `assertEquals` is deprecated, use `assertEqual` instead. (I was under the impression that `assertEquals` would stop working entirely, but it looks like it just raises a warning after 3.2, and in theory will stop working in the future.)

infra/buckets/s3.tf
xbrianh: fixed

tests/test_config.py
xbrianh: changed.

PR #2434, opened 2019-09-17
Use the flash-flood library to record bundle events

xbrianh:
This uses the [flash-flood](https://github.com/xbrianh/flash-flood) library to record event data to s3.

depends on  #2433 
connected to #2293


dss/events/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+    name = name.replace('.', '_')
+    if name.endswith('_json'):
+        name = name[:-5]
+        parts = name.rpartition("_")
+        if name != parts[2]:
+            name = parts[0]
+        name += "_json"
+    return name
+
+def datetime_from_timestamp(ts):
```
amarjandu: nit: keep this loaded in `dss.util.version` 🔫

xbrianh:
@amarjandu The event data is exactly the same as is used in the `dss-notify-v2` daemon. The notification data delivered to subscriber endpoints is different from the event data, but that's not what's being recorded here.

dss/events/__init__.py
xbrianh: done

PR #2435, opened 2019-09-17
Define the API to retrieve and playback events

xbrianh:
depends on #2433 #2434 
connected to #2293

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -2347,7 +2347,7 @@ paths:
                   code:
                     type: string
                     description: Machine-readable error code.  The types of return values should not be changed lightly.
-                    enum: [unhandled_exception, Forbidden, Unauthorized, OAuthResponseProblem, not_found]
+                    enum: [unhandled_exception, illegal_arguments, bad_request]
                 required:
                   - code
   /events/{uuid}:
@@ -2378,55 +2378,16 @@ paths:
           description: OK
           schema:
             type: object
-        400:
-          description: Bad request
-          schema:
-            allOf:
-              - $ref: '#/definitions/Error'
-              - type: object
-                properties:
-                  code:
-                    type: string
-                    description: >
-                      Machine-readable error code.  The types of return values should not be changed lightly.
-
-
-                      The code `illegal_arguments` is returned when required parameters are missing or parameters do not
-                      match the requirements in the specification.
-                    enum: [illegal_arguments]
-                required:
-                  - code
         500:
-          description: Server Error.
-          headers:
-            Retry-After:
-              description: Delay in seconds, service clients should retry after the delay.
-              type: integer
-              format: int64
-          schema:
-            allOf:
-              - $ref: '#/definitions/Error'
-              - type: object
-                properties:
-                  code:
-                    type: string
-                    description: >
-                      Machine-readable error code.  The types of return values should not be changed lightly.
-
-
-                      The code `unhandled_exception` is returned when an unexpected exception is encountered.
-                    enum: [unhandled_exception]
-                required:
-                  - code
+          $ref: '#/responses/ServerError'
         502:
           $ref: '#/responses/BadGateway'
         503:
-          description: Service Unavailable.
-          headers:
-            Retry-After:
-              description: Delay in seconds, service clients should retry after the delay.
-              type: integer
-              format: int64
+          $ref: '#/responses/ServiceUnavailable'
+        504:
+          $ref: '#/responses/GatewayTimeout'
+        default:
+          description: Unexpected error
           schema:
             allOf:
               - $ref: '#/definitions/Error'
@@ -2434,17 +2395,10 @@ paths:
                 properties:
                   code:
                     type: string
-                    description: >
-                      Machine-readable error code.  The types of return values should not be changed lightly.
-
-
-                      The code `service_unavailable` is returned when service is unavailable because of unusually high
-                      load/latency
-                    enum: [service_unavailable]
+                    description: Machine-readable error code.  The types of return values should not be changed lightly.
+                    enum: [unhandled_exception, illegal_arguments, not_found]
                 required:
                   - code
-        504:
-          $ref: '#/responses/GatewayTimeout'
 
 definitions:
   File:
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -2257,6 +2257,186 @@ paths:
                     description: Detailed error message.
                 required:
                   - code
+  /events:
+    get:
+      operationId: dss.api.events.list_events
+      summary: Replay events
+      description: >
+        Return urls where event data is available, with manifest of contents.
+      parameters:
+        - name: from_date
+          in: query
+          description: Timestamp to begin replaying events, in DSS_VERSION format.  If this is not provided, replay from the earliest event.
+          required: false
+          type: string
+        - name: to_date
+          in: query
+          description: Timestamp to stop replaying events, in DSS_VERSION format.  If this is not provided, replay to the latest event.
+          required: false
+          type: string
+        - name: replica
+          in: query
+          description: Replica to fetch from.
+          required: true
+          type: string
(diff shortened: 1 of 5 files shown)


dss-api.yml
```diff
@@ -2666,6 +2846,54 @@ definitions:
...
+            - event_id
+            - timestamp
+            - offset
+            - size
+    required:
+      - journal_id
+      - from_date
+      - to_date
+      - size
+      - events
```
amarjandu: nit: add `{dss_api:f'https://{os.environ.get("API_DOMAIN_NAME")}'`

amarjandu: nit: provide `{"page_count":len(events),"per_page":per_page}` for the users, just a nice to have

tests/test_events.py
```diff
@@ -0,0 +1,105 @@
...
+import json
+import unittest
+import logging
+import typing
+import boto3
+import time
+from uuid import uuid4
+from datetime import datetime
+
+from flashflood import replay_with_urls
```
amarjandu: does this need to be added to requirements?

dss-api.yml
amarjandu: nit: provide next_link for the user in the response of the body.

tests/test_events.py
amarjandu: ignore this, i saw the other PR.

dss/api/events.py
```diff
@@ -0,0 +1,39 @@
...
+@dss_handler
+def list_events(replica: str, from_date: str=None, to_date: str=None, per_page: int=1, start_at: int = 0):
+    ff = FlashFlood(resources.s3, Config.get_flashflood_bucket(), Replica[replica].flashflood_prefix)  # type: ignore
+    urls = ff.replay_urls(events.datetime_from_timestamp(from_date) if from_date else datetime.min,
+                          events.datetime_from_timestamp(to_date) if to_date else datetime.max,
+                          per_page)
+    if len(urls) < per_page:
+        response = make_response(jsonify(urls), requests.codes.ok)
+    else:
+        next_url = UrlBuilder(request.url)
```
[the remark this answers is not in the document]
xbrianh: thanks for catching this.  Headers added.

dss-api.yml
```diff
@@ -2666,6 +2846,54 @@ definitions:
...
+        description: DSS_VERSION format timestamp of the first event in the journal.
+        type: string
+      to_date:
+        type: string
+        format: DSS_VERSION
+        description: DSS_VERSION format timestamp of the last event in the journal.
+        type: string
+      size:
+        type: integer
+        description: Size of the journa in bytes.
```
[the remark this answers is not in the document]
xbrianh: fixed

dss-api.yml
xbrianh: I don't think this information belongs in the event manifest (which I can describe in greater detail), although it may be beneficial to include it in the paginated response to GET /events.

amarjandu: ^ i agree, apologies i was writing comments in the wrong space.

dss-api.yml
```diff
@@ -2257,6 +2257,194 @@ paths:
...
+          enum: [aws, gcp]
+        - name: per_page
+          in: query
+          description: Max number of results to return per page.
+          required: false
+          type: integer
+          format: int32
+          minimum: 1
+          maximum: 10
+          default: 1
```
DailyDreaming: Should this be increased?

dss-api.yml
```diff
@@ -2257,6 +2257,194 @@ paths:
...
+        502:
+          $ref: '#/responses/BadGateway'
+        503:
+          description: Service Unavailable.
+          headers:
+            Retry-After:
+              description: Delay in seconds, service clients should retry after the delay.
+              type: integer
+              format: int64
+          schema:
```
DailyDreaming: Would it be useful to add this schema to all 503 responses by adding it here?  https://github.com/HumanCellAtlas/data-store/blob/master/dss-api.yml#L2730

dss-api.yml
```diff
@@ -2683,6 +2871,54 @@ definitions:
       - type
       - uuid
       - version
+  EventsManifest:
+    type: object
+    properties:
+      journal_id:
```
DailyDreaming: This is returned but I don't see any way for the user to use the `journal_id` currently.  Are there future plans for users to be able to get a journal by id?

dss/api/events.py
```diff
@@ -0,0 +1,42 @@
...
+        response.headers['Link'] = link
+        response.headers['X-OpenAPI-Pagination'] = 'true'
+        response.headers['X-OpenAPI-Paginated-Content-Key'] = 'event_streams'
+    return response
+
+@dss_handler
+def get(uuid: str, replica: str, version: str = None):
+    key = f"bundles/{uuid}.{version}"
+    doc = events.get_bundle_metadata_document(Replica[replica], key)
+    return doc
```
DailyDreaming: It seems that if this doesn't exist, it will return `None`?  https://github.com/HumanCellAtlas/data-store/blob/8ba67879dedbb0b9bb3e60ffaf328b7320b989e5/dss/events/__init__.py#L36

Should this return a `404` instead?

dss-api.yml
```diff
@@ -2257,6 +2257,194 @@ paths:
...
+          description: Replica to fetch from.
+          required: true
+          type: string
+          enum: [aws, gcp]
+      responses:
+        200:
+          description: OK
+          schema:
+            type: object
+        400:
```
DailyDreaming: Should this include a `404` response?

tests/test_events.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def setUp(self):
+        Config.set_config(dss.BucketConfig.TEST)
+        events.record_event_for_bundle.cache_clear()
+        for replica in Replica:
+            name = f'DSS_{replica.name.upper()}_FLASHFLOOD_PREFIX'
+            os.environ[name] = f"flashflood-{uuid4()}"
+            events.record_event_for_bundle(replica, self.bundle[replica.name]['key'])
+
+    def test_list(self):
```
DailyDreaming: Could this test a bad request as well as the working case?

tests/test_events.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _test_list(self, replica):
+        res = self.app.get("/v1/events",
+                           params=dict(replica=replica.name))
+        self.assertIn(res.status_code, [requests.codes.ok, requests.codes.partial])
+        event = [e for e in replay_with_urls(res.json())][0]
+        event_doc = json.loads(event.data.decode("utf-8"))
+        self.assertEqual(events._build_bundle_metadata_document(replica, self.bundle[replica.name]['key']), event_doc)
+
+    def test_get(self):
```
DailyDreaming: Could this test when a GET `event` does not find the `event`?

dss-api.yml
xbrianh: The journals will typically be large files that are streamed from S3, so 1 seems reasonable for now. This may change if/when we add a CLI operation to facilitate event streaming.

dss-api.yml
xbrianh: Perhaps, but in a separate PR.

dss-api.yml
xbrianh: There are no plans provide a get journal endpoint. However, `journal_id` is still useful if users wish to store journals locally.

dss/api/events.py
xbrianh: yes, will fix

dss-api.yml
xbrianh: will fix

tests/test_events.py
xbrianh: fixed

tests/test_events.py
xbrianh: added test

PR #2436, opened 2019-09-18
Add lambda parameter functionality to dss-ops

### LINKED ISSUE #2389: Replace and remove populate_lambda_ssm_parameters script
This is a follow-on issue to issue #2299 (move lambda/SSM parameter handling into dss-operations), closed by PR #2324 (adding lambda/SSM parameter management to dss-ops script).

Now that the dss-ops.py script has functionality to update/modify parameters in the SSM environment and in lambda function environments (or, will after #2324 is closed), we should modify all of our devops tooling to remove references to the deprecated `populate_lambda_ssm_parameters.py` script.

chmreid:
This PR adds functionality to the dss-ops.py script that allows operators to list, set, and unset deployed lambda functions and their environments (environment variables).

This PR is part of #2299 and is a reworked version of (part of) a PR that was abandoned (#2324). This PR is much simpler and should be easier/faster to review.

This closes #2389

### DIFF
--- a/dss/operations/lambda_params.py
+++ b/dss/operations/lambda_params.py
@@ -98,7 +98,7 @@ def get_deployed_lambdas(quiet: bool = True):
             yield name
         except lambda_client.exceptions.ResourceNotFoundException:
             if not quiet:
-                logger.warning(f"{name} not deployed, or does not deploy a lambda function")
+                print(f"{name} not deployed, or does not deploy a lambda function")
 
 
 def get_deployed_lambda_environment(lambda_name: str, quiet: bool = True) -> dict:
@@ -108,7 +108,7 @@ def get_deployed_lambda_environment(lambda_name: str, quiet: bool = True) -> dic
         c = lambda_client.get_function_configuration(FunctionName=lambda_name)
     except lambda_client.exceptions.ResourceNotFoundException:
         if not quiet:
-            logger.warning(f"{lambda_name} is not a deployed lambda function")
+            print(f"{lambda_name} is not a deployed lambda function")
         return {}
     else:
         # above value is a dict, no need to convert
@@ -127,7 +127,7 @@ def get_local_lambda_environment(quiet: bool = True) -> dict:
     For each environment variable being set in deployed lambda functions, get the value of the
     environment variable from the local environment.
 
-    :param quiet: (boolean) if true, don't print warning messages
+    :param bool quiet: if true, don't print warning messages
     :returns: dict containing local environment's value of each variable exported to deployed lambda functions
     """
     env = dict()
@@ -136,7 +136,7 @@ def get_local_lambda_environment(quiet: bool = True) -> dict:
             env[name] = os.environ[name]
         except KeyError:
             if not quiet:
-                logger.warning(
+                print(
                     f"Warning: environment variable {name} is in the list of environment variables "
                     "to export to lambda functions, EXPORT_ENV_VARS_TO_LAMBDA, but variable is not "
                     "defined in the local environment, so there is no value to set."
--- a/dss/operations/lambda_params.py
+++ b/dss/operations/lambda_params.py
@@ -1,0 +1,202 @@
+"""
+Get and set environment variables in deployed lambda functions using the SSM param store
+variable named "environment".
+"""
+import os
+import sys
+import select
+import json
+import argparse
+import logging
+import typing
+
+from dss.operations import dispatch
+from dss.operations.ssm_params import get_ssm_variable_prefix, fix_ssm_variable_prefix
+
+from dss.util.aws.clients import ssm as ssm_client  # type: ignore
+from dss.util.aws.clients import secretsmanager as sm_client  # type: ignore
+from dss.util.aws.clients import es as es_client  # type: ignore
+import dss.util.aws.clients
+lambda_client = getattr(dss.util.aws.clients, "lambda")
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_elasticsearch_endpoint() -> str:
+    domain_name = os.environ["DSS_ES_DOMAIN"]
+    domain_info = es_client.describe_elasticsearch_domain(DomainName=domain_naame)
+    return domain_info["DomainStatus"]["Endpoint"]
+
+
+def get_admin_emails() -> str:
+
+    def get_secret_variable_prefix() -> str:
+        # TODO: this functionality should be moved to secrets.py
+        store_name = os.environ["DSS_SECRETS_STORE"]
+        stage_name = os.environ["DSS_DEPLOYMENT_STAGE"]
+        store_prefix = f"{store_name}/{stage_name}"
+        return store_prefix
+
+    def fix_secret_variable_prefix(secret_name: str) -> str:
+        # TODO: this functionality should be moved to secrets.py
+        prefix = get_secret_variable_prefix()
+        if not (secret_name.startswith(prefix) or secret_name.startswith("/" + prefix)):
+            secret_name = f"{prefix}/{secret_name}"
+        return secret_name
+
+    def fetch_secret_safely(secret_name: str) -> str:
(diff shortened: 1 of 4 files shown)


dss/operations/lambda_params.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def fix_secret_variable_prefix(secret_name: str) -> str:
+        # TODO: this functionality should be moved to secrets.py
+        prefix = get_secret_variable_prefix()
+        if not (secret_name.startswith(prefix) or secret_name.startswith("/" + prefix)):
+            secret_name = f"{prefix}/{secret_name}"
+        return secret_name
+    
+    def fetch_secret_safely(secret_name: str) -> dict:
+        # TODO:
```
xbrianh: Is there something to do?

dss/operations/lambda_params.py
```diff
@@ -0,0 +1,291 @@
...
+def get_elasticsearch_endpoint() -> str:
+    domain_name = os.environ["DSS_ES_DOMAIN"]
+    domain_info = es_client.describe_elasticsearch_domain(DomainName=domain_name)
+    return domain_info["DomainStatus"]["Endpoint"]
+
+
+def get_admin_emails() -> str:
+
+    def get_secret_variable_prefix() -> str:
+        # TODO: this functionality should be moved to secrets.py
```
xbrianh: consider expanding secrets.py to dss/operations/secrets.py

dss/operations/lambda_params.py
```diff
@@ -0,0 +1,291 @@
...
+from dss.util.aws.clients import secretsmanager as sm_client  # type: ignore
+from dss.util.aws.clients import es as es_client  # type: ignore
+import dss.util.aws.clients
+lambda_client = getattr(dss.util.aws.clients, "lambda")
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_elasticsearch_endpoint() -> str:
```
xbrianh: Is this method unused?

dss/operations/lambda_params.py
```diff
@@ -0,0 +1,291 @@
...
+import json
+import argparse
+import logging
+import typing
+
+from botocore.exceptions import ClientError
+
+from dss.operations import dispatch
+from dss.operations.ssm_params import get_ssm_variable_prefix, fix_ssm_variable_prefix
+from dss.operations.ssm_params import get_ssm_parameter, set_ssm_parameter
```
xbrianh: Looks like `get_ssm_parameter` needs to be removed.

dss/operations/lambda_params.py
```diff
@@ -0,0 +1,291 @@
...
+    if args.lambda_name:
+        lambda_names = [args.lambda_name]  # single lambda function
+    else:
+        lambda_names = list(get_deployed_lambdas())  # all lambda functions
+
+    # Iterate over lambda functions and get their environments
+    d = {}
+    for lambda_name in lambda_names:
+        lambda_env = get_deployed_lambda_environment(lambda_name, quiet=args.json)
+        d[lambda_name] = lambda_env
```
xbrianh: Consider skipping this assignment for non-deployed functions. Otherwise the output looks weird:
```
>scripts/dss-ops.py lambda environment --lambda-name dss-index-de
WARNING:dss.operations.lambda_params:dss-index-de is not a deployed lambda function

dss-index-de:
```

dss/operations/lambda_params.py
```diff
@@ -0,0 +1,291 @@
...
+from dss.operations.ssm_params import get_ssm_parameter, set_ssm_parameter
+
+from dss.util.aws.clients import ssm as ssm_client  # type: ignore
+from dss.util.aws.clients import secretsmanager as sm_client  # type: ignore
+from dss.util.aws.clients import es as es_client  # type: ignore
+import dss.util.aws.clients
+lambda_client = getattr(dss.util.aws.clients, "lambda")
+
+
+logger = logging.getLogger(__name__)
```
xbrianh: We might consider using `print` instead of the logger throughout this file since this script will always be executed on an operators machine.

Note that this is not the case for some of the other operational scripting, which may be executed in lambda functions and therefore needs to output log messages.

dss/operations/lambda_params.py
```diff
@@ -0,0 +1,290 @@
...
+        return secret_name
+    
+    def fetch_secret_safely(secret_name: str) -> dict:
+        # TODO: this functionality should be moved to secrets.py
+        try:
+            response = sm_client.get_secret_value(SecretId=secret_name)
+        except ClientError as e:
+            if 'Error' in e.response:
+                errtype = e.response['Error']['Code']
+                if errtype == 'ResourceNotFoundException':
```
natanlao: Should this be caught in the `try` block, i.e. `except ResourceNotFoundException as e`?

dss/operations/lambda_params.py
```diff
@@ -0,0 +1,290 @@
...
+    if gcp_service_account_email not in email_list:
+        email_list.append(gcp_service_account_email)
+    return ",".join(email_list)
+
+
+def get_deployed_lambdas(quiet: bool = True):
+    """
+    Generator returning names of lambda functions
+
+    :param quiet: (boolean) if true, don't print warnings about lambdas that can't be found
```
natanlao: Nit: this could be written instead as
```suggestion
    :param bool quiet: if true, don't print warnings about lambdas that can't be found
```

dss/operations/lambda_params.py
chmreid: No, because boto exceptions are weird. There are no explicitly-defined exceptions anywhere, so you cannot for example `import boto3.ResourceNotFoundException`. There is only one exception, which is the `ClientError`, and you have to pull apart the exception message to see what type of exception it was.

That being said, if you have ideas on how to make this more elegant, I would very much appreciate it.

chmreid:
I wanted to update this PR to incorporate work from #2432, so I did a rebase of this branch onto master. But this broke links for all the prior code review comments (as I've already experienced). It is still hard for me to figure out the appropriate time and way of doing rebases...

dss/operations/lambda_params.py
chmreid: D'oh, I forgot that I needed set_ssm_parameter and unset_ssm_parameter in this file. I inlined those functions per your suggestion in PR #2432 . Thanks for pointing out the superfluous import.

dss/operations/lambda_params.py
chmreid: This method is used in the update functionality, which was not yet implemented when your review came in.

dss/operations/lambda_params.py
chmreid: Related: https://gist.github.com/chmreid/1441361f70efac009a88be4def576be1

natanlao: I think you can do this by catching `sm_client.exceptions.ResourceNotFoundException` instead of `botocore.exceptions.ClientError`. Relevant issue is boto/boto3#1195. [I tried reproducing it locally](https://github.com/HumanCellAtlas/data-store/commit/b0188f848ca59a6e108cb30c31f61b7d55f20384) and it seems to work, though I had to remove references to `get_ssm_parameter` and `set_ssm_parameter` to get it to run.

dss/operations/lambda_params.py
chmreid: Using `print()` everywhere now, instead of mixing `print` and `logger`

dss/operations/lambda_params.py
chmreid: Can't get this to work, `sm_client.exceptions.ResourceNotFoundException` doesn't derive from `BaseException`. :-/

dss/operations/lambda_params.py
chmreid: Done

chmreid:
Rebasing onto master

PR #2439, opened 2019-09-19
Periodically check for replica consistency (again)

natanlao:
Current limitations:
- Only checks if GCP is missing anything on AWS (not if AWS is missing anything on GCP)

How it works:
- Adds a `dss-ops.py` target `verify-sync-all` that runs `verify-sync` against all keys in a given bucket (optionally newer than a given DSS_VERSION)
- The datetime of the last replica verification is stored in SSM
- If any inconsistencies are found, the pipeline will fail.
- `.gitlab-ci.yml` is configured with two jobs - one job that verifies all blobs/files/bundles, and a job that verifies only new blobs/files/bundles.
- Gitlab will be configured to notify someone if either of the verification jobs fail. (I would appreciate feedback on who should be notified.)
- Gitlab will be configured to run both of these jobs periodically. (I would appreciate feedback on how often each check should be run.)

Does not cl*se #1680; Gitlab will need to be configured after this PR is merged. See [this comment](https://github.com/HumanCellAtlas/data-store/pull/2412#issuecomment-528573935) for schedule

This is a follow up to #2412

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -214,14 +214,14 @@ verify_sync_partial:on-schedule:
       - schedules
   before_script:
     - VERIFY_SECRET_NAME="${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/${DSS_VERIFY_SECRETS_NAME}"
-    - LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names ${VERIFY_SECRET_NAME} --json | jq -r '.[]')"
+    - LAST_VERIFIED="$(python scripts/dss-ops.py secrets get ${VERIFY_SECRET_NAME})"
     - NEW_LAST_VERIFIED="$(python -c 'import datetime; from dss.util.version import datetime_to_version_format; \
                                       print(datetime_to_version_format(datetime.datetime.utcnow()))')"
   script:
     - python scripts/dss-ops.py sync verify-sync-all --source-replica aws --destination-replica gcp \
           --since "${LAST_VERIFIED}" | tee verify_log
   after_script:
-    - python scripts/dss-ops.py secrets set --secret-name "${VERIFY_SECRET_NAME}" --secret-value "${NEW_LAST_VERIFIED}"
+    - echo -n "${NEW_LAST_VERIFIED}" | python scripts/dss-ops.py secrets set "${VERIFY_SECRET_NAME}"
     - test "$(cat verify_log | wc -l)" -eq 0  # If any inconsistencies, fail pipeline
 
 verify_sync_full:on-schedule:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,7 @@
 image: humancellatlas/dss-build-box
 # The Docker image `humancellatlas/dss-build-box` is created through a manual process from
-# `${DSS_HOME}/allspark.Dockerfile`. See the contents of `${DSS_HOME}/allspark.Dockerfile` # creation and usage instructions.
+# `${DSS_HOME}/allspark.Dockerfile`. See the contents of `${DSS_HOME}/allspark.Dockerfile`
+# creation and usage instructions.
 
 variables:
   GIT_SUBMODULE_STRATEGY: normal
@@ -16,6 +17,7 @@ stages:
   - integration_test
   - scale_and_performance
   - release
+  - verify_sync
 
 before_script:
   - date && date -u
@@ -202,3 +204,31 @@ force_release_prod:
   only:
     - staging
 
+# This cron job needs to be manually configured using the Gitlab web UI to run
+# periodically. See here for discussion on the proposed schedule:
+# https://github.com/HumanCellAtlas/data-store/pull/2412#issuecomment-528573935
+verify_sync_partial:on-schedule:
+  stage: verify_sync
+  only:
+    refs:
+      - schedules
+  before_script:
+    - VERIFY_SECRET_NAME="${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/${DSS_VERIFY_SECRETS_NAME}"
+    - LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names ${VERIFY_SECRET_NAME} --json | jq -r '.[]')"
+    - NEW_LAST_VERIFIED="$(python -c 'import datetime; from dss.util.version import datetime_to_version_format; \
+                                      print(datetime_to_version_format(datetime.datetime.utcnow()))')"
+  script:
+    - python scripts/dss-ops.py sync verify-sync-all --source-replica aws --destination-replica gcp \
+          --since "${LAST_VERIFIED}" | tee verify_log
+  after_script:
+    - python scripts/dss-ops.py secrets set --secret-name "${VERIFY_SECRET_NAME}" --secret-value "${NEW_LAST_VERIFIED}"
+    - test "$(cat verify_log | wc -l)" -eq 0  # If any inconsistencies, fail pipeline
+
+verify_sync_full:on-schedule:
+  stage: verify_sync
+  only:
+    refs:
+      - schedules
+  extends: verify_sync_partial:on-schedule
+  script:
+    - python scripts/dss-ops.py sync verify-sync-all --source-replica aws --destination-replica gcp | tee verify_log
--- a/dss/operations/sync.py
+++ b/dss/operations/sync.py
@@ -2,6 +2,7 @@
 Replication consistency checks: verify and repair synchronization across replicas
 """
 import argparse
+import concurrent
 import datetime
 import itertools
 import json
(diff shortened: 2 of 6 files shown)


.gitlab-ci.yml
```diff
@@ -171,3 +173,30 @@ force_release_prod:
     RELEASE_COMMAND: scripts/release.sh staging prod --force
   only:
     - staging
+
+verify_sync_partial:on-schedule:
```
Bento007: There is a lot of duplicate code between `verify_sync_partial:on-schedule` and `verify_sync_full:on-schedule:` can you consolidate?

Bento007 (approved):
LGTM

.gitlab-ci.yml
natanlao: Good call, done!

natanlao:
Force pushed to rebase off of master.

.gitlab-ci.yml
```diff
@@ -190,3 +192,28 @@ force_release_prod:
   only:
     - staging
 
+verify_sync_partial:on-schedule:
```
chmreid: Where are the scheduling details set for the verify_sync tasks?

.gitlab-ci.yml
```diff
@@ -190,3 +192,28 @@ force_release_prod:
...
+  before_script:
+    - VERIFY_SECRET_NAME="${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/${DSS_VERIFY_SECRETS_NAME}"
+    - LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names ${VERIFY_SECRET_NAME} --json | jq -r '.[]')"
+    - NEW_LAST_VERIFIED="$(python -c 'import datetime; from dss.util.version import datetime_to_version_format; \
+                                      print(datetime_to_version_format(datetime.datetime.utcnow()))')"
+  script:
+    - python scripts/dss-ops.py sync verify-sync-all --source-replica aws --destination-replica gcp \
+          --since "${LAST_VERIFIED}" | tee verify_log
+  after_script:
+    - python scripts/dss-ops.py secrets set --secret-name "${VERIFY_SECRET_NAME}" --secret-value "${NEW_LAST_VERIFIED}"
```
chmreid:
```suggestion
    - echo -n "${NEW_LAST_VERIFIED}" | python scripts/dss-ops.py secrets set "${VERIFY_SECRET_NAME}"
```

.gitlab-ci.yml
```diff
@@ -190,3 +192,28 @@ force_release_prod:
...
     - staging
 
+verify_sync_partial:on-schedule:
+  stage: verify_sync
+  only:
+    refs:
+      - schedules
+  before_script:
+    - VERIFY_SECRET_NAME="${DSS_SECRETS_STORE}/${DSS_DEPLOYMENT_STAGE}/${DSS_VERIFY_SECRETS_NAME}"
+    - LAST_VERIFIED="$(python scripts/dss-ops.py secrets get --secret-names ${VERIFY_SECRET_NAME} --json | jq -r '.[]')"
```
chmreid:
```suggestion
    - LAST_VERIFIED="$(python scripts/dss-ops.py secrets get ${VERIFY_SECRET_NAME})"
```

chmreid: There was a PR recently that updated/improved the secrets handling in dss-ops. It did away with the `--secret-names` flag, and values can only be piped in or provided via an input file.

chmreid (changes requested):
Some calls to `dss-ops.py secrets` (whose interface was changed recently) that need to be updated, and a few questions, but other than that this PR is lookin good

.gitlab-ci.yml
natanlao: They must be set in Gitlab - to my understanding, they can't be set in `.gitlab-ci.yml`

infra/secretsmanager/main.tf
```diff
@@ -0,0 +1,27 @@
...
+  )}"
+}
+
+resource "aws_secretsmanager_secret" "last_verified" {
+  name        = local.verify_secret_name
+  description = "Time of last replica verification"
+  tags        = local.tags
+}
+
+resource "aws_secretsmanager_secret_version" "last_verified_init" {
```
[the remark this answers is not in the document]
natanlao: The `last_verified` and `last_verified_init` names you are referencing are local names for the AWS resources in terraform. They have different names because they need to be unique within a module. To create the secret, we have a resource `last_verified`. To initialize the value of that secret, we need another resource `last_verified_init`.

infra/secretsmanager/main.tf
```diff
@@ -0,0 +1,27 @@
...
+  "Name"      , "${var.DSS_INFRA_TAG_SERVICE}-secrets",
+  "owner"     , "${var.DSS_INFRA_TAG_OWNER}",
+  "managedBy" , "terraform",
+  "project"   , "${var.DSS_INFRA_TAG_PROJECT}",
+  "env"       , "${var.DSS_DEPLOYMENT_STAGE}",
+  "service"   , "${var.DSS_INFRA_TAG_SERVICE}"
+  )}"
+}
+
+resource "aws_secretsmanager_secret" "last_verified" {
```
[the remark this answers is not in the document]
natanlao: `last_verified` is the Terraform resource name and won't appear in a deployed resource. The name of the secret is specified on line 14.

.gitlab-ci.yml
natanlao: Thanks!

.gitlab-ci.yml
natanlao: Fixed!

.gitlab-ci.yml
chmreid: I would suggest adding a one line comment that mentions that

chmreid (approved):
only requesting one clarification comment, other than that LGTM

dss/util/version.py
```diff
@@ -6,4 +6,4 @@ def datetime_to_version_format(timestamp: datetime.datetime) -> str:
     return timestamp.strftime(_datetime_format)
 
 def datetime_from_timestamp(ts: str) -> datetime.datetime:
-    return datetime.datetime.strptime(ts, _datetime_format)
+    return datetime.datetime.strptime(ts, _datetime_format).replace(tzinfo=datetime.timezone.utc)
```
amarjandu: if this function is used in a lambda, or out in another function, changing the timezone might have an undesired consequence.

dss/operations/sync.py
```diff
@@ -26,6 +29,42 @@ def _log_warning(**kwargs):
...
+    """
+    Like verify-sync, but provides keys from the designated source replica
+    (so that keys don't need to be supplied individually).
+    """
+    def list_objects_since(replica, since: datetime.datetime):
+        replica = Replica[replica]
+        handle = Config.get_blobstore_handle(replica)
+        for prefix in (BUNDLE_PREFIX, FILE_PREFIX, BLOB_PREFIX):
+            for name, metadata in handle.list_v2(replica.bucket, prefix=prefix):
+                if metadata[BlobMetadataField.LAST_MODIFIED] > since:
```
amarjandu: might want to use .get(), can LM values be unset?

dss/operations/sync.py
```diff
@@ -26,6 +29,42 @@ def _log_warning(**kwargs):
     logger.warning(json.dumps(kwargs))
 
 
+@sync.action('verify-sync-all',
+             arguments={"--source-replica": {'choices': [r.name for r in Replica], 'required': True},
+                        "--destination-replica": {'choices': [r.name for r in Replica], 'required': True},
+                        "--since": {'required': False, 'help': "Only check objects newer than this (DSS_VERSION)"}})
```
amarjandu: can you include an example of a valid timestamp the help?

amarjandu (commented):
I dont think `verify sync` is a multi-threaded process, this could cause unexpected run-times. Please see 2 comments below.

dss/operations/sync.py
natanlao: `Last-Modified` is provided by S3 for every object. I think it makes sense for this code to fail if that data is missing - without it, we can't determine if it should be checked, and I think that is a problem worth breaking for (as opposed to always skipping or always verifying).

.gitlab-ci.yml
amarjandu: I think i was wrong in thinking it could only be set in the UI, https://docs.gitlab.com/ee/user/project/pipelines/schedules.html

amarjandu: We might want to differentiate scheduled by stage, I think it would be a waste to run it on dev; but as we advance through the stages we should have more frequent checks. integration every 10 days, staging 7 days, prod 3 days.

amarjandu (approved):
Just some small stuff about scheduling, otherwise LGTM.

dss/operations/sync.py
```diff
@@ -26,6 +30,47 @@ def _log_warning(**kwargs):
...
+def verify_sync_all(argv: typing.List[str], args: argparse.Namespace):
+    """
+    Like verify-sync, but provides keys from the designated source replica
+    (so that keys don't need to be supplied individually).
+    """
+    def list_objects_since(replica, since: datetime.datetime):
+        replica = Replica[replica]
+        handle = Config.get_blobstore_handle(replica)
+        for prefix in (BUNDLE_PREFIX, FILE_PREFIX, BLOB_PREFIX):
+            for name, metadata in handle.list_v2(replica.bucket, prefix=prefix):
```
amarjandu: can you change this to `key, metadata` its what the other scripts use, easier for an operator to keep it in context.

amarjandu (approved):
1 thing

amarjandu:
@natanlao should have merged it 😭

natanlao:
@amarjandu 😔

PR #2440, opened 2019-09-19
Fix exception being instantiated but not raised

natanlao:


natanlao:
[Passing on Allspark](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/pipelines/16313).

PR #2441, opened 2019-09-20
Error catching improvements from @natanlao

chmreid:
Implements better handling of AWS exceptions


chmreid:
This doesn't actually work - 

```
except sm_client.exceptions.ResourceNotFoundException as e:
TypeError: catching classes that do not inherit from BaseException is not allowed
```

:-/

PR #2449, opened 2019-09-20
Fix signature in daemons/dss-notify-v2/app.py

xbrianh:
Also add a test the code block that contained the incorrect method signature.


DailyDreaming:
LGTM, but I think Trent's changes might need to be removed.

PR #2463, opened 2019-09-25
[easy] Add GCP function error handling.

DailyDreaming:
Google cloud functions were failing for a few hours due to google cloud function flakiness.  This error handling should hopefully save time in the future trying to figure out why google cloud functions are breaking for no good reason.


DailyDreaming:
@chmreid Haha, I like that rule.  I think that PR will be coming soon.

Issue #2460, opened 2019-09-24
Set up and tear down users for tests, without use of setup_fusillade script


chmreid:
## User story

* As a dcp operator,
* I want to set up and tear down Fusillade test users (using dcplib.security)
* so that we can protect endpoints with Fusillade, and test them with the proper authorization headers (or with a mocked fusillade server response)

## Definition of done

The indexer test implements code to create a test user during the test setup, perform the test actions using the test user, and delete the test user during test cleanup.

chmreid:
Requirements for our Fusillade test strategy in the DSS:

- we don't want to create/destroy temporary Fusillade users for every test (this gets complicated/messy and is difficult to implement cleanly)
- we don't want to require the client (who runs the tests) to connect to a live, working Fusillade deployment - tests should be able to pass, independent of Fusillade's availability
- we want to mock the functionality of the Fusillade authorization endpoint, and retain the ability to mock _both_ access granted _and_ access denied responses from Fusillade
- we don't want to make all of our tests messy by requiring every test that touches a Fusillade-protected endpoint to mock the Fusillade functionality

chmreid:
Current Fusillade test strategy:

- in `tests/infra/servers.py` we set up a simple HTTP server that will provide minimal functionality to mock Fusillade
- The simple HTTP server will run on `127.0.0.1:54321` as part of the test setup method
- The Config (static) class in `dss/config.py` will be modified so that the AuthZ endpoint returned by the Config class will be `127.0.0.1:54321` during testing and the real AuthZ endpoint otherwise
- The security module in `dss/util/security.py` actually checks whether the requested action can be performed by the given user; it will hit the `/v1/policies/evaluate` Fusillade endpoint using the endpoint that Config returns, so this should go to the simple HTTP server at `127.0.0.1:54321` during testing
- To control responses from the simple HTTP server, we add a user whitelist and a user blacklist to the class and minimal logic to return the correct response (access granted if user is on whitelist, access denied if user is on blacklist)
- To test the functionality of endpoint access control, we change the user we use to perform the given action (as implemented, `valid@ucsc.edu` is on the whitelist and `invalid@ucsc.edu` is on the blacklist)

PR #2467, opened 2019-09-25
Incorporate mock Fusillade server into tests

chmreid:
This PR adds the ability to test endpoints protected by Fusillade, without using a deployed Fusillade server.

See #2460 (issue) for details about the strategy being used. In short, we're creating a mock Fusillade server to return a response granting or denying access. 

This PR is awesome. Here's why:
- requires _zero_ changes to deployed Fusillade servers
- requires _zero_ calls to external servers/services
- allows the ability to tailor the Fusillade response (approve or deny) during a test, by using different principals (users) during the authorization check, and implementing a whitelist for the mock Fusillade server
- does not require any mocking of the authorization check, it _just works_ (by transparently swapping out the authorization endpoint for tests)

This PR will close issue #2460.

### Test Plan

This PR adds testing infrastructure (a mock Fusillade server) in `tests/infra/server.py`, and it also adds a test for that mock Fusillade server (to ensure it is working as expected).

### Deployment Instructions & Migrations

This should not change any deployment procedures.  The mock Fusillade server only runs during testing and binds to a local port via the link-local IP address, so it will not require any ports to be opened.

### Release notes

Adding ability to run tests on Fusillade-protected endpoints


dss/config.py
```diff
@@ -416,9 +417,14 @@ def get_OIDC_email_claim():
 
     @staticmethod
     def get_authz_url():
-        if Config._AUTH_URL is None:
-            Config._AUTH_URL = Config._get_required_envvar("AUTH_URL")
-        return Config._AUTH_URL
+        if Config._CURRENT_CONFIG == BucketConfig.NORMAL:
```
DailyDreaming: Should probably be:
```
        if (Config._CURRENT_CONFIG == BucketConfig.TEST or Config._CURRENT_CONFIG == BucketConfig.TEST_FIXTURE):
            host = "127.0.0.1"
            port = "54321"
            return f"http://{host}:{port}"
        else:
            if Config._AUTH_URL is None:
                Config._AUTH_URL = Config._get_required_envvar("AUTH_URL")
            return Config._AUTH_URL
```

xbrianh: Another option:
```
if Config._CURRENT_CONFIG in [BucketConfig.TEST, BucketConfig.TEST_FIXTURE]:
    ...
```

chmreid:
All tests except the parallel test are passing on Travis. This PR breaks on the indexer tests because the mock Fusillade server (at 127.0.0.1:54321) is not responding during the tests (so bundle put operations fail). What's weird is, the mock Fusillade server works fine locally and in the other (non-parallel) tests.

Furthermore, local tests work on my machine, including parallel tests. I can run both of the following tests without problems locally:

```
make tests/test_indexer.py
make -j2 tests/test_indexer.py
```

I also added tests for the mock Fusillade server's functionality. These tests are also passing locally and in non-parallel tests, and breaking in parallel tests. Here is the traceback from the [allspark pipeline](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/48340):

```
coverage run -p --source=dss tests/test_server.py -v
INFO:dss:{
    "dispatch_info": {
        "content-length": "40",
        "content-type": "application/json",
        "method": "PUT",
        "path": "/v1/files/9bd4b327-581c-4823-bf6b-cef327a9543f",
        "query_params": "MultiDict({'version': ['2019-09-27T182154.967543Z']})",
        "started_at": "",
        "status_code": 201
    },
    "log-msg-type": "info",
    "system": "data-storage-service"
}
INFO:dss:{
    "log-msg-type": "info",
    "request_info": {
        "content_length": "281",
        "method": "PUT",
        "path": "/v1/bundles/baecb635-4fa1-46cd-9714-48a42a176431",
        "query_params": " MultiDict({'replica': ['aws'], 'version': ['2019-09-27T182155.156304Z']})",
        "source_ip": "127.0.0.1",
        "user_agent": "python-requests/2.22.0"
    },
    "system": "data-storage-service"
}
INFO:dss.util.security:{"valid": true, "token": {"iss": "travis-test@human-cell-atlas-travis-test.iam.gserviceaccount.com", "sub": "travis-test@human-cell-atlas-travis-test.iam.gserviceaccount.com", "aud": ["https://dev.data.humancellatlas.org/", "https://dev.data.humancellatlas.org/"], "iat": 1569608515.1567125, "exp": 1569612115.1567125, "scope": ["email", "openid", "offline_access"], "https://auth.data.humancellatlas.org/group": "hca", "email": "travis-test@human-cell-atlas-travis-test.iam.gserviceaccount.com"}}
ERROR:dss.error:Traceback (most recent call last):
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/connection.py", line 157, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/util/connection.py", line 84, in create_connection
    raise err
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/util/connection.py", line 74, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 672, in urlopen
    chunked=chunked,
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 387, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python3.6/http/client.py", line 1254, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1300, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1249, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.6/http/client.py", line 1036, in _send_output
    self.send(msg)
  File "/usr/lib/python3.6/http/client.py", line 974, in send
    self.connect()
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/connection.py", line 184, in connect
    conn = self._new_conn()
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/connection.py", line 169, in _new_conn
    self, "Failed to establish a new connection: %s" % e
urllib3.exceptions.NewConnectionError: <urllib3.connection.HTTPConnection object at 0x7f79c9758940>: Failed to establish a new connection: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

  File "/home/hca_cicd/venv/lib/python3.6/site-packages/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 720, in urlopen
    method, url, error=e, _pool=self, _stacktrace=sys.exc_info()[2]
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/urllib3/util/retry.py", line 436, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='127.0.0.1', port=54321): Max retries exceeded with url: /v1/policies/evaluate (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f79c9758940>: Failed to establish a new connection: [Errno 111] Connection refused',))

During handling of the above exception, another exception occurred:

  File "/builds/HumanCellAtlas/data-store/dss/error.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/builds/HumanCellAtlas/data-store/dss/api/bundles/__init__.py", line 194, in put
    [f'arn:hca:dss:{Config.deployment_stage()}:*:bundle/{uuid}/{version}'])
  File "/builds/HumanCellAtlas/data-store/dss/util/security.py", line 127, in assert_authorized
    "principal": principal})
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/requests/sessions.py", line 581, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "/home/hca_cicd/venv/lib/python3.6/site-packages/requests/adapters.py", line 516, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=54321): Max retries exceeded with url: /v1/policies/evaluate (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f79c9758940>: Failed to establish a new connection: [Errno 111] Connection refused',))
```

I can't think of a reason that the mock Fusillade server would fail this way...

chmreid:
This ended up unfolding into a pretty complicated tangle of problems.

- the call to start/stop the mock fusillade server was only added to one of the test classes in `tests/test_indexer.py`, so the fusillade calls from other test classes would have no mock fusillade server to reply to their requests to check authorization
- `dss.util.security` sends requests to the Fusillade API to make sure principal/resource/action are allowed for a given endpoint, but it also checks to make sure the group and OpenID provider domain match what's specified by environment variables. These environment variables had to be set such that tests can pass.
- the mock fusillade server was not being started/stopped for every test that calls the PUT endpoint (currently the only endpoint protected by fusillade), so some of the tests were making calls to the mock fusillade server after it had been torn down (for other tests)

chmreid:
More problems uncovered with this approach:
* In Unix, there is a timeout/waiting period before an address/socket is truly "released" by a service. A setting in the TCPServer class (base class of http.server) called `HTTPServer.allow_reuse_address` has to be set on the class (not the instance).
* The multi-threaded test is failing because each thread runs a separate test, and if two threads both run a test that requires setting up and tearing down a mock Fusillade server, they will both be trying to use the same port.
* The mock Fusillade servers are getting the address/port to serve on from the Config class, which in turn must hard-code the address and port. If we try to get Config to find and return an unused port, it will return a different endpoint (`127.0.0.1:<random port number>`) each time, and requests to have Fusillade do an auth check will all go to different ports.

Basically, we need one single mock Fusillade server to be started and stopped at the beginning/end of _all_ tests, since _all_ tests use the same Fusillade endpoint. We cannot start/stop mock Fusillade servers for each individual test or test module.

Am working through how to implement this now.

chmreid:
Ended up having to start the mock fusillade server as a standalone process using a wrapper script. That way, every test that needs it has it available, without having to start/stop multiple mock fusillade servers. Duplicate mock fusillade servers are prevented by putting the PID of the running server in a PID file. Cleanup happens when the script stops, which sends a sigterm to the mock fusillade server, which cleans up the PID file at the end of all the tests.

chmreid:
Rebasing onto master to resolve conflicts

dss/config.py
chmreid: good suggestions all. went with the `if X in [list]` syntax

chmreid:
The approach mentioned in [this comment](https://github.com/HumanCellAtlas/data-store/pull/2467#issuecomment-536768518) didn't end up working out due to problems with the tests getting stuck and timing out. 

The approach we went with instead was to modify the Config class at `dss/config.py` so that an authorization URL could be set by each test, which could find and use open ports dynamically, rather than having to hard-code a single address and port. This made it possible for each test to create and tear down its own mock Fusillade server.

tests/infra/server.py
```diff
@@ -79,3 +93,98 @@ def shutdown(self):
...
+
+class MockFusilladeHandler(BaseHTTPRequestHandler):
+    """
+    Create a mock Fusillade auth server endpoint so that any operation that tries to check
+    permissions with Fusillade will be handled correctly; we keep it simple and accept/reject
+    based on whether the principal (user) is on the whitelist or not.
+    """
+    _server = None
+    _thread = None
+    _whitelist = [
```
DailyDreaming: If these service accounts are needed to validate tests, then this list should probably be populate from the ssm environment whitelist (this will also take of adding the valid service accounts for `staging` and `prod` which are missing here, as well as anyone on the team running locally).

dss/config.py
```diff
@@ -414,6 +414,10 @@ def get_OIDC_group_claim():
     def get_OIDC_email_claim():
         return os.environ.get("OIDC_EMAIL_CLAIM", 'email')
 
+    @staticmethod
+    def _set_authz_url(mock_authz_url):
```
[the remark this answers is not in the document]
chmreid: fixed!

tests/infra/server.py
chmreid: Good idea - using `get_admin_emails()` from `dss/operations/lambda_params.py`

tests/infra/server.py
```diff
@@ -79,3 +95,94 @@ def shutdown(self):
...
+    """
+    Create a mock Fusillade auth server endpoint so that any operation that tries to check
+    permissions with Fusillade will be handled correctly; we keep it simple and accept/reject
+    based on whether the principal (user) is on the whitelist or not.
+    """
+    _server = None
+    _thread = None
+    _whitelist = get_admin_emails().split(",") + ["valid@ucsc.edu"]
+
+    @classmethod
```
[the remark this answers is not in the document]
chmreid: Not sure what you mean... this currently is a static method

chmreid:
@Bento007 good catch, added the two calls to restore variables back into the `stop_serving()` method

tests/infra/server.py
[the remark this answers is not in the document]
Bento007: it's a classmethod

PR #2472, opened 2019-09-30
Allow multiple flashflood writes

xbrianh:
This allows multiple flashflood write prefixes to facilitate the following "reindex" procedure, which provides uninterrupted event replay during reindex:
1. A new flashflood index is created
1. Past events are recorded into the new index
1. In parallel, Incoming events are recorded into both the live index and the new index
1. When the new index is complete, the new and live indexes are swapped.


tests/test_events.py
```diff
@@ -54,15 +80,18 @@ def setUp(self):
...
-            os.environ[f'DSS_{replica.name.upper()}_FLASHFLOOD_PREFIX'] = f"flashflood-{uuid4()}"
+            pfx = f"flashflood-{replica.name}-{uuid4()}"
+            os.environ[f'DSS_{replica.name.upper()}_FLASHFLOOD_PREFIX_READ'] = pfx
+            os.environ[f'DSS_{replica.name.upper()}_FLASHFLOOD_PREFIX_WRITE'] = pfx
             events.record_event_for_bundle(replica, self.bundle[replica.name]['key'])
 
     def test_list(self):
-        for replica in Replica:
-            self._test_list(replica)
+        # for replica in Replica:
```
DailyDreaming: Leftovers?

dss/config.py
```diff
@@ -541,8 +543,12 @@ def checkout_bucket(self) -> str:
...
 
     @property
-    def flashflood_prefix(self) -> str:
-        return os.environ[self._flashflood_prefix_envvar]
+    def flashflood_prefix_read(self) -> str:
+        return os.environ[self._flashflood_prefix_read_envvar]
+
+    @property
+    def flashflood_prefix_write(self) -> typing.Tuple[str, ...]:
+        return tuple(os.environ[self._flashflood_prefix_write_envvar].split(","))
```
DailyDreaming: I realize that you are using them differently, but it's unexplained here why `Replica.flashflood_prefix_write()` would return a `tuple` while `Replica.flashflood_prefix_read()` returns a `str`.  I would consider making them consistent.  If not, a short docstring here explaining it would aid someone like me in the future.

dss/config.py
```diff
@@ -511,14 +511,16 @@ def __init__(
             storage_schema: str,
             blobstore_class: typing.Type[BlobStore],
             hcablobstore_class: typing.Type[HCABlobStore],
-            flashflood_prefix_envvar: str,
+            flashflood_prefix_read_envvar: str,
```
DailyDreaming: Mind adding in some comments about the purpose of separating the read/write prefixes?  I'm not entirely clear.

dss/events/__init__.py
```diff
@@ -14,22 +15,15 @@
...
 
 @lru_cache(maxsize=2)
 def get_bundle_metadata_document(replica: Replica,
                                  key: str,
                                  flashflood_prefix: str=None) -> dict:
     fqid = key.split("/", 1)[1]
-    ff = FlashFlood(resources.s3,  # type: ignore
-                    Config.get_flashflood_bucket(),
-                    flashflood_prefix or replica.flashflood_prefix)
+    pfx = flashflood_prefix or replica.flashflood_prefix_read
```
DailyDreaming: Why would a user wish to set this?  (I'm probably just arguing for a docstring explaining the argument here)

environment
```diff
@@ -86,8 +88,10 @@ DSS_GS_CHECKOUT_BUCKET_TEST_USER=org-hca-dss-checkout-test-user  # for tests w/n
 DSS_GS_CHECKOUT_BUCKET_INTEGRATION=org-humancellatlas-dss-checkout-integration
 DSS_GS_CHECKOUT_BUCKET_STAGING=org-humancellatlas-dss-checkout-staging
 DSS_GS_CHECKOUT_BUCKET_PROD=org-humancellatlas-dss-checkout-prod
-DSS_AWS_FLASHFLOOD_PREFIX="aws"
-DSS_GCP_FLASHFLOOD_PREFIX="gcp"
+DSS_AWS_FLASHFLOOD_PREFIX_READ="aws"
+DSS_AWS_FLASHFLOOD_PREFIX_WRITE="aws"  # This can be a comma separated list
+DSS_GCP_FLASHFLOOD_PREFIX_READ="gcp"
+DSS_GCP_FLASHFLOOD_PREFIX_WRITE="gcp"  # This can be a comma separated list
```
DailyDreaming: https://github.com/HumanCellAtlas/data-store/tree/master/docs/environment should probably be updated.

tests/test_events.py
xbrianh: fixed :/

dss/config.py
xbrianh: docstrings added

dss/config.py
xbrianh: Not sure this is the appropriate place. Maybe this documentation should live in the script that performs a flashflood reindex (future PR).

dss/events/__init__.py
xbrianh: Operators may wish to choose the prefix during administrative actions.

environment
xbrianh: updated

dss/events/__init__.py
```diff
@@ -52,23 +46,25 @@ def get_deleted_bundle_metadata_document(replica: Replica, key: str) -> dict:
...
     if use_version_for_timestamp:
         _, version = fqid.split(".", 1)
         event_date = datetime_from_timestamp(version)
     else:
         event_date = datetime.utcnow()
-    ff.put(json.dumps(metadata_document).encode("utf-8"), event_id=fqid, date=event_date)
+    for pfx in flashflood_prefixes:
+        ff = FlashFlood(resources.s3, Config.get_flashflood_bucket(), pfx)
+        if not ff.event_exists(fqid):
+            ff.put(json.dumps(metadata_document).encode("utf-8"), event_id=fqid, date=event_date)
```
amarjandu: Should there be a catch to ensure all prefixes are attempted to be written to? 
if one of the first prefixes in the list fails, i assume we want the later ones to still be written to?

xbrianh: Merged before I saw this comment. But I don't know what the correct strategy is for that case.

PR #2475, opened 2019-09-30
Update infrastructure to use dss-ops.py and not set_secret/fetch_secret (redux)

chmreid:
This is a new version of #2370, which was merged and then reverted in #2474.

The old PR (#2370) had an extra `--json` flag in `.gitlab-ci.yml` and `.gitlab-ci-prod.yml` that prevented Google Cloud service account credentials from being correctly exported. This PR fixes that problem (and does a rebase onto the latest master branch).


README.md
```diff
@@ -139,7 +139,9 @@ The DSS uses the [Amazon S3 backend](https://www.terraform.io/docs/backends/type
 	1) Run the command
 
 	   ```
-	   cat $DSS_HOME/application_secrets.json | scripts/set_secret.py --secret-name $GOOGLE_APPLICATION_SECRETS_SECRETS_NAME
+       ### WARNING: RUNNING THIS COMMAND WILL
```
amarjandu: 👍

README.md
```diff
@@ -242,15 +244,17 @@ When deploying for the first time, a Google Cloud Platform service account must
 1.  Run the command
 
     ```
-    cat $DSS_HOME/gcp-credentials.json | scripts/set_secret.py --secret-name $GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME
+    cat $DSS_HOME/gcp-credentials.json | ./scripts/dss-ops.py secrets set --secret-name $GOOGLE_APPLICATION_CREDENTIALS_SECRETS_NAME
```
amarjandu: actual question here: should we assume that the user is currently in the `$DSS_HOME` dir, so using `./scripts/` would resolve? 
It's something that i also do; but others sometimes say to omit.

chmreid:
@amarjandu Re `./scripts`, we could replace with `$DSS_HOME/scripts`, but the reason I did not is, there are several other commands (e.g., all the `make` commands) that expect the user to be in `$DSS_HOME` first. I think we should mention in the readme that all commands are run from `$DSS_HOME`, which is the root of the `data-store` repo, then we can make the commands less verbose.

Re removing `fetch_secrets.py` and `get_secrets.py` - yes, the original plan was to replace those scripts with the `dss-ops.py` script. If you have concerns about that we can discuss with the team before merging this PR.

chmreid:
#2478 has a Readme clarification

amarjandu:
> @amarjandu Re `./scripts`, we could replace with `$DSS_HOME/scripts`, but the reason I did not is, there are several other commands (e.g., all the `make` commands) that expect the user to be in `$DSS_HOME` first. I think we should mention in the readme that all commands are run from `$DSS_HOME`, which is the root of the `data-store` repo, then we can make the commands less verbose.
> 
> Re removing `fetch_secrets.py` and `get_secrets.py` - yes, the original plan was to replace those scripts with the `dss-ops.py` script. If you have concerns about that we can discuss with the team before merging this PR.

No concerns, its just i dont see those old files marked for deletion, should they not be deleted?

chmreid:
Good catch, they were supposed to be deleted :)

PR #2476, opened 2019-10-01
[easy] Print response when error handling for security checks.

DailyDreaming:


dss/util/security.py
```diff
@@ -125,5 +125,6 @@ def assert_authorized(principal: str,
                         json={"action": actions,
                               "resource": resources,
                               "principal": principal})
-    if not resp.json()['result']:
-        raise DSSForbiddenException()
+    resp = resp.json()
```
xbrianh: What information is contained in `resp`?

DailyDreaming:
```
{'evaluation_results': [{'EvalActionName': 'dss:PutBundle', 'EvalDecision': 'implicitDeny', 'EvalResourceName': 'arn:hca:dss:integration:*:bundle/e527501b-12e0-4e34-bc14-a23b81768f7e/2019-10-01T164454.142322Z', 'MatchedStatements': [], 'MissingContextValues': ['fus:group_id']}], 'reason': 'Permission was implicitly denied.', 'result': False}
```

xbrianh: That looks confusing to me. Should we report a subset, or is it better to return all of it?

DailyDreaming: I think that until all of the fusillade bugs are worked out, it might save us time to print all of it for now and maybe think about reducing it in the future if it's too verbose.  Trent said that it should be safe to print.

Bento007: We should also raise if not 200. This would tell use something is wrong with fusillade backend

PR #2477, opened 2019-10-01
DSS Notifyv2: add logging [EZ]

amarjandu:
adds logging for s3 bundle events that don't have warnings.


daemons/dss-notify-v2/app.py
```diff
@@ -50,6 +50,7 @@ def launch_from_s3_event(event, context):
             if key.startswith("bundles"):
                 is_delete_event = (event_record['eventName'] == "ObjectRemoved:Delete")
                 _notify_subscribers(replica, key, is_delete_event)
+                logger.info(f"Attempting notifications for; replica: {replica}, key: {key} delete: {is_delete_event}")
```
xbrianh: Why not log this from `_notify_subscribers`? That would cover each replica.

amarjandu: Changed, thx :)

PR #2478, opened 2019-10-01
add readme note on where commands should be run from

chmreid:
Addressing @amarjandu code review comment for #2475, clarifying where commands should be run from


chmreid:
man you're fast 🐎

PR #2485, opened 2019-10-02
Update tombstone_bundles.py [EZ]

amarjandu:
data-ops team requested to be tagged on tombstone scripts, updating script as a reminder.

### DIFF
--- a/scripts/tombstone_bundles.py
+++ b/scripts/tombstone_bundles.py
@@ -2,6 +2,7 @@
 """
 This script will tombstone the bundles in `bundles_to_tombstone`
 Usage:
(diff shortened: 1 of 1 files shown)


scripts/tombstone_bundles.py
```diff
@@ -2,6 +2,7 @@
 """
 This script will tombstone the bundles in `bundles_to_tombstone`
 Usage:
+    - tag HCA/data-ops team in the issue comments with @HumanCellAtlas/data-ops
```
xbrianh: Is this on Slack or  GitHub?

PR #2488, opened 2019-10-03
Back up collections, subscriptions (closes #1661)

### LINKED ISSUE #1661: Store subscriptions as an object in replica
As an operator I would like to be able to rebuild the subscription database in the event elastic search fails.

# Definition of Done
* Subscriptions objects are stored as object in the data store replica when subscription/put is used.
* Subscription object are deleted when subscriptions/{uuid}/delete api is used
* Subscriptions objects can be used to rebuild the elasticsearch subscriptions index.

natanlao:
Closes #1661.

Backs up the collection and subscription DynamoDB tables by enabling point-in-time recovery (PITR).
The original ticket specified backing up v1 subscriptions, which are backed by ElasticSearch, which is being deprecated. Because DynamoDB is severless, the need to back up data in the event of a component failure is reduced.

Because PITR restores take are not fast (can take hours), the main use case of these backups seems to be to protect against unintentional wide-scale deletion of data.

### DIFF
--- a/infra/collections_db/main.tf
+++ b/infra/collections_db/main.tf
@@ -17,6 +17,10 @@ resource "aws_dynamodb_table" "collections-db-aws" {
   hash_key     = "hash_key"
   range_key    = "sort_key"
 
+  point_in_time_recovery {
+    enabled = true
+  }
+
   attribute {
     name = "hash_key"
     type = "S"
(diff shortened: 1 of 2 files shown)


amarjandu:
tagged @chmreid to put this on his radar too.

natanlao:
Deployed to dev and passing on Allspark.

PR #2495, opened 2019-10-07
DSS Bundle Enumeration: Test Refactor

amarjandu:
move complex enumeration tests out of `tests/test_smoketest.py`, introduces more testing into `tests/test_bundles.py`
marks file for `@testmode.standalone`
addresses change to header in test

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -988,8 +988,17 @@ class TestBundleApi(unittest.TestCase, TestAuthMixin, DSSAssertMixin, DSSUploadM
     def test_enumeration_bundles(self):
         bundle_uuid, bundle_version = self._put_bundle()
         res = self.app.get(f"/v1/bundles/all",
-                           params=dict(version=bundle_version, replica="aws", per_page=500))
+                           params=dict(version=bundle_version, replica="aws", per_page=10))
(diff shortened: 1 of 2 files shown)


amarjandu:
@xbrianh ive marked the file for `@testmode.standalone` i also added in a fix for changes to the header modification in #2497

tests/test_bundle.py
```diff
@@ -172,7 +172,7 @@ def _test_bundle_get_paging(self,
                 # Make sure we're getting the expected response status code
                 if link_header:
                     self.assertEqual(resp_obj.response.headers['X-OpenAPI-Pagination'], 'true')
-                    self.assertEqual(resp_obj.response.headers['X-OpenAPI-Paginated-Content-Key'], 'files')
+                    self.assertEqual(resp_obj.response.headers['X-OpenAPI-Paginated-Content-Key'], 'bundle.files')
```
natanlao: This LGTM, though in every case the `X-OpenAPI-Paginated-Content-Key` should be present and equal to `bundle.files`, so it could make sense to move L175 outside of the `if link_header` block.

amarjandu: addressed

PR #2496, opened 2019-10-07
Add simple bundle enumeration test

xbrianh:
This test should be expanded to catch bundle enumeration errors before deployment.

This also changes bundle API tests back to unit tests. These tests are expected to run before deployment.


xbrianh:
superseded by #2495

PR #2497, opened 2019-10-07
Send X-OpenAPI-Paginated-Content-Key even when only one page is present

kislyuk:
X-OpenAPI-Paginated-Content-Key needs to be sent with every response
for methods that support pagination, even if there is only one
page. The client needs the key name metadata in
X-OpenAPI-Paginated-Content-Key to stream results without hardcoding
the key name.

### DIFF
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -135,7 +135,7 @@ def get(
         response.headers['X-OpenAPI-Pagination'] = 'true'
         response.headers['Link'] = link
 
-    response.headers['X-OpenAPI-Paginated-Content-Key'] = 'bundles.files'
+    response.headers['X-OpenAPI-Paginated-Content-Key'] = 'bundle.files'
     return response
 
 @dss_handler
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -130,13 +130,13 @@ def get(
     if link is None:
         response = make_response(jsonify(response_body), requests.codes.ok)
         response.headers['X-OpenAPI-Pagination'] = 'false'
-        return response
     else:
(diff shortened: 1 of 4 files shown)


dss/api/bundles/__init__.py
```diff
@@ -130,13 +130,13 @@ def get(
     if link is None:
```
amarjandu: on line 188 please remove the indent to allow `response.headers['X-OpenAPI-Paginated-Content-Key'] = 'bundles'` to be applied to all responses

kislyuk:
Thanks for catching that!

dss/api/bundles/__init__.py
kislyuk: Thanks for catching that!

dss/api/bundles/__init__.py
```diff
@@ -130,13 +130,13 @@ def get(
...
         response.headers['X-OpenAPI-Pagination'] = 'false'
-        return response
     else:
         response = make_response(jsonify(response_body), requests.codes.partial)
         response.headers['X-OpenAPI-Pagination'] = 'true'
-        response.headers['X-OpenAPI-Paginated-Content-Key'] = 'files'
         response.headers['Link'] = link
-        return response
+
+    response.headers['X-OpenAPI-Paginated-Content-Key'] = 'bundle.files'
```
natanlao: This change could might cause a test failure here:

https://github.com/HumanCellAtlas/data-store/blob/8be83e45114ffbd090966463650c83639d8bde9f/tests/test_bundle.py#L175

If it does, the failure would only be picked up after merge, because those tests are mistakenly marked as integration tests (will be fixed in #2495)

amarjandu: I've added the required change to #2495

PR #2498, opened 2019-10-07
DSS Bundle Enumeration: bundle enumeration page breaks

amarjandu:
refactor bundle enumeration to handle tombstones more effectively, removes possibility for tombstone to occur across a page break.
Refactor Tests to ensure more thorough testing of tombstone cases.


tests/test_utils.py
```diff
@@ -248,38 +248,57 @@ def test_multipart_parallel_upload(self):
...
             self.assertNotIn('.'.join([x['uuid'], x['version']]), dead_bundles)
             self.assertNotIn(x['uuid'], dead_bundles)
-        self.assertEquals(resp['page_count'], 10)
+
+    @mock.patch("dss.Config.get_blobstore_handle")
+    def test_tombstone_pages(self, mock_list_v2):
+        mock_list_v2.return_value = self.MockStorageHandler()
+        resp = enumerate_available_bundles(replica='aws', per_page=5)  # pagesize tests for split page tombstones
+        dead_bundles = [x.split('bundles/')[1].rsplit('.', 1)[0] for x in self.MockStorageHandler.tombstoned_bundle_list
+                        if x.endswith(TOMBSTONE_SUFFIX)]
```
xbrianh: This misses the unversioned tombstones. For instance, the following keys are all tombstones even though the versioned fqids don't end with `.dead`:
```
bundles/0007edde-f22c-4858-bf17-513dc2d05863.2019-02-26T033907.789762Z
bundles/0007edde-f22c-4858-bf17-513dc2d05863.2019-05-23T220340.829000Z
bundles/0007edde-f22c-4858-bf17-513dc2d05863.dead
```

xbrianh: Consider using a hard coded list of tombstoned bundles.

tests/test_utils.py
```diff
@@ -248,38 +248,57 @@ def test_multipart_parallel_upload(self):
...
         for x in resp['bundles']:
             self.assertNotIn('.'.join([x['uuid'], x['version']]), dead_bundles)
             self.assertNotIn(x['uuid'], dead_bundles)
-        self.assertEquals(resp['page_count'], 10)
+
+    @mock.patch("dss.Config.get_blobstore_handle")
+    def test_tombstone_pages(self, mock_list_v2):
+        mock_list_v2.return_value = self.MockStorageHandler()
+        resp = enumerate_available_bundles(replica='aws', per_page=5)  # pagesize tests for split page tombstones
+        dead_bundles = [x.split('bundles/')[1].rsplit('.', 1)[0] for x in self.MockStorageHandler.tombstoned_bundle_list
```
xbrianh: You can split on `/` instead of `bundles/`

tests/test_utils.py
```diff
@@ -248,38 +248,57 @@ def test_multipart_parallel_upload(self):
...
                         if x.endswith(TOMBSTONE_SUFFIX)]
         for x in resp['bundles']:
             self.assertNotIn('.'.join([x['uuid'], x['version']]), dead_bundles)
             self.assertNotIn(x['uuid'], dead_bundles)
-        self.assertEquals(resp['page_count'], 10)
+
+    @mock.patch("dss.Config.get_blobstore_handle")
+    def test_tombstone_pages(self, mock_list_v2):
+        mock_list_v2.return_value = self.MockStorageHandler()
+        resp = enumerate_available_bundles(replica='aws', per_page=5)  # pagesize tests for split page tombstones
```
xbrianh: Currently we're testing page sizes of 5 and 2.

Consider adding a test with 4 and 2 to test the case where paging begins with a tombstone.

tests/test_utils.py
amarjandu: Yeah refactoring these all these functions.

tests/test_utils.py
amarjandu: Refactoring around this.

tests/test_utils.py
amarjandu: removed this usage, moved to explicit dead_bundle list

tests/test_utils.py
```diff
@@ -248,38 +248,65 @@ def test_multipart_parallel_upload(self):
...
+                                               "version": "2018-10-09T002318.558884Z"}},
+                         {"size": 4,
+                          "last_good_bundle": {"uuid": "00171d5a-f14f-48b3-a467-0b0ac8208c6b",
+                                               "version": "2019-02-26T033414.851898Z"}}]
+
+        mock_list_v2.return_value = self.MockStorageHandler()
+        for tests in test_per_page:
+            test_size = tests['size']
+            last_good_bundle = tests['last_good_bundle']
+            resp = enumerate_available_bundles(replica='aws', per_page=test_size)
```
xbrianh: It looks like the continuation tests have been removed, since `enumerate_available_bundles` is never called with `search_after` or `token`. I was thinking we could try something like the following strategy: 
1. list 5, and then use paging info to list 2 more
1. list 2, and then use paging info to list 2 more
1. list 4, and then use paging info to list 2 more

amarjandu: implemented.

tests/test_utils.py
```diff
@@ -248,38 +248,76 @@ def test_multipart_parallel_upload(self):
...
+            page_one = resp['bundles']
+            for x in resp['bundles']:
+                self.assertNotIn('.'.join([x['uuid'], x['version']]), self.MockStorageHandler.dead_bundles)
+            self.assertDictEqual(last_good_bundle, resp['bundles'][-1])
+            search_after = resp['search_after']
+            resp = enumerate_available_bundles(replica='aws', per_page=2,
+                                               search_after=search_after)
```
xbrianh: `token` doesn't need to be passed in?

amarjandu: refactored tests to support this

amarjandu: i had to change the page strategy to :

    list 5, and then use paging info to list 5 more
    list 2, and then use paging info to list 2 more
    list 4, and then use paging info to list 4 more

i changed the bundle list to support this, it was due to an enumeration error that would occur in the cloud_blobstore.

dss-api.yml
```diff
@@ -1496,7 +1496,7 @@ paths:
           required: false
           type: integer
           format: int32
-          minimum: 10
+          minimum: 2
```
xbrianh: I think the minimum should stay 10 (or higher) to prevent our service from being overwhelmed by many tiny requests.

These values should not be changed to accommodate our testing strategies.

tests/test_bundle.py
```diff
@@ -1009,6 +1012,34 @@ def test_enumeration_bundles(self):
...
 
+    def test_bundle_enumeration_paging(self):
+        #  Note this test requires a particular /bundle order in s3 test bucket, see test_utils for mocked version
+        with override_bucket_config(BucketConfig.TEST):
+            bucket = Config.get_s3_bucket()
+            key_list = MockStorageHandler.bundle_list
+            for key in key_list:
+                tf = tempfile.TemporaryFile()
+                tf.write(os.urandom(8))
+                s3.upload_fileobj(Fileobj=tf, Bucket=bucket, Key=key)
```
xbrianh: The temp file can be avoided, and you can upload a file with 0 size:
```
s3.upload_fileobj(Fileobj=io.BytesIO(b""), Bucket=bucket, Key=key)
```

amarjandu: sneaky

dss-api.yml
amarjandu: done

tests/test_bundle.py
```diff
@@ -4,6 +4,7 @@
 import datetime
 import hashlib
 import io
+import tempfile
```
xbrianh: This can be removed.

amarjandu: complete

PR #2482, opened 2019-10-01
DSS NotifyV2: Subsciprition URL Stats

amarjandu:
inserts statistics about DSS -> callback url interaction on a per subscription basis
Connected To: #2255

PR #2500, opened 2019-10-08
Fix typing for enums

chmreid:
Fixes mypy typing errors in #2482 due to introducing enums

cc @amarjandu

### DIFF
--- a/dss/events/handlers/notify_v2.py
+++ b/dss/events/handlers/notify_v2.py
@@ -1,5 +1,6 @@
 import os
 import re
+import typing
 import json
 import requests
 import urllib3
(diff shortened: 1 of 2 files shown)


chmreid:
Natan and I had a discussion about this PR, adding a note here for the record.

Ran into an issue, when I converted the "weak" enum classes to proper `Enum` classes, this led to a typing error where we use an enum value to retrieve a key from a dictionary (here: https://github.com/HumanCellAtlas/data-store/blob/amar-subscrip-info/dss/events/handlers/notify_v2.py#L113). The dictionary expected string keys but got an enum type key.

One solution was to specify the type of the dictionary keys as a union of strings and the enum class, `typing.Dict[typing.Union[str, MyEnumClass]]`.

- pros: allows us to use enums in subscriptions v2 without having to change much about the notification handler v2.
- cons: if you're using enums, you should be consistent. (i.e., mypy is "right" to raise the error.) if the dictionary doesn't have enum keys, you shouldn't retrieve keys using an enum.

A different solution is not to use enums.

- pros: no type error issues, no inconsistency in how we use enums.
- cons: requires using some magic strings instead of enums.

We decided it would be best to explicitly specify the type of the dictionary in `dss/events/handlers/notify_v2.py` (i.e., specify that the types of the keys can be the enum type or a string), since this type issue only appears in one location of one file, and does not represent a "breakdown" of the abstraction.

PR #2503, opened 2019-10-10
Add Elasticsearch reindex operation

xbrianh:
This adds an operation to the `dss/operations` framework to bulk queue SQS messages to the indexer lambda for each key in `bundles/{pfx}`.

Usage:
```
scripts/dss-ops.py elasticsearch index-prefix --replica {replica} --prefix 0
```

### DIFF
--- a/dss/operations/elasticsearch.py
+++ b/dss/operations/elasticsearch.py
@@ -54,8 +54,7 @@ def index(argv: typing.List[str], args: argparse.Namespace):
     def _forward_keys(pfx):
         with SQSMessenger(index_queue_url) as sqsm:
             for key in handle.list(replica.bucket, pfx):
-                # sqsm.send(json.dumps(dict(replica=args.replica, key=key)))
-                print(key)
+                sqsm.send(json.dumps(dict(replica=args.replica, key=key)))
 
     with ThreadPoolExecutor(max_workers=10) as e:
         futures = [e.submit(_forward_keys, f"bundles/{args.prefix}{c}") for c in hexdigits.lower()]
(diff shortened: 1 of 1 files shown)


dss/operations/elasticsearch.py
```diff
@@ -31,13 +32,34 @@
...
+    handle = Config.get_blobstore_handle(replica)
+    index_queue_url = get_queue_url("dss-index-operation-" + os.environ['DSS_DEPLOYMENT_STAGE'])
+
+    def _forward_keys(pfx):
+        with SQSMessenger(index_queue_url) as sqsm:
+            for key in handle.list(replica.bucket, pfx):
+                sqsm.send(json.dumps(dict(replica=args.replica, key=key)))
+
+    with ThreadPoolExecutor(max_workers=10) as e:
+        futures = [e.submit(_forward_keys, f"bundles/{args.prefix}{c}") for c in hexdigits.lower()]
```
amarjandu: Would using f"bundles/{args.prefix}{c}" mean that an operator would have to submit multiple calls to this script in order to cover all 22 possible starting values for a `uuid`?

xbrianh: Since UUIDs are hexadecimal digits the operator would need to issue 16 calls to cover the entire UUID space.

I'll extend the script to send a message for every bundle if `--prefix` is omitted.

dss/operations/elasticsearch.py
```diff
@@ -26,18 +27,40 @@
...
```
amarjandu:
```
In [4]: hexdigits.lower()
Out[4]: '0123456789abcdefabcdef'
```
I dont think it would matter, but if the prefix is omitted, the queue would fill up 2x for the `a-e` chars. I'm not sure how it would be handled by the SQS Queue Consumer, but just wanted to note it.

xbrianh: thanks for catching that! Will fix.

PR #2505, opened 2019-10-10
Fix optional reindex notifications

xbrianh:


daemons/dss-index/app.py
```diff
@@ -63,25 +63,25 @@ def launch_from_operator_queue(event, context):
...
         remaining_time = AdjustedRemainingTime(actual=RemainingLambdaContextTime(context),
                                                offset=-10)  # leave 10s for shutdown
-        backend = CompositeIndexBackend(executor, remaining_time, DEFAULT_BACKENDS)
+        backend = CompositeIndexBackend(executor, remaining_time, DEFAULT_BACKENDS, notify=send_notifications)
         indexer_cls = Indexer.for_replica(replica)
         indexer = indexer_cls(backend, remaining_time)
-        if send_notifications:
-            indexer.process_new_indexable_object(key)
-        else:
+        if operator_initiated:
```
amarjandu: did you use the indexer.index_object(key) for speed, being that the queue-producer has already validated the presence in stroage?

xbrianh: I just followed the existing reindex pattern found in `dss/stepfunctions/visitation/index.py`, but I like your rationale better.

PR #2509, opened 2019-10-15
Add IAM auditing functionality to dss-ops script (redux)

chmreid:
This is a refactor of a previous PR (#2492). The previous PR tried to keep things logically separated into their own PRs but then things got messy. 😟 

This PR reorganizes the proposed changes into logically consistent but separate PRs. Merges of all the branches together was verified using a throwaway integration branch.

This PR closes #2454 and replaces #2492.

This is a multi-part PR:

* AWS IAM functionality: #2512
* Fusillade IAM functionality: #2511

### Test Plan

Each part of this PR introduces its own tests for all of the functionality it adds.

### Release Notes

Add IAM auditing functionality to dss-ops.py script.

PR #2512, opened 2019-10-15
Add AWS IAM auditing functionality to dss-ops script (redux)

chmreid:
This PR adds AWS functionality to the dss-ops.py script's IAM action.

This PR is part of #2509

### Testing Plan

This adds both AWS IAM functionality and tests for the new functionality.

### DIFF
--- a/dss/operations/iam.py
+++ b/dss/operations/iam.py
@@ -166,22 +166,106 @@ def list_aws_policies_grouped(asset_type, client, managed: bool, do_headers: boo
     :param managed: (boolean) if true, include AWS-managed policies
     :returns: list of tuples of two strings in the form (asset_name, policy_name)
     """
-    pass
+    extracted_list = []
+
+    # Prepare to extract labels for JSON returned by API
+    def _make_aws_api_labels_dict_entry(*args) -> typing.Dict[str, str]:
+        """
+        Convenience function to unpack 4 values into a dictionary of labels, useful for processing
+        API results.
+
+        :params arg[0]: label of asset type
+        :params arg[1]: label of asset name
+        :params arg[2]: label of asset details
+        :params arg[3]: label of asset policy details
+        :returns: dictionary of organized labels
+        """
+        assert len(args) == 4, "Error: need 4 arguments!"
+        return dict(
+            extracted_list_label=args[0], name_label=args[1], detail_list_label=args[2], policy_list_label=args[3]
+        )
+
+    def _get_aws_api_labels_dict() -> typing.Dict[str, typing.Dict[str, str]]:
+        """Store the labels used to unwrap JSON results from the AWS API"""
+        labels = {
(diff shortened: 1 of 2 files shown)


dss/operations/iam.py
```diff
@@ -66,16 +67,51 @@ def get_fus_role_attached_policies(fus_client, action, role):
...
+                # first save as strings
+                master_list.append(json.dumps(policy, sort_keys=True, default=str))
+
+    if action == "list":
+        # Sort names, remove duplicates
+        master_list = sorted(list(set(master_list)))
+    elif action == "dump":
+        # Convert to strings, remove dupes, convert to back dicts
+        master_list = list(set(master_list))
+        master_list = [json.loads(j) for j in master_list]
```
amarjandu: Can the set(master_list) be used in the list comprehension?

amarjandu: Can the set(master_list) be used in the list comprehension?

dss/operations/iam.py
```diff
@@ -120,20 +160,112 @@ def dump_fus_policies(fus_client, do_headers):
...
+    # 2. managed policies
+
+    for asset_detail in response_detail_list:
+        asset_name = asset_detail[name_label]
+
+        # 1. Inline policies
+        # Check if any policies are present
+        if policy_list_label in asset_detail:
+            for inline_policy in asset_detail[policy_list_label]:
+                policy_name = inline_policy["PolicyName"]
```
amarjandu: should this be able to handle if the policy is missing? or is that case not possible.

chmreid: Line 227, `if policy_list_label in asset_detail`, handles the case where there are no policies attached to the asset; if there are any policies attached to the asset, it will be held under the keys "UserDetailList"/"GroupDetailList"/"RoleDetailList" (which is what `policy_list_label` stores).

Exploring the `get_account_authorization_details` API responses, I did not see any cases where policies were missing names. But I don't know if a policy name is _required_ for a policy to be attached to a user. Will double-check that...

dss/operations/iam.py
chmreid: 👍  good call

dss/operations/iam.py
chmreid: From [aws cli docs](https://docs.aws.amazon.com/cli/latest/reference/iam/get-account-authorization-details.html#output) for the `get_account_authorization_details` api endpoint, 

> UserPolicyList -> (list)
    A list of the inline policies embedded in the user.
    (structure)
        Contains information about an IAM policy, including the policy document.
        This data type is used as a response element in the GetAccountAuthorizationDetails operation.
        PolicyName -> (string)
            The name of the policy.
        PolicyDocument -> (string)
            The policy document.

Same is true for GroupPolicyList and RolesPolicyList return values - so we can expect the `PolicyName` field to be present.

Issue #2409, opened 2019-08-30
Add file deletion API


kozbo:
From RFC https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0004-dss-deletion-process.md

## User stories
As an operator of a secondary index, I would like to know what bundles have been deleted, so I can keep my personal index up-to-date.
As a data contributor, I need to remove data from the DSS as I have realised the consent it was collected under does not allow it to be hosted by the DCP.
As a data consumer of the DSS, I would like the API to return a status indicating bundle deletion, so that I know what happened to a bundle that I used previously.

## Definition of done
Only physical deletes exist for the file deletion API. The file deletion API is a two part process. The first request will return the bundles that will be logically deleted as a side effect and a confirmation code. The second request must include the confirmation code in order to begin the deletion process. The request follows this swagger document:

Path: DELETE /file/{uuid}

uuid	: A RFC4122-compliant ID for the bundle.
version : Timestamp of bundle creation in DSS_VERSION format.
reason : the reason for the deletion.
details : User-friendly reason for the bundle or timestamp-specfic bundle deletion.
Confirmation Code : A code used to confirm a deletion operation.

Response Codes
200	Deletion pending. A confirmation code is returned to confirm the deletion in the second request. A list of affect files and bundles is also returned
201	Deletion Confirmed with confirmation code.
403	Unauthorized user it attempting this action.
404	The files has already been deleted.
409	The confimation code used is invalid. Either the confirmation code was incorrect, or an effected bundle has changed state since the original request.

A user must have explicit permission to perform the file deletion operation.

kozbo:
@natanlao

kozbo:
@natanlao If the tasks have been identified for this effort then the **[SPIKE]** prefix can be removed from the description

PR #2516, opened 2019-10-15
Gentle refactor of tombstoning code

natanlao:
This is the first of several PRs needed to implement the file deletion API (#2409). This PR refactors a tombstone helper function to be consistent with the deletion RFC and relocates it in preparation for use with files (and not just bundles).

### DIFF
--- a/dss/storage/__init__.py
+++ b/dss/storage/__init__.py
@@ -1,0 +1,11 @@
+from typing import NamedTuple, Optional
+
+
+class Tombstone(NamedTuple):
+    """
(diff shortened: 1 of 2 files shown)


dss/storage/__init__.py
```diff
@@ -0,0 +1,8 @@
+from typing import NamedTuple, Optional
+
+
+class Tombstone(NamedTuple):
+    """Tombstone object compliant with RFC #4 (Deletion of data in the DCP)"""
```
chmreid: can you provide a permalink or mention the org/repo where this rfc can be found?

natanlao:
Passing on Allspark

PR #2526, opened 2019-10-16
Hotfix for lambda params

chmreid:
This fixes a problem with how lambda variable prefixes are set. The SSM store requires a leading /, while the secrets store does not. This update reflects that change in `dss/operations/lambda_params.py`.

### DIFF
--- a/dss/operations/lambda_params.py
+++ b/dss/operations/lambda_params.py
@@ -35,7 +35,7 @@ def get_ssm_variable_prefix() -> str:
     """
     store_name = os.environ["DSS_PARAMETER_STORE"]
     stage_name = os.environ["DSS_DEPLOYMENT_STAGE"]
-    store_prefix = f"{store_name}/{stage_name}"
+    store_prefix = f"/{store_name}/{stage_name}"
     return store_prefix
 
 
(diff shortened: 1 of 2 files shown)


dss/operations/lambda_params.py
```diff
@@ -35,21 +35,27 @@ def get_ssm_variable_prefix() -> str:
...
     prefix = get_ssm_variable_prefix()
 
-    # Strip leading/trailing slash
+    # Strip leading slash
     if param_name.startswith("/"):
         param_name = param_name[1:]
+
+    # Strip trailing slash
     if param_name.endswith("/"):
         param_name = param_name[:-1]
```
xbrianh: This will strip a trailing slash or do nothing:
```
param_name = param_name.rstrip("/")
```

dss/operations/lambda_params.py
```diff
@@ -35,21 +35,27 @@ def get_ssm_variable_prefix() -> str:
...
 
 
 def fix_ssm_variable_prefix(param_name: str) -> str:
     """Add (if necessary) the variable store and stage prefix to the front of the name of an SSM store parameter"""
     prefix = get_ssm_variable_prefix()
 
-    # Strip leading/trailing slash
+    # Strip leading slash
     if param_name.startswith("/"):
         param_name = param_name[1:]
```
xbrianh: This will strip a leading slash or do nothing:
```
param_name = param_name.lstrip("/")
```

dss/operations/lambda_params.py
```diff
@@ -35,21 +35,27 @@ def get_ssm_variable_prefix() -> str:
...
     if param_name.endswith("/"):
         param_name = param_name[:-1]
 
-    if not (param_name.startswith(prefix) or param_name.startswith("/" + prefix)):
+    if not (
+            param_name.startswith(prefix)
+            or param_name.startswith("/" + prefix)
+            or param_name.startswith(prefix[1:])
+    ):
         param_name = f"{prefix}/{param_name}"
```
xbrianh: Is this equivalent?:
```
return param_name.replace(prefix, "", 1).lstrip("/")
```

Note that `lstrip("/")` will strip multiple slashes off the left.

dss/operations/lambda_params.py
chmreid: Perfect!

dss/operations/lambda_params.py
chmreid: Excellent, thank you Brian

Gonna start calling you Mr. brevity is the soul of wit

PR #2527, opened 2019-10-16
Fix events paging, bump flashflood to 0.4.0

xbrianh:
This also adds a paging test asserting future CLI expectations of the paged response.

### DIFF
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -26,6 +26,7 @@ from dss import events
 from dss.config import BucketConfig, Config, Replica, override_bucket_config
 from dss.util.version import datetime_to_version_format
 from dss.util import UrlBuilder
+from dss.util.version import datetime_from_timestamp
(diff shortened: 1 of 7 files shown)


tests/test_events.py
```diff
@@ -135,6 +138,19 @@ def test_get(self):
...
+            resp = self.app.get(f"/v1/events?replica={replica.name}")
+            self.assertIn(resp.status_code, [200, 206])
+            content_key = resp.headers.get("X-OpenAPI-Paginated-Content-Key")
+            self.assertEqual("event_streams", content_key)
+            results = resp.json()
+            for key in content_key.split("."):
+                results = results[key]
+            for result in results:
+                for event in replay_event_stream(result):
+                    pass
```
amarjandu: Is there something to be added here? If we are testing that the paging works and that data is being passed back, the length of the results object could be checked.

PR #2532, opened 2019-10-17
[WIP] Daemons and ops to journal and update events

xbrianh:
This ads a daemon function to compile events into journals, and to apply updates.

Operational scripting is also updated to use the new flashflood interface methods.

depends on #2527

### DIFF
--- a/daemons/dss-notify-v2/app.py
+++ b/daemons/dss-notify-v2/app.py
@@ -9,6 +9,7 @@ from urllib.parse import unquote
 from concurrent.futures import ThreadPoolExecutor
 
 import domovoi
+from flashflood import FlashFloodJournalingError
 
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), 'domovoilib'))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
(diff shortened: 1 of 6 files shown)


daemons/dss-notify-v2/app.py
```diff
@@ -123,5 +126,24 @@ def _func(subscription):
     with ThreadPoolExecutor(max_workers=20) as e:
         e.map(_func, get_subscriptions_for_replica(replica))
 
+@app.scheduled_function("rate(5 minutes)")
```
amarjandu: I *think* this creates another lambda, it might be worthwhile to move it into its own daemon, so we can exert a little more control over it, without effecting changes to the notify_v2.

tests/test_utils.py
```diff
@@ -363,5 +335,51 @@ def test_tombstone_pages(self, mock_list_v2):
...
+        metadata = {'something': "foolish"}
+        part_size = 5 * 1024 * 1024
+        s3_client = Config.get_native_handle(Replica.aws)
+        bucket = os.environ['DSS_S3_BUCKET_TEST']
+        with self.subTest("copy multiple parts"):
+            with io.BytesIO(data) as fh:
+                multipart_parallel_upload(
+                    s3_client,
+                    bucket,
+                    "fake_key",
```
amarjandu: please use a key name that can link us back to this test.

dss/util/__init__.py
```diff
@@ -188,3 +190,23 @@ def _chunks():
...
+    Convenience generator to time limit iterated work.
+    """
+    start_time = python_time.time()
+    while True:
+        seconds_remaining = maximum_duration - (python_time.time() - start_time)
+        if 0 > seconds_remaining:
+            break
+        yield seconds_remaining
+
+def circular_generator(lst: typing.List[typing.Any]) -> typing.Any:
```
amarjandu: could itertools.cycle have not been used?

xbrianh: Thanks, I wasn't aware of itertools.cycle. Itertools is usually a good way to write cryptic code, but this looks like an appropriate use case.

tests/test_utils.py
xbrianh: I'm going to revert this change and address your comment in a different PR.

daemons/dss-notify-v2/app.py
xbrianh: Ok. I was hoping to avoid another daemon, but event journaling probably deserves its own process that we can monitor and throttle as necessary.

xbrianh:
Splitting this out into separate PRs

amarjandu:
Just just wanted to note that we should probably have Kharon delete this prefix.

PR #2533, opened 2019-10-18
Improve events operations, bump flashflood to 0.4.1

xbrianh:
This adds a Lambda-parallelized event journaling operation, and a journal listing operation. It also centralizes some flashflood access calls to `dss/events/__init__.py`

flashflood is bumped to 0.4.1

### DIFF
--- a/dss/operations/events.py
+++ b/dss/operations/events.py
@@ -6,20 +6,22 @@ import typing
 import argparse
 import json
 import logging
+import traceback
 from uuid import uuid4
 from string import hexdigits
 from datetime import datetime
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
 from cloud_blobstore import BlobNotFoundError
-from flashflood import FlashFlood, FlashFloodJournalingError
+from flashflood import FlashFlood, JournalID
 from dcplib.aws.sqs import SQSMessenger
 from dcplib.aws.clients import logs
 
 from dss.config import Config, Replica
 from dss.util.aws import resources
 from dss.operations import dispatch
-from dss.events import get_bundle_metadata_document, record_event_for_bundle
+from dss.events import (get_bundle_metadata_document, record_event_for_bundle, journal_flashflood,
+                        update_flashflood, list_new_flashflood_journals)
 from dss.events.handlers.notify_v2 import _versioned_tombstone_key_regex, _unversioned_tombstone_key_regex
 from dss.storage.bundles import Living
 from dss.storage.identifiers import TOMBSTONE_SUFFIX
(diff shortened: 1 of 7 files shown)


dss/operations/events.py
```diff
@@ -86,24 +88,49 @@ def forward_keys(bundle_fqids):
 @events.action("journal",
                arguments={"--prefix": dict(required=True,
                                            help="flashflood prefix to journal events"),
-                          "--minimum-number-of-events": dict(default=1000, type=int),
-                          "--minimum-size": dict(default=1024 * 1024, type=int)})
+                          "--number-of-events": dict(default=None, type=int),
+                          "--starting-journal-id": dict(default=None),
+                          "--job-id": dict(default=None)})
 def journal(argv: typing.List[str], args: argparse.Namespace):
```
chmreid: What is the high-level idea behind this `dss-ops.py events journal` action?

dss/operations/events.py
```diff
@@ -86,24 +88,49 @@ def forward_keys(bundle_fqids):
...
+                    sqsm.send(cmd_template.format(journals[0]))
+                    journals = list()
+        monitor_logs(logs, job_id, start_time)
+    else:
+        msg = dict(action="journal", job_id=job_id, prefix=args.prefix, key=args.starting_journal_id)
         try:
-            ff.journal(args.minimum_number_of_events, args.minimum_size)
-        except FlashFloodJournalingError:
-            break
+            journal_flashflood(args.prefix, args.number_of_events, JournalID(args.starting_journal_id))
```
chmreid: The `journal_flashflood()` method compiles a list of Journals, then assembles and returns a combined journal. But that value isn't assigned to a variable here, so it doesn't seem like this `journal_flashflood()` call actually does anything. Can you clarify what's happening with this function call?

dss/operations/events.py
```diff
@@ -86,24 +88,49 @@ def forward_keys(bundle_fqids):
...
+        print(json.dumps(msg))
 
 @events.action("update",
                arguments={"--prefix": dict(required=True,
                                            help="flashflood prefix to journal events"),
-                          "--minimum-number-of-events": dict(default=1000, type=int),
-                          "--minimum-size": dict(default=1024 * 1024, type=int)})
+                          "--number-of-updates-to-apply": dict(default=None, type=int)})
 def update(argv: typing.List[str], args: argparse.Namespace):
+    update_flashflood(args.prefix, args.number_of_updates_to_apply)
```
chmreid: Not to belabor this, but the `update_flashflood()` method's docstring says "Apply event updates to existing journals" and I am not following what it means to apply an event update to a journal.

Is the update action taking the command parameters from each SQS queue item (above) and running the dss-ops script with the arguments provided by the queue item?

dss/operations/events.py
xbrianh: When a new event occurs, the event data is stored as an s3 object. During journaling, these objects are compiled into a _journal_ of many events and stored as a single s3 object. This makes event replay faster.

dss/operations/events.py
```diff
@@ -86,24 +88,49 @@ def forward_keys(bundle_fqids):
...
-                          "--minimum-number-of-events": dict(default=1000, type=int),
-                          "--minimum-size": dict(default=1024 * 1024, type=int)})
+                          "--number-of-events": dict(default=None, type=int),
+                          "--starting-journal-id": dict(default=None),
+                          "--job-id": dict(default=None)})
 def journal(argv: typing.List[str], args: argparse.Namespace):
-    ff = FlashFlood(resources.s3, Config.get_flashflood_bucket(), args.prefix)
-    while True:
+    job_id = args.job_id or f"{uuid4()}"
+    cmd_template = (f"events journal --job-id {job_id} "
```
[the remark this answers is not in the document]
xbrianh: The command is processed if `args.starting_journal_id != None`, otherwise it is queued. You can see the control flow on line 101.

I'll add doc strings.

dss/operations/events.py
xbrianh: The heavy lifting is performed by `ff.combine_journals()`, which causes flashflood to save the new journal to s3.

dss/operations/events.py
xbrianh: This command does not use an SQS queue.

Flashflood supports the deletion or update of event data. However, modifying event journals as updates/deletes arrive is not concurrency safe. Instead they are saved and applied later by a non-concurrent process. This is done by calling `ff.update()`.

dss/operations/events.py
```diff
@@ -86,24 +88,49 @@ def forward_keys(bundle_fqids):
...
+
+    if args.starting_journal_id is None:
+        start_time = datetime.now()
+        with SQSMessenger(command_queue_url) as sqsm:
+            journals = list()
+            for journal_id in list_new_flashflood_journals(args.prefix):
+                journals.append(journal_id)
+                if args.number_of_events == len(journals):
+                    print(f"Journaling from {journals[0]} with {args.number_of_events} events.")
+                    sqsm.send(cmd_template.format(journals[0]))
```
amarjandu: nit: Can this string formatting be more explicit?
the build_deploy_config.py for the `infra` has a nice example where the `backend_template` is being formatted. https://github.com/HumanCellAtlas/data-store/blob/2530dbfca9080eb8392b4c0ced4fbd0dbc35a454/infra/build_deploy_config.py#L30

xbrianh: it can! fixed.

PR #2541, opened 2019-10-22
[test] Disable codecov upload on PR builds

### LINKED ISSUE #2692: Improve codecov reports
We have experienced intermittent problems with codecov reports, mainly related to misleading code coverage statistics. For example, very significant drops in coverage (>50%) caused by changing a single line in a Readme.

The problem we have identified is that we are running five separate tests for each commit on a pull request (because there are 5 test suites that each upload codecov reports). @natanlao has proposed bumping the "report after N builds" setting to 5, so that codecov will wait until all 5 tests have finished uploading to actually compute its code coverage statistics.

This issue is linked to PR #2541 from @natanlao.

natanlao:
Codecov has been returning weird results for some builds, usually
trending lower (saying that code coverage has dropped by some
ridiculous amount). I think this is because Codecov is getting
confused by all the different reports being uploaded - five from
the PR build, and another five from the branch build.

Here's an example of Codecov conflating reports from different
builds:

https://codecov.io/gh/HumanCellAtlas/data-store/commit/2fcdaa2c27de001e6a436371950d0caecdb40d3d/builds

Closes #2692

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,3 @@
+codecov:
+    notify:
+        after_n_builds: 5
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ script:
 - make $MAKE_ARGS
 
 after_success:
-- bash <(curl -s https://codecov.io/bash)
+- if [[ "$TRAVIS_EVENT_TYPE" == "push" ]]; then bash <(curl -s https://codecov.io/bash); fi
 
 if: tag IS blank # don't build tags
 
--- a/README.md
+++ b/README.md
@@ -525,3 +525,4 @@ See our [Security Policy](https://github.com/HumanCellAtlas/.github/blob/master/
 ## Contributing
 
 External contributions are welcome. Please review the [Contributing Guidelines](CONTRIBUTING.md)
+


natanlao:
The expected behavior is that Codecov reports CI failure (for the linked report).

natanlao:
I'm not entirely convinced that this fix works, I'm going to leave this PR open to do some more testing over the next week or so.

natanlao:
[Exhibit B](https://codecov.io/gh/HumanCellAtlas/data-store/commit/2f30094f738c6978a67e07ed145859068cca41a7/builds)

.codecov.yml
```diff
@@ -0,0 +1,3 @@
+codecov:
+    notify:
+        after_n_builds: 5
```
chmreid: Where does the 5 come from? The [codecov.io documentation of `after_n_builds`](https://docs.codecov.io/docs/pull-request-comments#section-after_n_builds) says it should be set greater than the number of reports uploaded per commit

chmreid:
Hmm, the old codecov reports that demonstrated the improved behavior seem to have disappeared

.codecov.yml
natanlao: I read

> Generally, you should set `after_n_builds` equal to the number of reports you upload per commit.

There are five codecov reports (one per job) uploaded per build, so I think this is right. One part I am unclear of is if it refers to number of reports uploaded per *commit* or per *build* - if it is the former, then, were we not to disable codecov on one of two builds per PR, `after_n_builds` should be set to ten. But I think with these changes it might not matter...

chmreid:
We revisited this during standup today, I think we're going to merge this in and keep an eye on the codecov reports to see if there is any improvement.

.codecov.yml
chmreid: Will keep an eye out for whether the reported statistics improve - if not we can try bumping from 5 to 10.

PR #2545, opened 2019-10-23
Update smoketest to new  FlashFlood API

xbrianh:


natanlao:
Merging at the direction of @amarjandu

PR #2549, opened 2019-10-28
Refactor checkout operations.

DailyDreaming:
Some mostly minor refactoring of the checkout operations that I felt would clutter up adding the sync operation.  There was one bug fixed by renaming `_file` to `file_metadata` which should allow verification of file keys.  Another one skips (and warns) about files that are not in the main bucket but previously errored when attempting to fetch the metadata.


dss/operations/checkout.py
```diff
@@ -151,37 +150,43 @@ def process_keys(self):
                 checkout_status[_key] = bundle_internal_status
             elif _key.startswith(FILE_PREFIX):
                 temp = collections.defaultdict(blob_checkout=False, should_be_cached=False)
-                file_metadata = self._verify_get(self.handle, self.replica.bucket, _key)
+                file_metadata = self._get_metadata(self.handle, self.replica.bucket, _key)
+                if not file_metadata:
+                    checkout_status[_key] = 'KEY_DOES_NOT_EXIST'
```
amarjandu: nit: I would want to set the key to  `None` to avoid parsing issues if the output gets piped into another script.

dss/operations/checkout.py
```diff
@@ -151,37 +150,43 @@ def process_keys(self):
...
+                if cache_flow.should_cache_file(file_metadata['content-type'], file_metadata['size']):
                     temp['should_be_cached'] = True
 
                 for x in ['name', 'uuid', 'version']:
-                    temp.update({x: _file[x]})
+                    temp.update({x: file_metadata[x]})
                 checkout_status[_key] = collections.defaultdict(uuid=temp)
-        print(json.dumps(checkout_status, sort_keys=True))
+            else:
+                checkout_status[_key] = 'INVALID_KEY_REGEX'
```
amarjandu: same comment as above

dss/operations/checkout.py
```diff
@@ -151,37 +150,43 @@ def process_keys(self):
                 checkout_status[_key] = bundle_internal_status
             elif _key.startswith(FILE_PREFIX):
                 temp = collections.defaultdict(blob_checkout=False, should_be_cached=False)
-                file_metadata = self._verify_get(self.handle, self.replica.bucket, _key)
+                file_metadata = self._get_metadata(self.handle, self.replica.bucket, _key)
```
amarjandu: If there is no file meta_data, or the file is not found, would this actually be set to `None`? it looks like it was removed from the command above.

dss/operations/checkout.py
```diff
@@ -63,72 +62,72 @@ def _sleepy_checkout(self, function=None, **kwargs):
...
-        Verify that keys are in the checkout bucket
-        """
+        """Verify that keys are in the checkout bucket."""
         checkout_status = dict(replica=self.replica.name)
         for _key in self.keys:
             if DSS_BUNDLE_KEY_REGEX.match(_key):  # handles bundles/fqid keys or fqid
                 uuid, version = self._parse_key(_key)
                 bundle_manifest = get_bundle_manifest(replica=self.replica, uuid=uuid, version=version)
-                checkout_bundle_contents = [x[0] for x in self.handle.list_v2(bucket=self.bucket,
+                print(bundle_manifest)
```
amarjandu: i would avoid printing stuff that does not need to be printed, unless you can mute it, If i want to pipe this into JQ, the print statement  causes problems.

dss/operations/checkout.py
```diff
@@ -63,72 +62,72 @@ def _sleepy_checkout(self, function=None, **kwargs):
...
-                    self._verify_delete(self.handle, self.bucket, key)
+                    self._verify_delete(self.handle, self.checkout_bucket, key)
             elif _key.startswith(FILE_PREFIX):
                 # should handle other keys, files/blobs
-                file_metadata = self._verify_get(self.handle, self.replica.bucket, _key)
-                self._verify_delete(self.handle, self.bucket, key=compose_blob_key(file_metadata))
+                file_metadata = self._get_metadata(self.handle, self.replica.bucket, _key)
+                self._verify_delete(self.handle, self.checkout_bucket, key=compose_blob_key(file_metadata))
+            else:
+                print(f'INVALID KEY REGEX: {_key}')
```
amarjandu: i would avoid printing stuff that does not need to be printed, unless you can mute it, If i want to pipe this into JQ, the print statement causes problems.

DailyDreaming: Hmmm... should we raise here and disallow nonsense keys?  I can see your point when wanting to pipe.  Inputting a key like `borkel/snorkel`, I expect to either see that it doesn't exist in the checkout bucket or that it raises an error because it does not match the expected key.

dss/operations/checkout.py
DailyDreaming: Ah, this was cruft.  Should be removed now.

dss/operations/checkout.py
amarjandu: perhaps https://github.com/HumanCellAtlas/data-store/pull/2549#discussion_r340356914

dss/operations/checkout.py
DailyDreaming: Setting it to `None` is and was the original behavior.  I just renamed the function from `_verify_get` to `_get_metadata` since I thought it was clearer.

DailyDreaming: You're right though, that's a bug (one of a few) whenever it eats any key that doesn't exist in the main bucket.  I've fixed it in `Verify` to `continue` when it hits this, but it ideally would have had tests to catch this.

dss/operations/checkout.py
xbrianh: Warnings written to [sys.stderr](https://docs.python.org/3/library/sys.html#sys.stderr) will not contaminate stdout.

dss/operations/checkout.py
```diff
@@ -24,35 +24,34 @@
...
+    def _get_metadata(handle, bucket, key):
+        """Helper Function to get file_metadata from cloud-blobstore."""
         try:
             file_metadata = json.loads(handle.get(bucket=bucket, key=key).decode('utf-8'))
+            return file_metadata
         except BlobNotFoundError:
-            logger.warning(f'unable to locate {bucket}/{key}')
-            return None
-        return file_metadata
+            logger.warning(f'Unable to locate: {bucket}/{key}')
```
amarjandu: return `None` was removed here, does json.loads(handle.get()) eval to  None if there is no blob found ?

dss/operations/checkout.py
DailyDreaming: This situation is different from that above in my mind since the key is malformed (and thus never could exist).  Should we raise to warn/disallow the user?

dss/operations/checkout.py
DailyDreaming: I could see that, though either could easily be parsed.  I chose this since it's unambiguous.

dss/operations/checkout.py
DailyDreaming: Yes.  The default of any function that does not explicitly return is to return `None`.

dss/operations/checkout.py
DailyDreaming: @xbrianh I like that idea, will implement.

dss/operations/checkout.py
amarjandu: `[k for k,v in checkout_status.items() if v is not None]` 
vs
`[k for k,v in checkout_status.items() if v !='KEY_DOES_NOT_EXIST ]`  

for the latter you would need to know that that's the value of they if  the file_metadata  is  not found. I don't think its an intuitive approach, it is explicit tho.  
🤷‍♂

dss/operations/checkout.py
amarjandu: https://github.com/HumanCellAtlas/data-store/pull/2549#discussion_r340363942 from the wizard himself. :male_mage

dss/operations/checkout.py
DailyDreaming: I can see your point.  I'd expect `None` when it's not in the checkout bucket, but this is when the file isn't even in the main bucket and so can't be checked out even if you wanted to.  I wanted to differentiate between those two cases, but I think I'll use `sys.stderr.write()` here too, as Brian suggested.

dss/operations/checkout.py
amarjandu: Does this need to get changed to sys.stderr.write() ?

PR #2552, opened 2019-10-29
[easy] Pin working version of boto3 for gcp cloud fn.

DailyDreaming:


daemons/dss-gs-event-relay/requirements.txt
```diff
@@ -1 +1 @@
-boto3
```
amarjandu: I think this would be ok, it seems that this is just publishing SNS topics, we should not see interference to the lambda requirements.txt or the layers. 

 i would rather we remove the boto3 http adapter and use https://3.python-requests.org/user/advanced/#transport-adapters

DailyDreaming: There's an issue for doing exactly that: https://github.com/HumanCellAtlas/data-store/issues/2553

PR #2557, opened 2019-10-30
Add checkout sync operation.

DailyDreaming:
This will check out all files meeting the caching criteria.

### DIFF
--- a/dss/operations/checkout.py
+++ b/dss/operations/checkout.py
@@ -1,13 +1,15 @@
 """
 Tools for managing checkout buckets
 """
-import sys
 import typing
 import argparse
 import logging
 import json
 import time
 import collections
+import multiprocessing
+from typing import Dict
+from concurrent.futures import ThreadPoolExecutor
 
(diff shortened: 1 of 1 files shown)


dss/operations/checkout.py
```diff
@@ -190,3 +194,57 @@ def process_keys(self):
...
+        if len(self.tombstone_cache) >= self.tombstone_cache_max_len:
+            self.tombstone_cache.popitem()
+
+        if uuid not in self.tombstone_cache:
+            try:
+                self.tombstone_cache[uuid] = self.handle.get(self.replica.bucket,
+                                                             key=f'{FILE_PREFIX}/{uuid}.{TOMBSTONE_SUFFIX}')
+            except BlobNotFoundError:
+                self.tombstone_cache[uuid] = None
+        return self.tombstone_cache[uuid]
```
xbrianh: What’s the purpose of the tombstone cache? Do we expect to list keys more than once?

PR #2561, opened 2019-10-31
Suppress test output on a per-test basis

natanlao:
Turns out that smoketests rely on stdout so those can't be buffered.

Sorry for breaking master, again.


chmreid:
Way to go @natanlao ! Output looks so much nicer now: https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/53523

PR #2565, opened 2019-11-07
Fix test typo with LICENSE

chmreid:
Yup, what it says

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -592,7 +592,7 @@ class TestBundleApi(unittest.TestCase, TestAuthMixin, DSSAssertMixin, DSSUploadM
                 resp = self.put_bundle(
                     replica,
                     bundle_uuid3,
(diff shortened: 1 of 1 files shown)


tests/test_bundle.py
```diff
@@ -592,7 +592,7 @@ def _test_bundle_put(self, replica: Replica, fixtures_bucket: str):
                 resp = self.put_bundle(
                     replica,
                     bundle_uuid3,
-                    [(file_uuid, file_version, "LICENSE"), (file_uuid, file_version, "LIasdfCENSE")],
+                    [(file_uuid, file_version, "LICENSE"), (file_uuid, file_version, "LICENSE")],
```
xbrianh: I don't think this is a typo (although it may be a poor choice of filename for this test).

Check out the other subtest for [duplicated filenames](https://github.com/HumanCellAtlas/data-store/blob/23396bc25dfa96e6b40105cdb34c4ccf6772e831/tests/test_bundle.py#L567)

chmreid: 🤔  thanks for pointing this out!

PR #2573, opened 2019-11-08
Add method to list bundle zombies

xbrianh:
This adds a method to list bundle keys associated with a tombstone key, in the following manner:

For this listing:
```
bundles/uuid.version1
bundles/uuid.version2
bundles/uuid.version2.dead
bundles/uuid.version3
bundles/uuid.version3.dead
bundles/uuid.dead
```

We should get these results:
`get_tombstoned_bundles(bundles/uuid.version2.dead)` -> `bundles/uuid.version2`
`get_tombstoned_bundles(bundles/uuid.dead)` -> `bundles/uuid.version1`

Note that the unversioned tombstone is associate with bundles _not previously tombstoned_.

This functionality will be useful for refactoring notify_v2 logic, and for forthcoming logic to remove tombstoned bundles from event streams.

### DIFF
--- a/dss/storage/bundles.py
+++ b/dss/storage/bundles.py
@@ -203,7 +203,7 @@ _unversioned_tombstone_key_regex = re.compile(f"^(bundles)/({UUID_PATTERN}).{TOM
 def get_tombstoned_bundles(replica: Replica, tombstone_key: str) -> typing.Iterator[str]:
     """
     Return the bundle fqid(s) associated with a versioned or unversioned tombstone, as verified on object storage.
-    Note that an unversioned tombstone returns keys associated with bundles not previosly, as show in the example
+    Note that an unversioned tombstone returns keys associated with bundles not previously, as show in the example
     below.
 
     bundles/uuid.version1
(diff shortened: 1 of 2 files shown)


dss/storage/bundles.py
```diff
@@ -196,3 +197,41 @@ def __iter__(self):
 
         for bundle_fqid in self._living_fqids_in_bundle_info():
             yield bundle_fqid
+
+_versioned_tombstone_key_regex = re.compile(f"^(bundles)/({UUID_PATTERN}).({VERSION_PATTERN}).{TOMBSTONE_SUFFIX}$")
+_unversioned_tombstone_key_regex = re.compile(f"^(bundles)/({UUID_PATTERN}).{TOMBSTONE_SUFFIX}$")
+def get_tombstoned_bundles(replica: Replica, tombstone_key: str) -> typing.Iterator[str]:
+    """
+    Return the bundle fqid(s) associated with a versioned or unversioned tombstone, as verified on object storage.
+    Note that an unversioned tombstone returns keys associated with bundles not previosly, as show in the example
```
Bento007:
```suggestion
    Note that an unversioned tombstone returns keys associated with bundles not previously, as show in the example
```

PR #2575, opened 2019-11-08
Move tombstone identifiers, update test

xbrianh:
This:
- Updates the test for `get_tombstoned_bundles`
- Moves tombstone regex into `dss.storage.identifiers`


dss/storage/identifiers.py
```diff
@@ -57,6 +57,9 @@
 BUNDLES_PATTERN = f'/v1/bundles/{UUID_PATTERN}'
 BUNDLES_URI_REGEX = re.compile(BUNDLES_PATTERN)
 
+VERSIONED_TOMBSTONE_KEY_REGEX = re.compile(f"^(bundles)/({UUID_PATTERN}).({VERSION_PATTERN}).{TOMBSTONE_SUFFIX}$")
```
amarjandu: is `bundles` supposed to have a value?

amarjandu: also is this different from `DSS_BUNDLE_TOMBSTONE_REGEX ` ln 47?

xbrianh: Not different. Thanks for catching that, will fix.

PR #2578, opened 2019-11-11
Fix syntax errors in Fusillade roles.json

chmreid:
This fixes some bad formatting in the `roles.json` file, which defines Fusillade roles for this component


chmreid:
Not an issue, was looking at the wrong branch...

PR #2580, opened 2019-11-11
Operational tooling: IAM fixups

chmreid:
A few assorted fixups for the dss-ops script and the IAM action.
* Fix argument checking
* Add `--quiet` and `--stage` flags to `dss-ops.py iam list` action

Partially addresses #2592


chmreid:
>  Is there a good reason to diverge?

It's necessary to add the `--stage` flag to IAM operations because fusillade's dev/staging/integration stages don't line up with DSS's dev/staging/integration stages.

It is also okay because it is a purely informational action and does not make any modifications to the Fusillade state.

This doesn't modify any make commands, so I'm not sure the `make` example you gave applies here.

DailyDreaming:
I think as the dss's interaction with fusillade develops more we should add those relationships to env vars and populate them as needed.  I feel like it would keep the env vars consistent.

Is there a downside to currently taking the stage from the `DSS_DEPLOYMENT_STAGE` env var?

chmreid:
We cannot pull the Fusillade stage from the DSS stage environment variable, because that would require the `dev` stage of DSS to point to the `dev` stage of Fusillade. Currently, the `dev` stage of DSS must point to the `testing` stage of Fusillade.

The alternatives are to: (1) add an environment variable for the Fusillade stage, which I don't like because it isn't very transparent, or (2) hard-code some logic like "if we are on the dev stage of DSS, point to the testing stage of Fusillade", or (3) swap the role that fusillade's `dev` and `testing` branches are playing, so that the fusillade `dev` deployment is used by other components' `dev` deployments, and fusillade's `testing` deployment is used for testing Fusillade (and is considered unstable). Currently, the fusillade `dev` deployment is used for testing Fusillade, so other components' `dev` deployments must point to Fusillade's `testing` deployment.

(3) is what Amar and I were suggesting at the stand-up meeting last week. I think it would be a less confusing approach for operators and would eliminate the need for the `--stage` flag.

chmreid:
As per suggestion from @xbrianh we will implement a mapping from dss stages to Fusillade stages, so that we can use the environment variables to determine stage while still respecting the difference between Fusillade stages and DSS stages.

chmreid:
@DailyDreaming eliminated the `--stage` flag and am now using a map to get the DSS->Fusillade stages correct, please re-review at your convenience

chmreid:
There are several other PRs waiting on this one and not sure when @DailyDreaming will get back to this so gonna merge this in

PR #2582, opened 2019-11-11
Fix API formatting issue

### LINKED ISSUE #443: Do not allow uninitialized indexes
A DSS user reported the following error when issuing a `GET /v1/subscriptions` against  both staging and dev:
```
{
  "code": "unhandled_exception",
  "stacktrace": "Traceback (most recent call last):\n  File \"/var/task/chalicelib/dss/error.py\", line 31, in wrapper\n    return func(*args, **kwargs)\n  File \"/var/task/chalicelib/dss/api/subscriptions.py\", line 70, in find\n    for hit in search.scan()]\n  File \"/var/task/chalicelib/dss/api/subscriptions.py\", line 64, in <listcomp>\n    responses = [{\n  File \"/var/task/elasticsearch_dsl/search.py\", line 642, in scan\n    **self._params\n  File \"/var/task/elasticsearch/helpers/__init__.py\", line 288, in scan\n    request_timeout=request_timeout, **kwargs)\n  File \"/var/task/elasticsearch/client/utils.py\", line 73, in _wrapped\n    return func(*args, params=params, **kwargs)\n  File \"/var/task/elasticsearch/client/__init__.py\", line 623, in search\n    doc_type, '_search'), params=params, body=body)\n  File \"/var/task/elasticsearch/transport.py\", line 312, in perform_request\n    status, headers, data = connection.perform_request(method, url, params, body, ignore=ignore, timeout=timeout)\n  File \"/var/task/elasticsearch/connection/http_requests.py\", line 90, in perform_request\n    self._raise_error(response.status_code, raw_data)\n  File \"/var/task/elasticsearch/connection/base.py\", line 125, in _raise_error\n    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)\nelasticsearch.exceptions.NotFoundError: TransportError(404, 'index_not_found_exception', 'no such index')\n",
  "status": 500,
  "title": "TransportError(404, 'index_not_found_exception', 'no such index')"
}
```

This issue is about ensuring that our deployed systems are never in this state. All relevant indexes need to be initialized/ensured in the process of deployment.

chmreid:
This fixes an issue with the API documentation. An improperly formatted string in the API documentation results in confusing output for the dcp-cli documentation:

Screen Shot 2019-11-11 at 12 06 25

Ref: "Named Arguments" section of dcp-cli documentation https://hca.readthedocs.io/en/latest/cli.html#put-subscription

Closes #443

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -762,13 +762,21 @@ paths:
                 description: >
                   A collection of static form fields to be supplied in the request body, alongside the actual
                   notification payload. The value of each field must be a string. For example, if the subscriptions has
-                  this property set to `{"foo" : "bar"}`, the corresponding notification HTTP request body will be
+                  this property set to `{"foo" : "bar"}`, the corresponding notification HTTP request body will consist
+                  of a multipart frame with two frames,
 
                   ```
 
+                  ----------------2769baffc4f24cbc83ced26aa0c2f712
+
                   Content-Disposition: form-data; name="foo"
+
                   bar
+
+                  ----------------2769baffc4f24cbc83ced26aa0c2f712
+
                   Content-Disposition: form-data; name="payload"
+
                   {"transaction_id": "301c9079-3b20-4311-a131-bcda9b7f08ba", "subscription_id": ...
 
                   ```
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -766,14 +766,10 @@ paths:
 
                   ```
 
-                  --2769baffc4f24cbc83ced26aa0c2f712
                   Content-Disposition: form-data; name="foo"
-
                   bar
                   Content-Disposition: form-data; name="payload"
-
                   {"transaction_id": "301c9079-3b20-4311-a131-bcda9b7f08ba", "subscription_id": ...
-                  --2769baffc4f24cbc83ced26aa0c2f712--
 
                   ```
 


dss-api.yml
```diff
@@ -766,14 +766,10 @@ paths:
 
                   ```
 
-                  --2769baffc4f24cbc83ced26aa0c2f712
                   Content-Disposition: form-data; name="foo"
-
                   bar
```
DailyDreaming: Should `bar` still be here?

chmreid: Yes - this block of code contains a header + body, and `bar` is the body

DailyDreaming: Ah, I didn't understand it.  Since this is documentation, could this be made clearer?

chmreid: For context: the request body shown has multiple header+body frames stacked together. This is perfectly acceptable (see these links from Mozilla: [HTTP/2 Frames section of this  page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages) and [multipart/form-data section of this page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#multipartform-data)). 

Reviewing the second link, I believe the `----------------2769baffc4f24cbc83ced26aa0c2f712` was intended to be a marker between frames, and was just improperly formatted.

chmreid:
View of documentation with new changes:

Screen Shot 2019-11-13 at 14 25 57

To test locally, I edited `~/.config/hca/config.json` and set the following in the config:

```
"DSSClient": {"swagger_url": "http://localhost:5000/v1/swagger.json"}
```

Then in the data store repo, I ran:

```
./dss-api --host localhost
```

which ran a local Swagger API server on localhost on port 5000. Then I was able to build the documentation locally, and the `hca` utility assembled the API-based documentation from the local, modified version of the data store swagger YML instead of the dev stage YML.

Issue #2583, opened 2019-11-12
Dropped v1 notifications.


DailyDreaming:
After the dcp scale test on Nov 2nd and Nov 3rd, many notifications seemed to be dropped.  After speaking with Hannes, who originally brought up the issue and suggested (correctly) that the problem might be that the response received might show an error and then a success and never make the attempt to index, I investigated the logs and found that the indexing lambda successfully creates notifications after a 400 error response when writing the index:
`[WARNING] 2019-11-02T15:12:24.990Z efb000c5-b4f0-58ab-a454-56f22be7b731 GET https://search-dss-index-integration-vx3ftz6xjooc2vr3ayd7qeyh3m.us-east-1.es.amazonaws.com:443/dss-integration-aws-docs-alias/_all/19f0ac23-a9d5-4962-b4f9-88fbbbcb1c13.2019-11-02T150811.723046Z?_source=false&stored_fields= [status:400 request:0.018s]`

`[INFO] 2019-11-02T15:12:46.454Z efb000c5-b4f0-58ab-a454-56f22be7b731 Queueing asynchronous notification Notification(notification_id='91269a05-6122-4ae1-b2cd-146d1bec9847', subscription_id='076dd1ad-e269-4f17-b1b1-07845f34a3ef', url='https://indexer.dev.explore.data.humancellatlas.org/', method='POST', encoding='application/json', attempts=12, hmac_key_id='267eba9a-2ae9-46c1-849a-cf3de97598e9', correlation_id='19f0ac23-a9d5-4962-b4f9-88fbbbcb1c13.2019-11-02T150811.723046Z') for bundle 19f0ac23-a9d5-4962-b4f9-88fbbbcb1c13.2019-11-02T150811.723046Z`

But sometimes _it makes a 409_ error response, it sees the index entry already exists, and doesn't create a notification:

`[WARNING] 2019-11-02T15:12:08.407Z c51c337e-03e3-56d7-8ab4-65499ce52fcc PUT https://search-dss-index-integration-vx3ftz6xjooc2vr3ayd7qeyh3m.us-east-1.es.amazonaws.com:443/dss-integration-gcp-docs-6be44a3c0ba535a1e731a2d0fb42811915b72daf/doc/88006a33-fe39-4fea-8309-8cfe193185c6.2019-11-02T150811.726673Z?op_type=create [status:409 request:0.016s]`

[...]

PR #2584, opened 2019-11-12
[Easy] Set dss-indexer lambda concurrency to 200; increase timeout by 1 minute.

DailyDreaming:
Partially addresses: https://github.com/HumanCellAtlas/data-store/issues/2583

### DIFF
--- a/daemons/dss-index/.chalice/config.json
+++ b/daemons/dss-index/.chalice/config.json
@@ -10,7 +10,7 @@
     }
   },
   "version": "2.0",
-  "lambda_timeout": 300,
+  "lambda_timeout": 360,
   "lambda_memory_size": 1536,
-  "reserved_concurrent_executions": 400
+  "reserved_concurrent_executions": 200
 }
--- a/daemons/dss-index/.chalice/config.json
+++ b/daemons/dss-index/.chalice/config.json
@@ -11,5 +11,6 @@
   },
   "version": "2.0",
   "lambda_timeout": 300,
-  "lambda_memory_size": 1536
+  "lambda_memory_size": 1536,
+  "reserved_concurrent_executions": 400
 }
--- a/daemons/dss-notify/.chalice/config.json
+++ b/daemons/dss-notify/.chalice/config.json
@@ -10,6 +10,7 @@
     }
   },
   "version": "2.0",
-  "lambda_timeout": 300,
-  "lambda_memory_size": 1536
+  "lambda_timeout": 360,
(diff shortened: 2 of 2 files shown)


daemons/dss-notify/.chalice/config.json
```diff
@@ -10,6 +10,7 @@
     }
   },
   "version": "2.0",
-  "lambda_timeout": 300,
-  "lambda_memory_size": 1536
+  "lambda_timeout": 360,
```
xbrianh: Why has the TTL been increased?

DailyDreaming: The indexer's been timing out so I thought it might help.  I can leave it at 300 if you feel it's unnecessary though.

daemons/dss-index/.chalice/config.json
```diff
@@ -11,5 +11,6 @@
   },
   "version": "2.0",
   "lambda_timeout": 300,
-  "lambda_memory_size": 1536
+  "lambda_memory_size": 1536,
+  "reserved_concurrent_executions": 400
```
xbrianh: `reserved_concurrent_executions` may need to be scaled with the ES cluster size, which is different for prod.

hannes-ucsc:
Given the size of your ES instance, 400 is probably still too high. For this to be a better experiment with a stronger outcome, you may want to limit concurrency more aggressively to confirm that this indeed resolves the issue and then relax the limit gradually until the issue occurs again.

hannes-ucsc:
Also, PR title is misleading. This is about indexer lambda, not notifier lambda.

DailyDreaming:
@hannes-ucsc I've lowered the limit to 200 and changed the title.  There will be a subsequent PR enabling indexing even when 409 errors occur.

daemons/dss-index/.chalice/config.json
DailyDreaming: That makes sense and I would spend time optimizing the scale for both configs, but feel I think that this will work and is "good enough" for a lambda that we intend to deprecate soon.  Let me know though if you feel like it's worth investing more time!

hannes-ucsc:
I was more thinking 40, an order of magnitude lower or even less. I don't think this is a good experiment. It will still drop notifications and we'll still don't know what the actual concurrency limit should be. From my experience with SQS trigger, the # of worker lambdas consuming a queue is scaled up rather conservatively. I actually doubt you will see even 100 concurrent workers with a queue length of 10000 or less, rendering the 200 limit ineffective.

PR #2539, opened 2019-10-22
Fix incorrect test for string equality

natanlao:
For example, `Ctrl-F` / `Cmd-F` for this in [this Allspark log](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/52364):

```
$ export COMMITS_URL=${GITHUB_API}/repos/HumanCellAtlas/data-store/commits
$ if not [[ CI_COMMIT_SHA == $(http GET $COMMITS_URL sha==$CI_COMMIT_REF_NAME | jq -r '.[0]["sha"]') ]]; then exit 1; fi
/bin/bash: line 91: not: command not found
```

(I can't figure out how to link to a log line number in Gitlab?)

Issue #2546, opened 2019-10-24
Malformed condition in gitlab build


natanlao:
This line incorrectly checks for string equality:

https://github.com/HumanCellAtlas/data-store/blob/0a4dfda6a354d90f4edc8e05306e1cb31551a586/.gitlab-ci.yml#L22

It should be:

```
- if [[ CI_COMMIT_SHA != $(http GET $COMMITS_URL sha==$CI_COMMIT_REF_NAME | jq -r '.[0]["sha"]') ]]; then exit 1; fi
```

The line has been failing silently on GItlab. For example, `Ctrl-F` / `Cmd-F` for this in [this Allspark log](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/52364) (I can't figure out how to link to a log line number in Gitlab?):

```
$ export COMMITS_URL=${GITHUB_API}/repos/HumanCellAtlas/data-store/commits
$ if not [[ CI_COMMIT_SHA == $(http GET $COMMITS_URL sha==$CI_COMMIT_REF_NAME | jq -r '.[0]["sha"]') ]]; then exit 1; fi
/bin/bash: line 91: not: command not found
```

But when the line is corrected as above, then any build on a non-master branch fails:

1. The line in question requests a list of the latest commits on `master` from the GitHub API.
2. If the requested build is for the latest commit on `master` on GitHub, then the build will continue.
3. If the requested build is for any other commit, the build will `exit 1`. As such, branch builds would fail almost immediately.

Looking at the blame, it looks like the intent of this line might be to prevent only deploy (but not testing) when the Gitlab mirror is behind GitHub. If so, then I think the correct fix would be to move the (fixed) line to the `before_script` block in the `deploy` stage.

Is this assessment correct? @xbrianh

natanlao:
Closed by #2591.

PR #1423, opened 2018-07-13
Deploy dev on the  Build Box

xbrianh:
* Run tests on GitLab, and deploy to `dev` if they pass.
* Create a custom Docker container to resolve Elasticsearch needing to run as a non-root user, and other permission issues.
* Turn off travis `dev` deployment

PR #2585, opened 2019-11-12
Move check of whether allspark is behind github to deploy stage only

chmreid:
This PR addresses #2546 and attempts to fix PR #2539 (which was abandoned).

It fixes this string in `gitlab-ci.yml`:

https://github.com/HumanCellAtlas/data-store/blob/0a4dfda6a354d90f4edc8e05306e1cb31551a586/.gitlab-ci.yml#L22

The bash statement above does the following:
- `http GET $COMMITS_URL sha==$CI_COMMIT_REF_NAME` passes the name of the current branch (in the Gitlab CI/CD) to the Github API
- `jq -r '.[0]["sha"]` parses out the SHA hash of the _latest_ commit (on Github) made to the branch with that name
- finally, checking whether this is equal to `$CI_COMMIT_SHA` is checking whether the Allspark CI/CD system is testing the newest/freshest commit on this branch (this assumes Github is the source of truth and Gitlab is just the mirror)
- the `not` bit is invalid bash syntax, so this check always fails

This PR does the following:
- fix the invalid syntax
- move the check out of the `before_script` for all stages, and move it into the `before_script` for the `deploy` stage
- @natanlao already tried fixing the invalid bash syntax and leaving it in place (see PR #2539), which caused non-deploy tests to fail every time. That's why this PR moves it to the `before_script` of the deploy stage only.
- Currently this check fails due to a syntax error, so regardless of its original intent, it does nothing
- This proposed fix may not be in line with the original intent of the check, so please suggest alternative approaches if you think they are more in line with the original thinking

Useful:
- List of environment variables defined in Gitlab CI/CD tests, plus example values: https://docs.gitlab.com/ee/ci/variables/

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,7 +19,6 @@ stages:
 
 before_script:
   - export COMMITS_URL=${GITHUB_API}/repos/HumanCellAtlas/data-store/commits
-  - if not [[ CI_COMMIT_SHA == $(http GET $COMMITS_URL sha==$CI_COMMIT_REF_NAME | jq -r '.[0]["sha"]') ]]; then exit 1; fi
 # TODO: figure out how to get the gitlab-runner to not clone the repo as root - Brian H
   - virtualenv ~/venv
   - source ~/venv/bin/activate
@@ -97,6 +96,8 @@ setup_fusillade:
 
 deploy:
   stage: deploy
+  before_script:
+    - if [[ CI_COMMIT_SHA != $(http GET $COMMITS_URL sha==$CI_COMMIT_REF_NAME | jq -r '.[0]["sha"]') ]]; then exit 1; fi
   script:
     - make plan-infra
     - make deploy


.gitlab-ci.yml
```diff
@@ -97,6 +96,8 @@ setup_fusillade:
 
 deploy:
   stage: deploy
+  before_script:
```
DailyDreaming: Might be worthwhile to move: `export COMMITS_URL=${GITHUB_API}/repos/HumanCellAtlas/data-store/commits` here too since it's only called for this function.

natanlao:
This is a little long, since I am not convinced I fully understand the original intent of the check. I might be totally off base here.

> This proposed fix may not be in line with the original intent of the check, so please suggest alternative approaches if you think they are more in line with the original thinking

Earlier, I thought it would make sense to keep the check intact - if there is a newer commit on GitHub, then it would make sense to skip deploy if Gitlab would be pulling in a new commit and deploying Very Very Soon anyway. Thinking about it more, though, I think it makes sense for this logic to be removed entirely because it could potentially obscure a deployment failure.

Take this hypothetical case:

* A commit `foo` is merged into `master` on Github.
* Gitlab syncs and picks up `foo`, kicking off the unit testing -> deploy -> integration testing sequence.
* Before Gitlab begins the deploy stage for `foo`, a commit `bar` is merged into `master` on GitHub.
* Gitlab unit tests on `foo` pass. Gitlab starts the deploy stage for `foo`, sees `bar` on GitHub, and aborts the deploy. Integration tests don't run.
* Gitlab syncs and picks up `bar`, kicking off testing / deploy of `bar`.
* Gitlab begins the deploy stage for `bar`. It fails.

In this case, because deploy was not run for `foo`, the deficiency could be present in either `foo` or `bar`. There could also be some integration test failure in `foo` not detected because it was not deployed. Deploy is cheap enough that I think it makes sense to run it even if there is a newer commit on Github.

chmreid:
For archaeological purposes, the PR that added this line to the .gitlab-ci.yml file is https://github.com/HumanCellAtlas/data-store/pull/1423

chmreid:
Commit message on 8b0b44c "Build-box runs tests on all branches, deploys dev" indicates this line was part of configuring the Allspark deployment process, so in that context, it appears to be making sure that deployment only happens when Allspark and Github are synced.

@natanlao makes a great point about a scenario where integration tests could be skipped for one or more commits, and makes another great point about the fact that deploy is cheap.

This check is also in `.gitlab-ci-prod.yml`, where it makes more sense - by the time changes make it to prod, probability of failures or branches getting out of sync between Github and Allspark before tests run should be very low or zero.

chmreid:
Opening #2591 as an alternative approach that only keeps/fixes this line in the prod CI/CD file.

PR #2586, opened 2019-11-12
Include error message in api logs

xbrianh:
This includes error information along with the stacktrace in API server logs. Previously only the stacktrace was logged, which does not perform string interpolation.

### DIFF
--- a/dss/error.py
+++ b/dss/error.py
@@ -75,20 +75,18 @@ def dss_handler(func):
                 code = ex.name
                 title = str(ex)
                 stacktrace = traceback.format_exc()
-                headers = None
             except DSSException as ex:
                 status = ex.status
                 code = ex.code
                 title = ex.message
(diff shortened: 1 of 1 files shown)


dss/error.py
```diff
@@ -75,20 +75,18 @@ def wrapper(*args, **kwargs):
...
-                headers = None
             except Exception as ex:
                 status = requests.codes.server_error
                 code = "unhandled_exception"
                 title = str(ex)
                 stacktrace = traceback.format_exc()
-                headers = None
-            logger.error(stacktrace)
+            headers = None
+            logger.error(json.dumps(dict(status=status, code=code, title=title, stacktrace=stacktrace)))
```
amarjandu:
```suggestion
            logger.error(json.dumps(dict(status=status, code=code, title=title, stacktrace=stacktrace), indent=4))
```

PR #2589, opened 2019-11-13
DSS GitLab: TimeStamps

amarjandu:
adds timestamps to gitlab logs, assists with pinpointing which cloudwatch log to inspect.


amarjandu:
We might want to permanently set the time in the `dss-buildbox` to PST. 






west-side

.gitlab-ci.yml
```diff
@@ -77,6 +78,7 @@ test_subscriptions:
 setup_fusillade:
   stage: fusillade
   script:
+    - date && date -u
```
xbrianh: Is there a reason to call `date` again here?

.gitlab-ci.yml
```diff
@@ -87,7 +89,13 @@ setup_fusillade:
...
     - fi
     - cat ./roles.json | sed "s/\${stage}/${DSS_DEPLOYMENT_STAGE}/g" > temp-roles.json
-    - dcp-fusillade/scripts/setup_fusillade.py --file temp-roles.json --force testing
+    - if [ "$DSS_DEPLOYMENT_STAGE" == "dev" ]; then
+    -   FUS_STAGE='testing'
+    - else
+    -   FUS_STAGE=$DSS_DEPLOYMENT_STAGE
+    - fi
+    - echo $FUS_STAGE
+    - dcp-fusillade/scripts/setup_fusillade.py --file temp-roles.json --force $FUS_STAGE
```
xbrianh: This block looks like it could be handled in a separate PR since it's unrelated to build timestamps.

.gitlab-ci.yml
amarjandu: I would like to have both local timezone, and UTC because our logging system records events in UTC.

.gitlab-ci.yml
amarjandu: Would you prefer a change of description, or a separate PR? @xbrianh 
I suppose the overall PR is to provide the operator more context around the gitlab job that is running, rather than the current Title -- `Timestamps`

.gitlab-ci.yml
xbrianh: The command `date && date -u` already called in before_script on line 21. Why call it again here?

amarjandu: Ah! i see what you mean; removing.

.gitlab-ci.yml
amarjandu: Ah i see what you mean, this was from a different PR that was merged into master already, correct this by re-basing.

PR #2596, opened 2019-11-13
Remove tombstoned bundles from event journals

xbrianh:


dss/events/__init__.py
```diff
@@ -46,9 +46,6 @@ def get_deleted_bundle_metadata_document(replica: Replica, key: str) -> dict:
         "version": version,
     }
 
-# TODO: Delete event from flashflood
-# TODO: Update event data in flashflood
```
chmreid: Just to check - is this TODO item finished/no longer needed?

xbrianh: Yep: it turned out that updates were not needed.

PR #2600, opened 2019-11-14
Confirm flashflood writes during journaling/update

xbrianh:
This configures flashflood to wait for s3 consistency during event journaling and updates, avoiding the possibility that the previous batch of journals/updates are unavailable for read before the next batch is
processed.

### DIFF
--- a/dss/config.py
+++ b/dss/config.py
@@ -8,6 +8,7 @@ from enum import Enum, EnumMeta, auto
 
 import boto3
 import botocore.config
+import flashflood
 from cloud_blobstore import BlobStore
 from cloud_blobstore.s3 import S3BlobStore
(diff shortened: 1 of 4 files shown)


dss/config.py
```diff
@@ -204,6 +206,15 @@ def get_blobstore_handle(replica: "Replica") -> BlobStore:
     def get_hcablobstore_handle(replica: "Replica") -> HCABlobStore:
         return replica.hcablobstore_class(Config.get_blobstore_handle(replica))
 
+    @staticmethod
+    @functools.lru_cache()
+    def get_flashflood_handle(prefix: str, confirm_writes: bool=False) -> flashflood.FlashFlood:
```
chmreid: This is a good pattern  👍  

(makes changing the constructor method signature easier too)

PR #2604, opened 2019-11-15
Include `bundle_info` in bundle metadata document

### LINKED ISSUE #2513: v2 subscriptions don't allow filtering by bundle UUID prefix
This ~was~ is possible in v1 and we need it for Azul.

The problem is that neither the bundle manifest nor the overarching ["bundle metadata document"](https://github.com/HumanCellAtlas/data-store/blob/23d38f7ded20f0581ee822cecaeb0a69006093a2/dss/events/__init__.py#L109) contain the actual bundle UUID.

xbrianh:
This adds `bundle_info` to the bundle metadata document, making it possible to filter JMESPath subscriptions on both bundle uuid and version.
```
{
  "event_type: ...
  "bundle_info" = {
    'uuid': ...,
    'version': ...
  },
  ...
}
```

closes #2513

### DIFF
--- a/daemons/dss-notify-v2/Readme.md
+++ b/daemons/dss-notify-v2/Readme.md
@@ -80,6 +80,10 @@ The bundle metadata document format for a new bundle or version is
 ```
 {
   "event_type": "CREATE",
+  "bundle_info": {
+    "uuid": "48b7bdf2-410a-4d56-969c-e42e91d1f1fe",
+    "version": "2019-02-12T224603.042173Z"
+  },
   "manifest": {
     "version": ...,
     "files": [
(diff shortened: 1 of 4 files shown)


daemons/dss-notify-v2/Readme.md
```diff
@@ -80,6 +80,10 @@ The bundle metadata document format for a new bundle or version is
 ```
 {
   "event_type": "CREATE",
```
amarjandu: Do you have to specify an event_type in order to get into bundle info?Would it be possible to not specify any event_type and get **any** event_type for some bundle_uuid?

xbrianh: I think you're asking about JMESPath filtering. Yes, it would be possible to filter on bundle_info.uuid without filtering on event_type.

PR #2607, opened 2019-11-18
Refactor notify v2 event handler

xbrianh:
This:
- verifies the existence of normal bundle keys and tombstoned bundle keys before attempting to build or retrieve the bundle metadata document.
- tests the refactored handler method

### DIFF
--- a/daemons/dss-notify-v2/app.py
+++ b/daemons/dss-notify-v2/app.py
@@ -110,24 +110,31 @@ def launch_from_notification_queue(event, context):
 def _handle_bucket_event(replica: Replica, key: str, is_delete_event: bool):
     if is_delete_event:
         metadata_document = get_deleted_bundle_metadata_document(replica, key)
-    elif key.endswith(TOMBSTONE_SUFFIX):
-        for zombie_key in get_tombstoned_bundles(replica, key):
-            delete_event_for_bundle(replica, zombie_key)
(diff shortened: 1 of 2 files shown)


daemons/dss-notify-v2/app.py
```diff
@@ -110,24 +110,31 @@ def launch_from_notification_queue(event, context):
...
             notify_or_queue(replica, subscription, metadata_document, key)
 
     # TODO: Consider scaling parallelization with Lambda size
-    logger.info(f"Attempting notifications for; replica: {replica}, key: {key} delete: {is_delete_event}")
-    with ThreadPoolExecutor(max_workers=20) as e:
+    logger.info(json.dumps(dict(message="Attempting notifications",
+                                replica=replica.name,
+                                key=key,
+                                event_type=metadata_document['event_type']), indent=4))
+    with ThreadPoolExecutor(max_workers=6) as e:
```
DailyDreaming: Why was `max_workers` decreased?

xbrianh: You know what, I don't have good justification for this change, so I'll revert it.

Notifications are probably either delivered immediately or doing nothing while waiting for timeouts. Perhaps this justifies a  higher level of parallelization than for more computationally expensive tasks (where previous experience indicates a lower count for max_worker).

Issue #2528, opened 2019-10-17
Tombstone Spatio-temporal kidney project on 4 envs


jahilton:
We have discovered GDPR issues for some of the data in a project. Solution is to wipe out the whole project ASAP and re-ingest only the data that is approved.

project_id: abe1a013-af7a-45ed-8c26-f3793c24a1f4

Needs to be wiped from all 4:
- [ ] prod
- [ ] dev
- [ ] integration
- [ ] staging

jahilton:
```
prod   bundle_uuid--bundle_version
--
f7081630-0d97-4550-8b16-77008ea868a3--2019-10-03T124443.721375Z
a0a78a4e-7891-4ef2-a22d-1bfab43ffe47--2019-10-03T124443.730326Z
3c3061a7-1dc5-422e-bc14-4760b47f6ad2--2019-10-03T124443.724307Z
e22e7f03-4917-4ab9-ba0c-31784fa62de9--2019-10-03T124443.700068Z
6086e0e1-c679-44ad-bc62-57875b216fc4--2019-10-03T124443.726155Z
f838b9fe-3e03-41cf-8e1f-1f17db456975--2019-10-03T124443.708244Z
9914ac5e-772f-4347-ae33-290bd67ddf11--2019-10-03T124443.735694Z
499b690e-feee-4675-8531-882a62b6020c--2019-10-03T124443.749980Z
2e373a6e-f531-449b-a989-eb163322ce83--2019-10-03T124443.751966Z
a36a76dd-3036-42e6-beb2-dd41e1e312b6--2019-10-03T124443.745589Z
d8f7f78d-1c77-49cc-b934-28d0da11c6d0--2019-10-03T124443.728149Z
1076c830-8986-4fee-9e7a-847fe0b227c7--2019-10-03T124443.734618Z
7b4fd692-c38d-43c6-bb5b-bf3dcbc837eb--2019-10-03T124443.768440Z
5c4b214d-ffd7-426b-88a2-cf611e5803f7--2019-10-03T124443.778799Z
fff16441-5212-4319-bf52-9e23d0eff7eb--2019-10-03T124443.766555Z
8b7635a4-88a4-4c76-9946-6e3bebd51bfa--2019-10-03T124443.718000Z
5a37f51b-3f66-4b22-95c1-d82842265f19--2019-10-03T124443.709143Z
10ae73ef-d2e1-4a1d-a192-cf332d850c4f--2019-10-03T124443.775778Z
[...]

PR #2611, opened 2019-11-19
DSS Operations: Bundle Reference List

amarjandu:
Allows reference lists to be built for bundle key

Enables us to build object references for tombstoned bundles

Connected to #2528

### DIFF
--- a/dss/operations/storage.py
+++ b/dss/operations/storage.py
@@ -225,7 +225,7 @@ class build_reference_list(StorageOperationHandler):
             storage_object_references.append(f"files/{files['uuid']}.{files['version']}")
             storage_object_references.append(compose_blob_key(files))
         pprint.pprint(storage_object_references)
-        return storage_object_references # returned for testing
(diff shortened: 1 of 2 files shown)


dss/operations/storage.py
```diff
@@ -206,6 +207,27 @@ def process_key(self, key):
...
+        storage_object_references = list()
+        storage_object_references.append(key)
+        try:
+            bundle_metadata = json.loads(self.handle.get(self.replica.bucket, key))
+        except BlobNotFoundError as e:
+            logger.error(f"bundle {key} not found {e}")
+        for files in bundle_metadata['files']:
+            storage_object_references.append(f"files/{files['uuid']}.{files['version']}")
+            storage_object_references.append(compose_blob_key(files))
+        pprint.pprint(storage_object_references)
```
amarjandu: @xbrianh is there a better way to print values out of the operation tooling?

PR #2628, opened 2019-11-25
Xbrianh update event daemon config

xbrianh:
- Increase the journaling daemon invocation schedule to 10 minutes (Journaling 1000 events with flashflood write verification enabled takes ~8 minutes.)
- Use a more descriptive name for handler method (this method name is used to construct the AWS CloudWatch rule name. Rule names should be descriptive of their source to increase discoverability.)
- Use a skeleton daemon configuration file instead of the version populated during deploy.

### DIFF
--- a/daemons/dss-events-scribe/.chalice/config.json
+++ b/daemons/dss-events-scribe/.chalice/config.json
@@ -7,57 +7,11 @@
   "stages": {
     "dev": {
       "api_gateway_stage": "dev",
-      "tags": {
-        "DSS_DEPLOY_ORIGIN": "dcp-b087432b5657-2019-10-01-15-33-51-integration.release-46-g6e447df-2019-11-04-17-30-43.deploy",
-        "Name": "dss-events-scribe",
-        "service": "dss",
-        "project": "dcp",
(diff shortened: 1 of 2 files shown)


daemons/dss-events-scribe/.chalice/config.json
```diff
@@ -7,57 +7,11 @@
...
-      "layers": [
-        "arn:aws:lambda:us-east-1:861229788715:layer:dss-dependencies-dev:45"
-      ]
+      "tags": {},
+      "environment_variables": {}
     }
   },
   "reserved_concurrency": 1,
-  "lambda_timeout": 300,
+  "lambda_timeout": 600,
```
amarjandu: if the lambda is processing events, and the lambda gets a timeout due to long duration, what happens to the events that it was working on? Is there a potential for malformed journals?

xbrianh: Yes, interrupting the journaling process will produce malformed journals. Currently, malformed journals will need to be created from the main bucket.

It is possible to build recovery capabilities into flashflood, which should be done in the future.

PR #2631, opened 2019-11-26
Use stage in dss-events-scribe CloudWatch rule name

xbrianh:
Unless the rule name is parameterized by stage, all stages will share the same rule.

### DIFF
--- a/daemons/dss-events-scribe/app.py
+++ b/daemons/dss-events-scribe/app.py
@@ -24,6 +24,7 @@ from dss.events import journal_flashflood, update_flashflood
 configure_lambda_logging()
 logger = logging.getLogger(__name__)
 dss.Config.set_config(dss.BucketConfig.NORMAL)
+stage = os.environ['DSS_DEPLOYMENT_STAGE']
 
 app = domovoi.Domovoi()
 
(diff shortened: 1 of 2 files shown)


daemons/dss-events-scribe/app.py
```diff
@@ -37,10 +38,10 @@ def __init__(self, prefix):
         self.prefix = prefix
 
 
-@app.scheduled_function("rate(10 minutes)")
-def dss_events_scribe_journal_and_update(event, context):
+@app.scheduled_function("rate(10 minutes)", rule_name=f"dss-events-scribe-journal-and-update-{stage}")
```
amarjandu: you might have to cleanup the old ones, so its not cluttered. :)

daemons/dss-events-scribe/app.py
```diff
@@ -37,10 +38,10 @@ def __init__(self, prefix):
...
+def journal_and_update(event, context):
     # TODO: Make this configurable
-    minimum_number_of_events = 10 if "dev" == os.environ['DSS_DEPLOYMENT_STAGE'] else 1000
+    minimum_number_of_events = 10 if "dev" == stage else 1000
```
chmreid: Just to clarify what is happening here, at a high level. Does `minimum_number_of_events` refer to the minimum number of events required to start writing a new journal? What is the reason for setting it to 10 for staging and 1000 for other stages?

If I look at the `journal_flashflood()` method, it seems to be listing all flashflood journals, and combining/returning the combined journal when the number of journals is equal to the minimum number of events. I am pretty confused about what's going on.

What happens if the number of events is less than the minimum number of events, does it skip creation of a journal altogether? What if the number of events is larger than the minimum number of events, does it keep creating new journals?

xbrianh: `journal_flashflood` either combines N new journals and returns a value, or lists fewer journals than N journals and returns `None`. Here, N = minimum_number_of_events. 

`minimum_number_of_events` is set to a smaller value for dev for a couple reasons:
  - Increase debugging opportunity
  - The dev bucket is wiped nightly, and rarely contains 1000 bundles

xbrianh:
Closed in favor of #2648

Issue #2643, opened 2019-12-02
GET /bundles/all ignores invalid prefix


jessebrennan:
```
jesse@vader ~$ http -b get "https://dss.data.humancellatlas.org/v1/bundles/all?replica=aws&prefix=F"
{
    "bundles": [],
    "dss_api": "https://dss.data.humancellatlas.org",
    "event_timestamp": "2019-12-02T214901.570541Z",
    "has_more": false,
    "object": "list",
    "page_count": 0,
    "per_page": 100,
    "search_prefix": "bundles/F"
}
```
but
```
jesse@vader ~$ http -b get "https://dss.data.humancellatlas.org/v1/bundles/all?replica=aws&prefix=f"
{
    "bundles": [
...
    ],
    "dss_api": "https://dss.data.humancellatlas.org",
    "event_timestamp": "2019-12-02T214929.423574Z",
    "has_more": true,
    "link": "https://dss.data.humancellatlas.org/v1/bundles/all?prefix=f&replica=aws&search_after=bundles%2Ff02cae0f-7cf3-474e-aa8b-cf0e09de7518.2019-08-01T200147.791889Z&token=1T0Nj%2FkyWssgaGbl%2FKM%2FfIrVMypR6Szwy3i0q5zkHckkrsAXMh2KJ92k2k9EtIHIbvNYLXeUNiiVc17aEj%2FE%2BUv%2F1tgZU9ST5wI4jrMEIC0x15e8PfDaF2VAHSrHqnA6pcUAx%2BgVt3Pw%3D",
    "object": "list",
    "page_count": 100,
    "per_page": 100,
    "search_after": "bundles/f02cae0f-7cf3-474e-aa8b-cf0e09de7518.2019-08-01T200147.791889Z",
    "search_prefix": "bundles/f",
    "token": "1T0Nj/kyWssgaGbl/KM/fIrVMypR6Szwy3i0q5zkHckkrsAXMh2KJ92k2k9EtIHIbvNYLXeUNiiVc17aEj/E+Uv/1tgZU9ST5wI4jrMEIC0x15e8PfDaF2VAHSrHqnA6pcUAx+gVt3Pw="
}

```
This isn't necessarily a bug, but I think there is room for improvement. Either
- the OpenAPI spec should warn that invalid prefixes will be ignored, or
- there should be some validation on the query parameter that lets the user know if the prefix is not one of the expected values ([a-f0-9][a-f0-9\\-]*).

[...]

PR #2649, opened 2019-12-02
[easy] Lowercase uuid prefixes for GET /bundles/all.

DailyDreaming:
When calling GET /bundles/all with the prefix `bundles/f` and `bundles/F`, the former gives search results and the latter does not.

Since we lowercase uuids for users when they submit them, we should also probably lowercase them when that user searches for that same uuid (or uuid prefix).

See https://github.com/HumanCellAtlas/data-store/issues/2643

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1486,7 +1486,10 @@ paths:
           enum: [aws, gcp]
         - name: prefix
           in: query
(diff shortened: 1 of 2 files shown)


DailyDreaming:
@amarjandu Done, good catch.

PR #2652, opened 2019-12-03
Remove check that errors on hot patches.

DailyDreaming:


.gitlab-ci-prod.yml
```diff
@@ -16,7 +16,6 @@ stages:
 before_script:
   - date && date -u
   - export COMMITS_URL=${GITHUB_API}/repos/HumanCellAtlas/data-store/commits
```
amarjandu: can line 18 also be removed?

DailyDreaming: Yar, good catch, thanks!

PR #2654, opened 2019-12-03
Define cloudwatch rule and SQS queue

xbrianh:
This defines infrastructure that will be used for the forthcoming refactor of the DSS event journaling and update daemon:
- A scheduled CloudWatch event rule
- An SQS queue
- CloudWatch-SQS integration

connected to #2648

Issue #2648, opened 2019-12-02
Refactor Event Journaling and Update Daemon


xbrianh:
The current schedule-drive invocation of the event journaling and update daemon is inelastic with respect to lambda TTLs. If the TTL is set too short, the journaling operation may be interrupted, possibly producing corrupted event journals. On the other hand, long TTLs may impact the performance of event playback.

An alternative design is lambda invocation via SQS, where messages are enqueued on a schedule, with lambda reserved concurrency set to 1.

PR #2655, opened 2019-12-03
Refactor the event journaling and update daemon dss-events-scribe

xbrianh:
This refactors the event journaling and update daemon using a more elastic rate-limiting algorithm.

depends on #2654 
connected to #2648

### DIFF
--- a/daemons/dss-events-scribe/app.py
+++ b/daemons/dss-events-scribe/app.py
@@ -1,14 +1,7 @@
-"""
-This is the "offline" event journaling and update daemon. The compilation of flashflood journals, and the application
-of flashflood event updates and event deletes, is not concurrency safe. This daemon should be executed on a schedule,
-and Lambda TTL should be configured such that it is never executed in a concurrent manner. Additionally,
-`reserved_concurrency` has been set to 1 to prevent parallel execution
-(see `dss/daemons/dss-events-scribe/.chalice/config.json`).
-"""
 import os
 import sys
+import json
 import logging
-from itertools import cycle
 
 import domovoi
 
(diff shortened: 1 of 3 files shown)


daemons/dss-events-scribe/app.py
```diff
@@ -28,48 +21,43 @@
 app = domovoi.Domovoi()
 
 
-class ReplicaStatus:
-    prefix = None
-    finished_journaling = False
-    finished_updating = False
+# TODO: Make this configurable
+NUMBER_OF_EVENTS_PER_JOURNAL = 10 if "dev" == os.environ['DSS_DEPLOYMENT_STAGE'] else 1000
```
DailyDreaming: If `dev` is different, could this potentially cause unexpected behavior when releasing to `integration`?

amarjandu: I think it has to do with the amount of activity on integration, but i could be wrong.....

xbrianh: I don't think so. `NUMBER_OF_EVENTS_PER_JOURNAL` is set dynamically at runtime.

daemons/dss-events-scribe/app.py
```diff
@@ -28,48 +21,43 @@
...
 
-    def __init__(self, prefix):
-        self.prefix = prefix
 
+@app.sqs_queue_subscriber("dss-events-scribe-" + os.environ['DSS_DEPLOYMENT_STAGE'],
+                          batch_size=1,
+                          queue_attributes=dict(VisibilityTimeout="600"))
+def handle_sqs_message(event, context):
+    if 1 != len(event['Records']):
+        logger.error(f"Received {len(event['Records'])}, expected 1")
```
amarjandu: if the batch_size changes will there be an error? Might want to throw in a comment to that these values can be changed from the lambda console, under lambda triggers

xbrianh: This check ensures there is an error if `batch_size != 1`. I'll add a comment.

PR #2656, opened 2019-12-03
Documentation for the refactored dss-events-scribe

### LINKED ISSUE #2648: Refactor Event Journaling and Update Daemon

xbrianh:
depends on #2654, #2655 
closes #2648

### DIFF
--- a/daemons/dss-events-scribe/Readme.md
+++ b/daemons/dss-events-scribe/Readme.md
@@ -1,0 +1,37 @@
+# DSS Event Journaling and Update Daemon
+
+The dss-events-scribe daemon compiles event data into journals, and applies event updates and deletes to existing
+journals.
+
+DSS events are managed with the [flashflood library](https://github.com/HumanCellAtlas/flash-flood).
+
+## Concurrency
+
+The flashflood event journaling and update API is not concurrency safe, so this daemon uses a synchronous
+execution model by setting Lambda reserved concurrency to 1. For more information about Lambda reserved concurrency,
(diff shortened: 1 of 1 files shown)


daemons/dss-events-scribe/Readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Concurrency
+
+The flashflood event journaling and update API is not concurrency safe, so this daemon uses a synchronous
+execution model by setting Lambda reserved concurrency to 1. For more information about Lambda reserved concurrency,
+see [AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/per-function-concurrency.html).
+
+## Rate limiting
+
+Daemon invication is rate limited similarly to the [token bucket algorithm](https://en.wikipedia.org/wiki/Token_bucket) as follows:
```
chmreid:
```suggestion
Daemon invocation is rate limited similarly to the [token bucket algorithm](https://en.wikipedia.org/wiki/Token_bucket) as follows:
```

chmreid:
A bit more clarification around what cloudwatch does would be useful  - from the terraform files, I see cloudwatch is configured with permission to create queue items, and scheduled to run every 10 minutes, but it's not clear what it is actually doing.

PR #2664, opened 2019-12-03
[easy] Add tests for GET /bundles/all prefixes and syntax checking.

DailyDreaming:
After speaking with @jessebrennan , this further checks syntax and adds tests for prefixes supplied to `GET /bundles/all`.

### DIFF
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1488,9 +1488,12 @@ paths:
           in: query
           description: >
             Used to specify the beginning of a particular bundle UUID.  Capitalized letters will be lower-cased as is
-            done when users submit a uuid (all uuids have lower-cased letters upon ingestion into the dss).
-            Specifying a character(s) would provide all avaliable bundles starting with that character(s).
+            done when users submit a uuid (all uuids have lower-cased letters upon ingestion into the dss).  Characters
+            other than letters, numbers, and dashes are not allowed and will error.
+
+            The specified character(s) will return all available bundle uuids starting with that character(s).
           required: false
+          pattern: '^[a-zA-Z0-9-]+$'
           type: string
         - name: token
           in: query
--- a/dss-api.yml
+++ b/dss-api.yml
@@ -1493,7 +1493,7 @@ paths:
 
             The specified character(s) will return all available bundle uuids starting with that character(s).
           required: false
-          pattern: '^[a-zA-Z0-9-]+$'
+          pattern: '^[a-zA-Z0-9][a-zA-Z0-9-]*$'
           type: string
         - name: token
           in: query
--- a/dss/api/bundles/__init__.py
+++ b/dss/api/bundles/__init__.py
@@ -152,9 +152,6 @@ def enumerate(replica: str,
     :param search_after: used to page searches, should not be set by the user.
     """
     if prefix:
-        if any(c.isupper() for c in prefix):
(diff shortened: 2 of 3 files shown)


tests/test_bundle.py
```diff
@@ -1011,6 +1011,28 @@ def test_enumeration_bundles(self):
...
+
+        with self.subTest('Test successful case finds full uuid prefix.'):
+            res = self.app.get(f"/v1/bundles/all", params=dict(replica="aws", per_page=10, prefix=bundle_uuid))
+            self.assertEquals(res.json()['bundles'][0]['uuid'], bundle_uuid)
+            self.assertIn(res.json()['bundles'][0]['uuid'], res.json()['search_prefix'])
+            self.assertEquals(res.status_code, requests.codes.okay)
+
+        with self.subTest('Test successful case finds partial uuid prefix.'):
+            partial_uuid = bundle_uuid[:8]
+            res = self.app.get(f"/v1/bundles/all", params=dict(replica="aws", per_page=10, prefix=partial_uuid))
```
chmreid: This should test both uppercase prefix and lowercase prefix scenarios

dss/api/bundles/__init__.py
```diff
@@ -152,6 +152,9 @@ def enumerate(replica: str,
     :param search_after: used to page searches, should not be set by the user.
     """
     if prefix:
+        if any(c.isupper() for c in prefix):
+            sys.stderr.write('Note: prefix contains capital letters!  All data-store UUIDs are lower-cased (even if '
```
chmreid: Would it be better to use `logger.warning()` or `logger.error()` here? 

Also, AFAICT this error message never makes it to the user, it happens on the API server - so is this warning actually necessary?

DailyDreaming: That's a good point, should probably be added to `dcp-cli` instead.

tests/test_bundle.py
DailyDreaming: Will add.

DailyDreaming: Done.

dss/api/bundles/__init__.py
DailyDreaming: There isn't an easy way to add it to the CLI, and I think the description of the argument is explicit enough so I'm tabling this unless I get feedback that it's a must-have in the CLI.

dss-api.yml
```diff
@@ -1488,9 +1488,12 @@ paths:
...
           description: >
             Used to specify the beginning of a particular bundle UUID.  Capitalized letters will be lower-cased as is
-            done when users submit a uuid (all uuids have lower-cased letters upon ingestion into the dss).
-            Specifying a character(s) would provide all avaliable bundles starting with that character(s).
+            done when users submit a uuid (all uuids have lower-cased letters upon ingestion into the dss).  Characters
+            other than letters, numbers, and dashes are not allowed and will error.
+
+            The specified character(s) will return all available bundle uuids starting with that character(s).
           required: false
+          pattern: '^[a-zA-Z0-9-]+$'
```
amarjandu: `.` are also in fqids

jessebrennan: `-` will never occur as the first character, so `^[a-zA-Z0-9][a-zA-Z0-9-]*$` is slightly more correct.

DailyDreaming:
@jessebrennan Updated.

PR #2666, opened 2019-12-03
Eventually see dynamodb result for subscription owner.

DailyDreaming:
These pass locally for me so I'd like to see if this fixes our recent test failures:
https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/58183
https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/data-store/-/jobs/58162

### DIFF
--- a/tests/test_subscriptions.py
+++ b/tests/test_subscriptions.py
@@ -100,7 +100,7 @@ class TestSubscriptionsBase(ElasticsearchTestCase, TestAuthMixin, DSSAssertMixin
 
     @eventually(timeout=5, interval=1, errors={AssertionError})
     def assert_owner_subscription_count(self, owner, replica, subscription_count):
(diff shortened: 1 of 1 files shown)


tests/test_subscriptions.py
```diff
@@ -98,6 +98,10 @@ def test_put(self):
         finally:
             self._cleanup_subscription(uuid_)
 
+    @eventually(timeout=5, interval=1, errors={AssertionError})
+    def assert_owner_subscription_count(self, owner, replica, subscription_count):
+        assert count_subscriptions_for_owner(Replica[replica], owner) == subscription_count
```
chmreid:
```suggestion
        self.assertEqual(count_subscriptions_for_owner(Replica[replica], owner), subscription_count)
```

chmreid: This will improve output in the case of failure

Issue #2634, opened 2019-11-27
DSS Infra: Remove TF Warnings  


amarjandu:
`make deploy-infra` shows this warning message multiple times (15+)

```
Warning: Interpolation-only expressions are deprecated

  on s3.tf line 156, in resource "aws_s3_bucket" "dss_s3_events_bucket":
 156:   bucket = "${var.DSS_FLASHFLOOD_BUCKET}"

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.


Warning: Interpolation-only expressions are deprecated

  on s3.tf line 174, in resource "aws_s3_bucket" "dss_s3_events_bucket":
 174:   tags = "${merge(local.common_tags, local.aws_tags)}"

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.
```
We should remove them now before their support get dropped.

DailyDreaming:
Duplicate of #2671.

PR #2670, opened 2019-12-04
Update terraform files to address deprecation warnings.

DailyDreaming:
Terraform generates 100 billion warnings since we updated to `0.12.16`.  This should address them and bring our scripts up to date.

Issue: #2671


amarjandu:
Please connect to #2634

DailyDreaming:
Ah, missed #2634, thanks for pointing that out Amar!

PR #2675, opened 2019-12-04
Migrate check_deployment_secrets.py script to dss-ops.py script

### LINKED ISSUE #2380: DSS operations: Migrate check_deployment_secrets.py into dss-ops.py

### LINKED ISSUE #2668: Using test mixins disables integration tests
## Software version (git hash, date, branch)

Commit hash eee134a

Master branch

12/3/2019

## Expected behavior

I expect that if a test is marked as an integration test using the `@testmode.integration` decorator, and the environment variable `DSS_TEST_MODE="integration"`, the integration test will be run.

I also expect that if there are limitations on when `@testmode.integration` can be used (i.e., certain scenarios where the integration test will never be run), it would be documented somewhere, most likely in the file `tests/infra/testmode.py` that defines the integration decorator.

## Actual behavior

If I write a unit test and decorate it with `@testmode.integration`, than let it inherit from all four `TestBundleApi, TestAuthMixin, DSSAssertMixin, DSSUploadMixin` mixin classes, the integration test does not run.

**EDIT:** the problem is only with the `TestBundleApi` class, the other mixins do not break integration tests that inherit from them.

If I define an equivalent test class that does not inherit from any of those four mixin classes, the integration test runs.

## Steps to reproduce the behavior

See branch `chmreid-mixins-break-integration-tests` and associated pull request for an example of a test that reproduces this problem.

To run the integration test that works (no mixins), run `TestFoobar1`:

```
$ DSS_TEST_MODE="integration" python tests/test_foobar.py TestFoobar1
ohai foobar 1 integration test ran okay
.
Ran 1 test in 0.000s

OK
```

To run the integration test that fails (because of mixins), run `TestFoobar2`:

```
$ DSS_TEST_MODE="integration" python tests/test_foobar.py TestFoobar2
ssssssssssssssssssssssss
Ran 24 tests in 0.000s

OK (skipped=24)
```

There are also 4 sub-tests that isolate each parent class to see which one breaks the integration tests. It turns out it is `TestBundleApi` class that breaks integration tests, as it is the class associated with `TestFoobar2A`.

```
$ DSS_TEST_MODE="integration" python tests/test_foobar.py TestFoobar2A
ssssssssssssssssssssssss
Ran 24 tests in 0.000s

OK (skipped=24)

$ DSS_TEST_MODE="integration" python tests/test_foobar.py TestFoobar2B
ohai foobar 2B integration test ran okay
.
Ran 1 test in 0.000s

OK

$ DSS_TEST_MODE="integration" python tests/test_foobar.py TestFoobar2C
ohai foobar 2C integration test ran okay
.
Ran 1 test in 0.000s

OK

$ DSS_TEST_MODE="integration" python tests/test_foobar.py TestFoobar2D
ohai foobar 2D integration test ran okay
.
Ran 1 test in 0.000s

OK
```

## Further notes

* Based on the behavior noted above, the TestOperationsIntegration class in `tests/test_operations.py` will never run, because it is both an integration test and inherits from the mixin classes specified above.

* The order of mixin classes `TestBundleApi, TestAuthMixin, DSSAssertMixin, DSSUploadMixin` can be simplified - `TestAuthMixin` inherits from `DSSAssertMixin`

chmreid:
This pull request migrates the `check_deployment_secrets.py` script (which basically runs `make -C infra` and then compares the resulting (secret) environment variable values with the values stored in the AWS secrets manager.

### How to run

The check secrets script can be run as follows:

```
./scripts/dss-ops.py secrets check
```

This will run `make -C infra` from the root of the repo (and therefore requires that terraform be installed), then compare the variable values generated to values stored in the AWS secrets manager.

### Testing strategy

I would appreciate some feedback on the test strategy.

* **Current test strategy:** The test of `dss-ops.py secrets check` is an integration test only (which will ensure terraform is present when it is run). Currently test strategy is to test the CheckSecrets class, and then to just run the secrets check with no mocking or anything. (The rReason is, when we run integration tests, we would expect the secrets check to pass - if not, deploy process would fail). Adding the secrets check to the operations tests just causes that failure to happen during test stage, not during deploy stage.

* **Alternative test strategies:** Alternatively we could mock the secrets manager and mock the secret value to be whatever the terraform infra files contain for the secret value. Not sure if it makes sense to put in the work to do that - but open to alternatives.

Closes #2380 

Closes #2668


dss/operations/secrets.py
```diff
@@ -291,3 +293,163 @@ def del_secret(argv: typing.List[str], args: argparse.Namespace):
...
+        `type` should be `service_account`
+        `client_email` should be the user account returned from the terraform output "service_account".
+                       For example: dev should be `travis-test@human-cell-atlas-travis-test.iam.gserviceaccount.com`.
+    """
+    def __init__(self, stage):
+        """Initialize variables useful to the secrets checker"""
+        self.stage = stage
+        self.stages = {'dev': 'environment',
+                       'integration': 'environment.integration',
+                       'staging': 'environment.staging'}
```
amarjandu: what about `prod` ?

dss/operations/secrets.py
```diff
@@ -291,3 +293,163 @@ def del_secret(argv: typing.List[str], args: argparse.Namespace):
...
+        if p.returncode != 0:
+            raise RuntimeError(f'While checking secrets, an error occured:\n'
+                               f'stdout: {stdout}\n\n'
+                               f'stderr: {stderr}\n')
+        return stdout.decode('utf-8')
+
+    def get_stage_env(self, env_file):
+        """Source the environment file for this stage, then export the current environment as a dict"""
+        dump = 'python -c "import os, json; print(json.dumps(dict(os.environ)))"'
+        cmd = ['bash', '-c', f'source {env_file} && {dump}']
```
amarjandu: For sourcing `dev`
`source environment`

For sourcing `integration`
`source environment && source environment.integration`

dss/operations/secrets.py
```diff
@@ -291,3 +293,163 @@ def del_secret(argv: typing.List[str], args: argparse.Namespace):
...
+    def fetch_terraform_output(self, output_name, output_infra_dir):
+        """
+        Use the makefiles in the data-store repository to create terraform resource files and use these to
+        check environment variables. See: https://www.terraform.io/docs/commands/output.html
+        """
+        dss_root_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..'))
+        output_infra_dir = os.path.abspath(os.path.join(dss_root_dir, 'infra', output_infra_dir))
+
+        # populate infra vars for the current stage using terraform
+        self.run_cmd(cmd=f'make -C infra', cwd=dss_root_dir)
```
amarjandu: if `make -C infra` gets held up on waiting for a `yes` to apply changes what happens? I've never seen it happen with the current version, just playing devils advocate :)

dss/operations/secrets.py
amarjandu: Does the CI pipeline that runs the stage already have the first environment file sourced? if so i think all we would need is a comment about this.....

dss/operations/secrets.py
amarjandu: I've noticed that this check_secrets script takes a long time, it might be due to it attempting to build out every single terraform file, if all we need is `gcp_service_account` we could use `make -C infra COMPONENT=gcp_service_account` which would get us exactly what we need, and faster CI deployments :)

dss/operations/secrets.py
chmreid: This class was migrated from the prior check deployment secrets script, which did not have a prod stage originally either. (I'm not sure if we're running this script on prod.)

chmreid: To clarify: `make deploy` will invoke this script, so it will always run on a `make deploy`, but because no prod stage is defined in `self.stages`, the secrets checker will exit without doing anything when run on the prod stage. (Not sure why that decision was made, it makes sense to me to add the prod stage here.)

DailyDreaming: This was written by me before I had access to `prod`.  I could probably add it in after this is merged in a follow-up PR.

dss/operations/secrets.py
chmreid: The `environment` and `environment.$STAGE` files should be sourced before this test is run (see https://github.com/HumanCellAtlas/data-store/blob/master/.gitlab-ci.yml#L26-L29), so this is not an issue when running this as a test. 

However, we should make sure that if `scripts/dss-ops.py secrets check` is run independent of tests, it still sources environment files in the correct order...

chmreid: As it is written currently, it will not source environment files in the correct order - non-dev stages will only source `environment.$STAGE` and will not source `environment` first

dss/operations/secrets.py
chmreid: The `run_command()` function will wait for the command to finish, and will raise a RuntimeError if it does not complete successfully. If `make -C infra` were held up, the CI system would would sit and wait for the check secrets test to finish; if the `make -C infra` command finished with a failure, the tests would also fail. If the command never finished, the test would time out on Allspark (after 30 minutes, I think).

DailyDreaming:
Just a suggestion, but I notice that you use if args.quiet a lot.  It might be worth considering wrapping it as its own print function to declutter code.  Maybe something like:

```
def super_courteous_print(msg, quiet):
    if not quiet:
        print(msg)
```

Then something like

```
if args.dry_run:
    # Delete it for fakes
    if not args.quiet:
        print(f"Secret variable {secret_name} found in secrets manager, dry-run deleting it")
else:
    # Delete it for real
    if not args.quiet:
        print(f"Secret variable {secret_name} found in secrets manager, deleting it")
    sm_client.delete_secret(SecretId=secret_name)
```

Would become something like:

```
if args.dry_run:
    # Delete it for fakes
    super_courteous_print(f"Secret variable {secret_name} found in secrets manager, dry-run deleting it", quiet=args.quiet)
else:
    # Delete it for real
    super_courteous_print(f"Secret variable {secret_name} found in secrets manager, deleting it", quiet=args.quiet)
    sm_client.delete_secret(SecretId=secret_name)
```

chmreid:
> It looks like self.app_secret is being set to None?
@DailyDreaming fixed, this was due to leaving off the secret store prefix `/dcp/dss/STAGE/`.

chmreid:
> It looks like self.app_secret is being set to None?

@DailyDreaming fixed, this was due to leaving off the secret store prefix `/dcp/dss/STAGE/` and JSON that was a string, not a dict.

dss/operations/secrets.py
chmreid: Fixed by changing the command from `source {env_file} && {dump}` to `source environment && source {env_file} && {dump}`

dss/operations/secrets.py
chmreid: re adding `COMPONENT=gcp_service_account` doesn't seem to give much of a speedup. Changing the make command to `make -C infra COMPONENT=gcp_service_account` takes almost a minute:

```
$ time GOOGLE_APPLICATION_CREDENTIALS="${PWD}/gcp-credentials.json" ./scripts/dss-ops.py secrets check
Now checking the secrets for dev...
Secret check complete for dev.

real	0m54.767s
user	0m12.624s
sys	0m3.506s
```

while the original script using `make -C infra` (no component flag) takes ~50 seconds:

```
$ time GOOGLE_APPLICATION_CREDENTIALS="${PWD}/gcp-credentials.json" ./scripts/dss-ops.py secrets check
Now checking the secrets for dev...
Secret check complete for dev.

real	0m48.812s
user	0m12.521s
sys	0m3.221s
```

Passing an env var to subprocess this way may not actually be setting the env var, though, so I need to check whether the command is really exporting the component variable.

dss/operations/secrets.py
chmreid: Fixed by changing the command from `source {env_file} && {dump}` to `source environment && source {env_file} && {dump}`

chmreid:
codecov says none of the SecretsChecker class is being tested ([link](https://codecov.io/gh/HumanCellAtlas/data-store/compare/9891489189e31d1012f7e2e792398a36489c2e32...92e466f499294587299b9b3e7f8ac8fab05b3a3e/diff)), but there are several tests of the SecretsChecker. Looking into this...

chmreid:
Rebasing onto master, hopefully this fixes the codecov report

chmreid:
Ah, right, the code coverage report does not run integration tests, and the SecretsChecker test is an integration-only test.

Issue #2681, opened 2019-12-04
DSS Infra: CI IAM policy mismatch


amarjandu:
its possible for an IAM policy that is attached to a lambda role to become out of sync with what chalice uses for an operator.

This is caused by setting `mange_iam_role: false` in https://github.com/HumanCellAtlas/data-store/blob/5a0eb22718be10e5f025bba13dd4f1dcab6518cd/chalice/build_deploy_config.sh#L66

Although the policy is created on line 70, its not uploaded/updated with chalice, this can cause issues when policies are modified or changed, due to chalice using a different policy.

a fix for this would be finding out if we can change the name for the chalice policy that is created by an operator to be `dss-stage` rather than `dss-stage-api_handler`. 
Doing so enable the operator to always update the IAM policy as needed, and the CI deployment to just follow along and use the latest version of the policy.

amarjandu:
Issue addressed #2683

Issue #1973, opened 2019-03-13
Ensure deployed and expected IAM policies match


xbrianh:
This should be performed after each Lambda.

Linked to #1955

Issue #1955, opened 2019-03-06
Ensure consistent IAM role for deployed Lambdas


xbrianh:
Chalice occasionally deploys Lambdas to inconsistent roles. For instance, deploying the API server Lambda locally results in IAM role `dss-dev`. However, deployments from allspark use IAM role `dss-dev-api_handler`. This frequently causes IAM errors and broken deployments.

Root causes and workaround should be investigated.

xbrianh:
Is it possible to force allspark deploys to use current roles for Lambdas?

xbrianh:
see #1973

PR #2683, opened 2019-12-05
DSS Infra: Chalice IAM Policy

amarjandu:
Removes different IAM policies depending on deployment source.
Only updates IAM policies from non-CI environments (operator machines).
Addresses malformed IAM policy for dss-{stage} lambda. 

Connected To #2681 , #1973 , #1955

### DIFF
--- a/chalice/build_deploy_config.sh
+++ b/chalice/build_deploy_config.sh
@@ -60,11 +60,14 @@ for var in $(echo $env_json | jq -r keys[]); do
     cat "$config_json" | jq .stages.$stage.environment_variables.$var="$val" | sponge "$config_json"
 done
 
-if [[ ${CI:-} == true ]]; then
-    account_id=$(aws sts get-caller-identity | jq -r .Account)
-    export iam_role_arn="arn:aws:iam::${account_id}:role/dss-${stage}"
-    cat "$config_json" | jq .manage_iam_role=false | jq .iam_role_arn=env.iam_role_arn | sponge "$config_json"
-fi
-
+account_id=$(aws sts get-caller-identity | jq -r .Account)
+export iam_role_arn="arn:aws:iam::${account_id}:role/dss-${stage}"
+cat "$config_json" | jq .manage_iam_role=false | jq .iam_role_arn=env.iam_role_arn | sponge "$config_json"
 cat "$iam_policy_template" | envsubst '$DSS_S3_BUCKET $DSS_S3_CHECKOUT_BUCKET $dss_es_domain $account_id $stage' > "$policy_json"
 cp "$policy_json" "$stage_policy_json"
+
+if [[ ${CI:-} != true ]]; then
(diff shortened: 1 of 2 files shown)


iam/policy-templates/dss-lambda.json
```diff
@@ -69,7 +69,6 @@
         "Action": [
             "dynamodb:*"
         ],
-        "Resource": "arn:aws:dynamodb:*:$account_id:table/dss-async-state-$stage",
```
xbrianh: Consider making this change in a separate PR to disentangle deploy script changes from IAM policy updates.

chalice/build_deploy_config.sh
```diff
@@ -60,11 +60,14 @@ for var in $(echo $env_json | jq -r keys[]); do
...
+account_id=$(aws sts get-caller-identity | jq -r .Account)
+export iam_role_arn="arn:aws:iam::${account_id}:role/dss-${stage}"
+cat "$config_json" | jq .manage_iam_role=false | jq .iam_role_arn=env.iam_role_arn | sponge "$config_json"
 cat "$iam_policy_template" | envsubst '$DSS_S3_BUCKET $DSS_S3_CHECKOUT_BUCKET $dss_es_domain $account_id $stage' > "$policy_json"
 cp "$policy_json" "$stage_policy_json"
+
+if [[ ${CI:-} != true ]]; then
+	# IAM policies must be updated from an operators machine, this will not run on CI environments.
+	echo "updating $iam_role_arn with $stage_policy_json"
+	aws iam put-role-policy --role-name $lambda_name --policy-name $lambda_name --policy-document file://$stage_policy_json
```
xbrianh: What's the advantage of this over `cat "$config_json" | jq .manage_iam_role=true`?

iam/policy-templates/dss-lambda.json
amarjandu: shall do, ill merge it in first, as it needs to be fixed for the syntax to work in `aws  iam put-role-policy`

chalice/build_deploy_config.sh
amarjandu: For a given stage `dev`;
we dont use `auto-gen-policies` in our deployments script, by setting `manage_iam_role:true` we would end up with a policy that is called `dss-dev-function_handler`; there is no problem with this, the only issue is that when the CI uploads its manually managed IAM policy it uses `policy-dev.json`, but calls uses the arn name `dss-{dev}`. 

By always having a manually managed IAM Role we can keep using the same name `dss-dev` but only update the policy when we are not in a CI system. 

:)

iam/policy-templates/dss-lambda.json
amarjandu: @xbrianh please see https://github.com/HumanCellAtlas/data-store/pull/2685

PR #2690, opened 2019-12-07
user env var for user home dir

xbrianh:
Abstract user home into env var

### DIFF
--- a/build_image.Dockerfile
+++ b/build_image.Dockerfile
@@ -45,22 +45,23 @@ RUN python3 -m pip install virtualenv==16.0.0
 RUN ln -s /usr/bin/python3.6 /usr/bin/python
 RUN ln -s /usr/bin/pip3 /usr/bin/pip
 
-RUN useradd -d /home/hca_cicd -ms /bin/bash -g root -G sudo hca_cicd
+ENV USER_HOME ${USER_HOME}
+RUN useradd -d ${USER_HOME} -ms /bin/bash -g root -G sudo hca_cicd
 RUN mkdir /HumanCellAtlas && chown hca_cicd /HumanCellAtlas
 USER hca_cicd
-WORKDIR /home/hca_cicd
+WORKDIR ${USER_HOME}
 
-ENV PATH /home/hca_cicd/bin:${PATH}
-RUN mkdir -p /home/hca_cicd/bin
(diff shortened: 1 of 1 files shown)


build_image.Dockerfile
```diff
@@ -45,22 +45,23 @@ RUN python3 -m pip install virtualenv==16.0.0
 RUN ln -s /usr/bin/python3.6 /usr/bin/python
 RUN ln -s /usr/bin/pip3 /usr/bin/pip
 
-RUN useradd -d /home/hca_cicd -ms /bin/bash -g root -G sudo hca_cicd
+ENV USER_HOME ${USER_HOME}
```
amarjandu: Is ${USER_HOME} set by default on the base image we use?

build_image.Dockerfile
```diff
@@ -45,22 +45,23 @@ RUN python3 -m pip install virtualenv==16.0.0
...
+RUN useradd -d ${USER_HOME} -ms /bin/bash -g root -G sudo hca_cicd
```
chmreid: Regarding the use of the `-g` and `-G` flags here, `-g` is for providing a single group name, while `-G` is for providing a list of groups, so I would expect one or the other but not both

chmreid: (Just noticed that these flags weren't added by you, they were carried over from the previous useradd command...)

chmreid: (Just noticed that these flags weren't added by you, they were carried over from the previous useradd command...)

chmreid: I think I understand the logic - specifying `-g root` means that `root` will be the new user's initial login group. If you changed this from `-g root -G sudo` to `-G root,sudo` then the user's initial login group would be a new group that is the same as their username.

```
vagrant@ubuntu-bionic:~$ useradd -g root -G sudo batman
vagrant@ubuntu-bionic:~$ useradd -G root,sudo batman2
vagrant@ubuntu-bionic:~$ groups batman
batman : root sudo
vagrant@ubuntu-bionic:~$ groups batman2
batman2 : batman2 root sudo
```

build_image.Dockerfile
natanlao: To add on to this, it's not just bash - any POSIX-compliant system will have `$HOME` set by default.

PR #2708, opened 2020-01-09
Gentle refactor of tombstoning code, second try

natanlao:
Continuation of reverted PR #2516 with bug fixes.

This is the first of several PRs needed to implement the file deletion API (#2409). This PR refactors a tombstone helper function to be consistent with the deletion RFC and relocates it in preparation for use with files (and not just bundles).


natanlao:
Closed in favor of DataBiosphere/data-store#13

PR #2717, opened 2020-03-11
DSS INFRA: Update Google Cloud Function API

amarjandu:
changes gcp deployment from `v1beta2` to `v1` to avoid deprecation issues.


chmreid:
Should there also be a `API_VERSION = "v1"` changed somewhere?

amarjandu:
> Should there also be a `API_VERSION = "v1"` changed somewhere?

yes! it dropped that commit when i was rebasing with this remote, thanks! ill add the link too.

